PR #69, opened 2015-01-26
Change parentheses of sigma type of precategories

benediktahrens:
The function mapping an object to its identity morphism is unique, i.e. a property rather than a structure, hence should be in the "property component" of that sigma type.
This PR implements that change.

However, proofs of "is_precategory" should now be transparent, in order for identity to compute, e.g., 
(identity F) x ~> identity (F x) .


UniMath/Ktheory/AbelianGroup.v
```diff
@@ -474,12 +474,18 @@ Module Category.
     intros F G. apply (monoidfun F G).
   Defined.
   Definition Data : precategory_data.
-    exists ObMor. split. intro G. exists (idfun (G : abgr)). split. 
-    split. reflexivity. intros a b c.  exact monoidfuncomp. Defined.
+    exists ObMor. (* split. intro G. exists (idfun (G : abgr)). split.
```
DanGrayson: Why bother leaving old code in comments, here and elsewhere?

UniMath/RezkCompletion/ess_alg_categories.v
```diff
@@ -0,0 +1,52 @@
...
+    forall p : target f = source g, source (comp f g p) = source f.
+Definition comp_target := forall f g : mor C,
+    forall p : target f = source g, target (comp f g p) = target g.
+Definition assoc := forall f g h : mor C, 
+    forall p : target f = source g, forall q : target g = source h,
+      forall p' : target f = source (comp g h q),
+        forall q' : target (comp f g p) = source h,
+          comp f (comp g h q ) p' = comp (comp f g p) h q' .
+
+End axioms.
```
DanGrayson: missing newline at end of file

DanGrayson:
What does "identity function is an hprop" mean?

JasonGross:
I'd be curious to know if this change impacts the compilation time of any files.

DanGrayson:
How about arranging for travis to report the times?  It's an option in the
Makefile, already.

benediktahrens:
On 01/26/2015 06:02 PM, Daniel R. Grayson wrote:
> What does "identity function is an hprop" mean?

It means that it is unique, i.e., a property rather than a structure.

DanGrayson:
Oh, then the word "function" should be replaced by "arrow" or "morphism".
Then one
could say: given associativity, the statement that a mapping from objects
to endomorphisms
satisfies the two identity axioms is a proposition.  So moving it from
precategory_data to
is_precategory seems to make sense, and it would make the construction
parallel to what
Vladimir does in algebra1a in "isunital", "ismonoidop", and
"isapropismonoidop".

benediktahrens:
Oh, I have realized only just now that confusion might arise with the 
identity function "x \mapsto x".
My formulation was very unfortunate, sorry for that. I updated the comment.

DanGrayson:
Merging is okay with me.

DanGrayson:
Yes, "identity function" means "x \mapsto x".

benediktahrens:
On 01/26/2015 06:16 PM, Daniel R. Grayson wrote:
> How about arranging for travis to report the times?  It's an option in the
> Makefile, already.

I'd be happy to see it done, but I am a bit busy with other stuff right 
now, so won't be able to do it myself.

DanGrayson:
Me, too.  I've added it as issue
https://github.com/UniMath/UniMath/issues/71

PR #85, opened 2015-03-13
changed parameters of type of natural transformations

benediktahrens:
This change of parameters enables is necessary to work with rank 2 functors, that is, functors on (pre)categories of functor, in a reasonable way.
I have not seen any inconvenience of tightening the parameters in this way.

### DIFF
--- a/UniMath/RezkCompletion/equivalences.v
+++ b/UniMath/RezkCompletion/equivalences.v
@@ -54,8 +54,8 @@ Notation "G 'O' F [ hsB , hsC ]" := (functor_composite hsB hsC F G) (at level 20
 Definition form_adjunction (A B : precategory) 
    (hsA: has_homsets A) (hsB: has_homsets B)  (F : ob [A, B, hsB])
        (G : ob [B, A, hsA]) 
-       (eta : nat_trans (functor_identity A) (pr1 (FC hsB hsA F G)))  
-       (eps : nat_trans (pr1 (FC hsA hsB G F)) (functor_identity B)) : UU :=
+       (eta : nat_trans (functor_identity A) (FC hsB hsA F G))  
+       (eps : nat_trans (FC hsA hsB G F) (functor_identity B)) : UU :=
 dirprod 
   (forall a : ob A,
        # (pr1 F) (pr1 eta a) ;;   pr1 eps (pr1 F a) = identity (pr1 F a))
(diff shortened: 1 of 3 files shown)


vladimirias:
Can you point out where this change makes things easier?

benediktahrens:
On 03/14/2015 02:17 AM, Vladimir Voevodsky wrote:
> Can you point out where this change makes things easier?

The following situation occurs in a project I am working on with Ralph
Matthes:


Local Notation "F ⟶ G" := (nat_trans F G) (at level 39).
Local Notation "G ∙ F" := (functor_composite _ _ _ F G) (at level 35).

Variable C : precategory.
Variable hs : has_homsets C.

Variable H : functor [C, C, hs] [C, C, hs].

Local Notation "'U'" := (functor_ptd_forget C hs).

Variable θ : ∀ (F : functor C C) (Z : precategory_Ptd C hs),
      (H F) ∙ (U Z) ⟶ H (F ∙ (U Z)).


In particular, we have

H (F ∙ (U Z)) : ob [C, C, hs]

Without the modification I suggest in this PR, I get a type checking
error message when typing θ, complaining that it would expect

H (F ∙ (U Z)) : functor_data ? ?

The reason for this message is that the coercion

functor >-> functor_data

is not triggered when having a thing of type

ob [C, C, hs]

instead of type "functor C C", even though these are convertible.


So the problem seems to be that coercions are only searched with
unnormalized types, whereas here normalization would be needed to
"discover" the possibility to coerce.


Does that make sense?

benediktahrens:
I forgot to say: 
without this suggested change, to make the thing typecheck, I need to insert "pr1" occassionally, as in

Variable θ : ∀ (F : functor C C) (Z : precategory_Ptd C hs), 
   (H F) ∙ (U Z) ⟶ pr1 (H (F ∙ (U Z))).

and similarly everywhere where functors between functor categories are involved, e.g.,
for working with the forgetful functor from pointed endofunctors to endofunctors, here called U.

benediktahrens:
The following piece of code illustrates the point that I am trying to make:

Definition nat2 := {n : nat & True}.
Coercion coerc (n : nat2) : nat := projT1 n.
Variable x : nat2.
Check (x : nat). (* works *)
Definition constnat2 : bool -> Type := fun _ => nat2.
Variable y : constnat2 true.
Definition type_of {T} (t : T) := T.
Eval compute in (type_of y). (* { _ : nat & True } *)
Check (y : nat). (* fails *)

benediktahrens:
Are there any objections to this PR? I am still having the problem described above, which this change should remedy. So far no reason against the PR has come up.

DanGrayson:
It doesn't change any of my files, so it's okay with me automatically.

vladimirias:
No serious objection. I am Ok with merging.

benediktahrens:
I have come to the conclusion that this PR was not a good idea. More generally, I would like to rework RC, gearing it towards minimal hypotheses in the sense of removing dependency on the axiom part of a structure wherever possible.
Are you ok with me reverting this PR? 
Sorry for the hassle.

DanGrayson:
Sure, no hassle for me!

vladimirias:
If it is Ok with your applications then I am definitely for it - for reversing the PR and going in the direction of step-wise introduction of structures and axioms such that objects with intermediate combinations could be used when only these structures/axioms are needed. In my developments on B-systems (which are stalled now because I want to finish two papers for the usual publication first) I also use a lot statements where the structures required are separate parameters and then have a separate file where those of these statements that are needed for external use are re-stated in the way where the parameters are packaged together into structures. 

Vladimir.

PR #89, opened 2015-04-22
Fewer assumptions in constructions on functors and natural transformations

benediktahrens:
in particular in the definition of whiskering


DanGrayson:
okay with me

vladimirias:
Fine with me.

PR #90, opened 2015-05-06
Issue 40

DanGrayson:
fix it we don't have to run make twice

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,5 +5,4 @@ before_install:
   - sudo apt-get update -qq
   - sudo apt-get install -y ocaml ocaml-nox ocaml-native-compilers camlp4-extra 
 
-# run make twice to work around bug in Makefile
-script: make && make
+script: make
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,10 @@
 # -*- makefile-gmake -*-
 UMAKEFILES += Makefile
+ifneq "$(INCLUDE)" "no"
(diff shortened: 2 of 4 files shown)


benediktahrens:
On 05/06/2015 10:49 PM, Daniel R. Grayson wrote:
> fix it we don't have to run make twice

Awesome. I have tested it on my machine (Debian Wheezy, GNU make 3.81),
it works here.

DanGrayson:
We should modify the travis script, too, so it runs "make" only once.

benediktahrens:
On 05/07/2015 07:47 AM, Daniel R. Grayson wrote:
> We should modify the travis script, too, so it runs "make" only once.

Would you mind making this part of your PR? The file is 
UniMath/.travis.yml .


Would it be good to have a test file the compilation of which fails in 
case "make" does not compile the whole of UniMath?

At the moment, it just checks for exit status 0 of make, which would not 
discover problems we had with incomplete compilation.

For instance, one could explicitly compile a file test.v which imports a 
file from Ktheory.

Explicitly, I am thinking about amending the travis script with 
something like

make && make install
echo "Require Import UniMath.Ktheory.Elements." > test.v
./sub/coq/bin/coqc test.v

DanGrayson:
oops, hold on, I checked in too much.

DanGrayson:
transferring to https://github.com/UniMath/UniMath/pull/91

PR #93, opened 2015-05-12
Update Makefile

mortberg:
This fixes the issue I had with cloning and then making (without this change the modified version of Coq was never fetched).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,7 @@ ifeq ($(BUILD_COQ),yes)
 export PATH:=$(shell pwd)/sub/coq/bin:$(PATH)
 sub/coq/configure:
 	git submodule update --init sub/coq
-sub/coq/config/coq_config.ml: sub/coq/configure.ml sub/coq/configure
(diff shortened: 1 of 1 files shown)


DanGrayson:
It's best not to depend much on ordering of prerequisites, because people might run ```make``` in parallel.   What about instead adding ```sub/coq/configure.ml``` as another target of the command ```git submodule update --init sub/coq```?  That ought to work for you.

mortberg:
I hope I didn't misunderstand what you meant, but this works fine for me as well.

DanGrayson:
Travis will probably like that, in which case it looks good.  Thanks!

benediktahrens:
Anders, you have added some category theory stuff to that PR. Was that on purpose?

mortberg:
No that was not intentional. I didn't know that if I push to my fork it will automatically be added to the pull request...

mortberg:
Ok, now it should be fine and I have read up on how to properly develop using pull requests. Sorry for the confusion! :)

DanGrayson:
Thanks!

Issue #92, opened 2015-05-10
Switch to Coq 8.5


benediktahrens:
Is it too early to think about switching to Coq 8.5? 
Using the type-in-type flag (and -indices-matter), it should be possible to stick to vanilla Coq.
Maybe we should start a branch to gather some work towards that switch? It might take a while to adapt all the packages.

DanGrayson:
I tried that at https://github.com/DanGrayson/UniMath/commits/update-for-coq-trunk-unpatched and coq crashed, and I wasn't able to debug it.  But of course, we should return to that.

benediktahrens:
On 05/11/2015 03:23 AM, Daniel R. Grayson wrote:
> I tried that at
> https://github.com/DanGrayson/UniMath/commits/update-for-coq-trunk-unpatched
> and coq crashed, and I wasn't able to debug it.  But of course, we
> should return to that.

I have tried to change to vanilla Coq for the git submodule, see branch 
for-8.5 of my fork [1].
However, I did not get far:
$ make
Makefile:66: *** target file `clean' has both : and :: entries.  Stop.

A related question: In case we can use vanilla Coq 8.5, should we still 
carry it around as a submodule?
Maybe it makes sense to modularize the system a bit, by making Coq 8.5 a 
dependency rather than part of the installation.
1) People might already have a suitable installation of Coq on their 
machine, and
2) we might get a less complicated Makefile.


[1] https://github.com/benediktahrens/UniMath

vladimirias:
I am very much for it. I think it is a major milestone for us to be able to use of-the-shelf Coq.

Vladimir.

benediktahrens:
@DanGrayson : the tip of the branch you indicate fails even to start comiling: 
```
$ make 
Makefile:28: build/CoqMakefile.make: No such file or directory
making .coq_makefile_input
make: *** No rule to make target `sub/coq/configure.ml', needed by `sub/coq/config/coq_config.ml'.  Stop.
```
Do you remember when coq crashed in your tests?


I have made a branch "for-8.5-std-makefile" where I only use a simple wrapper script around the standard Coq Makefile.
Having Coq8.5beta2 [1] in my path, the first problem seems to be that a lot of "induction e" have to be replaced by "destruct e" for e being some path.
[1] https://github.com/coq/coq/releases/tag/V8.5beta2


One thing that I couldn't get to work: passing options "indices-matter" and "type-in-type" to coqc via coq_makefile. So for now I have to call make via
```
[...]

PR #94, opened 2015-05-13
For coq trunk unpatched

DanGrayson:
We take the work Benedikt did getting UniMath to compile with coq v8.5, unpatched
by us, but we keep the current makefile system working.  This will take care
of https://github.com/UniMath/UniMath/issues/92

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "sub/coq"]
 	path = sub/coq
 	url = https://github.com/coq/coq.git
-	branch = trunk
+	branch = v8.5
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "sub/coq"]
 	path = sub/coq
-	url = https://github.com/DanGrayson/coq.git
-	branch = trunk-patched
+	url = https://github.com/coq/coq.git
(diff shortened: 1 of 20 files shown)


vladimirias:
What happened in UniMath/RezkCompletion/precomp_ess_surj.v l. 361 and below. Why suddenly two of the arguments of something called Y need not be typed in anymore?

benediktahrens:
On 05/14/2015 12:51 PM, Vladimir Voevodsky wrote:
> What happened in UniMath/RezkCompletion/precomp_ess_surj.v l. 361 and
> below. Why suddenly two of the arguments of something called Y need
> not be typed in anymore?

The first two arguments of Y had been marked as implicit since the 
beginning, but Coq did not take that into account previously: despite 
having them marked as implicit, I still had to give those arguments 
later on.

The new Coq version behaves correctly, and so I could (and needed to) 
remove the two implicit arguments of Y.

(Those two arguments indicate source and target of the third argument, 
which is a morphism in some precategory.)

benediktahrens:
LGTM

DanGrayson:
Travis is very picky about white space.

Okay to merge, now that travis approves?

vladimirias:
Ok with me.

PR #95, opened 2015-05-18
added some comments about the two definitions of iso in a precategory

benediktahrens:
Mention both "iso" and the more intuitive "z_iso" in header of precategories.v

Changes only concern comments, not actual code

### DIFF
--- a/UniMath/RezkCompletion/precategories.v
+++ b/UniMath/RezkCompletion/precategories.v
@@ -15,12 +15,21 @@ Contents :  Definition of
 	        Categories (aka saturated precategories)         	
                 Setcategories
                 
-                Isomorphisms
+                Isomorphisms I: [iso]
+                  Definition: [isiso f := isweq (precomp_with f)]
                   various lemmas:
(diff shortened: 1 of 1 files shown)


DanGrayson:
Is it useful to have these two equivalent definitions?

benediktahrens:
They are only equivalent when hom-types are hsets.

Am 19. Mai 2015 16:14:21 OESZ, schrieb "Daniel R. Grayson" <notifications@github.com>:

DanGrayson:
Can you show that precomposition with f is an equivalence if and only postcomposition with f is?

benediktahrens:
On 05/19/2015 09:07 PM, Daniel R. Grayson wrote:
> Can you show that precomposition with f is an equivalence if and only postcomposition with f is?

They are both logically equivalent to having a two-sided inverse (a.k.a. 
being a z_isomorphism), it seems.
That might be worth adding to the library - I don't have time right now, 
but I'll file an issue as to not forget.

PR #97, opened 2015-05-25
Put opposite category in a separate file and minor changes

mortberg:
I put the opposite category in a separate file and defined the opposite of a functor. I also added a format rule to the "f ;; g" notation so that it gets printed nicely. Finally I prove that functor_identity is the left and right inverse of functor_composite, and that functor composite is associative.

### DIFF
--- a/UniMath/Ktheory/Precategories.v
+++ b/UniMath/Ktheory/Precategories.v
@@ -2,6 +2,7 @@
 
 Require Import Ktheory.Utilities.
 Require Import RezkCompletion.precategories 
+               RezkCompletion.opp_precat
                RezkCompletion.yoneda
                RezkCompletion.category_hset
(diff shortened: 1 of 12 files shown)


vladimirias:
The lemmas functor_idenity_left and functor_idenity_right should be provable without has_homsets assumption and this should be easy. The lemma functor_assoc it seems to me is also provable without has_homsets assumption but this might be more difficult. But it is worth trying and I can help if you get stuck.

benediktahrens:
I suggest merging this PR now - Vladimir's improved versions of the lemmas about functor composition can replace Anders' later.

vladimirias:
Fine with me. My PR will be ready very soon but there is a lot in Anders’s PR that is not in mine.

PR #98, opened 2015-05-26
Some updates to hnat.v

vladimirias:
Along with this updates to hnat there appeared in my fork many old copies of various files of Foundations that where sitting on a computer that I have not used. I don't know why they appeared and made my best to delete them. So they are not on the surface of the commit. Maybe there is a better way of getting rid of them so that they don't clutter up the history?


JasonGross:
They appeared because your repository is synced with Dropbox, and Dropbox and git don't always play nicely together.  If you want to not clutter up the history, you can run `git rebase -i upstream/master` or `git rebase -i 5f02a02`, and change some of the `pick`s to `squash`es to combine the relevant commits.  See also https://help.github.com/articles/about-git-rebase/.  You'll have to add the `--force` flag when you push, as in, e.g., `git push origin master:master --force`, if you rebase, in order to tell git that it's ok to rewrite history.

DanGrayson:
It seems to me that all the commits after 5f02a2 could simply be discarded.  In fact, one of them deleted sub/coq, so we can't permit that one to stay.  Hence rebasing is not a good idea.  It might be better to check out 5f02a2, give it a branch name, push the branch, and issue a pull request for that branch.  Eventually rename "master" to something else (in case you need it), and reinstitute the master branch.

vladimirias:
Travis says that it failed and points out file 05c8547 (funextfun (Fuld Hall ...)) that was deleted. Why does it try to compile deleted files?

vladimirias:
One of them deleted sub/coq, right. After that I tried to make pull request from UniMath to my branch to get a clean version of sub/coq but GitHub was (and is) telling me that everything is up to date and there is nothing to compare. 

Why is it so?

DanGrayson:
Jason, what is the precise problem with Dropbox and Git?  The discussion at http://stackoverflow.com/questions/1960799/using-git-and-dropbox-together-effectively seems ambiguous.

DanGrayson:
... Even if Dropbox makes files with names such as "uu0 (Fuld Hall iMac's conflicted copy 2015-04-01).v ", they would not end up getting added automatically to the repository.

PR #99, opened 2015-05-27
Changes to hnat and changes to functors_transformations

vladimirias:
I have re-done three additions that Anders proposed - that the composition of functors is unital on each side and that it is associative. I did it without the assumption that hom-types are sets. It was fun!

In the process I also changed a bit some things about functors in general. I think more can be done there to make future proofs more convenient.


vladimirias:
Something is wrong here. There are my changes that are adding new things and then a whole bunch of changes that take things back to an earlier stage. Dan, could you please sort it out?? It is very annoying.

DanGrayson:
Yes, let's sort that out first together.  Closing.

PR #100, opened 2015-05-28
Vladimir update

DanGrayson:
I'm under the impression that this pull request is an uncluttered version of what Vladimir submitted in #98 and #99, but since there were various merge conflicts, it should be checked carefully.  I don't understand why I had to totally replace three files.

### DIFF
--- a/UniMath/Foundations/hlevel2/hnat.v
+++ b/UniMath/Foundations/hlevel2/hnat.v
@@ -400,11 +400,13 @@ where "n + m" := (plus n m) : nat_scope.
 (** *** The structure of the additive ablelian monoid on [ nat ] *) 
 
 
-Lemma natplusl0 ( n : nat ) : paths ( 0 + n ) n .
+Lemma natplusl0 ( n : nat ) : 0 + n = n .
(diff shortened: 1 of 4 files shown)


vladimirias:
Actually it is not. This would reverse some of the work that Benedikt has done. I suggest the following: we drop this pull request completely. I move my Coq repos from Dropbox (already done) to somewhere. Jason, what cloud service is friendly to GitHub? I re-install clean UniMath (how can it be done without wiping out the history of my fork?) and then add the changes that I have made to hnat and to functors_transformations.

DanGrayson:
Sounds good to me -- I had a lot of trouble trying to figure out what your changes actually were.

By the way, with the newer Coq version, you have to fill in some implicit arguments with "hs" to get the proofs to work.

The only problem with your fork is that now the branch "master" is pointing at stuff you don't want.  If you could rename "master" to "old-master" on github, then you could checkout a fresh "master" branch from Unimath/Unimath.  Here is a discussion about renaming branches: http://stackoverflow.com/questions/9524933/renaming-a-branch-in-github .

PR #102, opened 2015-05-29
Slice categories

mortberg:
Definition of slice precategories with a proof that C/x is a category if C is and a prooof that any morphism x --> y in C induces a functor from C/x to C/y.


benediktahrens:
Two comments:
a) I would like the code in RezkCompletion to be CC0-licensed. Is that ok with you?
b) Would you mind adding information on authorship at the top of the file? 
(In the README for RezkCompletion it says that the authors are Ahrens, Kapulkin, and Shulman. I should add something like "unless mentioned otherwise", or remove that sentence altogether, since authorship info is in the individual files anyway.)

mortberg:
CC0 is fine with me. 

I added both our names to the header of the file as you helped me a lot with finding the proofs. I was not sure if I should order the names wrt first or last name? In the other files it there's no difference how one orders it... Also, should there really be a month and year information? It will be quite hard to remember to update it when changes are added and all of this information can be found in the git logs.

benediktahrens:
It's very good as it is now, thanks. No need to put dates.

At some suitable moment I will update the README to reflect the fact that Dan, Vladimir and you have contributed to RezkCompletion.

vladimirias:
Can we have a license that would require someone who wants to use the code to provide a general reference to UniMath? I would prefer this to a completely blank one like CC0 seems to be.

benediktahrens:
I suggest we have the discussion about licensing in a dedicated thread, see #103.

PR #106, opened 2015-05-29
speed up Integers.v

DanGrayson:


benediktahrens:
I originally unset this option only where strictly necessary, but it is a good idea to do some experiments to see where it is beneficial.

PR #107, opened 2015-06-03
Remove hom-set assumption in construction of equiv from ff ess.surj. functor

benediktahrens:


DanGrayson:
okay with me.

vladimirias:
Ok with me as well.

PR #108, opened 2015-06-18
Update coq version

DanGrayson:
Housekeeping.

The main change is to update coq to the latest commit on the trunk, so we can keep track of developments.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,7 @@ ifneq "$(INCLUDE)" "no"
 include build/CoqMakefile.make
 endif
 everything: TAGS all html install
+OTHERFLAGS += $(MOREFLAGS)
 OTHERFLAGS += -indices-matter -type-in-type
 ifeq ($(VERBOSE),yes)
 OTHERFLAGS += -verbose
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ time as ```gtime``` by running ```brew install gnu-time```.
 
 Timing of execution of individual tactics and vernacular commands can be obtained by
 ```bash
-$ make OTHERFLAGS=-time
+$ make MOREFLAGS=-time
 ```
 For postprocessing of the (huge) output, direct the output into a file as in
 ```bash


benediktahrens:
On 06/18/2015 11:31 PM, Daniel R. Grayson wrote:
> Housekeeping.
> 
> The main change is to update coq to the latest commit on the trunk, so we can keep track of developments.

I am a bit confused: I thought the plan was to stick with V8.5, that is,
to track the V8.5 branch?
But maybe we have never discussed that, and I just implicitly assumed
so. Anyway, maybe that's the moment to discuss it: would it not be good
(better?) to track a stable version instead of trunk?

DanGrayson:
Ah: a good reason to track v8.5 is so the forthcoming release of 8.5 doesn't break our code, so I'd be happy to prepare a pull request that does that.  But then after 8.5 is released, and our code works on it, we should follow the trunk to make sure subsequent changes don't break our code.

benediktahrens:
On 06/19/2015 12:22 AM, Daniel R. Grayson wrote:
> Ah: a good reason to track v8.5 is so the forthcoming release of 8.5
> doesn't break our code, so I'd be happy to prepare a pull request
> that does that.  But then after 8.5 is released, and our code works
> on it, we should follow the trunk to make sure subsequent changes
> don't break our code.

Couldn't we have a stable branch in which we track the latest released 
Coq version, and a trunk branch where we track Coq trunk?

I agree that it would be good to check continuously whether UniMath (or 
at least Foundations) works with Coq trunk.
But having this as the *only* version would require us to work around 
temporary bugs or incompletenesses in Coq to keep having a working UniMath.
However, if we have a stable working branch, we are not obliged to fix 
every problem that might occur with trunk (if the problem is stupid), 
and we are not obliged to fix problems immediately (if the problem is a 
big one).

I also think that it is easier to develop new libraries against a stable 
version of Coq, and to adapt them to new releases afterwards.

Both solutions, tracking only trunk and having two version, come with 
some overhead as compared to only having the stable branch, but I think 
that the overhead of having both branches comes with more benefits than 
the overhead of having only trunk.

DanGrayson:
I anticipate the procedure being a little different.  Every few weeks we try to update the coq version.  If everything works, we commit it we all start using it.  If something doesn't work we report the bug and stick with the commit version that does work.  Hence every commit of UniMath would work.

benediktahrens:
On 06/19/2015 07:22 PM, Daniel R. Grayson wrote:
> I anticipate the procedure being a little different.  Every few weeks
> we try to update the coq version.  If everything works, we commit it
> we all start using it.  If something doesn't work we report the bug
> and stick with the commit version that does work.  Hence every commit
> of UniMath would work.

Ok, sounds great!

PR #110, opened 2015-06-24
Draft license and copyright

DanGrayson:
for your consideration

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,28 @@
+UniMath copyright and license
+=============================
+
+UniMath consists of the files in the UniMath github repository at
+https://github.com/UniMath/UniMath .
+
+UniMath is copyright 2015 by the UniMath Development Team, whose members are
+listed at https://github.com/UniMath/UniMath in the file README.md .
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of
+UniMath, to deal in UniMath without restriction, including without limitation
+the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+sell copies of UniMath, and to permit persons to whom UniMath is furnished to
+do so, subject to the following conditions:
+
+* This copyright and license shall be included in all copies or substantial
+  portions of the Software.
+
+* If UniMath is used in a paper, please cite it.  If you use latex with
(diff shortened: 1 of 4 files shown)


benediktahrens:
ok from A,K,S

vladimirias:
I suggest that we change: "This copyright and license notice shall be included in all copies or
+  substantial portions of the Software" to make it read "This copyright and license notice as well as the paragraph about the UniMath development team shall be included in all copies or
+  substantial portions of the Software"

DanGrayson:
I purposely left the paragraph about the membership of the team out of the text to be copied, because we may want to change the membership of the team from time to time.  It would be bad if the membership changed, and we edited the README file to change it, but then many copied statement with conflicting information remained out there in the world.

How about instead including a link in the copyright notice to our github site stating that the membership list is always visible there?

I have a second suggestion: let's put the license and copyright information in a separate file called "LICENSE".  Then that makes it much easier for others to propagate the text of it, since all they have to do is to propagate the file.

vladimirias:
I think this is Ok. There will be many copies of this header (about the development team and the license and copyright) in the world. They will differ in the composition of the development team section depending on when a particular transaction involving taking a part of UniMath was made.

In fact, I am not even sure how it should look like in practice. Suppose that I want to copy some code from UniMath for my own use (say after modification). Where will I have to put the copy of the “header” (with or without the team composition info)?

DanGrayson:
I still think it's best to have the team membership info in one place only.  Actually, some of those who copy UniMath will copy LICENSE as well as README, so they will get both bits of information.

Someone who copies 1000 lines and puts it all into one file would be free to add the LICENSE to the same file.  Or they could make two files and put them in the same subdirectory, with an indication somewhere that the 1000 lines are governed by the LICENSE file.

vladimirias:
Ok.

DanGrayson:
I've made the changes discussed above, so perhaps we are almost ready to accept them.

vladimirias:
LGTM

benediktahrens:
lgtm

Am 4. Juli 2015 18:25:54 MESZ, schrieb "Daniel R. Grayson" <notifications@github.com>:

PR #114, opened 2015-07-03
Update coq 8.5

DanGrayson:
Here we update to the latest coq on the 8.5 branch, after a bug we encountered
(involving Print Assumptions) was fixed for us.


benediktahrens:
I have tested the 'substitution' repo with it as well, works fine.

Issue #83, opened 2015-03-09
add a target for building coqide


benediktahrens:
Add a target to the UniMath Makefile that builds coqide.

DanGrayson:
I've started working on this in https://github.com/DanGrayson/UniMath/tree/issue-83

benediktahrens:
Resolved in #115 .

PR #115, opened 2015-07-04
Issue 83

DanGrayson:
Modify makefile so we can optionally build coqide, as well as coq.
This involves a new submodule: lablgtk.

This will resolve issue https://github.com/UniMath/UniMath/issues/83.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,8 @@
 [submodule "sub/coq"]
 	path = sub/coq
-	url = https://github.com/DanGrayson/coq.git
-	branch = trunk-patched
+	url = https://github.com/coq/coq.git
+	branch = v8.5
+[submodule "sub/lablgtk"]
+	path = sub/lablgtk
+	url = http://forge.ocamlcore.org/anonscm/git/lablgtk/lablgtk.git
+	branch = master
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language:       ocaml
 before_install: sudo apt-get update -qq
-install:        sudo apt-get install -y ocaml ocaml-nox ocaml-native-compilers camlp4-extra time
+install:        sudo apt-get install -y ocaml ocaml-nox ocaml-native-compilers camlp4-extra time libgtk2.0 libgtksourceview2.0
 script:
-  - time make build-coq
+  - time make BUILD_COQIDE=yes build-coq
   - time make TIMECMD=time all
   - time make install
 after_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,10 @@
-language: ocaml
-
-# install build dependencies before attempting to build Coq
-before_install:
-  - sudo apt-get update -qq
-  - sudo apt-get install -y ocaml ocaml-nox ocaml-native-compilers camlp4-extra 
-
-# run make twice to work around bug in Makefile
-script: make && make
+language:       ocaml
+before_install: sudo apt-get update -qq
+install:        sudo apt-get install -y ocaml ocaml-nox ocaml-native-compilers camlp4-extra time
+script:
+  - time make build-coq
+  - time make TIMECMD=time all
+  - time make install
+after_script:
+  - test -f UniMath/Ktheory/Circle.vo
+  - test -f sub/coq/user-contrib/UniMath/Ktheory/Circle.vo
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,28 @@
+UniMath copyright and license
+=============================
+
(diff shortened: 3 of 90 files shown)


benediktahrens:
Compilation failed for me because of missing camlp4, see below. After installing camlp4, installation worked well. camlp4 is not needed for compilation of current master, instead one can use camlp5.

Is there a reason why for Linux we recommend installing camlp4-extra, whereas for MacOS we recommend installing camlp5?

Is it necessary to build lablgtk from source? I have built Coq8.5 in the past with the lablgtk included in my (rather old) Debian distribution. So for most Linux distributions it does not seem necessary to build lablgtk from source, and for MacOS you have added it to the installation instruction for brew.


ahrens@alien:~/git/benediktahrens/UniMath$ make BUILD_COQIDE=yes
Makefile:34: build/CoqMakefile.make: No such file or directory
git submodule update --init sub/lablgtk
Submodule 'sub/lablgtk' (http://forge.ocamlcore.org/anonscm/git/lablgtk/lablgtk.git) registered for path 'sub/lablgtk'
Cloning into 'sub/lablgtk'...
Submodule path 'sub/lablgtk': checked out 'c6345f655c0558f8c606477d8fea5f108f9a54bb'
cd sub/lablgtk && ./configure
checking for ocamlc... ocamlc
ocaml version is 3.12.1
ocaml library path is /usr/lib/ocaml
Default install dirs are : /usr/lib/ocaml/lablgtk2 and /usr/lib/ocaml/stublibs
Compile with ocamlc -I +lablgtk2
checking for ocamlopt... ocamlopt
checking ocamlopt version... ok
checking for ocamlc.opt... ocamlc.opt
checking ocamlc.opt version... ok
checking for ocamlopt.opt... ocamlopt.opt
checking ocamlc.opt version... ok
checking for ocamlrun... ocamlrun
checking for ocamldep... ocamldep
checking for ocamldoc... ocamldoc
checking for ocamllex... ocamllex
checking for ocamlyacc... ocamlyacc
checking for ocamlmktop... ocamlmktop
checking for ocamlmklib... ocamlmklib
checking for camlp4o... no
configure: error: Cannot find camlp4o.
make: *** [sub/lablgtk/src/gSourceView2.cmi] Error 1
ahrens@alien:~/git/benediktahrens/UniMath$

DanGrayson:
Good questions.

I think camlp4 and camlp5 are interchangeable, from our point of view, so
whichever one
we can get most easily is the one we should get.  A side effect of
installing camlp4-extra
under ubuntu is that camlp4 is installed, so I don't see why installing
camlp4 separately
would have helped.  A side effect of installing camlp5 is that camlp4 is
installed, too.
Perhaps we should just tell the user to install all three?

I just tried again with "brew" under Mac OS X, and when it installs
lablgtk, it leaves out
a file that coq wants to find:
```
cd sub/coq && ./configure -coqide "opt" -lablgtkdir /usr/local -opt
-no-native-compiler -with-doc no -annotate -debug -local
Warning: obsolete -opt option
You have GNU Make 4.1. Good!
You have OCaml 4.02.1. Good!
You have Camlp5 6.12. Good!
You have native-code compilation. Good!
Incomplete LablGtk2 (manually provided): no /usr/local/gSourceView2.cmi.
Configuration script failed!
```
So either we have to build lablgtk ourselves, or we have to get homebrew to
modify their recipe
somehow so the whole thing gets built (?).  The latter would certainly be
cleaner.

DanGrayson:
How much do we care about making coqide, anyway?  Emacs is better.

DanGrayson:
Re: "Compilation failed for me because of missing camlp4"

What OS was that under?

DanGrayson:
I propose we just ask users to install camlp4.  I've updated the README to reflect that.  The latest homebrew recipe for lablgtk still doesn't provide the file we need, and there is no way to tell lablgtk to use camlp5 instead of camlp4.

benediktahrens:
On 08/13/2015 03:39 PM, Daniel R. Grayson wrote:
> Re: "Compilation failed for me because of missing camlp4"
> 
> What OS was that under?

That was under Debian 7.
I have recently upgraded to Debian 8, but not checked if the error persists.

DanGrayson:
Okay, I've tested it under Debian 8.  It turns out two more packages need to be installed, so I've added them to the list in INSTALL.

DanGrayson:
Shall we modify the travis script so it always builds coqide, too?

benediktahrens:
On 08/13/2015 11:06 PM, Daniel R. Grayson wrote:
> Shall we modify the travis script so it always builds coqide, too?

Yes, that would be good, I think.

DanGrayson:
Okay, travis seems to have built coqide.

PR #118, opened 2015-07-12
some improvements to FunctorAlgebras

benediktahrens:
small modifications:
- introducing a name for functor algebra morphism diagram
- opacifying some proofs

### DIFF
--- a/UniMath/RezkCompletion/FunctorAlgebras.v
+++ b/UniMath/RezkCompletion/FunctorAlgebras.v
@@ -40,12 +40,19 @@ Variable F : functor C C.
 
 Definition algebra_ob : UU := Σ X : C, F X ⇒ X.
 
-Coercion ob_from_algebra_ob (X : algebra_ob) : C := pr1 X.
+(* this coercion causes confusion, and it is not inserted when parsing most of the time
+   thus removing coercion globally
+*)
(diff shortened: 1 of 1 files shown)


vladimirias:
LGTM with a question: what is abstract ( … ) ?

benediktahrens:
On 07/12/2015 10:48 PM, Vladimir Voevodsky wrote:
> LGTM with a question: what is abstract ( … ) ?

It is an opacifier, i.e., it makes the subterm created between those
parentheses opaque.

It is the first time that I have been using it, but it seems quite
useful when one would like to opacify part of a term, but does not want
to write an extra lemma for this term, in order to close it with Qed.

For instance, the lemma might need a lot of context to be copied, or
might be of too little general use to deserve being a lemma.

Actually, internally a lemma is created for the term created by the
tactics in parentheses, see [1]. But it is still more convenient.

One inconvenience of abstract (...) is that only one tactic can be
between (...), so one has to combine them with semicolons.


[1] https://coq.inria.fr/refman/Reference-Manual011.html#hevea_tactic189

vladimirias:
Another inconvenience maybe that there is probably no “transparent” for this “abstract”. But it is worth a try to see how it works.

PR #120, opened 2015-07-29
Some tactics.

mawarren:
A few tactics for manipulating various algebraic structures and natural
numbers in UniMath.

### DIFF
--- a/UniMath/Tactics/Monoids_Tactics.v
+++ b/UniMath/Tactics/Monoids_Tactics.v
@@ -3,7 +3,7 @@
 (** Description: Some tactics for monoids.*)
 
 Require Import Foundations.hlevel2.algebra1b
-               UniMath.Utilities.
+               UniMath.Tactics.Utilities.
 
 Ltac op_strip f :=
   repeat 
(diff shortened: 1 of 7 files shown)


benediktahrens:
I cannot comment on the content of the PR yet, but I want to point out a problem that is exhibited by the PR: there are some unqualified Imports in Ktheory and in Tactics, which lead to at least one clash, when importing a file called "Utilities" - this exists in both Ktheory and in Tactics.

@mawarren: could you please qualify all Imports, as in "Require Import UniMath.Tactics.Utilities."?
@DanGrayson: In file "Equivalences" and maybe elsewhere, there remain unqualified Imports.

UniMath/Tactics/Monoids_Tactics.v
```diff
@@ -0,0 +1,153 @@
+(** Author: Michael A. Warren.*)
+(** Date: Spring 2015.*)
+(** Description: Some tactics for monoids.*)
+
+Require Import Foundations.hlevel2.algebra1b
+               UniMath.Utilities.
```
benediktahrens: Here it should read "UniMath.Tactics.Utilities" instead of "UniMath.Utilities". 
(The corresponding error message can be seen at the bottom of https://travis-ci.org/UniMath/UniMath/builds/73476873 ).

benediktahrens:
We should integrate #121 before integrating this PR.

vladimirias:
LGTM

PR #121, opened 2015-07-31
added qualifications for imports in Ktheory

benediktahrens:
Add qualifications to imports in Ktheory, which is necessary before PR #120 can be integrated.


benediktahrens:
Partially resolved by #119 already, which I had overlooked.
Some missing qualifications in imports should be addressed in a new PR.

Issue #124, opened 2015-08-02
UniMath/Foundations/Generalities/uuu.v: references to dependent sum as record


nyraghu:
There are comments in uuu.v saying:

"It contains some new notations for constructions defined in Coq.Init library as well as the definition of dependent sum as a record."

"One could use "Inductive" instead of "Record" here but using "Record" which is equivalent to "Structure" allows us later to use the mechanism of canonical structures with total2."

They give the impression that dependent sums will be defined as records. However the current definition of total2 uses Inductive, not Record.  Perhaps these comments can be removed or corrected.

nyraghu:
I just now saw that these comments are mentioned in issue #111 too.  Apologies for reporting a known issue!

benediktahrens:
Thanks for your comment! I updated the comment in #125 .

nyraghu:
Thank you for the correction.

There's still one comment in line 3 which says that the file defines dependent sums as records:

"It contains some new notations for constructions defined in Coq.Init library as well as the definition of dependent sum as a record."

benediktahrens:
Right, thanks. I updated #125.

vladimirias:
Thanks for reminding us about this inconsistency anyway. I suggest that we remove the comment.

benediktahrens:
Fixed with PR #125 .

PR #125, opened 2015-08-02
adapted comment for total2, following #124

benediktahrens:
responding to #124  .

### DIFF
--- a/UniMath/Foundations/Generalities/uuu.v
+++ b/UniMath/Foundations/Generalities/uuu.v
@@ -84,9 +84,9 @@ canonical structures with total2.
 By using "Structure", we could also get eta for dependent pairs, by adding the option 
 "Set Primitive Projections.".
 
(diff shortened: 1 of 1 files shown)


vladimirias:
We should also add:

“However by “inductive” allows us to obtain proof terms that are expressed in terms of the eliminator total2_rect that, unlike the “match" construct that would appear in the proof terms if we used “record”, has a known interpretation in the framework of the univalent model.”

benediktahrens:
On 08/02/2015 08:19 PM, Vladimir Voevodsky wrote:
> We should also add:
> 
> “However by “inductive” allows us to obtain proof terms that are
> expressed in terms of the eliminator total2_rect that, unlike the
> “match" construct that would appear in the proof terms if we used
> “record”, has a known interpretation in the framework of the
> univalent model.”

Done.

PR #134, opened 2015-08-14
remove trailing white space from all lines

### LINKED ISSUE #131: Several files: terminal white space
There are several files which contain instances of terminal white space, i.e., white space after periods, etc.  Perhaps these can be removed.

DanGrayson:
resolves #131

### DIFF
--- a/UniMath/Foundations/Generalities/uu0.v
+++ b/UniMath/Foundations/Generalities/uu0.v
@@ -1,6 +1,6 @@
-(** * Univalent Basics. 
+(** * Univalent Basics.
 
-Vladimir Voevodsky. 
+Vladimir Voevodsky.
 Feb. 2010 - Sep. 2011.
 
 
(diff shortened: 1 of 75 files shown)


benediktahrens:
A general remark: this is a rather big patch for not much benefit. 
For the sake of keeping the repository small, I would like to avoid this kind of patch.
Also, it is very difficult to check this patch, because of missing line wrapping in the Github code view.

That being said, I don't want to oppose integration of this patch.

vladimirias:
I would very much prefer not to add anything that is hard to understand to the repository.

White spaces at the end of the lines is not something that a common user would be concerned with is it?

vladimirias:
Sorry, I take my comment back, just too tired (it’s 10:58pm in Singapore).

DanGrayson:
The benefit from removing white space is nil, as far as I can see.
Checking the patch is easy, though, using the "-b" option to "git diff" --
you'll be informed that, modulo changes to white space, there are no
changes.  Repository size is no problem, though, since git stores slight
changes efficiently, and since computers are gigantic.

benediktahrens:
LGTM

DanGrayson:
Let's not bother with removing white space.  Closing.

Issue #129, opened 2015-08-12
UniMath/Foundations/Generalities/uu0a.v: pretty-printing of ~


nyraghu:
Coqdoc wrongly pretty-prints the homotopy infix operator ~ as the negation symbol ¬ in the HTML documentation created by "make doc".  [Section 15.4.1 of the Coq manual](https://coq.inria.fr/distrib/V8.5beta2/refman/Reference-Manual017.html#coqdoc) says that this translation is specified by the current pretty-printing table.  To correct it, it appears that one has to write

(** printing ~ #~# *)

or perhaps

(** remove printing ~ *)

in each file that uses the ~ notation.

benediktahrens:
I find all the entries of the whole pretty-printing table unnecessary and confusing. More specifically:

* it breaks copy-and-paste from HTML into vernacular files
* the benefit is minimal
* the right solution to this problem (if one considers it to be one) is Unicode, in my opinion
* in particular, the "star" is also used in proof mode as one of the "bullet" levels, and there should not be translated to "times" at all anyway, but is nevertheless.

I am thus suggesting to remove all pretty-printing, not just the one for tilde.

Do you know of a global option to remove all pretty-printing?

DanGrayson:
There doesn't seem to be such a command line option.  I bet they'd accept a pull request from us that implemented one.

nyraghu:
A heavy-handed way to do this is to remove "-utf8" from COQDOCFLAGS and from COQDOC in the UniMath Makefile.  But this would destroy all Unicode text in comments, e.g., non-breaking space between two identifiers in a Coq code snippets like [foo&nbsp;x].

benediktahrens:
On 08/14/2015 04:56 PM, N. Raghavendra wrote:
> A heavy-handed way to do this is to remove "-utf8" from COQDOCFLAGS
> and from COQDOC in the UniMath Makefile.  But this would destroy all
> Unicode text in comments, e.g., non-breaking space between two
> identifiers in a Coq code snippets like [foo&nbsp;x].

Note that there are a few files in UniMath/RezkCompletion that make use
of Unicode not just in comments, but in the actual code.

nyraghu:
Sorry, I didn't know that.  So anyway this isn't a way out.

nyraghu:
As an experiment, I've generated the docs with the command

[...]

PR #135, opened 2015-08-14
remove -utf8 option from coqdoc command

DanGrayson:
resolve issue #129

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,10 @@ COQIDE_OPTION := opt
 LABLGTK := -lablgtkdir "$(shell pwd)"/sub/lablgtk/src
 endif
 endif
+
+# override the definition in build/CoqMakefile.make, to eliminate the -utf8 option
+COQDOCFLAGS := -interpolate
+
 ifneq "$(INCLUDE)" "no"
 include build/CoqMakefile.make
 endif
@@ -42,7 +46,7 @@ OTHERFLAGS += -verbose
 endif
 ENHANCEDDOCTARGET = enhanced-html
 ENHANCEDDOCSOURCE = util/enhanced-doc
-COQDOC := $(COQDOC) -utf8
+COQDOC := $(COQDOC)
 COQDEFS := --language=none -r '/^[[:space:]]*\(Axiom\|Theorem\|Class\|Instance\|Let\|Ltac\|Definition\|Lemma\|Record\|Remark\|Structure\|Fixpoint\|Fact\|Corollary\|Let\|Inductive\|Coinductive\|Notation\|Proposition\|Module[[:space:]]+Import\|Module\)[[:space:]]+\([[:alnum:]'\''_]+\)/\2/'
 TAGS : $(VFILES); etags $(COQDEFS) $^
 install:all


benediktahrens:
Do you understand why this works?
If not, I would prefer asking a question about the deactivation of pretty-printing on Coq-club over using this hack.
Can we postpone this until then?

DanGrayson:
No, I don't understand it, because the usage message says this about the utf8 option:
```
  --utf8               set UTF-8 input language
```
We can postpone it while you inquire.

benediktahrens:
I had a superficial look at the coqdoc source code. Below are some of the functions involved:

in tools/coqdoc/main.ml, 

| ("-utf8" | "--utf8") :: rem ->
        Cdglobals.set_utf8 (); parse_rec rem

which sets, in tools/coqdoc/cdglobals.ml

let set_utf8 () =
  charset := "utf-8";
  inputenc := "utf8x";
  utf8 := true

In tools/coqdoc/output.ml, we translate to Unicode when utf8 is true:

let initialize_tex_html () =
  let if_utf8 = if !Cdglobals.utf8 then fun x -> Some x else fun _ -> None in
  let (tree_latex, tree_html) = List.fold_right (fun (s,l,l') (tt,tt') ->
      (Tokens.ttree_add tt s l,
       match l' with None -> tt' | Some l' -> Tokens.ttree_add tt' s l'))
      [ "*" ,  "\\ensuremath{\\times}", if_utf8 "×";
        "|", "\\ensuremath{|}", None;
        "->",  "\\ensuremath{\\rightarrow}", if_utf8 "→";
        "->~",  "\\ensuremath{\\rightarrow\\lnot}", None;
        "->~~",  "\\ensuremath{\\rightarrow\\lnot\\lnot}", None;
        "<-",  "\\ensuremath{\\leftarrow}", None;
        "<->", "\\ensuremath{\\leftrightarrow}", if_utf8 "↔";
        "=>",  "\\ensuremath{\\Rightarrow}", if_utf8 "⇒";
        "<=",  "\\ensuremath{\\le}", if_utf8 "≤";
        ">=",  "\\ensuremath{\\ge}", if_utf8 "≥";
        "<>",  "\\ensuremath{\\not=}", if_utf8 "≠";
        "~",   "\\ensuremath{\\lnot}", if_utf8 "¬";
        "/\\", "\\ensuremath{\\land}", if_utf8 "∧";
        "\\/", "\\ensuremath{\\lor}", if_utf8 "∨";
        "|-",  "\\ensuremath{\\vdash}", None;
        "forall", "\\ensuremath{\\forall}", if_utf8 "∀";
        "exists", "\\ensuremath{\\exists}", if_utf8 "∃";
        "Π", "\\ensuremath{\\Pi}", if_utf8 "Π";
        "λ", "\\ensuremath{\\lambda}", if_utf8 "λ";
        (* "fun", "\\ensuremath{\\lambda}" ? *)
      ] (Tokens.empty_ttree,Tokens.empty_ttree) in
  token_tree_latex := tree_latex;
  token_tree_html  := tree_html


On the other hand, omitting --utf8 will result in a wrong html header, see function header () in output.ml .

DanGrayson:
Then all we have to do, in addition to removing --utf8, is to add --charset utf-8, to get the character
set correct in the html.  Does that work?

For consistency, if coqdoc can output → instead of -> optionally, than coqtop should be able to do
the same on input.

DanGrayson:
No, we're also using the option ```--with-header util/enhanced-doc/header.html```, which overrides the function header () in output.ml.

DanGrayson:
... so nothing further is needed : the charset is always utf-8, no matter what.

benediktahrens:
Actually, we are using those custom headers when making the enhanced html (with toggable proofs), but we are using the standard headers when doing "make html", i.e., when producing the standard html.

But anyway, browsers can generally cope with wrong information in the header, so we just simply use your suggested solution.

DanGrayson:
I think we make only the enhanced version, with "make doc".  But I have added the
charset option to the Makefile in the recent commit, which will be better, anyway.

How about that?

benediktahrens:
Perfect, thanks a lot!

Issue #148, opened 2015-09-03
Undefined reference to weqgf and weqgf00 in uu0a.v


nyraghu:
uu0a.v has the comment

`We now define different homotopies and maps between the paths spaces corresponding to a weak equivalence. What may look like unnecessary complexity in the definition of weqgf is due to the fact that the "naive" definition, that of weqgf00 , needs to be corrected in order for lemma weqfgf to hold. '

The identifiers weqgf and weqgf00 don't seem to occur anywhere in the library, and can perhaps be removed.

DanGrayson:
resolved in https://github.com/UniMath/UniMath/pull/149

PR #149, opened 2015-09-03
guess how to fix a comment of VV

DanGrayson:
addresses #148 

Vladimir should check

### DIFF
--- a/UniMath/Foundations/Generalities/uu0a.v
+++ b/UniMath/Foundations/Generalities/uu0a.v
@@ -784,9 +784,9 @@ Definition invmap {X Y : UU} (w : weq X Y) : Y -> X :=
 
 (** We now define different homotopies and maps between the paths
     spaces corresponding to a weak equivalence. What may look like
-    unnecessary complexity in the definition of [ weqgf ] is due to the
-    fact that the "naive" definition, that of [ weqgf00 ], needs to be
-    corrected in order for lemma [ weqfgf ] to hold. *)
+    unnecessary complexity in the definition of [ homotinvweqweq ] is due to the
+    fact that the "naive" definition needs to be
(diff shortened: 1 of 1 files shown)


vladimirias:
I introduced that other definition to show what it was precisely that I meant by saying "naive definition".  I would suggest not to remove it but maybe make an extra comment saying that it is there for clarification only and is not used anywhere later.

DanGrayson:
I didn't see anything sensible to replace ```weqgf00``` with, which is why I removed it.  Old versions of the code where the other two routines had their old names had nothing with that name.  What would you suggest replacing it with?

vladimirias:
The same name that you gave to weqgf extended with _naive .

DanGrayson:
The name ```homotinvweqweq_naive``` isn't currently being used, so it doesn't make sense to refer to it.  What's needed is the name of an existing lemma that formalizes the naive proof.

vladimirias:
I meant the old lemma (what was it weqfg00?) renamed into homotinvweqweq_naive . Or there was no such lemma? Then I misunderstood.

DanGrayson:
As far as I can tell, the old lemma weqfg00, presumed to exist by the
comment,
never did.  Hence the dilemma.

On Sat, Sep 5, 2015 at 2:35 PM, Vladimir Voevodsky <notifications@github.com

vladimirias:
In the version that I have the names weqgf and weqfg are replaced in the main body of the code but not in comments.

vladimirias:
I see now. Yes, your PR is what is needed - just please update it to have the right names in the comments.

DanGrayson:
Re: "just please update it to have the right names in the comments."

So the pull request is not good enough yet?  If so, could you be more precise?  (I think
the pull request does provide the right names in the comments.)

PR #154, opened 2015-09-24
makefile target for making a single package

DanGrayson:
add a makefile target for making all the files in a given package,
together with just the files needed by them from the previous packages.

Example: make CategoryTheory


benediktahrens:
Works nicely, thanks!

benediktahrens:
Actually, one problem is that "make install" still builds all the files.

DanGrayson:
As it should.  Why not?

PR #157, opened 2015-09-27
README and automatic use of correct Coq program in ProofGeneral

DanGrayson:


DanGrayson:
this is included in my next pull request ,closing

PR #161, opened 2015-09-27
Nonrecursive module search path

DanGrayson:


benediktahrens:
Editing files in Emacs works fine.

PR #162, opened 2015-09-28
removed duplicate file UnicodeNotations (present in CatThe and SubstSys)

benediktahrens:


DanGrayson:
By the way, I've started converting Foundations to unicode notations.

benediktahrens:
On 09/27/2015 10:12 PM, Daniel R. Grayson wrote:
> By the way, I've started converting Foundations to unicode
> notations.

Great! From what I have seen, you are for now replacing "forall",
"total2", and "weq" by their Unicode symbols. These changes seem pretty
harmless, but make the code much more beautiful already.

HTML output for coqdoc works fine with Unicode. We should maybe check
that coqdoc --latex gives nice output (and a compilable document), too,
and if not, fix it. There are some Unicode symbols that don't work out
of the box, and I fixed this in the SubstitutionSystems package by
adding some coqdoc options for latex, see [1].

I'll put it on the issues list, and hope to work on it on "UniMath day".

[1] https://github.com/benediktahrens/substitution/blob/master/Makefile#L48

DanGrayson:
The tests succeeded, merging.

PR #164, opened 2015-09-28
Real numbers using Dedekind cuts

cathlelay:
Contents:
*  hnnq.v
  * Definition of non-negative rational numbers
  * Equality on non-negative rational numbers
*  hnnr.v
  * Definition of Dedekind cuts
  * Strict and large order on Dedekind cuts
  * Equality on Dedekind cuts

Work in progress: Apartness relation on Dedekind cuts


vladimirias:
LGTM except one thing - could you please avoid the use of modules? We prefer not to use them because of the poor documentation of what they actually do in Coq.

cathlelay:
The module is removed and I rename hnnr into hnnr_def.
hnnr will be the name of the final structure for non negative real numbers.

UniMath/Dedekind/README.md
```diff
@@ -0,0 +1,19 @@
+Dedekind
+===============
+
+Authors: Catherine Lelay
+
+This Coq library provide a formalisation of real numbers built using
```
DanGrayson: "provides"

UniMath/Dedekind/hnnq.v
```diff
@@ -0,0 +1,146 @@
...
+
+Unset Kernel Term Sharing.
+
+Require Export UniMath.Foundations.RationalNumbers.
+
+Opaque hq.
+
+Open Scope hq_scope.
+
+(** ** Non-negative ratrional numbers *)
```
DanGrayson: "rational"

DanGrayson:
I tested it at home and everything works, so modulo those spelling corrections, it looks fine.

cathlelay:
Spelling corrections done.

PR #165, opened 2015-09-28
	augment coding style description

DanGrayson:


DanGrayson:
This involves only a readme file, so merging ...

PR #168, opened 2015-09-28
To unicode

DanGrayson:
Move some unicode notations from CategoryTheory to Foundations and get a start
converting Preamble and PartA to unicode.

I simplified a couple of proofs and fixed an incorrectly named lemma, too.


DanGrayson:
That was a timeout at 50 minutes.

benediktahrens:
LGTM

Issue #167, opened 2015-09-28
make less coq stuff


benediktahrens:
can we replace "make coqbinaries" by something less, e.g., by 
make coqbinaries && make tools ?

DanGrayson:
Hmm?  How is that less?  It looks like more.

benediktahrens:
On 09/28/2015 03:12 PM, Daniel R. Grayson wrote:
> Hmm?  How is that less?  It looks like more.

 From Makefile.build:

coqlight: theories-light tools coqbinaries

DanGrayson:
I suppose you're talking about changing these lines in our Makefile:
```
sub/coq/bin/coq_makefile sub/coq/bin/coqc: sub/coq/config/coq_config.ml
	$(MAKE) -C sub/coq KEEP_ML4_PREPROCESSED=true VERBOSE=true READABLE_ML4=yes coqlight
sub/coq/bin/coqide: sub/lablgtk/README sub/coq/config/coq_config.ml
	$(MAKE) -C sub/coq KEEP_ML4_PREPROCESSED=true VERBOSE=true READABLE_ML4=yes coqide-binaries bin/coqide
```

Are you suggesting replacing ```coqlight``` by ```coqbinaries tools```?  If so, did you test it?

I remember looking for something more minimal than coqlight and not finding a good target.

benediktahrens:
On 09/28/2015 04:08 PM, Daniel R. Grayson wrote:
> I suppose you're talking about changing these lines in our Makefile:
> ```
> sub/coq/bin/coq_makefile sub/coq/bin/coqc: sub/coq/config/coq_config.ml
> 	$(MAKE) -C sub/coq KEEP_ML4_PREPROCESSED=true VERBOSE=true READABLE_ML4=yes coqlight
> sub/coq/bin/coqide: sub/lablgtk/README sub/coq/config/coq_config.ml
> 	$(MAKE) -C sub/coq KEEP_ML4_PREPROCESSED=true VERBOSE=true READABLE_ML4=yes coqide-binaries bin/coqide
> ```
>
> Are you suggesting replacing ```coqlight``` by ```coqbinaries tools```?  If so, did you test it?

No testing done yet, I was merely recording an idea, maybe to be 
implemented during a UniMath day.

benediktahrens:
Actually, the three
* coqbinaries 
* tools 
* theories/Init/Prelude.vo (to provide the prelude)
are necessary, which is a bit less than coqlight.

[...]

PR #170, opened 2015-09-28
compile less of Coq

benediktahrens:
see #167


DanGrayson:
That saves a lot of time (at home).  I bet travis will finish.

DanGrayson:
Looks good, merging!  (40 minute build time, this probably made the difference.)

benediktahrens:
The more abstract interface to making the Prelude seems to be the target "states". Should I correct this?

DanGrayson:
Yes, that would be good.

PR #178, opened 2015-10-01
Conform

DanGrayson:
start making Ktheory files conform to the standard

also a few housekeeping issues and a few unicode notations and conversions, UniMath wide


UniMath/Ktheory/Utilities.v
```diff
@@ -423,6 +412,10 @@ Proof. trivial. Qed.
 Definition isaprop_hProp (X:hProp) : isaprop X.
 Proof. intro. exact (pr2 X). Qed.
 
+Definition pr1hSet_injectivity (X Y:hSet) : weq (X = Y) (pr1hSet X = pr1hSet Y).
```
benediktahrens: That's an instance of total2_paths_isaprop_equiv, which can maybe simplify the proof or make the lemma superfluous.

DanGrayson: Thanks, that shortens the proof, but I want to keep the lemma.  I'll do that on another branch.

benediktahrens:
lgtm

PR #180, opened 2015-10-01
NonnegativeReals is a complete ordered set

cathlelay:
* definition of large and strict orders of non-negative real numbers
* definition of the LeastUpperBound for Dedekind cuts
* rename files and export versions of definitions and theorems


DanGrayson:
LGTM

cathlelay:
I proved that the equality relation eq_Dcuts imply the equality. The main consequence is that Dedekind cuts on non-negative rational numbers are non-negative real numbers.

1 lemma to prove : forall x y : hq, x < y -> total2 (fun t : hq => dirprod (x < t) (t < y)).

UniMath/Dedekind/Complements.v
```diff
@@ -0,0 +1,74 @@
...
+Definition isPartialOrder {X : UU} := ispo (X := X).
+Definition PartialOrder := po.
+Definition pairPartialOrder {X : UU} (R : hrel X) (is : isPartialOrder R) : PartialOrder X :=
+  popair R is.
+Definition pr1PartialOrder {X : UU} (R : PartialOrder X) : hrel X := carrierofpo X R.
+Coercion  pr1PartialOrder : PartialOrder >-> hrel.
+
+(** *** Strict Partial Order *)
+
+Definition isStrictPartialOrder {X : UU} (R : hrel X) := dirprod ( istrans R ) ( isirrefl R ).
```
DanGrayson: This condition is usually called "ordered", and it use to be called "totally ordered".  The word "partial"
in this context means "not total".

I've gotten a start on orderings on another branch, see https://github.com/DanGrayson/UniMath/blob/finite-sums/UniMath/Ktheory/OrderedSet.v .  Such basic things should be done just once, in the Foundations, and I'm happy for you to be the one to do them.

UniMath/Dedekind/Complements.v
```diff
@@ -0,0 +1,74 @@
...
+Proof.
+  intros (Htrans,Hirrefl).
+  split.
+  now apply istransquotrel.
+  now apply isirreflquotrel.
+Defined.
+
+(** *** Complete Ordered Space *)
+
+Definition isUpperBound {X : UU} (le : PartialOrder X) (E : X -> hProp) (ub : X) :=
```
DanGrayson: X->Prop has a name, see "hsubtypes".

UniMath/Dedekind/DedekindCuts.v
```diff
@@ -0,0 +1,645 @@
...
+  destruct X as [X (Hbot,(Hopen,Hbound))] ; simpl.
+  intros r Xr.
+  revert Hopen ; apply (hinhuniv (P := ishinh _)) ; intros Hopen.
+  now apply Hopen.
+Qed.
+Lemma is_Dcuts_bounded (X : Dcuts) :
+  hexists (fun ub : NonnegativeRationals => neg (X ub)).
+Proof.
+  destruct X as [X (Hbot,(Hopen,Hbound))] ; simpl.
+  now apply Hbound.
```
DanGrayson: What is "now"?  I can't find that in the manual.

UniMath/Dedekind/Complements.v
```diff
@@ -0,0 +1,74 @@
...
+
+(** ** for RationalNumbers.v *)
+
+Require Export UniMath.Foundations.RationalNumbers.
+
+Open Scope hq_scope.
+
+Lemma hq0lehandplus:
+  forall n m : hq,
+    hqleh 0 n -> hqleh 0 m -> hqleh 0 (n + m).
```
DanGrayson: Perhaps introduce notations <, <=, for rationals.

UniMath/Dedekind/Complements.v
cathlelay: About the order, your formalization is fine, but I have a problem with the idea of a predefined strict order.
Indeed, the definition "x < y := x <= y \/ x != y" for a strict order may not be useful when we don't have a decidable equality.

Another solution for ordered sets may be :
- an order (e.g. le)
- a strict order (e.g. gt)
- a proof of incompatibility : "forall x y, le x y -> gt x y -> empty"

UniMath/Dedekind/Complements.v
cathlelay: I use "hsubtypes" to define Dedekind cuts, but I fear of a lack of legibility if I use it in the definition of upper bounds.
Perhaps should we add "Definition subset := hsubtypes".

UniMath/Dedekind/DedekindCuts.v
cathlelay: The tactic "now" is use when all remaining goals can be automatically proved by Coq.
"now apply Hopen" can be understand as a strong version of "apply Hopen ; easy" because there is no remaining goals and then is more robust when the proof change.

UniMath/Dedekind/Complements.v
cathlelay: Good idea.
I will add these notations in "Complements.v"

UniMath/Dedekind/Complements.v
DanGrayson: Yes, as you say, it may not be useful in some contexts, but it's the only definition of x<y that you can make on an ordered set.

Perhaps invent a new name for the notion you propose.  Biordered set?  Doubly ordered set?

UniMath/Dedekind/Complements.v
DanGrayson: That would be good.  Or better, define the type of ```subset``` so it applies only to sets, not to all types.

UniMath/Dedekind/DedekindCuts.v
DanGrayson: How do I find it in the manual?

mortberg: It is defined in:

https://coq.inria.fr/library/Coq.Init.Tactics.html

I think it is good practice to always close a subgoal using "now" or "exact" (it is also faster to type "now apply H" compared to "apply H; assumption" :) ).

mortberg: Let me elaborate a bit: I think subgoals should be closed by tactics that will fail if it does not solve the subgoal. This is the behavior of for example "assumption" or "contradiction". The reason for this is that if one changes something earlier in the file one can faster find the exact place where the script fails just by replaying it, hence giving more robust proofs scripts as Catherine says. By looking at the definition of "easy" I'm not sure if this will always be the behavior of "now", however it should still be better than not using it. I would also like to point out that the SSReflect tactic language has a tactical called "by" which has exactly the desired behavior.

cathlelay: I am wrong saying that "now tactic" is stronger than "tactic ; easy"... It is the definition of now.

DanGrayson: Oh.  Maybe they don't document "now" and "easy" in the manual, because they don't want us to use it, since "easy" is experimental?

DanGrayson: It's a good idea to make proofs fail if they don't complete the subgoal when expected, but wouldn't it be simpler to advise delimiting the proof of the goal with '{' ... '}'?  I always do that.

mortberg: Consistently using {...} and bullets is an alternative. However I don't like it very much because it makes the proofs look like imperative code (like Java) which I don't like. Also, the "now" tactical is is colored red in emacs which I think is quite nice as you easily see which tactic/sequence of tactics closes the goal (however if one uses good indentation and bullets this is usually very clear so this is not a very strong argument). 

I don't think there is an issue with using easy because it is experimental. If the Coq developers decides to remove it we can just copy it to a file Tactics and import this file everywhere where we use it. However I think the best solution would be to start using the SSReflect tactic language which is a better and simpler tactic language specifically designed for doing mathematical proofs. :)

DanGrayson: I don't think there is an issue, either.  I just wonder whether we should submit a bug report about the documentation, or lack thereof.

What would be the first step toward using ssreflect?  Add the code to unimath?  If so, get it from where?

mortberg: I have access to their private repository containing the development version that we need to compile it on the version of Coq that UniMath relies on. I can contact them and ask if we can copy the files we need to UniMath/Tactics. However, this will also rely on what kind of license they have... Anyway I can look into it and if it works out give you a tutorial.

DanGrayson: Oh, right, you told me that before, and we gave up on the idea.  But I'd be happy to have the idea revived.

DanGrayson: But wait a minute -- when coq 8.5 comes out, we want unimath users not to have to compile coq.  If installation of ssreflect involves compiling coq, that might be a problem.

benediktahrens: On 10/04/2015 12:55 PM, Anders Mörtberg wrote:

> However I don't like it very much because it makes the proofs look 
> like imperative code (like Java) which I don't like.

The proof code we write is inherently imperative, not using bullets does
not make it anything else. It just makes it *worse imperative code*.

Please use bullets and {}, and also "now" if you desire.

> However I think the best solution would be to start using the 
> SSReflect tactic language which is a better and simpler tactic 
> language specifically designed for doing mathematical proofs. :)

Would it be less deceiving to say that the tactics of SSReflect were
designed for doing a particular kind of mathematical proofs?

I would be happy if we could keep the external dependencies small, at
least when it comes to compilation of code. (For documentation and
editing I don't mind more dependencies.)

mortberg: Sorry, I was not clear. I was only referring to the use of {..} which I think looks ugly and not bullets which I think are good. Is there any gain in using both {..} and bullets compared to using only bullets?

I would say that the SSReflect tactics were designed for doing any kind of mathematical proofs more pleasantly in Coq. For example they have tactics like "suff" (it suffices to prove that) and "wlog" (without loss of generality). I also think that learning the ssr tactics is easier for newcomers simply because there are fewer, but more powerful, tactics to learn. For example the case and destruct tactics have been merged into only one tactic for doing case analysis. Also the much more powerful rewrite tactic could be very useful for doing univalent mathematics because we are transforming equivalences into equalities that we want to rewrite with. Another thing to point out is that if a file import the ssreflect tactics one can still use all of the standard Coq tactics, so nothing would break if we include the possibility of using the ssr tactics as well. 

I agree with the problem with having more dependencies though. I think that it would not be worth bothering about using ssr unless we can distribute the few files we need (the Ocaml plugin and one or two .v files) together with UniMath. The better alternative to this would be to have ssr as a git submodule if they ever put their code on github (which I think they have been discussing...).

DanGrayson: I use {..} without bullets.

Can you build ssr without also building Coq from source?

benediktahrens: On 10/04/2015 07:59 PM, Anders Mörtberg wrote:
> 
> Sorry, I was not clear. I was only referring to the use of {..} which
> I think looks ugly and not bullets which I think are good. Is there
> any gain in using both {..} and bullets compared to using only
> bullets?

I see two uses:

1. When running out of levels of bulletting, you can use already used
levels inside {} again.

2. When using "assert", enclosing that subproof in {} seems more
reasonable. That is a matter of taste, though.

mortberg: The SSReflect tactic language is only an Ocaml plugin (which is necessary for doing more advanced things that you cannot do with only Ltac) and a little bit of notations and Ltac. So to compile it you need ocaml installed and coqc so that you can compile the ssreflect.v file. So I don't think there is any need to build coq from sources to have it. 

Regarding {..} I'll stick to Benedikt's two uses for using them.

DanGrayson: If ocaml is required to compile ssr, then that puts an additional burden on
future
unimath users, who might otherwise have to install just coq.

mortberg: I believe that this burden will be inevitable in the future, especially if all of Coq's libraries will be distributed with opam, which I hope is what will happen. If this ever happens I hope someone will package ocaml, coq and opam together so that they can be easily installed on any system. Anyway before bothering about all of this lets just see what the ssr people answers to my email about whether it is possible at all to do something with ssr+unimath now.

DanGrayson: Okay.

UniMath/Dedekind/Complements.v
DanGrayson: Or: an "effectively ordered" set?

UniMath/Dedekind/Complements.v
cathlelay: Done.

UniMath/Dedekind/Complements.v
cathlelay: I also add notations for multiplicative inverse and division.

cathlelay:
Last proofs has been done.
NonnegativeReals has now the type hSet.

DanGrayson:
Looks good to me.

vladimirias:
So now we can do things like Banach spaces? And Hilbert space?

cathlelay:
As soon as I have an addition and a 0, we can define a distance, and then a
metric spaces.
For Banach spaces, I also need a multiplication.

I think it can be done this week.

Catherine.

2015-10-06 19:43 GMT-04:00 Vladimir Voevodsky <notifications@github.com>:

> So now we can do things like Banach spaces? And Hilbert space?
>

DanGrayson:
Catherine, you may ignore my old comments about "totally ordered" sets, for I misread [isirrefl] as [isantisymm], sorry.

cathlelay:
I have a problem to choose the good definition of "totally ordered".
As addition and multiplication seem interested for other applications, I
will define them first.

Catherine

2015-10-06 20:53 GMT-04:00 Daniel R. Grayson <notifications@github.com>:

cathlelay:
I am doing a better version with effectively ordered sets.

2015-10-09 13:15 GMT-04:00 Daniel R. Grayson <notifications@github.com>:

DanGrayson:
Oh, okay, closing ...

DanGrayson:
Please don't submit a pull request that contains a change to the release number of sub/coq.  Do you know how to use "git submodule update"?

PR #182, opened 2015-10-04
reorganization and unicode conversion

DanGrayson:


UniMath/Ktheory/Utilities.v
```diff
@@ -4,42 +4,39 @@ Global Unset Automatic Introduction.
...
+Notation "f ;; g" := (funcomp f g) (at level 50).
+(* funcomp' is like funcomp, but with the arguments in the other order *)
+Definition funcomp' { X Y Z : UU } ( g : Y -> Z ) ( f : X -> Y ) := fun x : X => g ( f x ) . 
+Notation "p # x" := (transportf _ p x) (right associativity, at level 65) : transport_scope.
+Notation "p #' x" := (transportb _ p x) (right associativity, at level 65) : transport_scope.
+Notation "p # x" := (transportf _ p x) (right associativity, at level 65) : transport_scope.
+Notation "p #' x" := (transportb _ p x) (right associativity, at level 65) : transport_scope.
+Open Scope transport_scope.
+
+Notation totalSpace := total2.
```
benediktahrens: That is the third name we introduce, after total2 and unicode \Sigma. Is it helpful to have 3 different names for the same concept?

UniMath/Foundations/Propositions.v
```diff
@@ -46,6 +46,12 @@ Definition hProppair ( X : UU ) ( is : isaprop X ) : hProp := tpair (fun X : UU
 Definition hProptoType := @pr1 _ _ : hProp -> UU .
 Coercion hProptoType: hProp >-> UU.
 
+Definition propproperty (P:hProp) := pr2 P : isaprop (pr1 P).
+
+(* this lemma can be used with "apply" when the target is to show that an hProp coerced to UU yields a propostion *)
+Definition isaprop_hProp (X:hProp) : isaprop X.
```
benediktahrens: The name of this lemma is misleading: one thinks that it shows that hProp is a prop, which of course it is not.
I can't think of a better name right now, though.

UniMath/Ktheory/Utilities.v
DanGrayson: I think of it as a sensible replacement for total2, not a third name.

DanGrayson: I'd be happy to make it local.

UniMath/Foundations/Propositions.v
DanGrayson: Good point.  How about isaprop_an_hProp?

UniMath/Ktheory/Utilities.v
benediktahrens: On 10/04/2015 12:36 PM, Daniel R. Grayson wrote:
>> +Notation totalSpace := total2.
> 
> I'd be happy to make it local.

No, I don't mind.

UniMath/Foundations/Propositions.v
benediktahrens: That's good!

benediktahrens:
LGTM

UniMath/Foundations/Propositions.v
DanGrayson: Actually, isaprop_hProp and propproperty are identical.  By analogy with setproperty, perhaps we prefer propproperty.  So I'l just remove it.

DanGrayson:
I'll merge this after the test runs ...

PR #184, opened 2015-10-09
Finite sums

DanGrayson:
This includes the general associativity theorem, whose proof I am still working to improve,
and it also includes moving some tactics and the notion of equivalence upstream, where
equivalence is defined as follows:

```
Definition Equivalence X Y :=
  Σ (f:X->Y) (g:Y->X) (p:∀ y, f(g y) = y) (q:∀ x, g(f x) = x),
      ∀ x, ap f (q x) = p(f x).
```

Here "ap" is notation for ```maponpaths```.


UniMath/Foundations/Basics/PartA.v
```diff
@@ -246,11 +251,27 @@ Proof.
 Defined. 
 
 Lemma maponpathscomp {X Y Z : UU} {x x' : X} (f : X -> Y) (g : Y -> Z)
```
benediktahrens: There is also maponpathscomp0 in Basics.partA, which is about maponpaths on a composition of paths.
The naming is a bit confusing now, with this lemma having the same name except for the "0" at the end.

UniMath/Foundations/Basics/PartA.v
```diff
@@ -364,6 +385,14 @@ Proof.
   intros. induction e. apply idpath.
 Defined.
 
+Definition transportf_maponpaths_pr1 {X} {P:X->UU} {w w':total2 P} (p : w = w') :
```
benediktahrens: This is called fiber_paths in Basics.PartA.

vladimirias:
Dan, why do you need another definition of an equivalence? There is one standard definition and there is a theorem saying that isweq is a proposition that implies that the projection weq X Y -> ( X -> Y ) is of h-level 1, i.e., that two equivalences are equal if and only if their first components (function f:X->Y) are equal.

Why or where may you need another definition of an equivalence?

Vladimir.

UniMath/Foundations/Basics/PartA.v
DanGrayson: Good point, but it's Vladimir's name for it, from last December, so perhaps it is best dealt with separately.

UniMath/Foundations/Basics/PartA.v
DanGrayson: Thanks!  Changing ...

DanGrayson:
I use it as a convenient way to organize proofs when making objects of
type "weq" -- first I make one of these things, which is easily converted
to a "weq", and it seems
to me I get more control over the homotopies that are recorded in a "weq".
See, for example,
my lemma at
https://github.com/UniMath/UniMath/blob/master/UniMath/Ktheory/Equivalences.v#L379
,
which contains the lines:
```
       apply Equiv_to_weq.
       refine (makeEquiv _ _ _ _ _ _ _).
```
Such control was tricky to achieve in my situation (working with types that
may not be sets)
and crucial for the proof to go through, or at least, I found it crucial.

I also think it will be useful in my current project to prove that the type
Sequence X
( ```Definition Sequence X := Σ n, stn n -> X.``` ) has the same induction
principle that
the standard type ```List``` has.  Proofs by induction over the
"constructors" of sequences
ought to be simpler than proofs by induction over the length of a
sequence.  I think
that proving ```Equivalence (Sequence X) (coprod unit (X × Sequence X))```
will be the
essential point.  Perhaps it's a conceptual crutch, since the proof of the
adjointness
property (the last part of an object of type Equivalence) for a weq is just
one line long,
and could be formulated as a lemma, but packaging this way seems to be a
useful
crutch, at least for me, and perhaps for others.

Another point that ought to prove useful in future proofs is that, with
this definition, the
inverse of an inverse of an Equivalence agrees, on the nose, for 4 of the 5
components.

A final point is that if I make an Equivalence, then use it and its inverse
to produce two
weq's, inverse to each other, I have more control of the homotopies between
the two
composites with the identities than I would by making one weq and then
applying invweq
to it.  It seems to me this was also crucial for me, if memory serves me
well.
I could also package that as a
lemma without introducing the concept: accept certain inputs and provide a
pair of inverse
weq's, but such an arrangement would seem artificial.



On Fri, Oct 9, 2015 at 12:51 PM, Vladimir Voevodsky <
notifications@github.com> wrote:
>
> Dan, why do you need another definition of an equivalence? There is one
standard definition and there is a theorem saying that isweq is a
proposition that implies that the projection weq X Y -> ( X -> Y ) is of
h-level 1, i.e., that two equivalences are equal if and only if their first
components (function f:X->Y) are equal.
>
> Why or where may you need another definition of an equivalence?
>
> Vladimir.

DanGrayson:
For a concrete example, see the lemma at https://github.com/DanGrayson/UniMath/blob/finite-sums/UniMath/Ktheory/Equivalences.v#L170 .  It's a certain identity involving your homotinvweqweq that I couldn't prove.  I worked around it by introducing a different definition of homotinvweqweq at https://github.com/DanGrayson/UniMath/blob/finite-sums/UniMath/Ktheory/Equivalences.v#L153 , which made the proof of the desired identity trivial at https://github.com/DanGrayson/UniMath/blob/finite-sums/UniMath/Ktheory/Equivalences.v#L158 .

DanGrayson:
Vladimir, now it seems premature to me to put the new concept in Foundations, so I've moved it back to Ktheory.

vladimirias:
Ok.

PR #185, opened 2015-10-09
Non-negative real numbers are a Complete Set

cathlelay:


DanGrayson:
Github doesn't recognize you as the same committer in all cases.  I bet on one of your
machines you haven't configured git yet with your name and email address, or perhaps with
a different email address.

DanGrayson:
It seems that a strict partial order is not a partial order.  Is that standard terminology?  It could lead to confusion.

sub/coq
```diff
@@ -1 +1 @@
-Subproject commit b3bd2696c31ad2cb544f3436ddb5a237fe7fa6fe
+Subproject commit 76d6a8cfb0448ade98e1a7abc92ff1bf5075fe8f
```
benediktahrens: Do not change the Coq commit to an older one.

cathlelay: I will try to correct.

2015-10-09 16:19 GMT-04:00 Benedikt Ahrens <notifications@github.com>:

cathlelay:
We don't have reflexivity in a strict partial order

2015-10-09 16:11 GMT-04:00 Daniel R. Grayson <notifications@github.com>:

DanGrayson:
Yes, not having reflexivity is exactly why a strict partial order is not a partial order.  But that makes the term "strict partial order" confusing, because a general principal of language is that a strict gadget is a gadget.

vladimirias:
the proof 2=1+1 should work with “apply idpath”.

vladimirias:
Or does it?

DanGrayson:
It certainly does.

vladimirias:
We are about 2=1+1 in hq.

cathlelay:
It is in hq.

2015-10-09 17:13 GMT-04:00 Vladimir Voevodsky <notifications@github.com>:

DanGrayson:
Oh, never mind!

On Fri, Oct 9, 2015 at 5:13 PM, Vladimir Voevodsky <notifications@github.com

vladimirias:
In hz it works directly. It does not work in hq because an integer n is mapped to hq as (n, (1, p)) where p is the proof that 1\ne 0 in hz. Because of this n1 + n2 has (1, (p and p)) in denominator instead of (1, p).

The only way to resolve this issue completely is to have a “strict prop” (a Prop where proof irrelevance is up to substitutional equality and where one can place some of the objects of h-Prop). Matthieu is coming to Princeton at the end of September and as far as I know building such a strict prop into Coq is one of the goals of the work that they will do on their new ERC grant.

I think we should keep this example in mind and talk to him about it when he arrives.

Matthieu and

UniMath/Dedekind/decDcuts.v
```diff
@@ -0,0 +1,84 @@
...
+(** without decidability *)
+
+Require Import UniMath.Dedekind.Complements.
+Require Import UniMath.Dedekind.NonnegativeRationals.
+Require Export UniMath.Dedekind.DedekindCuts.
+
+(** ** Definition *)
+
+Definition isdecDcuts : hsubtypes Dcuts :=
+  (fun x : Dcuts => ishinh (forall r, coprod (r ∈ x) (neg (r ∈ x)))).
```
benediktahrens: This is a question rather than a remark: what happens if you put the truncation (ishinh) after the quantification over "r", that is, when you define instead
Definition isdecDcuts : hsubtypes Dcuts :=
+  (fun x : Dcuts => forall r, ishinh (coprod (r ∈ x) (neg (r ∈ x))).
?

UniMath/Dedekind/DedekindCuts.v
```diff
@@ -0,0 +1,605 @@
...
+  intros r Xr n Xn.
+  revert Xr ; apply hinhfun ; intros (X,(Ex,Xr)).
+  exists X ; split.
+  exact Ex.
+  now apply is_Dcuts_bot with r.
+Qed.
+Lemma Dcuts_lub_open :
+  forall (x : NonnegativeRationals),
+    Dcuts_lub_val x ->
+    hexists (fun y : NonnegativeRationals => dirprod (Dcuts_lub_val y) (x < y)%NnR).
```
benediktahrens: There really should be a notation for hexists in Foundations. That should be in a different PR, though.

DanGrayson:
Adding such a "strict prop" feature sounds unsafe.  It might be easier to
just add a lemma for doing equations between explicit rationals.

UniMath/Dedekind/DedekindCuts.v
DanGrayson: How about this one:
```
Notation "'∃'  x .. y , P" := (hexists (Σ x , .. (Σ y , P) .. )) (at level 200, x binder, y binder, right associativity) : type_scope.
```

benediktahrens: On 10/10/2015 05:34 PM, Daniel R. Grayson wrote:
>> +Definition Dcuts_lub_val : NonnegativeRationals -> hProp := +  fun
>> r : NonnegativeRationals => hexists (fun X : Dcuts => dirprod (E X)
>> (r ∈ X)). +Lemma Dcuts_lub_bot : +  forall (x :
>> NonnegativeRationals), +    Dcuts_lub_val x -> forall y :
>> NonnegativeRationals, (y <= x)%NnR -> Dcuts_lub_val y. +Proof. +
>> intros r Xr n Xn. +  revert Xr ; apply hinhfun ; intros
>> (X,(Ex,Xr)). +  exists X ; split. +  exact Ex. +  now apply
>> is_Dcuts_bot with r. +Qed. +Lemma Dcuts_lub_open : +  forall (x :
>> NonnegativeRationals), +    Dcuts_lub_val x -> +    hexists (fun y
>> : NonnegativeRationals => dirprod (Dcuts_lub_val y) (x < y)%NnR).
> 
> How about this one: ``` Notation "'∃'  x .. y , P" := (hexists (Σ x ,
> .. (Σ y , P) .. )) (at level 200, x binder, y binder, right
> associativity) : type_scope.

Looks good!

benediktahrens:
lgtm. 
in order to understand the overall ongoings, i would like to have a blackboard session at some point, though.

vladimirias:
Adding such a strict prop feature is one of the most desirable innovations that everybody wants to have. It is unsafe if done without clear understanding of what is and what is not allowed with such a strict prop. The univalent model gives us a way to specify a set of constructions that should be safe.

UniMath/Dedekind/DedekindCuts.v
DanGrayson: Oh, it has to be ishinh in the notation.  I'll submit a PR.  Also one for neg.

DanGrayson:
Probably hz does not qualify as a strict prop in the univalent model, since
ishinh (dirprod nat nat) doesn't either: it seems to me that it has
gigantic contractible components.

DanGrayson:
Catherine, perhaps "strict weak ordering" is the standard term, not "strict partial ordering", see https://www.sgi.com/tech/stl/StrictWeakOrdering.html .  The latter term might be confusing, because a reader would naturally assume that a strict partial ordering is a partial ordering.

UniMath/Dedekind/decDcuts.v
cathlelay: "forall r, ishinh (coprod (r ∈ x) (neg (r ∈ x))" has type Type and not hProp.
Then, isdecDcuts cannot be a hsubtypes.

cathlelay:
I will rename "strict partial ordering" into  "strict weak ordering".

But I have a problem to name if "weak" because, there is more informations in "less than" than in "less or equal" : for Dedekind cuts, we have (in hProp) a rational number which is in a cut and not in the other for "less than". It seams me stronger than the implication of "less or equal".

UniMath/Dedekind/decDcuts.v
DanGrayson: No, see this lemma:
```

Corollary impred_isaprop { T : UU } (P:T -> UU): (forall t:T, isaprop (P t)) -> (isaprop (forall t:T, P t)).
Proof. apply impred. Defined.
```

DanGrayson:
Well, then, instead of "strict partial ordering" or "strict weak ordering", what about "strong ordering"?

UniMath/Dedekind/decDcuts.v
benediktahrens: On 10/11/2015 09:12 AM, Daniel R. Grayson wrote:
>> @@ -0,0 +1,84 @@
>> +(** * A library about decidable Dedekind Cuts *)
>> +(** Author: Catherine LELAY. Oct 2015 - *)
>> +(** Additional results about Dedekind cuts which cannot be proved *)
>> +(** without decidability *)
>> +
>> +Require Import UniMath.Dedekind.Complements.
>> +Require Import UniMath.Dedekind.NonnegativeRationals.
>> +Require Export UniMath.Dedekind.DedekindCuts.
>> +
>> +(** ** Definition *)
>> +
>> +Definition isdecDcuts : hsubtypes Dcuts :=
>> +  (fun x : Dcuts => ishinh (forall r, coprod (r ∈ x) (neg (r ∈ x)))).
> 
> No, see this lemma:
> ```
> 
> Corollary impred_isaprop { T : UU } (P:T -> UU): (forall t:T, isaprop (P t)) -> (isaprop (forall t:T, P t)).
> Proof. apply impred. Defined.
> ```

Exactly, that's why I was asking.

vladimirias:
I am not sure what you mean. hz should qualify as a strict set. How to make sure that it works I do not yet know. If hz is a strict set then [ paths hz x y ] are strict props.

More importantly for the applications we have been discussing emptyset is a strict prop and therefore the negation of any type is a strict prop. So the type of the statement 1\ne 0 is a strict prop in any ring and this should be sufficient for having the function from hz to hq to commute with the addition definitionally (on particular elements such 1,2,3 etc. ).

DanGrayson:
Oops, I said it all wrong.  Let me try again.

Let's consider whether hz is a strict set.  It's a quotient of [dirprod nat
nat].  One of the
equivalence classes is the set E of pairs (m,m) -- it corresponds to 0 in
hz.  If there's
an argument that hz is a strict set, probably the same argument shows that
the quotient of
E by the relation that makes everything equivalent is a strict set.  And
probably that quotient
bears a lot of similarity to [ishinh E] and to [ishinh nat], whose
representations in the
model are big, since [ishinh] involves a product over all small
propositions.

One of the things I show in my formalization of the circle is that there is
a map from
[ishinh nat] to the circle that sends every m to the base point by
definition but wraps the paths
from m to m+1 around the circle (see the tail end of Halfline.v).  That map
cannot
factor through a one-point space in the model.

vladimirias:
In the model there is a retraction from hProp to sProp. It is not clear to me how such a retraction will affect the constructive character of the theory but it is consistent. There is also the point that one should be able to make strict set quotients probably in exactly the same way as the usual ones only using sProp instead of hProp.

Some objects naturally land to sProp such as [ paths nat n m ] or [ paths p a b ] where [ p : sProp ] . Also [ paths bool a b ] is in sProp and therefore the comparisons in nat and therefore in strict hz and strict hq are in sProp.

The main example of something that is not in sProp is a coconus (or whatever we call it now). It is contractible and therefore in hProp but, unless the ambient time is a strict set, it is not in sProp. For example, [ @coconusfromt UU nat ] is in hProp but not in sProp.

DanGrayson:
Perhaps you want to enforce, in the Coq language, some difference between
[strict hz] and [hz], so we don't run into contradictions, like the one I
fear?

vladimirias:
Surely there would have to be such a difference because the type of strict subsets (functions to sProp) and “homotopy subsets” (functions to hProp) are different - one is a strict set and another one is not.

benediktahrens:
Can we pull this now?

PR #188, opened 2015-10-12
Colimits

mortberg:
* Definition of colimits
* Colimits in functor categories from colimits in target category
* Colimits in HSET


DanGrayson:
Looks good to me.

UniMath/CategoryTheory/category_hset.v
```diff
@@ -204,22 +258,148 @@ Lemma is_category_HSET : is_category HSET.
...
+Variable D : diagram g HSET.
+
+Definition cobase : UU := Σ j : vertex g, pr1hSet (dob D j).
 
+(* Theory about hprop is in UniMath.Foundations.Propositions *)
+Definition rel0 : hrel cobase := λ (ia jb : cobase),
+  hProppair (ishinh (Σ f : pr1 ia --> pr1 jb, dmor D f (pr2 ia) = pr2 jb))
+            (isapropishinh _).
 
+Definition rel : hrel cobase := eqrel_from_hrel rel0.
```
benediktahrens: The definition of "rel" should be local, and probably other definitions in this file as well, such as "eqr".
The names are too short (and hence precious) to be global.

benediktahrens: We will fix this in a subsequent PR. Actually, this particular construction should probably go into Foundations anyway, and get a suitable name there.

mortberg: Yes, this should be a good thing for me to do on a UniMath day...

PR #189, opened 2015-10-14
Redefine multiplication of natural numbers

DanGrayson:
... so that m*n is a right associated sum copies of m, and so 1*n reduces to n.

Fix all the proofs that were broken by the new definition.

### DIFF
--- a/UniMath/Foundations/Basics/Preamble.v
+++ b/UniMath/Foundations/Basics/Preamble.v
@@ -144,6 +144,27 @@ Inductive Phant ( T : Type ) := phant : Phant T .
 Check (O = O) .
 
 
+(* confirm and repair some aspects of multiplication on [nat] *)
+Goal ∀ n, 0*n = 0.     reflexivity. Qed.
+Goal ∀ n, n*1 = n. try reflexivity. Abort.
+Goal ∀ n, 1*n = n. try reflexivity. Abort.
 
+(* Those two failures are not good.  The cause is that Nat.mul defines
+   multiplication as a right-associative sum.  Note also that the parser treats
+   "+" and "*" as left associative.  We redefine multiplication in nat here,
+   avoiding "Fixpoint", too. *)
 
-(* End of the file uuu.v *)
+Definition mul : nat -> nat -> nat.
+Proof.
+  intros n m.
+  induction n as [|p pm].
+  - exact O.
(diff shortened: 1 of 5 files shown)


UniMath/Foundations/Basics/Preamble.v
```diff
@@ -144,6 +144,29 @@ Inductive Phant ( T : Type ) := phant : Phant T .
...
+Goal ∀ n, n*1 = n. try reflexivity. Abort.
+Goal ∀ n, 1*n = n. try reflexivity. Abort.
+
+(* Those two failures are not good.  The cause is that Nat.mul defines
+   multiplication as a right-associative sum.  Note also that the parser treats
+   "+" and "*" as left associative.  We redefine multiplication in nat here,
+   avoiding "Fixpoint", too. *)
+
+Definition mul : nat -> nat -> nat.
+(* we name it "mult" to override Coq's name for Nat.mul *)
```
benediktahrens: You say "we name it "mult"", but it is named "mul".

UniMath/Foundations/StandardFiniteSets.v
```diff
@@ -237,22 +237,53 @@ Defined.
...
+      * change (S i * m) with (i*m + m). 
+        rewrite natpluscomm.
+        exact (natgthandplusl m j ( i * m ) lj ).
+      * exact ( natlehandmultr ( S i ) n m ( natgthtogehsn _ _ li ) ).
+    + set ( f := fun ij : stn n × stn m =>
+                   match ij
+                   with tpair _ i j =>
+                        stnpair ( n * m ) ( j + i * m ) ( i1 i j ( pr2 i ) ( pr2 j ) )
+                   end ).
+      Check f.
```
benediktahrens: Could you remove the "Check" command? Or is there a reason for it to be there?

benediktahrens:
lgtm, modulo the "Check" command.

UniMath/Foundations/Basics/Preamble.v
DanGrayson: No, it's a good suggestion.  I've taken care of it.

UniMath/Foundations/StandardFiniteSets.v
DanGrayson: ok

PR #193, opened 2015-10-16
Non-negative real numbers are an abelian monoid

cathlelay:
This is preliminary version.

Several definitions will change the next week, in partcular:
- Definition of large order on non-negative reals
- Evolution of half-fields

Goal of the next week: generalization of many theorems proved in NonnegativeRationals.v


UniMath/Dedekind/Complements.v
```diff
@@ -27,6 +27,26 @@ Proof.
   exact hzone_neg_hzzero.
 Defined.
 
+Opaque hz.
+Lemma hqinv0 : / 0 = 1.
```
cathlelay: Will probably be removed.

UniMath/Dedekind/Complements.v
```diff
@@ -115,4 +135,26 @@ Proof.
   now rewrite hqplusr0.
 Qed.
 
+Lemma hq0leminus :
+  forall r q : hq, r <= q -> 0 <= q - r.
```
cathlelay: To generalize in fld

UniMath/Dedekind/Complements.v
```diff
@@ -115,4 +135,26 @@ Proof.
...
+Proof.
+  intros r q Hr.
+  apply hqlehandplusrinv with r.
+  unfold hqminus.
+  rewrite hqplusassoc, hqlminus.
+  now rewrite hqplusl0, hqplusr0.
+Qed.
+
+Lemma hqinv_gt0 (x : hq) : 0 < x -> 0 < / x.
```
cathlelay: To generalize in fld

UniMath/Dedekind/HalfField.v
```diff
@@ -0,0 +1,405 @@
...
+Unset Kernel Term Sharing.
+
+Require Import UniMath.Foundations.Algebra.Monoids_and_Groups.
+Require Import UniMath.Dedekind.Sets_comp.
+
+(** ** More About Monoids *)
+
+(** monoids *)
+
+Definition ismonoid {X : hSet} (x0 : X) (op : binop X) : UU :=
```
DanGrayson: Use ismonoidop instead, which already exists.

UniMath/Dedekind/HalfField.v
```diff
@@ -0,0 +1,405 @@
...
+Definition ismonoid_isassoc {X : hSet} {x0 : X} {op : binop X} (is : ismonoid x0 op) : isassoc op :=
+  pr1 is.
+Definition ismonoid_islunit {X : hSet} {x0 : X} {op : binop X} (is : ismonoid x0 op) : islunit op x0 :=
+  pr1 (pr2 is).
+Definition ismonoid_isrunit {X : hSet} {x0 : X} {op : binop X} (is : ismonoid x0 op) : isrunit op x0 :=
+  pr2 (pr2 is).
+
+(** abelian monoids *)
+
+Definition isabmonoid {X : hSet} (x0 : X) (op : binop X) : UU :=
```
DanGrayson: Use isabmonoidop, which already exists.

UniMath/Dedekind/HalfField.v
```diff
@@ -0,0 +1,405 @@
...
+Definition isabmonoid_islunit {X : hSet} {x0 : X} {op : binop X} (is : isabmonoid x0 op) : islunit op x0 :=
+  ismonoid_islunit (isabmonoid_ismonoid is).
+Definition isabmonoid_isrunit {X : hSet} {x0 : X} {op : binop X} (is : isabmonoid x0 op) : isrunit op x0 :=
+  ismonoid_isrunit (isabmonoid_ismonoid is).
+Definition isabmonoid_iscomm {X : hSet} {x0 : X} {op : binop X} (is : isabmonoid x0 op) : iscomm op :=
+  pr2 is.
+
+(** ** More About Groups *)
+
+Definition unop (X : UU) := X -> X.
```
DanGrayson: move this upstream, near binop

UniMath/Dedekind/HalfField.v
```diff
@@ -0,0 +1,405 @@
...
+
+(** "additive" group *)
+
+Definition isgr {X : hSet} (x0 : X) (op : binop X) (inv : unop X) : UU :=
```
DanGrayson: use isgrop instead

Do the same for all other definitions that repeat things already done.

UniMath/Dedekind/HalfField.v
```diff
@@ -0,0 +1,405 @@
...
+           (is : isgr' x1 op exinv inv)
+           (Hx1 : exinv x1) (Hop : ∀ x y : X, exinv x -> exinv y -> exinv (op x y))
+           (Hinv : ∀ (x : X) (Hx : exinv x), exinv (inv (x,, Hx))) : isgr (x1' Hx1) (op' Hop) (inv' Hinv) :=
+  (isassoc_op' is Hop,, islunit_op'_x1' is Hx1 Hop,, isrunit_op'_x1' is Hx1 Hop)
+    ,, (islinv_op'_x1'_inv' is Hx1 Hop Hinv,, isrinv_op'_x1'_inv' is Hx1 Hop Hinv).*)
+
+(** ** Definition of a HalField *)
+(** to be a HalfField *)
+
+Definition isHalfField {X : hSet} (x0 x1 : X) (plus mult : binop X) (Hnz : subsetcond X) (inv : subset Hnz -> X) : UU :=
```
DanGrayson: If you drop commutativity from this definition, you could use the name DivisionRig, which would be easier
to guess the meaning of.

UniMath/Dedekind/Sets_comp.v
```diff
@@ -8,8 +8,16 @@ Require Export UniMath.Foundations.Sets
 
 (** ** Subsets *)
 
-Definition subset (X : hSet) : hSet :=
-  hSetpair _ (isasethsubtypes X).
+Definition subsetcond (X : hSet) : UU := X -> hProp.
```
DanGrayson: Consider removing this definition and replacing it, where used, by hsubtypes, which already exists.

UniMath/Dedekind/Sets_comp.v
```diff
@@ -8,8 +8,16 @@ Require Export UniMath.Foundations.Sets
...
+Lemma isaset_subsetcond {X : hSet} (Hsub : subsetcond X) : isaset (carrier Hsub).
+Proof.
+  intros.
+  apply (isasetsubset pr1 (pr2 X) (isinclpr1 (λ x : X, Hsub x) (λ x : X, pr2 (Hsub x)))).
+Qed.
+Definition subset {X : hSet} (Hsub : subsetcond X) : hSet :=
+  hSetpair (carrier Hsub) (isaset_subsetcond Hsub).
+Definition set_to_subset {X : hSet} {Hsub : subsetcond X} (x : X) (Hx : Hsub x) : subset Hsub :=
```
DanGrayson: The name makes it sound like it converts a set to a subset.  What about "makeSubset" instead?

UniMath/Dedekind/HalfField.v
cathlelay: My problem with ismonoidop is that we do not know which is the neutral element.
This definition seems me more useful for ordered monoids.

UniMath/Dedekind/HalfField.v
vladimirias: DivRig might be more convenient to use.

UniMath/Dedekind/HalfField.v
DanGrayson: What do you mean?  Here is the definition:
```
Definition isunital { X : hSet} ( opp : binop X ) := total2 ( fun un0 : X
=> isunit opp un0 ) .
```
If  [p:isunital op], then [pr1 op] is the neutral element.  Notice that the
neutral element is
unique, so [isunital op] is a proposition.

cathlelay: It seems to me that theorems such as:
isabmonoid x0 op -> forall x y : X, x0 <= x -> x0 <= y -> x0 <= op x y
are easier to read than a version were x0 is hidden.

DanGrayson: Maybe not.  We're trying hard to distinguish added structure from mere
properties, on the basis
that mere properties come just with proofs, and it's the added structure
that one has to be aware
of.  Before formalizing proofs, I believed that the best way to present the
definition of a monoid
was by listing all the ingredients first: set, operation, and unit element,
and then providing all
the proofs of the identities they satisfy.  But now that proofs are first
order mathematical objects, the
list of ingredients is longer: set, operation, unit, left unit proof, right
unit proof, associativity proof.  The
new principled way to partition the set into structure and properties is to
recognize which ingredients
or combinations of ingredients are propositions, and bring the rest of the
structure to the fore.  The unit
element is still there in a monoid, it's just combined with the unit laws
into a logical substructure.

Another general stylistic principle I'd like to propose is that properties,
such as "ismonoidop", should take
just one argument (together with its implicit prerequisites).  Otherwise,
of what is it a property?
I.e., a property of a pair of things should be formulated as a property of
pairs.

We can't have a proliferation of definitions of what it means to be a
monoid operation in the system --
there has to be some coherence and uniformity for the user.  We don't want
to write
conversion functions between Lelay-monoids and Foundations-monoids and
force the user to use
them all the time.

cathlelay: OK.

I remove these definitions.

benediktahrens:
LGTM

DanGrayson:
Travis doesn't seem to be able to show me the build errors, but I assume it just ran out of time.  I've checked the most recent version of the pull request, and it builds.

This looks better, thanks for removing the alternative development of monoid.

I wonder about one thing though.  I understand that in a constructive field, inverses are only defined on the subset of elements apart from 0, not those unequal to 0, so you can't copy the Foundations  definition of field, but is there a good reason for having a notion of "group" with an inverse defined only on a subset of the group, as your ```gr'```?  In other words, wouldn't it suffice to consider the usual notion of group on the set of elements apart from 0, if you need to discuss groups at all?

cathlelay:
The multiplication should be a total function, in particular it should be defined at 0.
This new definition seems to me more useful than using "isgr (fun x y : subset exinv => mult (pr1 x) (pr1 y),,(Hmult (pr2 x) (pr2 y)))" where Hmult is a proof that exinv is stable for the multiplication.

DanGrayson:
When you say the multiplication should be defined at 0, if you are speaking of the multiplication in a ring, then I agree with you, as that's part of the definition of ring.  Probably your entire formalization concerns rings and fields, and probably you have no reason to make the multiplicative group of real numbers with inverses.  Or do you?  

If you want a set with just one operation, the multiplication, and you want it to include the 0 from the field, then you could use the multiplicative monoid of the ring, which has already been implemented as ```rngmultmonoid```.  Perhaps it's true that a real number is invertible if and only if it's apart from zero.  In that case, there is no reason to provide the invertibility predicate as part of the structure, since the predicate is already available as the statement that an element has an inverse.  It's a proposition because the ring is a set.

Even if you do have occasion to make the multiplicative group, why not make it a group in the traditional sense?  Even in ordinary rings, when making the multiplicative group, the way we would do it is by making a group whose underlying set is the set of nonzero elements, so it would be natural for you to proceed the same way, taking the underlying set to be the set of numbers apart from 0.

We should be sparing in our introduction of new algebraic concepts at this low level, since the existing ones have been well tested over decades.

cathlelay:
And, what about this definition inspired by `isafield`:
`Definition isDivRig ( X : commrig ) := (forall x : subset (fun x => x != 0) , multinvpair X x).`

I don't like the fact that `X` is a commrig and not a hSet, but it seems closer to the library style.

DanGrayson:
That definition of isDivRig looks good to me.  So the real numbers would not be one, constructively.

Why do you say X is not an hSet?  Every commrig is an hSet -- it's part of the definition.

DanGrayson:
Oh, you should include isnonzerorng X in your definition of isDivRig, to exclude the case where 1=0, as we do with division rings and fields.

That goes all the way back to why we don't regard 1 as a prime number, for it would spoil the clean statement of unique factorization into prime factors.

DanGrayson:
Oh, and your definition would be a lot simpler this way:
```
Definition isDivRig ( X : commrig ) := dirprod (isnonzerorig X) (forall x, x != 0 -> multinvpair X x)).
```
Or by even closer analogy with isafield, it would say
```
Definition isDivRig ( X : commrig ) := dirprod (isnonzerorig X) (forall x, coprod (multinvpair X x) (x=0))).
```
, which builds in some decidability.  I wonder whether the decidability of equality in a division ring is important for later proofs.

But why bother with these definitions, if the real numbers will not be an example of such a thing, constructively?

cathlelay:
For the second definition, we need to be able to decide if x is equal to 0 or not.
I have this property for non-negative rational numbers, but not for Dedekind cuts.

I have to try the first one.

DanGrayson:
Yes, decidability seems not good in your context.

By the way, it seems the set of numbers not apart from zero is not provably closed under addition.  Is that true?

DanGrayson:
Still, for real numbers, x!=0 is not the criterion for invertibility, constructively.  So why consider the first definition at all?

cathlelay:
I can use a general `exinv : hsubtypes X` instred of `x != 0`.

DanGrayson:
Why bother using exinv?  Isn't exinv x equivalent to an inverse existing?
If inverses
don't always exist, why call it a division ring, or something similar to
that?  It's just a ring.

cathlelay:
An other solution is to introduce the appartness relation.
`x != 0` will be replaced by `x # 0`.

DanGrayson:
Yes, but what problem is that a solution to?  There's no point defining a class of objects that have an extra relation if the only implication involving it is that x#0 implies x has an inverse, for it will have no use.  

I'm becoming convinced that the correct thing is to not use terms in this context such as "field" or "division ring".

What about the following idea?  Prove, of the real numbers:

     forall x:Reals, x!=0 -> LEM -> x has an inverse

Here LEM is the law of excluded middle.  Then define a constructive field to be one
with that property.

cathlelay:
LEM seems to be exactly the predicate `isdecDcuts x` defined in decDcuts.v.

vladimirias:
Maybe this should be isDivCommRig .

DanGrayson:
No, the definition of isdecDcuts depends on isdecDcuts_isaprop, which uses hdisj instead of coprod.  So decidability is not a consequence, and including "isdec" in the name is inappropriate, because decidability has to do with knowing what the answer is.

The new definition of DivRig looks fine, but probably the theorem to come will be that LEM implies that the real numbers give one.  Is that right?

DanGrayson:
It might make more sense to remove commutativity from isDivRig.

DanGrayson:
I've edited my previous response, which was wrong.

PR #199, opened 2015-10-20
Initial algebra as colimit of initial chain

benediktahrens:
- repackage cocones
- show that colimit of initial chain of omega-cocont functor is initial algebra


mortberg:
Maybe we should move chains from the colimits folder to the main category theory folder?

benediktahrens:
On 10/20/2015 03:34 PM, Anders Mörtberg wrote:
> Maybe we should move chains from the colimits folder to the main category theory folder?

I think that it is fine where it is.

PR #201, opened 2015-10-21
Proof of Lambek's lemma

mortberg:
- Proof of Lambek's lemma: If (A,a) is an initial F-algebra then a is an iso
- Notation for precategory_FunctorAlg and make the precategory implicit in FunctorAlg.

### DIFF
--- a/UniMath/CategoryTheory/FunctorAlgebras.v
+++ b/UniMath/CategoryTheory/FunctorAlgebras.v
@@ -15,7 +15,7 @@ Contents :
 
 - Saturated if base precategory is
 
-- Lambek's lemma: if (A,a) is inital F-algebra then a is an iso
+- Lambek's lemma: if (A,a) is an inital F-algebra then a is an iso
 
 ******************************************************************)
 
@@ -35,8 +35,7 @@ Local Notation "G □ F" := (functor_composite _ _ _ F G) (at level 35).
 
 Section Algebra_Definition.
 
-Variable C : precategory.
-Variable F : functor C C.
+Context {C : precategory} (F : functor C C).
 
 Definition algebra_ob : UU := Σ X : C, F X ⇒ X.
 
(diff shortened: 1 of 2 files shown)


UniMath/CategoryTheory/FunctorAlgebras.v
```diff
@@ -280,8 +281,41 @@ Proof.
...
+
+Local Definition AaInitial : Initial (FunctorAlg F hsC) := tpair _ _ AaIsInitial.
+
+Local Notation A := (alg_carrier _ Aa).
+Local Notation a := (alg_map _ Aa).
+
+(* (FA,Fa) is an F-algebra *)
+Local Definition FAa : algebra_ob F := tpair (λ X, C ⟦F X,X⟧) (F A) (# F a).
+
+Lemma initialAlg_is_iso : is_iso a.
```
benediktahrens: In this lemma, please use the projections instead of destructing the pairs. You will get better computation that way.

UniMath/CategoryTheory/FunctorAlgebras.v
```diff
@@ -280,8 +281,41 @@ Proof.
...
+Proof.
+case (AaIsInitial FAa); simpl; intros Fa' HFa'.
+destruct Fa' as [a' Ha']; unfold is_algebra_mor in Ha'; simpl in *.
+assert (Ha'a : a' ;; a = identity A).
+  assert (algMor_a'a : is_algebra_mor _ _ _ (a' ;; a)).
+    unfold is_algebra_mor.
+    now rewrite functor_comp, <- Ha', assoc.
```
benediktahrens: Can you hide the terms produced via "rewrite"?

mortberg:
Done. I think I haven't hidden too much of the term now.

UniMath/CategoryTheory/FunctorAlgebras.v
mortberg: See my patch below.

benediktahrens:
Also, when you change the status of arguments in a file, you need to adapt the whole of UniMath to it, in particular, SubstitutionSystems.

mortberg:
Ah yes of course, and travis of course reported it. Anyway, should I wait for your patch on SubstitutionSystems to be merged and then rebase and patch with my changes?

benediktahrens:
On 10/21/2015 11:55 AM, Anders Mörtberg wrote:
> Ah yes of course, and travis of course reported it. Anyway, should I
> wait for your patch on SubstitutionSystems to be merged and then
> rebase and patch with my changes?

Yes, sounds good.
Feel free to review my PR, btw.

PR #205, opened 2015-10-22
coproducts as instances of colimits

benediktahrens:
- some more constructor functions for convenience
- comment out a lemma about coproducts for now, which should be an instance of a lemma about colimits


mortberg:
LGTM, modulo travis

benediktahrens:
Should some of the definitions on coproducts be opaque, as to avoid unfolding down to colimits?

PR #207, opened 2015-10-26
Cleaning SubstitutionSystems

benediktahrens:
* using bullets in proofs more systematically
* replacing Admitted by Qed in order to increase compile time

### DIFF
--- a/UniMath/SubstitutionSystems/Lam.v
+++ b/UniMath/SubstitutionSystems/Lam.v
@@ -253,29 +253,27 @@ Lemma bracket_property_for_LamE_algebra_on_Lam (Z : Ptd)
 Proof.
   (* Could we have this in a more declarative style? *)
   assert (Hyp := pr2 (pr1 (pr2 LamHSS _ (f;; bla)))).
-  
   apply parts_from_whole in Hyp.
   apply whole_from_parts.
   split.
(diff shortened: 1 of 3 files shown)


mortberg:
LGTM after a quick look. There are still some "math goal with ...", but maybe they are not so slow?

Also, how much does this reduce the compilation time?

benediktahrens:
On 10/26/2015 08:38 PM, Anders Mörtberg wrote:
> LGTM after a quick look. There are still some "math goal with ...",
> but maybe they are not so slow?
> 
> Also, how much does this reduce the compilation time?

I have not done any measurements of compilation time - the cleaning was
to make the proofs maintainable, not to make them compile faster.

One of the Qed's takes more than 10 minutes, and thus breaks Travis. I
have replaced it by Admitted in one of the commits.

mortberg:
Ok, so it should be at least 10 min faster now then. I vote for merging it.

benediktahrens:
On 10/26/2015 10:45 PM, Anders Mörtberg wrote:
> Ok, so it should be at least 10 min faster now then. I vote for merging it. 

No, it had been Admitted all the time. I changed to Qed just for one
commit in this PR, but Travis bailed out then, so the last commit of
this PR changes Qed back to Admitted.

PR #208, opened 2015-10-27
This contains my renaming proposal for total2_paths_second_isaprop, etc.

DanGrayson:


mortberg:
This renames both `total2_paths_second_isaprop` and `total2_paths_isaprop` into `subtypeEquality`, and `total2_paths_isaprop_equiv` into `subtypeInjectivity`?

It seems like you also had to change some proofs by adding some `intro` (in colimits.v for example)?

benediktahrens:
I fear that this patch will conflict with our work on limits (not submitted as a PR yet), and with my cleaning of SubstitutionSystems, only to remove whitespaces...

DanGrayson:
Let's pull this one after we pull your future PR.  It's easier to redo this one.

DanGrayson:
No, there are 4 lemmas now, 2 of them with a " ' " at the end of the name.  One of those is
used 0 times, and the other is used 1 time.

The added "intro"s are due to changing the hypothesis to one that checks every member of
the family is a proposition, instead of just 1.

Some of the proofs had the order of goals changed, with the check that things are propositions
coming sooner rather than later, so for those, I just interchanged the parts.

DanGrayson:
Re: "I fear that this patch will conflict with our work on limits (not submitted as a PR yet), and with my cleaning of SubstitutionSystems, only to remove whitespaces..."

Oh, wait, I thought this was a comment about the white space pull request.  Yes, some white spaces are removed.  Still, you're welcome to submit yours first and let me redo this one.  White spaces at ends of lines are now automatically removed when files are saved, but this PR is not only to remove white space.

benediktahrens:
On 10/27/2015 08:42 AM, Daniel R. Grayson wrote:
> Let's pull this one after we pull your future PR.  It's easier to
> redo this one.

The future PR I was referring to is now #211. I am in the process of
updating it.

benediktahrens:
On 10/27/2015 08:44 AM, Daniel R. Grayson wrote:
> 
> The added "intro"s are due to changing the hypothesis to one that
> checks every member of the family is a proposition, instead of just
> 1.

Is there a reason you chose that lemma over the one that had been used
before?

DanGrayson:
Yes: because then we really are talking about a subtype, explicitly, i.e.,
a function X->UU
whose values are all propositions.  It happens to be the case every single
time you used it.

DanGrayson:
Also, when reading a proof, it's easier for the reader to see a variable than a
larger subexpression.

DanGrayson:
All right, that should fix things.  Let's see if the test works.

DanGrayson:
This now also resolves #216.

benediktahrens:
lgtm

Issue #197, opened 2015-10-20
Inductive type bool from Coq's Init.Datatypes


benediktahrens:
We are using the booleans of Coq.Init.Datatypes, which are defined as an inductive type:

```
Inductive bool : Set :=
  | true : bool
  | false : bool.
```

Should we try replacing it by unit \coprod unit?
Is the alternative definition of bool as coproduct already given somewhere?

DanGrayson:
Before we do that, I'd like to try replacing ```coprod``` by ```total2``` over ```stn 2```.  Since many analogous theorems are proved for both ```coprod``` and ```total2```, we'd be able to shorten many proofs.  An example: a coproduct of sets is a set, and a total2 of a family of sets indexed by a set is a set.

arnoudvanderleer:
@DanGrayson I think it would be unwise to define `coprod` via `total2`, because then we can no longer use the `induction` tactic for elements of `coprod A B` for case distinction.
And of course, we still could prove an induction principle lemma, but personally, I prefer being able to use the `induction` tactic.

arnoudvanderleer:
Come to think of it, we have the same problem with `coprod unit unit`, because then `induction x` will give two cases: `inl x'` and `inr x'`. These are not definitionally equal to `true` and `false`, not everything in `unit` is definitionally equal to `tt`. So we would have to do `induction x as [x' x']; induction x'.` or something like that. And that feels cumbersome when one neesd to to it very often.
So is this a thing we still want?

arnoudvanderleer:
On the other hand, it may be possible to write a tactic that can apply an induction principle lemma for some custom types. Something like
```coq
Tactic old_induction = induction.

Tactic induction [some parameter x] :=
  match typeof x with
    coprod => refine (coprod_ind _ _)
    bool => refine (bool_ind _ _)
    _ => induction x
  end.
```

arnoudvanderleer:
Oh, I just realized that we can provide our own induction principles to coq with `induction ... using some_bool_rect`, so it is slightly more of a hassle, but still doable.

Regardless, @benediktahrens , how does this tie into the freeze of Foundations?

arnoudvanderleer:
Closing this as no longer relevant. Feel free to reopen if there is still interest in having/doing this.

PR #211, opened 2015-10-27
Limits

benediktahrens:
Limits as duals of colimits
some specific limits:
* products
* pullbacks


mortberg:
I think we should add:

Arguments f : simpl never.

for all definitions f in limits, and in all of the files that are implemented using limits or colimits. This will make sure that simpl don't unfold too much and show the underlying implementation to the user which will make these definitions more pleasant to work with. 


We should also change three in pullbacks to "unit + unit + unit" or some other finite set with 3 elements...

benediktahrens:
On 10/26/2015 10:42 PM, Anders Mörtberg wrote:
> I think we should add:
> 
> Arguments f : simpl never.
> 
> for all definitions f in limits, and in all of the files that are
> implemented using limits or colimits. This will make sure that simpl
> don't unfold too much and show the underlying implementation to the
> user which will make these definitions more pleasant to work with.

Good idea!

> We should also change three in pullbacks to "unit + unit + unit" or
> some other finite set with 3 elements...

We will need a layer on top of that to make programming with it
convenient, that is, to simulate the induction principle of our type
"three".

DanGrayson:
Yes:
```
Inductive three := One | Two | Three.
```
goes against the guidelines.  Another suitable set would be ```stn 3```.  It's easy to
make its elements with expressions like ```stnpair 3 1 (idpath _)```.

DanGrayson:
Another possibility would be coprod bool unit, and think of ii2 tt as being for the vertex of pushouts
and pullbacks.

mortberg:
Yes, but defining functions by case on `stn 3` is quite painful. I tried yesterday with natbooleq (or what it is called) and if-then-else, but the proofs got much longer and I gave up. I think using `bool + unit` could be easier, but I will have to try. 

Something that would be cool would be to have a dependently typed function that gives you the recursor for `stn n` (we need dependent types as the arity will depend on n).

mortberg:
I didn't see your second reply before replying, but we are thinking along the same lines.

DanGrayson:
No, defining a function on stn 3 is not so bad:

```

Definition three_cases {X} (a b c : X) : stn 3 -> X.
Proof.
  intros ? ? ? ? [i lt].
  induction i.
  - exact a.
  - induction i.
    * exact b.
    * induction i.
      + exact c.
      + simpl in lt. induction (nopathsfalsetotrue lt).
Defined.

Goal stnsum (three_cases 4 5 6) = 15. reflexivity. Qed.
```

DanGrayson:
We could get rid of four inductive definitions by replacing empty by stn 0,
unit by stn 1, bool by stn 2, and coprod by total2 over bool.

mortberg:
Ok, not too bad. But then when I do proofs I need to do a bit of extra work to expose the underlying nats to do induction on them (this work is trivial but tedious). For `stn 3` this is ok, but imagine doing it for a bigger n...  I think the best would be to define the recursor and then prove a general induction lemma for `stn n`. Maybe something to try for UniMath day.

DanGrayson:
How about just packaging up introduction and elimination for a three
element type "three", implemented
with stn 3, and with proofs locked under Qed so you never have to see the
numbers?  I could try that
exercise.

mortberg:
Do that if you want to, I could also do it. For this problem `bool + unit` might still be easier to work with though. Anyway this will be a good test to see if `stn 3` or `bool + unit` is easier to work with (ie which one gives the simplest and shortest definitions and proofs).

benediktahrens:
On 10/27/2015 10:16 AM, Anders Mörtberg wrote:
> Do that if you want to, I could also do it. For this problem `bool +
> unit` might still be easier to work with though. Anyway this will be
> a good test to see if `stn 3` or `bool + unit` is easier to work with
> (ie which one gives the simplest and shortest definitions and
> proofs).

If you are against three (which I am), you should also be against bool,
as it is the same: an inductive datatype - see #197 .
So replacing three by bool + unit is not solving the issue.

One could still argue that having only bool is better than having bool
and three, but in my opinion, that tiny advantage does not justify the
hassle of defining three = bool + unit.

So, put differently, if you want to eliminate three, please also
eliminate bool.


I am not very familiar with the parts of Foundations that use bool, but
I fear that it might be a gigantic undertaking.

mortberg:
Getting rid of bool seems like an incredibly painful endeavour and nothing I want to work on. I think it makes sense to remove three now as we are just adding new stuff. Removing bool will require much more work as it is used everywhere, while three is only used in one file. So lets keep three for now and then get rid of it once Dan has packaged `stn 3` nicely?

DanGrayson:
If we could figure out how to get Coq to have "induction" call something we
make
by hand, as Anders has told me is possible, then replacing "bool", "unit",
"empty",
and "coprod" would all be easy.

PR #212, opened 2015-10-27
setup travis for parallel builds

benediktahrens:
Every UniMath package (plus dependencies) built by Travis in a separate machine.

Pro: 
* faster build time
* less likely to run over the allowed 50 min

Con:
* cannot check "install" target any more


benediktahrens:
Variants are probably possible, e.g, building all the packages except for SubstitutionSystems in one build, and SubstitutionSystems + dependencies in a second build.

DanGrayson:
I notice the elapsed time is 45 minutes greater than usual.  Probably the
jobs are
running sequentially.

PR #213, opened 2015-10-27
Fix coq buffer syntax table in emacs

DanGrayson:
... so that ' and _ count as characters in words


mortberg:
That seems cool! Now, if I press M-t at _ in "apa_bepa cepa" it will swap "apa_bepa" with "cepa" and not just "apa" with "bepa"?

DanGrayson:
Exactly.

DanGrayson:
And here is something cool I just now discovered:
```
C-M-. runs the command find-tag-regexp, which is an interactive autoloaded
compiled Lisp function in `etags.el'.

It is bound to C-M-..

(find-tag-regexp REGEXP &optional NEXT-P OTHER-WINDOW)

Find tag (in current tags table) whose name matches REGEXP.
Select the buffer containing the tag's definition and move point there.

If second arg NEXT-P is t (interactively, with prefix arg), search for
another tag that matches the last tagname or regexp used.  When there are
multiple matches for a tag, more exact matches are found first.  If NEXT-P
is negative (interactively, with prefix arg that is a negative number or
just M--), pop back to the previous tag gone to.

If third arg OTHER-WINDOW is non-nil, select the buffer in another window.

A marker representing the point when this command is invoked is pushed
onto a ring and may be popped back to with M-*.
Contrast this with the ring of marks gone to by the command.

See documentation of variable `tags-file-name'.

[back]
```

mortberg:
Wow, emacs is great! :)

PR #218, opened 2015-10-28
Fix ispo

### LINKED ISSUE #216: posets should have an antisymmetric relation
... so this definition needs to be changed:

```
Definition ispo { X : UU } ( R : hrel X ) := istrans R × isrefl R .
```
Bourbaki Set Theory calls something with this property a preorder relation.  

I can infer that "ispo" does not mean "is preorder" because of this:
```
Definition Poset := total2 ( fun X : hSet => po X ) .
```

If we fix that, we'll have to modify this one, too:
```
Definition iseqrel { X : UU } ( R : hrel X ) := ispo R × issymm R .
```

If no one objects, I'll fix it.

DanGrayson:
resolves #216

### DIFF
--- a/UniMath/Dedekind/DedekindCuts.v
+++ b/UniMath/Dedekind/DedekindCuts.v
@@ -101,7 +101,7 @@ Proof.
   now intros x P HP ; apply HP ; clear P HP.
 Qed.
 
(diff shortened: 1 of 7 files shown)


DanGrayson:
This got included in #208 by accident.

PR #225, opened 2015-11-01
Finite sums using coprod associativity pull

DanGrayson:


UniMath/Foundations/OrderedSets.v
```diff
@@ -440,6 +378,239 @@ Defined.
...
+Proof.
+  intros.
+  apply (isofhleveltotal2 2).
+  - apply setproperty.
+  - intro x. apply setproperty.
+Defined.
+
+Definition total2_hSet {X:hSet} (Y:X->hSet) : hSet := hSetpair (Σ x, Y x) (isaset_total2 X Y).
+
+Notation "'Σ'  x .. y , P" := (total2_hSet (fun x => .. (total2_hSet (fun y => P)) ..))
```
benediktahrens: Will this notation conflict with the same notation for general total2's?

UniMath/Foundations/OrderedSets.v
```diff
@@ -440,6 +378,239 @@ Defined.
...
+    apply lex_istotal.
+    { apply FiniteOrderedSet_isdeceq. }
+    { apply istotal_FiniteOrderedSet. }
+    intro; apply istotal_FiniteOrderedSet. }
+  apply isfinitetotal2.
+  { apply finitenessProperty. }
+  intro; apply finitenessProperty.
+Defined.
+
+Notation "'Σ'  x .. y , P" := (concatenateFiniteOrderedSets (fun x => .. (concatenateFiniteOrderedSets (fun y => P)) ..))
```
benediktahrens: Same here: no conflicts with total2?

benediktahrens:
lgtm modulo question about having the \Sigma notation overloaded

UniMath/Foundations/OrderedSets.v
DanGrayson: There are two reasons it should not conflict.  One is, it's in a named
scope called "set",
which is not open by default.  If it is open, for some reason, and the
notation gets applied
to non-sets, there will be an error, and if it gets applied to sets, it'll
work just as well,
because of the automatic coercion from sets to types.

DanGrayson:
Oops, I added a little more stuff, but it doesn't affect anything.

Benedikt, did you like me answer?  Can we pull?

benediktahrens:
Yes, please pull!

Am 2. November 2015 07:17:16 EST, schrieb "Daniel R. Grayson" <notifications@github.com>:

PR #228, opened 2015-11-02
Apartness

cathlelay:
Definitions of apartness and constructive structures.


UniMath/Foundations/Algebra/Apartness.v
```diff
@@ -0,0 +1,122 @@
...
+Qed.
+
+Definition aprel (X : UU) := Σ ap : hrel X, isaprel ap.
+Definition aprel_pr1 {X : UU} (ap : aprel X) : hrel X := pr1 ap.
+Coercion aprel_pr1 : aprel >-> hrel.
+
+Definition apSet := Σ X : hSet, aprel X.
+Definition apSet_pr1 (X : apSet) : hSet := pr1 X.
+Coercion apSet_pr1 : apSet >-> hSet.
+Notation "x # y" := (pr2 (_ : apSet) x y).
```
DanGrayson: A special notation like this would be improved by putting it in a named scope, such as "apartness" or "constructive_logic" or something similar.

UniMath/Foundations/Algebra/Apartness.v
```diff
@@ -0,0 +1,122 @@
...
+  apply impred_isaprop ; intro x.
+  apply impred_isaprop ; intro y.
+  apply impred_isaprop ; intro z.
+  apply isapropimpl.
+  now apply pr2.
+Qed.
+
+(** ** Apartness *)
+
+Definition isaprel {X : UU} (ap : hrel X) :=
```
DanGrayson: I wonder why such a thing should be called "apartness" if it has no relationship with x != y.

DanGrayson: Perhaps add tightness:  `¬ (x#y) -> x=y` ?  Then together with irreflexivity ` x=y -> ! x#y ` and LEM we would have `x≠y <-> x#y'.

UniMath/Foundations/Algebra/ConstructiveStructures.v
```diff
@@ -0,0 +1,85 @@
...
+  λ X : ConstructiveCommutativeRing, pr1 X,, pr1 (pr2 X).
+Coercion ConstructiveCommutativeRing_CommutativeRing : ConstructiveCommutativeRing >-> ConstructiveRing.
+Definition ConstructiveCommutativeRing_ConstructiveCommutativeRig :
+  ConstructiveCommutativeRing -> ConstructiveCommutativeRig :=
+  λ X : ConstructiveCommutativeRing, pr1 X,, (iscommrngopstoiscommrigops _ _ _ (pr2 X)).
+Coercion ConstructiveCommutativeRing_ConstructiveCommutativeRig : ConstructiveCommutativeRing >-> ConstructiveCommutativeRig.
+
+(** ** Constructive rig with division *)
+
+Definition isConstrDivRig (X : ConstructiveCommutativeRig) :=
```
DanGrayson: If it's commutative, it should be called a constructive field, since "division ring" implies no commutativity is assumed.

DanGrayson: ... except it has no subtraction, so perhaps take out the commutativity and leave the name.

UniMath/Foundations/Algebra/ConstructiveStructures.v
```diff
@@ -0,0 +1,91 @@
...
+  Σ X : apsetwithbinop, @ismonoidop X op.
+
+(** ** Constructive abelian monoid *)
+
+Definition ConstructiveAbelianMonoid :=
+  Σ X : apsetwithbinop, @isabmonoidop X op.
+
+(** ** Constructive group *)
+
+Definition ConstructiveGroup :=
```
andrejbauer: I wonder if there's any point in labeling *constructive* those structures which are foundation-agnostic. The definition of groups really does not depend on what foundation we use. It's a different matter when you get ot zero-divisors and invertible elements of a ring, though.

UniMath/Foundations/Algebra/Apartness.v
andrejbauer: One reason for calling this "apartness" is that it is the standard definition of apartness.

UniMath/Foundations/Algebra/ConstructiveStructures.v
DanGrayson: That's a good point.  There seems to be no need for this definition, conceptually.

UniMath/Foundations/Algebra/Apartness.v
DanGrayson: Well, if it's standard, one must go with that.

UniMath/Foundations/Algebra/ConstructiveStructures.v
cathlelay: Done. There is now only structures needed for constructive division rigs and constructive fields.

DanGrayson:
In this definition : 
```
Definition isConstrDivRig (X : ConstructiveRig) :=
+  isnonzeroCR X × (∀ x : X, x # CRigzero -> multinvpair X x).
```
is there anything that prevents me from making `nat` into a constructive division rig by
defining `m#n?=empty` ?  If so, the definition's name seems inappropriate, since it conveys
that thought that division should be often possible, in which case adding tightness would
improve things: `neg (x#y) -> x=y`.  Maybe also add `1#0`, unless it's provable, and consider
whether the lemma `x*y=1 -> x#0` is provable from it.

cathlelay:
2015-11-05 10:49 GMT-05:00 Daniel R. Grayson <notifications@github.com>:

cathlelay:
`x*y=1 -> x#0` is a corollary of :
`Lemma CRigmultapCRigzero :
  forall x y : X, x * y # (0%rig : X) -> x # (0%rig : X) ∧ y # (0%rig : X).`
under the hypothesis `1 # 0`.

DanGrayson:
Thank you.  Now is there anything that prevents the existence of a constructive division rig that has a nonzero noninvertible element?  If not, the name "division rig" would be better applied to a concept where tightness is part of the definition.

DanGrayson:
Indeed, consider a local commutative ring.  That's a ring with a unique maximal ideal m.  Define r # s in the ring to mean r-s is not in m, i.e., that r-s is invertible.  Since noninvertible elements are closed under addition, the cotransitive property holds, so it's an apartness relation.  But there are many local rings that are not fields (or division rings), so calling it a constructive division ring would be confusing.  I suggest adding tightness as a condition.

andrejbauer:
In general, you should call a gadget `constructive X` if its definition looks different from the classical definition of `X`, but is equivalent to it if we assume LEM. By the way, what do you do in UniMath about maximal idelas, do you just assume classical logic?

DanGrayson:
There is nothing in UniMath yet about maximal ideals.  For fields, assuming that x=0 is decidable one could develop a lot of algebra for finitely presented algebras over the field without using LEM.  Anyone who wants to apply that theory to the complex numbers would have to assume LEM, but it would be fine for computable fields.

DanGrayson:
Here's a constructive theorem of commutative algebra one could aim for: let k be a field with decidable equality.  Then it is decidable whether a system of polynomial equations in several variables has a solution in some finite extension of k.

DanGrayson:
I'm closing this for now, until you have some more commits for us to consider, Catherine.  Then just reopen it.

PR #231, opened 2015-11-04
Tex make rules

benediktahrens:
preliminary version with too much verbosity, meant for testing only


DanGrayson:
How about removing the obsolete targets "pdf" and "latex"?

PR #239, opened 2015-11-16
Non-negative real numbers are a Constructive Division Rig

cathlelay:


andrejbauer:
By the way, why are you doing the nonnegative reals (as opposed to all reals)?

andrejbauer:
Oh, you want a rig. Is the plan to get reals from the rig later?

vladimirias:
Yes. Cath has planned it this way. I see it as making sense in part because there are so many areas (norms, metrics) where non-negative reals play a role. And if it is easier to define non-negative ones then it seems like an interesting approach.

Vladimir.

PS There is a construction of the ring of differences of a rig in the algebra files, so getting all reals from non-negative reals is just one line.

UniMath/Dedekind/.package/files
```diff
@@ -3,3 +3,4 @@ Fields_comp.v
 Complements.v
 NonnegativeRationals.v
 DedekindCuts.v
+decDcuts.v
```
DanGrayson: A more readable name for this file would be DecidableDedekindCuts.v.

DanGrayson:
The file `my-commit.sh` is only for you, so it should not be in the repository.

DanGrayson:
There seems to be no reason to offer a definition of ConstructiveRig, since the notion of rig can be used instead.  The time to add an apartness relation is one step later, as part of the definition of constructive division rig, when the apartness relation is useful as the condition under which multiplicative inverses exist.  My comment is analogous to the comment Andrej made formerly about there being no need to have a notion of constructive group.

DanGrayson:
Better filenames than `Fields_comp.v` and `Sets_comp.v` would be `Fields.v` and `Sets.v`, as "comp" conveys no useful information.

DanGrayson:
Tightness should be part of the definition of constructive division rig, for otherwise we can make many rings that are not division rings into constructive division rigs, even using classical logic, as I mentioned before. Indeed, consider a local commutative ring. That's a ring with a unique maximal ideal m. Define r # s in the ring to mean r-s is not in m, i.e., that r-s is invertible. Since noninvertible elements are closed under addition, the cotransitive property holds, so it's an apartness relation.  In general, one should label an object a constructive version of some classical concept only if the introduction of the law of excluded middle renders the two notions equivalent.

Here is an example of a local commutative ring that is not a field: the ring R of rational numbers whose denominator is odd.  Its maximal ideal is generated by 2, which is not invertible.  Under the apartness relation defined above, 2 is unequal to 0 but apart from 0.

cathlelay:
I chose to define ConstructiveRig and ConstructiveRing to have the same
structure than DivisionRig and fld.
I will remove these to definitions.

2015-11-24 7:57 GMT-05:00 Daniel R. Grayson <notifications@github.com>:

cathlelay:
Tightness was included in apsetwith2binop.
With the last corrections, it is easier to see.

2015-11-24 8:23 GMT-05:00 Daniel R. Grayson <notifications@github.com>:

DanGrayson:
Here's a question for my own edification.  Suppose we try to define what a constructive
division ring is by saying that x is "apart" from y if x-y is invertible.  We postulate that
if x is not invertible, then x=0, so that gives tightness.  We also assume 1 is not equal to 0.
Which of the other properties of apartness follow from these assumptions, and which do we 
have to postulate?

andrejbauer:
Irreflexvity and symmetry of apartness are obvious. Cotransitivity comes down to: if x - y is invertible then x - z is invertible or z - y is invertible. This looks like a piece of honest algebra and it must amount to a known algebraic condition. Do we have any algebraists lurking by any chance? Or former algebraists :-)

andrejbauer:
Trying to do some algebra: contransitivity seems to be equivalent to: if a + b is invertible then a is invertible or b is invertible.

DanGrayson:
That sounds sensible, for if invertibility were the negation of a proposition P, then P would be
transitive in the sense that P a -> P b -> P (a+b).  If we also knew P 0 and P a -> P (a b) (do we?),
then P would describe an ideal, and since the elements not in the ideal are invertible, we would 
say the ideal is the maximal ideal in a local ring.

DanGrayson:
All this raises a question: if apartness can be defined in terms of invertibility, why introduce it as a separate element of the structure?

vladimirias:
Can one prove for reals that if X is invertible then x>0 or x<0 ?

DanGrayson:
I think so...

DanGrayson:
Catherine, should this lemma be stronger?:
```
Lemma apCFmult :
  forall x x' y y' : X,
    x * y ≠ x' * y' -> x ≠ x' ∨ y ≠ y'.
```
Can't we prove `x ≠ x' ∧ y ≠ y'`?

DanGrayson:
This is a strange looking lemma:
```
Lemma hqinv0 : / 0 = 1.
```
It's hard to imagine ever needing to know it is true, as it's an accident of implementation that
`/ 0` is even defined.  Do you need it?

cathlelay:
We cannot prove x ≠ x' ∧ y ≠ y', because is it false : this theorem is a corollary of the cotransitivity of apartness with "islapbinop" and "israpbinop" which `forall x y y' : X,
    x * y ≠ x * y' -> y ≠ y'` (and the similar property for y).

DanGrayson:
Re: "We cannot prove x ≠ x' ∧ y ≠ y', because is it false".

That's interesting.  What's a good counterexample?

cathlelay:
About hqinv0, even if this lemma is strange, it is a consequence of the definition of hq. As inverse function can be defined as a total function is decidable sets, it is more useful to assign an arbitrary value to / 0.

For NonnegativeRationals, I choose to define / 0 as 0 because there are better properties (e.g. forall x : NonnegativeRationals, / / x = x).

It seams to me that I do not use this theorem, it can be remove.

cathlelay:
" 1 * x ≠ 1 * y " seams a good counterexample.

DanGrayson:
Oh, of course, thank you!

DanGrayson:
In the definition of `isEffectiveOrder` we see the condition `(¬ lt x y) = (le y x)`.  It would be better
if it were phrased as `(¬ lt x y) <-> (le y x)`, for then no axioms would be required to prove it.

DanGrayson:
Axioms perhaps can also be avoided if the following is weakened:
```
Lemma Dcuts_zero_empty :
  forall r : NonnegativeRationals, r ∈ 0 = hProppair empty isapropempty.
```
Wouldn't it suffice to prove this?:
```
Lemma Dcuts_zero_empty :
  forall r : NonnegativeRationals, ¬ (r ∈ 0).
```

cathlelay:
In many theorems, I chose the equality instead of logical equivalence because of my usage : in Coq the logical equivalence can be write only between two elements of type Prop.

DanGrayson:
I see other instances where you prove two propositions are equal, when it would probably suffice to prove each implies the other, or just that one implies the other.

In UniMath, logical equivalence can be written for any two types, see:
```
Definition logeq (X Y : UU) := dirprod (X -> Y)  (Y -> X) .
Notation " X <-> Y " := (logeq X Y) : type_scope .
```

cathlelay:
OK. I will change it.

DanGrayson:
I suggest examining every use of `uahp`, `uahp'`, or `isapropneg` with an eye toward removal.

DanGrayson:
By the way, the rationals have decidable equality, so there is a proposition logically equivalent to x != y when x and y are rational that does not require any axioms to prove that it is a proposition.  One of my near term goals is to install that proposition in Foundations/RationalNumbers.v.

cathlelay:
I replaced many equalities by logical equivalences.

For the last ones, I will need to split statements because "apply" is not able to choose the appropriate sens of the equivalence.

cathlelay:
About "x != y", it should be easier to change the definition of neg to have "neg : UU -> hProp".

DanGrayson:
I'm not sure what this should be easier than.  The change you propose would
involve appealing to an axiom to accomplish
the proof, and we should avoid appealing to axioms unnecessarily.  For
rational numbers, integers, and natural numbers, decidablity
ensures that there are axiom-free hProps available that are logically
equivalent to x != y.

In any case, for what you want, if you really want it see:
```
Definition hneg ( P : UU ) : hProp := hProppair ( ¬ P ) ( isapropneg P ) .
```
But avoid using it, if possible.

andrejbauer:
Just to confirm that for reals apartness `x # y` is equivalent to `x < y \/ y < x` and is also equivalent to `(x - y)` being invertible. If you're going to rip out apartness from the definition of a constructive field (or division ring or whatever) then you'll need to put in more axioms about invertibility. Your choice.

DanGrayson:
You did a good job removing uahp and isapropneg.  The single use of isapropneg can
be removed after we install an axiom-free inequality for the rationals, so let's keep that in mind.

I think it's good to go.

Issue #227, opened 2015-11-02
Apartness in sets with decidable equality (axiom removal)


DanGrayson:
Consider the natural problem of proving that `Σ i, i≠0` and `Σ i, i>0` are
equivalent sets.  The former set is the complement of `0` in `nat`, as encapsulated
in the following definition from PartC:

```
Definition compl (X:UU) (x:X):= Σ x', x ≠ x'.
```

One easily defines functions in both directions:

```
p : Σ i, i≠0 -> Σ i, i>0
q : Σ i, i≠0 <- Σ i, i>0
```

, but then one has to prove that the two composite functions are homotopic to
identity functions, and there one uses the lemmas that show that `i≠0` and
`i>0` are propositions.  The proof that `i≠0` is a proposition involves an
axiom, because `i≠0` is defined as `i=0 -> empty`.  The axiom `funextempty`
provides the needed statement:

```
Axiom funextempty : ∀ (X:UU) (f g : X->empty), f = g. 
```

Use of that axiom makes its way into the definition of the homotopy
`q∘p ~ idfun _`, rendering it incomputable.  Such homotopies are provided from
any weak equivalence by these functions:

```
Definition homotweqinvweq {X Y : UU} (w : X ≃ Y) : ∀ y : Y, w (invmap w y) = y.
Definition homotinvweqweq {X Y : UU} (w : X ≃ Y) : ∀ x : X, x = invmap w (w x).
```

Those homotopies, in turn, make their way into definitions of other maps
and equivalences, rendering them incomputable.  For example, consider `weqfp`,
a useful theorem:

```
Definition weqfp {X Y : UU} (w : X ≃ Y) (P:Y->UU) : (Σ x : X, P (w x)) ≃ (Σ y, P y).
```

, where one of the two maps in the equivalence it provides incorporates one of
those potentially incomputable homotopies:

```
Definition weqfp_invmap { X Y : UU } ( w : X ≃ Y ) (P:Y->UU) : (Σ y, P y) -> (Σ x,P(w x)).
Proof.
  intros ? ? ? ? yp.
  exact (invmap w (pr1 yp),,transportf P (! homotweqinvweq w (pr1 yp)) (pr2 yp)).
Defined.
```

The route for the propagation of incomputability passes through the following
basic lemma:

```
Theorem isapropneg (X:UU): isaprop (neg X).
Proof. intro.  apply invproofirrelevance . intros x x' .   apply ( funextempty X x x' ) . Defined .  
```

[...]

PR #240, opened 2015-11-17
Postponing use of funextempty

DanGrayson:
Here are some changes that remove early uses of the axiom funextempty, in order
to improve computation.  The axiom is used, for example, to show that a negation is
a proposition, so I've rooted out various propositions that were phrased as negations and
replaced them by axiom-free propositions.  This includes m ≤ n on nat.  We downplay the
use of x != y, and introduce on nat the notation m ≠ n for an equivalent proposition.
We should do the same for hz and hq, which I left alone.


vladimirias:
Dan,

I disagree with your change of the definition of \le and \ge on nat. There is a systematic way in which these relations are defined in Foundations that I have come to after many tries. It is not an accident but a result of much work that these definitions are what they they are for all four relations. The fact that formulated in this way they require funext for the proof of the isaprop component causes no difficulties anywhere.

Please reverse the changes for nat <= and >= they really contradict the systematic approach to these relations that is consistently used in Foundations.

If you would like you could introduce two new relations and prove lemmas about them but please do it in a separate file so that the main files of the library remain uncluttered.

I may have more requests of the same kind before this PR can me merged.

Vladimir.

DanGrayson:
I documented the trouble I had with the overuse of negation in issue https://github.com/UniMath/UniMath/issues/227 -- I ran into that problem while trying implement sequences as functions from [stn n] to some other type X, and it bugged me for weeks until I tracked it down.  And if you look in the new test files for comments with the word "fixed" you will see numerous computations that work now and used to not work.  For me this will make the system much more usable.

vladimirias:
Ok. I understood the issue and agree with Dan that the new definitions of >= and <= are better.

UniMath/Foundations/Basics/PartA.v
```diff
@@ -1752,10 +1755,45 @@ assert ( egf : ∀ xy : _ , ( g ( f xy ) ) = xy ) . intro . induction xy . apply
...
+  - intros yxp. induction yxp as [x yp]. induction yp as [y p]. reflexivity.
+Defined.
 
-
+Definition weqtotal2comm12 {X} (P Q : X -> UU) :
+  (Σ (w : Σ x, P x), Q (pr1 w)) ≃ (Σ (w : Σ x, Q x), P (pr1 w)).
+Proof.
+  intros.
+  refine (weqgradth _ _ _ _).
+  - intros [[x p] q]. exact ((x,,q),,p).
```
benediktahrens: I seem to remember that nested brackets in "intros ..." were forbidden?

UniMath/Foundations/Basics/PartA.v
```diff
@@ -1752,10 +1755,45 @@ assert ( egf : ∀ xy : _ , ( g ( f xy ) ) = xy ) . intro . induction xy . apply
...
+  - intros xyp. induction xyp as [x yp]. induction yp as [y p]. exact ((x,,y),,p).
+  - intros xyp. induction xyp as [xy p]. induction xy as [x y]. reflexivity.
+  - intros xyp. induction xyp as [x yp]. induction yp as [y p]. reflexivity.
+Defined.
 
+Definition weqtotal2dirprodassoc' {X Y:UU} (P: X × Y -> UU) : (Σ xy : X×Y, P xy) ≃ (Σ (y:Y) (x:X), P (x,,y)).
+Proof.
+  intros.
+  refine (weqgradth _ _ _ _).
+  - intros xyp. induction xyp as [xy p]. induction xy as [x y]. exact (y,,x,,p).
```
benediktahrens: I seem to remember that we wanted to use projections instead of destructing pairs, when defining maps on pairs.

UniMath/Foundations/Basics/PartA.v
DanGrayson: Right, but then Vladimir said that when he tried to remove his, the code slowed down, so he didn't want to remove them.  After that I began to feel entitled to use them in a few cases when it was late at night and I was tired of typing.

UniMath/Foundations/Basics/PartA.v
DanGrayson: Well, in some cases it might make a difference, so yes, one wants to keep that in mind.  I can change it if you want, but it takes a lot of energy to remember which projection does what.

UniMath/Foundations/Basics/PartA.v
benediktahrens: On 11/19/2015 06:42 PM, Daniel R. Grayson wrote:
>> 
>> - +Definition weqtotal2comm12 {X} (P Q : X -> UU) : +  (Σ (w : Σ x,
>> P x), Q (pr1 w)) ≃ (Σ (w : Σ x, Q x), P (pr1 w)). +Proof. +
>> intros. +  refine (weqgradth _ _ _ _). +  - intros [[x p] q]. exact
>> ((x,,q),,p).
> 
> Right, but then Vladimir said that when he tried to remove his, the
> code slowed down, so he didn't want to remove them.  After that I
> began to feel entitled to use them in a few cases when it was late at
> night and I was tired of typing.

Ok. I don't have a strong opinion, but I noticed it and did not want to
leave it without mention.

UniMath/Foundations/Basics/PartA.v
benediktahrens: On 11/19/2015 06:43 PM, Daniel R. Grayson wrote:
>> 
>> +Definition weqtotal2dirprodassoc' {X Y:UU} (P: X × Y -> UU) : (Σ
>> xy : X×Y, P xy) ≃ (Σ (y:Y) (x:X), P (x,,y)). +Proof. +  intros. +
>> refine (weqgradth _ _ _ _). +  - intros xyp. induction xyp as [xy
>> p]. induction xy as [x y]. exact (y,,x,,p).
> 
> Well, in some cases it might make a difference, so yes, one wants to
> keep that in mind.  I can change it if you want, but it takes a lot
> of energy to remember which projection does what.

Again, I don't have a strong opinion. But since your whole undertaking
is about getting more conversion, this choice seems weird unless we want
to activate eta for pairs.

benediktahrens:
lgtm

UniMath/Foundations/Basics/PartA.v
andrejbauer: What's the reasoning behind not using bested brackets in intros? It seems like a useful feature. Is it secretly inserting `match` statements?

DanGrayson: Yes, it inserts match statements.

vladimirias: I recall that the code slowed down when I removed destructs for coproducts and replaced them by inductions. When was the case with projections?

DanGrayson: I don't know what your question means.

DanGrayson:
Vladimir, do you have further comments for me on this pull request?

vladimirias:
No. I did not have time to see all of it but other than what we have discussed I did not see anything to complain about :). LGTM.

PR #245, opened 2015-11-30
add a lemma to yoneda

benediktahrens:


UniMath/CategoryTheory/yoneda.v
```diff
@@ -238,38 +241,45 @@ Proof.
...
 
+Lemma isweq_yoneda_map_1 (C : precategory) (hs: has_homsets C) (c : C)
+   (F : functor C^op HSET) :
+  isweq
+     (*a := hSetpair (hom _ ((yoneda C) hs c) F) (isaset_nat_trans_yoneda C hs c F)*)
+     (*b := F c*)
+     (yoneda_map_1 C hs c F).
+Proof.
+  set (T:=yoneda_map_2 C hs c F). simpl in T.
+  refine (gradth _ _ _ _ ).
```
andrejbauer: Out of curiosity, is it known how often `gradth` actually does anything?

DanGrayson: For equivalence between sets, as here, it's not really needed, because the extra adjointness
relation comes for free.  Often one can avoid its use between general types, i.e., sometimes the
maps you propose satisfy the adjointness easily, but in general, it's tricky.  In particular, it's tricky 
to get the adjointness for the homotopies between fgf and f from the adjointness between gfg and g,
hence the need for gradth in making the inverse of an equivalence.

vladimirias: What do you mean by “does anything”?

andrejbauer: I meant what you answered: how often are the original maps already adjoint?

PR #246, opened 2015-11-30
NonnegativeReals is a complete set

cathlelay:
I succeed to prove that Cauchy sequences of non-negative reals are convergent.

The next step will be the definition of real numbers (within few days).


andrejbauer:
Is there a good reason that you are using single-sided Dedekind cuts? In my experience the double-sided cuts are easier because they are more symmetric.

cathlelay:
Has I first define non-negative real numbers, symmetries implied by double-sided cuts are broken because I do not need to deal with negative numbers.
With this definition, I only need 3 hypotheses to define a Dedekind cut.

vladimirias:
Congratulations!

DanGrayson:
The diff displayed at https://github.com/UniMath/UniMath/pull/246/files appears to be incorrect, displaying  just some of the changes.

cathlelay:
I changed the order of some theorems to prove the convergence of Cauchy sequences.
I tried to merge all the modifications of my previous pull request, but I could miss some of them.

andrejbauer:
There's a lot of code here I am not familiar with. Could you just tell whether you're using classical logic? The definition itself looks like it ought to work constructively.

UniMath/Dedekind/Complements.v
```diff
@@ -1,5 +1,26 @@
 (** * Additionals theorems and definitions *)
 
+(** ** About nat *)
+
+Require Export UniMath.Foundations.NaturalNumbers.
+
+Lemma max_le_l : ∀ n m : nat, (n <= Nat.max n m)%nat.
+Proof.
+  induction n ; simpl Nat.max.
```
DanGrayson: You can use "max" for "Nat.max".

DanGrayson:
LEM isn't used, but these axioms are used:
```
uahp : ∀ P P' : hProp, (P -> P') -> (P' -> P) -> P = P'
funextfunax : ∀ (X Y : UU) (f g : X -> Y), (∀ x : X, f x = g x) -> f = g
funextempty : ∀ (X : UU) (f g : X -> ∅), f = g
```

andrejbauer:
Isn't `funextempty` a special case of `funextfunax`? You're probably keeping it separate for book-keeping purposes, yes?

mortberg:
Yes, and as you know they are all special cases of univalence. One reason that it is nice to keep them separate is that they all have different proofs in cubicaltt (this was probably not the original reason for keeping them separate in UniMath though). For instance funextfunax has a very short direct proofs, while the proof from univalence is a little longer, so in order to keep term sizes down when formalizing in cubicaltt it is good to use the direct proofs. The normal forms of these proofs are (propositionally) equal though.

PR #248, opened 2015-12-03
Cleaning set quot

mortberg:
When doing the set quotient exercise in cubicaltt I first had a look at the formalization in UniMath. To make it easier to read the proofs I cleaned them a bit (shorter lines, indentation and bullets, etc.).


benediktahrens:
ok w me

Issue #243, opened 2015-11-21
Subdirectories


vladimirias:
We have many files that are in the UniMath directory instead of a subdirectory of it. I suggest some subdirectory names:

1. Number_systems
for naturals, integers and rationals. Maybe it makes sense to put Dedekind there as well.
2. Combinatorics
for standard finite sets, finite sets and sequences (which would be better to call finite_sequences)
3. Functionalextensionality should probably be renamed into the Univalence_Axiom and put into Basics.

I don't have a good idea about the rest yet.

cathlelay:
The number_systems sub-directory seams to me a good idea.
Perhaps, p-adic numbers should also go in this sub-directory.

DanGrayson:
The only file there is the README:
```bash
$ pwd
/Users/dan/src/ProofChecking/UniMath/UniMath
$ ls -F
CategoryTheory/  Foundations/  PAdics/	  SubstitutionSystems/
Dedekind/	 Ktheory/      README.md  Tactics/
```

vladimirias:
I meant UniMath/UniMath/

DanGrayson:
I like the idea, with minor amendments to suggest: NumberSystems instead of Number_systems, UnivalenceAxiom instead of Univalence_Axiom, FiniteSequences instead of finite_sequences.

vladimirias:
Yes, it is better this way.

DanGrayson:
I'll also put Sets and Propositions in Basics and remove All.v.

benediktahrens:
All points seem to have been addressed, closing.

PR #252, opened 2015-12-05
Reorganization

DanGrayson:
this resolves issue #243


benediktahrens:
ok

vladimirias:
I suggest to put the files Associativity and Commutativity together and call the resulting file SequenceProducts.

DanGrayson:
Okay, but maybe it should resemble BinaryOperations more, so SequenceOperations?

vladimirias:
I am not sure that it is the right name. We do not introduce sequence operations there as a primitive notion, but only as iterated binary operations.

DanGrayson:
How about IteratedBinaryOperations?

On Sat, Dec 5, 2015 at 3:52 PM, Vladimir Voevodsky <notifications@github.com

vladimirias:
This sounds good.

DanGrayson:
The error travis found was due to a pull request being granted that didn't contain the same renaming in a new file.  All should be well now ...

PR #253, opened 2015-12-06
added substitution systems summary file

benediktahrens:
purpose of the added file is to link formalized theorems to their counterparts in a paper.
this file should not change name once it has been accepted, since it will be referred to in the paper.


DanGrayson:
How about explaining the purpose of the file in a comment in the file, along with
the desire for the name not to change?

benediktahrens:
On 12/06/2015 06:56 PM, Daniel R. Grayson wrote:
> How about explaining the purpose of the file in a comment in the file, along with
> the desire for the name not to change?

Good thinking, thanks.
See my updated PR.

Issue #254, opened 2015-12-07
definition of field


DanGrayson:
Here is our definition of "field".

```
Definition invpair (X:monoid) (x:X) := Σ x':X, ( x' * x = 1 ) × ( x * x' = 1 ).
Definition multinvpair ( X : rig ) ( x : X ) := invpair ( rigmultmonoid X ) x .
Definition isafield (X:commrng) := isnonzerorng X  ×  ∀ x:X, multinvpair X x ⨿ (x = 0).
Definition fld := Σ X, isafield X.
```

Are there any examples worthy of being considered as fields that satisfy only the weaker condition 
that ` x != 0 -> multinvpair X x ` ?

The construction of the field of fractions for an integral domain, as a result of this, requires the
additional hypothesis that the ring has decidable equality.  I can imagine that being counterintuitive
when we start formalizing commutative algebra, and we want to associate a field to each
prime ideal of a commutative ring.

I'm tempted to weaken the definition to allow more natural developments, even if there are
no interesting examples.

DanGrayson:
Here is a related comment.  In formalization of facts about integral domains
and fields, it is important to consider the multiplicative monoid of nonzero
elements.  But the proof that x!=0 is a proposition uses the axiom
`funextempty`, spoiling computability, so we will always want a proposition x≠0
to replace it that can be shown to be a proposition without using axioms.  When
equality is decidable, one is available, but in general, the development of the
real numbers with its apartness relation suggests that the right approach is
instead to seek a property x≠0 such that `(¬x≠0) <-> x=0`.  (With decidability,
one of these is always available, also.)

Currently the definition of "integral domain" is unsatisfying, as it uses
propositional truncation, thereby also introducing axioms:

```
Definition isintdom ( X : commrng ) :=
   dirprod ( isnonzerorng X )
           ( forall a1 a2 : X ,
	       paths ( a1 * a2 ) 0 -> hdisj ( eqset a1 0 ) ( eqset a2 0 ) ) .
```

Simpler looking would be:
```
Definition isintdom ( X : commrng ) := isnonzerorng X ×
	   ∀ x y, x ≠ 0 -> y ≠ 0 -> x*y ≠ 0.
```

If we adopt this approach, then our definition of "field" would be the same as
Catherine's definition of "constructive field", provided we add various
properties that say that the relation x≠y is compatible with addition and
multiplication.

Merging the two definitions seems worthwhile, at first glance.

vladimirias:
[...]

PR #257, opened 2015-12-10
Rational number computation

### LINKED ISSUE #244: computability of the rational numbers
My theory about why computations don't work well with rational numbers is that the following
definition uses `funextempty` (via `isapropneg`) unnecessarily:

```
Definition intdomnonzerosubmonoid ( X : intdom ) : @subabmonoids ( rngmultabmonoid X ) .
Proof . intros . split with ( fun x : X => hProppair _ ( isapropneg ( paths x 0 ) ) ) . split .
```

I will work on a replacement for this.

DanGrayson:
This resolves #244.

Note: this does not make 1/2 = 3/6 in Q provable by reflexivity.

It also renames "hsubtypes" to "subtype" and "submonoids" to "submonoid".  (The plural is
counterintuitive.)

I also introduce a way to use a proof of decidable equality to prove equations, and apply it to
hq in Dedekind:

```
Lemma hq2eq1plus1 : 2 = 1 + 1.
  Proof.
 -  rewrite <- hztohqand1, <- nattohzand1.
 -  now rewrite <- hztohqandplus, <- nattohzandplus.
 +  exact (confirm_pos (hqdeceq, 2, 1+1)).
  Qed.
```


DanGrayson:
Do people find these changes acceptable?

vladimirias:
Hi Dan,

I don’t find it acceptable.

I suggest that we call the new concepts using “ap” instead of “nonzero” and preserve the old definitions and lemmas about the non-zero objects in a separate file. I think that choosing one at the expense of another would make our library less useful in the future.

Vladimir.

DanGrayson:
You mean, you want to preserve the situation where 1+1=2 in hq is not true by definition?  And preserving that behavior somehow makes the library more useful?

DanGrayson:
Perhaps you think  you're commenting on https://github.com/UniMath/UniMath/issues/254 ?

vladimirias:
Let’s put this PR on hold for now. I need more time to read through the changes that you have made since the start of your fight with negation.

Vladimir.

DanGrayson:
Okay.

DanGrayson:
Re : "I suggest that we call the new concepts using “ap” instead of “nonzero”"

There is no new concept.  I've simply adopted a different implementation of an integer being nonzero, equivalent to the old one.  So calling it "ap" is not appropriate.

DanGrayson:
Perhaps we can decide this soon.

DanGrayson:
I've updated this pull request about rational numbers to remove the recent conflicts.

UniMath/Foundations/Basics/Sets.v
```diff
@@ -1191,15 +1260,37 @@ intros p1 p2 r12 q1 . apply ( ( pr1 ( pr2 is ) ) _ _ r12 q1 ) .
...
 Proof . intros . split with ( fun rp : carrier (funcomp (setquotpr R) P) => ( pr1 p ) ( pr1 rp ) ) .  apply ( iseqclassresrel R ( funcomp ( setquotpr R ) P ) _ ( pr2 ( pr1 p ) ) ) . intros x px .  set ( e := setquotl0 R _ ( carrierpair _ x px ) ) .  (* *) simpl in e . unfold funcomp . rewrite e . apply ( pr2 p ) . Defined .
 
-Definition tosubquot { X : UU } ( R : eqrel X ) ( P : hsubtypes ( setquot R ) ) : setquot ( resrel R ( funcomp ( setquotpr R ) P ) ) -> P .
+Definition tosubquot { X : UU } ( R : eqrel X ) ( P : subtype ( setquot R ) ) : setquot ( resrel R ( funcomp ( setquotpr R ) P ) ) -> P .
 Proof . intros X R P . assert ( int : isaset P ) . apply ( isasetsubset ( @pr1 _ P ) ) . apply ( setproperty ( setquotinset R ) ) . refine (isinclpr1carrier _) . apply ( setquotuniv _ ( hSetpair _ int ) ( fun xp => carrierpair P ( setquotpr R ( pr1 xp ) ) ( pr2 xp ) ) ) .  intros xp1 xp2 rp12 . apply ( invmaponpathsincl _ ( isinclpr1carrier P ) _ _ ) . simpl .  apply ( iscompsetquotpr ) . apply rp12 . Defined .
 
-Definition weqsubquot { X : UU } ( R : eqrel X ) ( P : hsubtypes ( setquot R ) ) : weq P ( setquot ( resrel R ( funcomp ( setquotpr R ) P ) ) ) .
+(* <<<<<<< HEAD *)
```
benediktahrens: There are commented-out merge conflict indicators here - no real problem, of course.

DanGrayson:
Speaking of old pull requests, what about this one?

vladimirias:
Dan, you have two different inequalities one is != and another one is, I guess, \neq.

It is a very large PR and I could not find where they are defined. Could you point me out please?

DanGrayson:
See https://github.com/UniMath/UniMath/pull/257/files#diff-6660fc0a8a1da41feed0c159301baefeR81 for the definition of `a \ne b` when a and b are rational.  For `a != b`, it's the older definition for any type, the negation of `a=b`.

UniMath/Foundations/NumberSystems/Integers.v
```diff
@@ -72,23 +72,21 @@ Definition hzdeceq : decrel hz := decrelpair isdecrelhzeq .
...
-Definition hzdecneq : decrel hz := decrelpair isdecrelhzneq .
+Open Local Scope hz_scope .
 
-(* Canonical Structure hzdecneq. *)
+(* [funextempty]-free replacement proposition for hzneq *)
 
-Definition hzboolneq := decreltobrel hzdecneq .
+Definition hzneq : neqReln hz := deceq_to_neqReln isdeceqhz.
 
+Notation " x ≠ y " := ( hzneq x y ) (at level 70, no associativity) : hz_scope.
```
DanGrayson: Oops, Vladimir, the link I gave in the conversation should have gone here.

vladimirias:
And where neqReln is defined?

DanGrayson:
See https://github.com/UniMath/UniMath/blob/9777dd3f6cb759c2211e4440a9a40bb1c21386d5/UniMath/Foundations/Basics/PartC.v#L100

vladimirias:
These definitions and the theory of complements that is based on them should be in PartB. The principle on which PartB and PartC are separated is that *all* of the results in PartC require funextempty. May be it should be stressed more in the opening comment to PartC.

DanGrayson:
Will do.  I'll add such a comment to the top of PartC and to README.

vladimirias:
While you are at it could you please remove the comment "The coconus behaves like the image of a function." from the line 854 of PartA. The coconus does not behave like the image.

DanGrayson:
Okay.

DanGrayson:
That comment was from Alex Kavvos.

vladimirias:
Maybe someone should have a look at his PR’s again.

DanGrayson:
Okay, I've done those things requested.   I've even moved a bit more from PartC to PartB.  However, since the main point of my pull request is to show how the results in PartC that use funextempty can be replaced by results that do not use it, I leave the new ones in place, side by side with the old ones.

Here's why the replacement is possible.  The only use of funextempty in PartC is to prove `isapropneg`, which says that a negation of a type is a proposition.  That gets used in lots of results having to do with complements of isolated points, because one needs to show that x != y is a proposition, but the main source of isolated points is from sets with decidable equality.  But once equality is decidable, it's easy to define a type logically equivalent to x != y that can be proved to be a proposition without using an axiom.

If this PR is accepted it would make sense to delete all the old results that no longer have any important application.  (The new results are side by side with them and have the suffix `_ne`.)

vladimirias:
Why the word “Definition” in PartB line 645 and similar words below it are not colored?

vladimirias:
> If this PR is accepted it would make sense to delete all the old results that no longer have any important application. (The new results are side by side with them and have the suffix _ne.)

I do not understand what you propose. Here is an example of a pair old definition/new definition:

> +Definition isolated ( T : UU ) := Σ t:T, isisolated _ t.
> +Definition isolated_ne ( T : UU ) (neq:neqReln T) := Σ t:T, isisolated_ne _ t (neq t).

the new definition requires a choice of an element neq in "neqReln T”. I understand that under the funextempty assumption there is a canonical element in neqReln for any T but adding this everywhere “isolated” is used would make the code much heavier with no gain at all.

What you did is very interesting. In some sense it is a little step from the UniMath in the direction of the George Gontier approach while preserving the UniMath spirit.

Instead of repeating it twice and so making any changes to it difficult I suggest that you put your development in a separate file (in Foundations) and when you want to use some lines from PartC there for illustration to use them commented out. This will free your code and open the possibility of developing it further.

Vladimir.

DanGrayson:
I think the comment should have said that it was the old definitino of
isdecprop.

But, no, the old definition of "isdecprop" as contractibility of X \coprod
\neg X was not as good as the new one, as
it could never be proven without the use of an axiom, such as
funextemptyAxiom.

One way to leave "isaprop" out of it, even in the new approach, would be to
posit the existence of a "complementary"
type Y so that X \coprod Y is contractible.  But to insist that the
complementary type be the negation is to
unnecessarily force users to invoke funextemptyAxiom.

DanGrayson:
Here's another way to put it.  The statement `iscontr (X \coprod Y)` is equivalent to the conjunction of four statements:

-  `X \coprod Y`
-  ` X -> Y -> empty `
- `isaprop X`
- `isaprop Y`

Any proof will essentially have to reduce to proving these four things, if it goes the usual way.

When Y is `neg X` then `isaprop Y` follows from `funextempty`, for any `Y`.  So including `isaprop Y` is redundant, since it's always true, and potentially prevents computations, since it forces the invocation of an axiom.

DanGrayson:
Re: " adding this everywhere “isolated” is used would make the code much heavier with no gain at all."

The gain is that one no longer needs to appeal to `funextempty`, and thus computations start to work.  It's a substantial gain.

vladimirias:
Dan, could you remind again this example with the dependent sum, the one that we discussed for a while.

DanGrayson:
I don't know what example you're trying to recall.

DanGrayson:
I don't understand what has happened to this pull request of mine -- it seems that part of it
has gotten incorporated into the master branch.  For example, the old definition
of Vladimir of `isdecprop`:
```
Definition isdecprop ( X : UU ) := iscontr ( coprod X ( neg X ) ) .
```
in PartC is now supplanted by the one included originally in my pull request:
```
Definition isdecprop (P:UU) := (P ⨿ ¬P) × isaprop P.
```
Also, many of the new functions ending with "_ne", such as `compl_ne`, are now present in
PartC.  However, some of the changes from my original pull request are not present on the master branch now, such as the simplification of Catherine's proof that 1+1=2 in the rationals or the renaming of `hsubtypes` to `subtype` everywhere.  I thought at first that I must have pushed my branch to UniMath master unintentionally in the process of making those additional changes or on one of those occasions when I was dealing with conflicts caused by acceptance of other pull requests, but if not all of those original changes of mine ended up on the master branch, then that can't be true.  I've tried using `gitk` to display the tree of commits that touch just PartC.v, but I don't understand the display, so I can't deduce what I must have odne.  So I'm just perplexed about what might have happened.  If I screwed it up somehow, I apologize.

(To make it highly unlikely that I would use `git` from the command line to push directly to the UniMath repository by accident, I have just changed the URL of that "remote" from `git@github.com:UniMath/UniMath` to `https://github.com/UniMath/UniMath`, for then git would demand a password from me when pushing.  That might be a good practice to recommend for those of us with write access.)

The current diffs for my pull request now show just a strange fraction of the changes proposed in the original pull request.

What would you like me to do?  It will be hard for you to consider the pull request without proper diffs.  I could try editing the master branch with the aim putting things back to an approximation of the way they were before my pull request, and submit that as a pull request.  (I don't see immediately how to do things in a automated way, since I don't understand what subset of changes got into the master branch.)

benediktahrens:
It looks like PR #240 incorporated that stuff into master.

DanGrayson:
Oh, well that explains it, thanks!

benediktahrens:
Can we decide on this PR soon?

benediktahrens:
Sorry, Dan, by merging another PR, I messed up this one again.
Would it be possible to split this PR into a part that people agree on (renaming hsubtypes into subtype, and maybe other stuff), and a part that is controversial?

DanGrayson:
Yes, that renaming was an overreach, since it means I always get conflicts.  I'll do as you suggest.

vladimirias:
I like hsubtype better. There hopefully will be an exact equality in the future and then there will be exact subtypes. hsubtype is not a sub-object in the model only a sub-object up to homotopy.

DanGrayson:
Oh, okay, good point.

vladimirias:
What is the reason for the change in l. 683 in BinaryOperations.v?

vladimirias:
I suggest that you separate all of the renamings from other changes and make a separate PR from them. The reason that this PR is stuck is because it contains so many different changes in different files.

DanGrayson:
Re: "What is the reason for the change in l. 683 in BinaryOperations.v?"

I assume you mean this one:

```
 -Definition ispartbinophrel { X : setwithbinop } ( S : hsubtypes X ) ( R : hrel X ) := dirprod ( Π a b c : X , S c -> R a b -> R ( op c a ) ( op c b ) ) ( Π a b c : X , S c -> R a b -> R ( op a c ) ( op b c ) ) .
 +Definition ispartbinophrel { X : setwithbinop } ( S : hsubtypes X ) ( R : hrel X ) := dirprod ( forall a b c : X , S c -> R a b -> R ( op c a ) ( op c b ) ) ( forall a b c : X , S c -> R a b -> R ( op a c ) ( op b c ) ) .
```

I looked into it.  What happened was that my commit 79ce273d8c changed "forall" to "Π" was pulled into master.  Then in my commit 489e2b173 on my branch I merged in "master" and resolved conflicts.  I guess I must have resolved a conflict on this line incorrectly.  The reason a conflict would have occurred is that on the same line I had changed 'hsubtypes' to 'subtype' on my branch.  That change was undone in the same commit on my branch in an attempt to simplify the PR.

I will simplify the PR further, as you suggest in your subsequent message.

DanGrayson:
Closing temporarily, so it doesn't keep running tests.

DanGrayson:
The purpose of this old pull request is to remove unnecessary uses of `funextemptyAxiom` from the construction of the rational numbers, so that some equalities between constants become true by definition.  Those proofs are slow, though, so it's better to use the decidability of equality to prove such equations.

DanGrayson:
This pull request was reopened a month ago, and no comments have been made.

benediktahrens:
I would love to comment on this PR, but: I don't understand it. 
In the comments, many things have been discussed, also things that have long been merged (e.g., new definition of `isdecprop`). Could you point me to the crucial change that this PR introduces?

DanGrayson:
The main point of this pull request is to make more terms in hq compute.  This is done by changing the definition of `hzneq` from
```
Definition hzneq ( x y : hz ) : hProp := hProppair ( neg ( x = y ) ) ( isapropneg _  )  .
```
to
```
Definition hzneq : neqReln hz := deceq_to_neqReln isdeceqhz.
```
Note that `isapropneg` uses an axiom:
```
Theorem isapropneg (X : UU) : isaprop (neg X).
Proof.
  intro. apply invproofirrelevance.
  intros x x'. apply (funextempty X x x').
Defined.
```
The new definition uses no axioms to show that the nonzeroness of the denominator is a proposition.   It does this by taking advantage of decidability of equality in `hz`.

It's important to notice that the new definition of `hzneq` is equivalent to the old one.  This is not a new mathematically, it's just implemented in a different way.

The main advantage of the code is that computations no longer get stuck on axioms, so that now some identities can be proven by reflexivity, such as this new one in the Tests file:
```
+  Goal nattohq 1%nat + nattohq 1%nat = nattohq 2.
 +    reflexivity.                  (* fixed, 11 seconds *)
 +  Defined.                        (* fixed, 11 seconds *)
```
But it's slow.  It's still preferable to use the "confirm" tactics I wrote previously, as here in RealNumber/Prelim:
```
Lemma hq2eq1plus1 : 2 = 1 + 1.
Proof.
  confirm_equal isdeceqhq.
Qed.
```

The main interest in the code, from my point of view, is that it shows yet another example where elements in a set quotient are computable, even though `setquot` uses axioms, if one takes care of what goes in the front end.

benediktahrens:
Could the problem with the original version
```
Definition hzneq ( x y : hz ) : hProp := hProppair ( neg ( x = y ) ) ( isapropneg _  )  .
```
be avoided by not bundling the type and the proof that the type is a proposition together into one type? I.e., by mostly only using `neg (x = y)` and only referring to it being a proposition when that fact is indeed needed?

DanGrayson:
No.  The problem is that the proof that it's a proposition gets bundled up into homotopies in equivalences that are used later, and those homotopies get promoted into more prominent locations by various lemmas involving equivalences.  I looked into it thoroughly more than a year ago, and recorded the details somewhere ...

benediktahrens:
lgtm

DanGrayson:
Benedikt, here's a link to my previous comment explaining how the use of funextempty blocks computation:

https://github.com/UniMath/UniMath/issues/227#issue-114592100

benediktahrens:
On 02/22/2017 08:32 PM, Daniel R. Grayson wrote:
> Benedikt, here's a link to my previous comment explaining how the use of funextempty blocks computation:
> 
> https://github.com/UniMath/UniMath/issues/227#issue-114592100

Interesting! Maybe it is worth finding a better spot for this
explanation than in a difficult to find issue comment.

DanGrayson:
I've added it to the wiki: https://github.com/UniMath/UniMath/wiki/Topics-of-discussion

vladimirias:
> Interesting! Maybe it is worth finding a better spot for this
> explanation than in a difficult to find issue comment.

Right.

DanGrayson:
Closing, see https://github.com/UniMath/UniMath/pull/592#issuecomment-281873412

Issue #229, opened 2015-11-02
Slowness when formalizing coproducts as special colimits


benediktahrens:
more precisely, when implementing the fact that coproducts lift to functor categories (from the target category) via a general result of lifting of colimits, instead of a direct proof.

For code, see https://github.com/benediktahrens/UniMath/tree/coprod-lift-functors and observe slowness in the first files that use this lifting, SubstitutionSystems.v and SumOfSignatures.v .
The lifting of coproducts is formalized in FunctorsPointwiseCoproducts.v.

DanGrayson:
The first slow point in SubstitutionSystems.v is in the proof of `bracket_property`, where it
takes 3 seconds to unify the types of the two sides of the equation.  That's the same
point where it thinks forever if we replace your coproducts with mine.  One type is
```
[C, C, hs] ⟦ (pre_composition_functor_data C C C hs hs (U Z)) (Id_H `T), `T ⟧
```
and the other type is
```
[C, C, hs] ⟦ CoproductObject ([C, C] hs) (CPEndC (U Z) ((θ_source C hs H) (`T ⊗ Z))), U p T ⟧
```

Is it possible to code things so they turn out to be the same?

benediktahrens:
On 12/14/2015 04:14 PM, Daniel R. Grayson wrote:
> The first slow point in SubstitutionSystems.v is in the proof of `bracket_property`, where it
> takes 3 seconds to unify the types of the two sides of the equation.  That's the same
> point where it thinks forever if we replace your coproducts with mine.  One type is
> ```
> [C, C, hs] ⟦ (pre_composition_functor_data C C C hs hs (U Z)) (Id_H `T), `T ⟧
> ```
> and the other type is
> ```
> [C, C, hs] ⟦ CoproductObject ([C, C] hs) (CPEndC (U Z) ((θ_source C hs H) (`T ⊗ Z))), U p T ⟧
> ```
>
> Is it possible to code things so they turn out to be the same?


The targets should be convertible.



I am less sure about the sources.


The first should reduce to

(Id_H `T) \bu (U Z)

where \bu is composition of functors ("U Z applied before (Id_H `T)"),
which should reduce to

( Id + H (`T) ) \bu (U Z)

which maybe does not reduce to

U Z + ( H (`T) \bu (U Z) )

which is the source of the second thing.


However, on every object those functors should be (definitionally) the same.


Why are you asking that question?

DanGrayson:
[...]

PR #258, opened 2015-12-11
Limits and Colimits

DanGrayson:
This addresses the slowness issue #229.  It contains various facts about representable
functors, including
```
Theorem functorPrecategoryBinaryProduct (B C:Precategory) :
  hasBinaryProducts C -> hasBinaryProducts [B,C].
```
, which is proved from a general theorem about family of representable functors, bypassing
colimits (which can be dealt with in the same way):
```
Lemma bifunctor_assoc_repn {B C:Precategory} (X : [B, [C^op,SET]]) :
  (∀ b, Representation (X ◾ b)) -> Representation (bifunctor_assoc X).
```
No serious slowness is observed.


mortberg:
Did you try to adapt SubstitutionSystems to use this definition? As far as I remember we had no problem with slowness before trying to use our formalization of colimits in functor categories there. 

Also why did you end has_homsets_opp with Defined instead of Qed?

benediktahrens:
To test this, it seems best to have a file that introduces the same definitions and lemmas as in Coproducts.v .

Am 11. Dezember 2015 10:01:59 EST, schrieb "Anders Mörtberg" <notifications@github.com>:

vladimirias:
LGTM.

Dan, do you have, for a functor eta:C->D and an object X of C the morphism Yo(X) -> eta\circ Yo(eta(X)) and a lemma saying that it is an equivalence (isomorphism) when eta is fully faithful? I think we may also need that it is natural in X (here Yo is the Yoneda to contravariant functors and \circ is in the diagrammatic order).

DanGrayson:
Re : "Did you try to adapt SubstitutionSystems to use this definition? " 

No, I haven't looked at the code.

Re: "Also why did you end has_homsets_opp with Defined instead of Qed?"

I don't usually use Qed unless I have a reason to.

Re: "To test this, it seems best to have a file that introduces the same definitions and lemmas as in Coproducts.v ."

I'll take a look.

Re: "do you have, for a functor eta:C->D and an object X of C the morphism Yo(X) -> eta\circ Yo(eta(X)) and a lemma saying that it is an equivalence (isomorphism) when eta is fully faithful?"

I have this theorem, which ought to suffice and says that representability and embeddings:
```
Definition embeddingRepresentability {C D:Precategory}
           {i:C==>D} (emb:fully_faithful i) {Y:D^op==>SET} (s:Representation Y) :
           (Σ c, universalObject s = i c) -> Representation (Y □ functor_opp i).
```
One may use it, for example, to show that abelian groups have products by showing first that groups have products and then showing that the group product of a family of abelian groups is abelian.

PR #261, opened 2015-12-14
upstream auxiliary content of SubstitutionSystems

benediktahrens:
contents of SubstSys/Aux upstreamed into CT and Foundations


UniMath/Foundations/Basics/PartA.v
```diff
@@ -280,6 +280,17 @@ Proof.
   intros. destruct ex. destruct ey. apply idpath.
 Defined.
 
+Lemma dirprodeq (A B : UU) (ab ab' : A × B) :
```
DanGrayson: This is the same as dirprod_eq, already in PartA.

benediktahrens: On 12/14/2015 07:00 AM, Daniel R. Grayson wrote:
>> @@ -280,6 +280,17 @@ Proof.
>>    intros. destruct ex. destruct ey. apply idpath.
>>  Defined.
>>  
>> +Lemma dirprodeq (A B : UU) (ab ab' : A × B) :
> 
> This is the same as dirprod_eq, already in PartA.

I can't find that one in my version of PartA.

DanGrayson: You're right, it's just on my branch.

DanGrayson:
looks good to me

PR #262, opened 2015-12-14
universal property for opposite of Rezk completion

benediktahrens:
by showing that eta opp is ff and ess surj


DanGrayson:
okay with me

PR #264, opened 2015-12-16
Add Jason-Gross' coq-tools as a submodule

DanGrayson:
The Makefile now has a convenient way to use it to isolate a bug


Makefile
```diff
@@ -134,6 +134,28 @@ doc: $(GLOBFILES) $(VFILES)
 	--with-header $(ENHANCEDDOCSOURCE)/header.html $(VFILES)
 	sed -i'.bk' -f $(ENHANCEDDOCSOURCE)/proofs-toggle.sed $(ENHANCEDDOCTARGET)/*html
 
+# Jason Gross' coq-tools bug isolator:
+# Override this on the command line; this is just an example:
```
benediktahrens: Could you add your example call here, please?

benediktahrens: Oh, I see that that is also part of PR #266 . Is this PR superseded completely by #266 ?

DanGrayson: Either way, thanks for merging.

Issue #111, opened 2015-06-29
total2 as record, use of "primitive projections" ?


benediktahrens:
As far as I recall, total2 was, at some point, defined as a record type. At the moment, it is defined as an inductive type. (The comments in uuu.v still say that a record type is used.)
What was the reason for changing from record to an ordinary inductive? 

If we switched back to total2 being defined as a record type, we could use the "Set Primitive Projections" option which, as @JasonGross told me, would provide significant speedup of compilation (at least it did for HoTT/HoTT).
It would also provide eta for dependent pairs, which might or might not be something we want.

vladimirias:
Here is from an email I wrote in Oct. 2014 when I changed from "record” to ‘Inductive” and removed “primitive projections”:

> Just a comment - the whole of UniMath (including RezkCompletion) compiles on my laptop 
> in about 5min30sec with primitive projections and for about 6min30sec without.
> 


I did it then with the trunk.  We should check how it is now with 8.5 before deciding anything.

benediktahrens:
On 06/29/2015 02:35 PM, Vladimir Voevodsky wrote:
> Here is from an email I wrote in Oct. 2014 when I changed from
> "record” to ‘Inductive” and removed “primitive projections”:
>
>> Just a comment - the whole of UniMath (including RezkCompletion)
>> compiles on my laptop in about 5min30sec with primitive projections
>> and for about 6min30sec without.
>>
>
>
> I did it then with the trunk.  We should check how it is now with 8.5
> before deciding anything.

It is great to have these measures, thanks. It looks like the benefit 
would be small, then, judging from those numbers.

The context of my question is the following: in a development on top of 
UniMath [1], we are hitting a point where Qed (or Defined) takes very 
long: we have measured 34 minutes for one of them, before starting to 
refactor the code significantly. Now the longest Qed takes 11 minutes.

I am thus looking for ways to make UniMath and [1] faster - the 
primitive projections are one thing I should look at, is what Jason 
Gross said.
There are a few other things I should test, including Opacifying lemmas 
temporarily.


Testing primitive projections will need a few changes to UniMath - at 
least compilation fails at some point in uu0a for the first time.

Maybe what I would like to suggest is *not* having primitive projections 
in the master branch, but in another branch for testing purposes, with 
at least a significant part of UniMath compiling there.

[1] https://github.com/benediktahrens/substitution

DanGrayson:
What about turning off sharing?  Would that speed up your substitution
project?

[...]

PR #269, opened 2015-12-21
Primitive projections

DanGrayson:
We change the definition of "total2" from an inductive definition to a record and make some trivial
changes to get UniMath to work with primitive projections.  (One advantage of turning on primitive
projections is eta expansion for pairs.)  Most changes involve replacing "apply" by "refine", which
works better.  Some changes involve working around failures of "fold".

We don't completely succeed in working around all the problems, but the Foundations package
compiles completely.  The current obstacle is in the Ktheory package.

It's worthwhile checking this in now to incorporate the changes into our development, allowing
others to experiment with primitive projections, and relieving me of the need to maintain the branch,
for the time being.  To experiment, just uncomment the command "Set Primitive Projections" in
Preamble.v.

This represents partial progress on #111


benediktahrens:
You say 'Most changes involve replacing "apply" by "refine", which
works better.'. What does "better" mean here? You mean that one works while the other doesn't?

DanGrayson:
Yes, "refine" works better because it works more often.

benediktahrens:
On 12/23/2015 08:23 AM, Daniel R. Grayson wrote:
> Yes, "refine" works better because it works more often.

So all the replacements of "apply H" by "refine (H _ _ _ _ )" that you
did in this PR were necessary?

mortberg:
Is it just me or isn't it very painful to have to write "refine (H _ _ _ _ _)" instead of simply "apply H"? It is very annoying when Coq develops in the wrong direction... 

Maybe we could make a tactic "rapply H" that tries "refine H | refine (H _) | refine (H _ _) | ..." so that one doesn't have to guess the number of implicit arguments all the time?

DanGrayson:
Yes, every replacement of apply by refine was necessary.  "Apply" is sort of buggy.

In "refine (H _ _ _)" the blanks represent explicit arguments.  But often "apply" does work.

DanGrayson:
Unless we switch to primitive projections, none of you will have the pain of having to write
"refine (H _ _ _ _ _ )", for "apply" will probably continue to work more often.

benediktahrens:
On 12/23/2015 03:20 PM, Daniel R. Grayson wrote:
> Yes, every replacement of apply by refine was necessary.  "Apply" is
> sort of buggy.

Is there a bug report on that? Is Matthieu aware of that problem?

DanGrayson:
It might have been Matthieu who told me...

mortberg:
Oops, I meant explicit arguments instead of implicits. Anyway I still really don't like to always write all these underscores...

benediktahrens:
What should we do with this PR? To summarize what I have understood:
If we pull it, then we might end up having to use refine instead of apply in many places.
But if we ever want to have eta for pairs, there is no way around this change. 
Is that correct?

DanGrayson:
No, that's not right.  Primitive projections are not turned on by this PR, because I couldn't fix everything, so you can keep coding new code as you did before, with refine instead of apply; I think I didn't say that clearly.  Notice that all the tests have passed.

But yes, if we ever want to use eta for pairs, these changes must be made, so it seems highly desirable to go ahead, so as to incorporate the work already done in that direction.

By the way, Matthieu gave some further hints about speeding things up when primitive projections are on, which I will ultimately implement.

benediktahrens:
On 01/19/2016 09:56 AM, Daniel R. Grayson wrote:
> No, that's not right.  Primitive projections are not turned on by
> this PR, because I couldn't fix everything, so you can keep coding
> new code as you did before, with refine instead of apply; I think I
> didn't say that clearly.  

That's whay I meant: even without primitive projections, but only using
a Record for total2 instead of an Inductive, we need to replace some
"apply" with "refine".
Right?

> Notice that all the tests have passed.

There's a merge conflict now, unfortunately.

> But yes, if we ever want to use eta for pairs, these changes must be
> made, so it seems highly desirable to go ahead, so as to incorporate
> the work already done in that direction.
> 
> By the way, Matthieu gave some further hints about speeding things up
> when primitive projections are on, which I will ultimately
> implement.

Brilliant. If you share those hints with us, we can maybe help.

DanGrayson:
No, this PR leaves "total2" as an inductive type.

To see Matthieu's suggestions, click on the second bug report link in my comment at https://github.com/UniMath/UniMath/issues/111#issuecomment-155904433

DanGrayson:
I fixed the merge conflicts.

PR #270, opened 2015-12-22
Update coq

DanGrayson:
We update Coq to the latest 8.5 version after the announcement of
v8.5 RC 1.

This update implements "simple refine", which is the recommended way to
get the new behavior of the "refine" tactic, so we replace "unshelve
refine" by "simple refine" everywhere except in those cases where it
doesn't work.  (It was just a few days ago that we had to replace almost
all of our uses of "refine" by "unshelve refine".)

### DIFF
--- a/UniMath/CategoryTheory/category_hset.v
+++ b/UniMath/CategoryTheory/category_hset.v
@@ -373,7 +373,7 @@ End from_colim.
 
 Definition colimCoconeHSET : cocone D colimHSET.
 Proof.
-unshelve refine (mk_cocone _ _).
+simple refine (mk_cocone _ _).
 - now apply injections.
 - abstract (intros u v e;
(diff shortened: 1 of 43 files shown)


DanGrayson:
See https://coq.inria.fr/news/127.html

benediktahrens:
ok

DanGrayson:
Here's a comment from the Coq CHANGES file, that we should take into account, since we
want to avoid using "match":
```
- The command "Record foo ..." does not generate induction principles
  (foo_rect, foo_rec, foo_ind) anymore by default (feature wish
  #2693). The command "Variant foo ..." does not either. A flag
  "Set/Unset Nonrecursive Elimination Schemes" allows changing this.
  The tactic "induction" on a "Record" or a "Variant" is now actually
  doing "destruct".
```

PR #280, opened 2016-01-25
Cocontinuous functors

mortberg:
Definition of cocontinuity of a functor and some examples


UniMath/CategoryTheory/category_hset.v
```diff
@@ -274,247 +216,18 @@ Defined.
...
+exists empty; apply isasetempty.
 Defined.
 
-Local Definition from_cobase_rel : hrel cobase.
+Definition unitHSET : HSET.
 Proof.
-intros x x'; exists (from_cobase x = from_cobase x').
-now apply setproperty.
+exists unit.
+abstract (apply isasetaprop, isapropifcontr, iscontrunit).
```
DanGrayson: just use isasetunit here

DanGrayson:
Modulo my suggestion above, it looks good to me.

mortberg:
I fixed your comment now

PR #281, opened 2016-01-25
some experiments on ff functors

benediktahrens:
provides two terms of type

∀ (C D : precategory) (F : functor C D),
fully_faithful F -> ∀ a b : C, iso a b ≃ iso (F a) (F b)

called  weq_ff_functor_on_iso_weqbandf and  weq_ff_functor_on_iso, respectively.

The former does not evaluate as expected, see 
weq_ff_functor_on_iso_weqbandf_compute.


benediktahrens:
Travis error due to changes in travis build vms, see #282 for opting into using travis legacy structure.

DanGrayson:
The changes look good to me once the check succeeds.

Issue #278, opened 2016-01-24
equivalence obtained using weqbandf does not compute 


benediktahrens:
[ all of that in file functor_categories ]
In PR #277 I prove an equivalence 
weq (iso a b) (iso (F a) (F b))
using weqbandf, called weq_ff_functor_on_iso. 
An equivalence of same type has previously been defined more directly.
In lemma "ff_is_inclusion_on_objects" the direct definition works well, whereas the definition using weqbandf does not: indeed, the inverse seems to not compute as I had hoped.
If anybody has a thought on that, I'd love to hear it.

DanGrayson:
I'd be happy to take a look if you would provide code for the two computations, one that works
well and one that doesn't.

benediktahrens:
On 01/24/2016 03:22 AM, Daniel R. Grayson wrote:
> I'd be happy to take a look if you would provide code for the two computations, one that works
> well and one that doesn't.

Right. See PR #281, and more specifically compare

weq_ff_functor_on_iso_compute

and

weq_ff_functor_on_iso_weqbandf_compute .

PR #286, opened 2016-01-27
improve computability of weqbandf

DanGrayson:
resolves https://github.com/UniMath/UniMath/issues/278


benediktahrens:
Now that's a nice example of what I have been trying to explain, that we should avoid destruct on pairs.

benediktahrens:
lgtm

mortberg:
Or that Coq really should have eta for pairs...

DanGrayson:
Which it will when we switch to using primitive projections.

mortberg:
Yes! :+1:

PR #293, opened 2016-02-01
Compilation of sub-directories with CoqMakefile

clarus:
Scripts to build each sub-directory of `UniMath` using the [CoqMakefile](https://coq.inria.fr/refman/Reference-Manual017.html#Makefile) generator.

The aim is to be able to build each of the six sub-directories independently, in order to make six OPAM packages `coq-unimath-category-theory`, `coq-unimath-dedekind`, `coq-unimath-foundations`, `coq-unimath-ktheory`, `coq-unimath-substitution-systems` and `coq-unimath-tactics`. The `coq_makefile` command generates a Makefile which infers the right build order according to the file dependencies and a `make install` rule.

There is also the alternative to have just one package `coq-unimath` and a single build system to build all UniMath at once.


benediktahrens:
On 02/01/2016 05:17 PM, Guillaume Claret wrote:
> 
> There is also the alternative to have just one package `coq-unimath`
> and a single build system to build all UniMath at once.

The different packages cover different areas of mathematics, so that I
could imagine someone wanting to install only a specific package (and
its dependencies).

Is there the concept of a meta package in OPAM, one that depends on a
selection of (real) packages and installs all of them?
I.e., is there the possibility of making a `coq-unimath` package the
installation of which causes the six OPAM packages to be installed?

clarus:
There are no specific concepts of meta package in OPAM. However, it is possible to make a package with dependencies but a "void" source code, in order to make a meta `coq-unimath` package.

benediktahrens:
Great. I'll pull your patch tomorrow, Tuesday, if there are no more comments.

clarus:
OK, thanks!

DanGrayson:
No, I'm not sure a good idea.  The current makefile already has targets for building just a single package.  For example, one can use "make Ktheory", which will also make the appropriate prerequisites from the other packages.  

The current pull request contains duplication of code, with identical lines scattered in various files.  That will increase the work in maintaining the code and keeping all the copies identical.

Why is it a good idea to have multiple OPAM packages instead of one?  I think the idea of UniMath is to have it be a unified repository of mathematics.  @vladimirias may wish to comment on that.

Why should our OPAM package be called coq-unimath instead of unimath?

vladimirias:
Thank you Dan. I agree with your comments.

As for the name of the package I also think it should be called unimath. If we have unimath implemented in a different language then we can rename the package but for now there is only one unimath.

Also, UniMath is under development and this development may and will change the subdirectories at the UniMath/UniMath level. This may be the strongest argument not to introduce OPAM packages for subdirectories.

benediktahrens:
On 02/02/2016 04:27 AM, Daniel R. Grayson wrote:

> No, I'm not sure a good idea.  The current makefile already has 
> targets for building just a single package.  For example, one can use
> "make Ktheory", which will also make the appropriate > prerequisites
> from the other packages.

The current makefile does not have targets to *install* individual
packages, which is the point of OPAM.

> The current pull request contains duplication of code, with identical
> lines scattered in various files.  That will increase the work in
> maintaining the code and keeping all the copies identical.

Considered the amount of code involved, that is a very academic argument.

> Why is it a good idea to have multiple OPAM packages instead of one?
>  I think the idea of UniMath is to have it be a unified repository of
>  mathematics.  @vladimirias may wish to comment on that.

In the Debian repositories, (almost?) all packages should be compatible.
That does not mean that I do install all of them.

UniMath will grow, and people won't always want to install all the packages.

For instance, I personally would not compile all the packages if I
wouldn't have the need to use the "install" make target, which depends
on compilation of all packages.

> Why should our OPAM package be called coq-unimath instead of 
> unimath?

OPAM is an Ocaml package manager, not just a Coq package manager. Their
policy seems to be to prefix every Coq-related package with "coq-", see
http://coq.io/opam/

benediktahrens:
On 02/02/2016 07:24 AM, Vladimir Voevodsky wrote:
> Thank you Dan. I agree with your comments.
> 
> As for the name of the package I also think it should be called
> unimath. If we have unimath implemented in a different language then
> we can rename the package but for now there is only one unimath.

See my answer to Dan's comment.

> Also, UniMath is under development and this development may and will
> change the subdirectories at the UniMath/UniMath level. This may be
> the strongest argument not to introduce OPAM packages for
> subdirectories.

OPAM is release-based, and points to a git tag (that is, a certain git
commit that has been given a name). It does not point to the latest
commit in, e.g., the master branch.

Every once in a while, one would create a new tag and update the
pointers in the OPAM packages, and introduce new OPAM packages in case
there is a new one in UniMath compared to the previous tag.

This means that people who install UniMath via OPAM would not get the
latest (and therefore the best) UniMath.
On the plus side, it is easier to use/install a UniMath package
alongside an existing Coq installation in this way.

One can keep the difference between latest UniMath and the OPAM packages
small by frequently updating the OPAM packages.

vladimirias:
I am not sure what we gain from it. Managing, in addition to all that we have to think about, OPAM packages is extra work for us. What do we get for this work? Extra visibility? We have plenty of visibility. This is not something that we should be willing to pay for by extra work.

benediktahrens:
On 02/02/2016 09:40 AM, Vladimir Voevodsky wrote:
> I am not sure what we gain from it. Managing, in addition to all that
> we have to think about, OPAM packages is extra work for us. What do
> we get for this work? Extra visibility? We have plenty of visibility.
> This is not something that we should be willing to pay for by extra
> work.

Well, clarus has agreed to do the work. That's why he filed that pull
request.

If you do not want clarus' code in the main repository, then I suggest
clarus build OPAM packages from his own fork of UniMath --- UniMath is
free software, after all.

vladimirias:
Maybe this is the best solution for now. Dan?

DanGrayson:
It depends on what the meaning of "this" is, but I'm happy to defer to you on this, Vladimir.

benediktahrens:
The UniMath team would like to keep packaging artefacts out of the repository.
I am thus closing this PR.

clarus:
OK, that's fine to me, I will make a package for all UniMath.

The general hope with OPAM for Coq is to help people make math libraries which depend on other libraries, by copying the model of open source softwares using packages managers for programming languages.

PR #296, opened 2016-02-05
more pleasant way to use refine

benediktahrens:
- a tactic to build pairs
- a tactic to avoid having to give the underscores to "simple refine (foo...)"


DanGrayson:
Okay by me.

PR #297, opened 2016-02-05
transparent assert

benediktahrens:
Introduce tactic "transparent assert", as in

transparent assert (t : T).

Note that type T sometimes needs parentheses as in

transparent assert (t : (T)).

See the patch for such an example.
I am not sure if this is due to T containing a colon or to something else.


DanGrayson:
I think its due to having another colon, as in 'assert (t:unit:UU).'

Looks good to me.

JasonGross:
> I think its due to having another colon

I don't think so; I think `transparent assert (H : 1 = 1)` also fails, if I recall correctly.  I'm not sure exactly how tactic notation precedences work.

PR #298, opened 2016-02-05
a few more lemmas about pullbacks

benediktahrens:
- square in functor cat is pb if it is pointwise pb
- comparing two morphisms into a pb object
- diagonal morphisms in pb are equivalent to sections of projection


mortberg:
LGTM

Issue #284, opened 2016-01-25
TAGS file in wrong order


DanGrayson:
The files named in the TAGS file are now in the wrong order.  Perhaps something changed about
the way VFILES is set in the coq makefile.  Fixing it will involve changing the order of the packages
in the Makefile and changing the order of the lines in each file `UniMath/*/.package/files`.

benediktahrens:
This issue seems to have been fixed with PR #300, thus closing.

PR #300, opened 2016-02-12
reorder lists of files and packages

DanGrayson:
For some reason the coq makefile maker has changed, so now we get to
list packages and files in forward order, and they will appear in
the correct order in the TAGS file.  This is important when using tags
search and find in emacs.

This resolves issue #284.


benediktahrens:
Ok!

PR #303, opened 2016-02-17
move files from hss to ct package

benediktahrens:
files, that have been written for SubstitutionSystems, but can be of more general use, are moved from SubstSys to CT


DanGrayson:
Fine with me, except this will conflict with pull request 300, which should be done first.

DanGrayson:
Fine with me, but it will conflict with #300, which should be pulled first.

benediktahrens:
Updated the PR

PR #307, opened 2016-02-26
yoneda iso is natural (2nd direction), two-pb lemma

benediktahrens:
two simple but useful lemmas

### DIFF
--- a/UniMath/CategoryTheory/limits/more_on_pullbacks.v
+++ b/UniMath/CategoryTheory/limits/more_on_pullbacks.v
@@ -113,8 +113,65 @@ Proof.
                 apply idpath] ).
 Defined.
 
+(**  Diagram for next lemma
 
+              i'           k
+         d' -------> d --------> c
+         |           |           |
+       h'|           |h          | g
+         v           v           v
+         b'--------> b --------> a
+              i            f
+*)
 
+Lemma isPullback_two_pullback
+     (b' d' : C) (h' : C⟦d', b'⟧)
+     (i : C⟦b', b⟧) (i' : C⟦d', d⟧)
+     (Houter : h' ;; (i ;; f) = (i' ;; k) ;; g)
+     (Houterpb : isPullback _ _ _ _ Houter)
+     (Hinner : h ;; f = k ;; g)
+     (Hinnerpb : isPullback _ _ _ _ Hinner)
+     (Hleft : h' ;; i = i' ;; h)
+     :isPullback _ _ _ _ Hleft.
+Proof.
+  apply (mk_isPullback).
+  intros e x y Hxy.
+  mkpair.
+  - mkpair.
+    use (PullbackArrow (mk_Pullback _ _ _ _ _ _ Houterpb)).
+    + apply x.
+    + apply (y ;; k).
+    + abstract (rewrite assoc; rewrite Hxy;
+                repeat rewrite <- assoc; rewrite Hinner; apply idpath).
+    + abstract (
+      split ;
+      [
+      apply (PullbackArrow_PullbackPr1 (mk_Pullback (i ;; f) g d' h' (i' ;; k) Houter Houterpb))
+      |
+      idtac
+      ];
+      apply (MorphismsIntoPullbackEqual Hinnerpb);
+      [rewrite <- assoc;
+        match goal with |[ |- ?KK ;; ( _ ;; _ ) = _ ] => pathvia (KK ;; (h' ;; i)) end;
+        [ apply maponpaths; apply (!Hleft) |
+          rewrite assoc;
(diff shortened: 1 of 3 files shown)


UniMath/CategoryTheory/limits/more_on_pullbacks.v
```diff
@@ -113,8 +113,65 @@ Proof.
...
+       h'|           |h          | g
+         v           v           v
+         b'--------> b --------> a
+              i            f
+*)
 
+Lemma isPullback_two_pullback
+     (b' d' : C) (h' : C⟦d', b'⟧)
+     (i : C⟦b', b⟧) (i' : C⟦d', d⟧)
+     (Houter : h' ;; (i ;; f) = (i' ;; k) ;; g)
```
DanGrayson: It would be better not to provide Houter as an argument, but to deduce it from the other two.

DanGrayson:
Modulo the one comment, okay with me.

UniMath/CategoryTheory/limits/more_on_pullbacks.v
benediktahrens: On 02/26/2016 03:25 PM, Daniel R. Grayson wrote:
>>
>> +Lemma isPullback_two_pullback
>> +     (b' d' : C) (h' : C⟦d', b'⟧)
>> +     (i : C⟦b', b⟧) (i' : C⟦d', d⟧)
>> +     (Houter : h' ;; (i ;; f) = (i' ;; k) ;; g)
>
> It would be better not to provide Houter as an argument, but to deduce it from the other two.

I disagree. I prefer my more general form, in particular for the use I 
have in mind, where Houter (or something similar) is given as a hypothesis.

If Houter were a side-condition instead of an argument, like in 
set-theoretic foundations, you'd be right, of course.

DanGrayson: In what sense is your formulation more general?  Since C has homsets, Houter turns out to be equal to the proof deduced from the other two, so it's simply redundant.

benediktahrens: On 02/26/2016 04:14 PM, Daniel R. Grayson wrote:
>>
>> +Lemma isPullback_two_pullback +     (b' d' : C) (h' : C⟦d', b'⟧) +
>> (i : C⟦b', b⟧) (i' : C⟦d', d⟧) +     (Houter : h' ;; (i ;; f) = (i'
>> ;; k) ;; g)
>
> In what sense is your formulation more general?  Since C has homsets,
> Houter turns out to be equal to the proof deduced from the other two,
> so it's simply redundant.

I said "more general form", not "more general formulation". To make it 
more clear, I should have said "shape" or whatever.

It is not more general in the mathematical sense, but it is more 
generally applicable, since I do not force the hypothesis Houterpb to be 
of a certain shape which is computed from the goal.

When doing backwards reasoning, I often find more general shapes more 
useful to work with.

benediktahrens: On 02/26/2016 04:14 PM, Daniel R. Grayson wrote:
>>
>> +Lemma isPullback_two_pullback
>> +     (b' d' : C) (h' : C⟦d', b'⟧)
>> +     (i : C⟦b', b⟧) (i' : C⟦d', d⟧)
>> +     (Houter : h' ;; (i ;; f) = (i' ;; k) ;; g)
>
> In what sense is your formulation more general?  Since C has homsets, Houter turns out to be equal to the proof deduced from the other two, so it's simply redundant.

Put differently, since Houter is not only an argument to the 
construction, but also determines the type of another argument to the 
construction, I want some flexibility here, as to choose the type of 
that other argument as is convenient for me, without having to rewrite 
with the homs-are-sets assumption.

DanGrayson: Since the definition of isPullback does not refer to the final argument, simple
unfolding will show that `isPullback _ _ _ _ Houter1` and `isPullback _ _ _ _ Houter2` are
identical types.  That might imply there is no problem with presenting your function with
a proof Houterpb based on a different Houter.

benediktahrens: On 02/26/2016 04:35 PM, Daniel R. Grayson wrote:
>>
>> +Lemma isPullback_two_pullback
>> +     (b' d' : C) (h' : C⟦d', b'⟧)
>> +     (i : C⟦b', b⟧) (i' : C⟦d', d⟧)
>> +     (Houter : h' ;; (i ;; f) = (i' ;; k) ;; g)
>
> Since the definition of isPullback does not refer to the final argument, simple
> unfolding will show that `isPullback _ _ _ _ Houter1` and `isPullback _ _ _ _ Houter2` are
> identical types.  That might imply there is no problem with presenting your function with
> a proof Houterpb based on a different Houter.

Good thinking. PR updated.

DanGrayson:
ok with me

PR #310, opened 2016-03-05
Add package "Folds"

benediktahrens:
a formalization of categories à la FOLDS, following notes by Mike Shulman

- tex files by Mike have been removed, but remain in the history
- package has been added to travis config file


DanGrayson:
Fine with me, if the test works.

benediktahrens:
Any more comments on including or not the Folds code as a regular package?

DanGrayson:
Fine with me.

PR #314, opened 2016-03-08
better transparency/opacification for pullbacks

benediktahrens:
not actually needed anywhere yet, but certainly an improvement


DanGrayson:
fine with me

mortberg:
LGTM

PR #318, opened 2016-03-09
On adjunctions and cocontinuity

mortberg:
Proofs that some more functors are (omega) cocontinuous.

Proof that left adjoints preserve colimits.

The equivalence between Voevodsky's definition of lists (as iterated products) and our lists over sets (as the initial algebra of the list functor).


benediktahrens:
lgtm

DanGrayson:
lgtm

PR #326, opened 2016-03-18
remove unicode arrow '⇒' for morphisms in CT

benediktahrens:
and remove the notation from the file CT/UnicodeNotations.v

The same notation '⇒' is still used in package Folds, but since this is a leaf package for now, that's less disturbing.


DanGrayson:
g

Issue #322, opened 2016-03-17
[CategoryTheory.UnicodeNotations] incompatible with [Ktheory.Precategories], due to notation level/associativity clashes


peterlefanulumsdaine:
It is currently impossible to simultaneously import these two libraries together:
```
Require Import UniMath.CategoryTheory.UnicodeNotations.
Require Import UniMath.Ktheory.Precategories.
```
fails, with error
```
Error: Notation _ ;; _ is already defined at level 50 with arguments
at level 50, at next level while it is now required to be at level 50
with arguments at next level, at next level.
```

In other words, they define this notation with different associativities.  Importing them in the other order gives a similar problem with a different notation; I guess there are other points they clash on.

If desired, I’m happy to try to prepare a pull request making them compatible — I’ll probably have time to do so this weekend, or early next week. However, I don’t know the conventions of (much of) the library very well, so if the developers would prefer me to leave this to someone who knows the library better, I’ll stay off it!

DanGrayson:
I'll take a look at it...

benediktahrens:
That error message is certainly amusing when one has some time to kill, but for everyone else the Coq developers should implement a more useful message.
Have you checked if there is a bug report for this message? How did you find out what the problem is?

JasonGross:
The issue is that Coq's parser cannot handle the same notation being declared at different levels / with different associativity; this seems to be an inherent limitation in the camlpX parsers.  You have the following declarations of `;;`:
```bash
$ git grep 'Notation' | grep ';;'
UniMath/CategoryTheory/UnicodeNotations.v:Notation "f ;; g" := (compose f g) (at level 50, format "f  ;;  g", left associativity).
UniMath/CategoryTheory/equivalences.v:Local Notation "f ;; g" := (compose f g) (at level 50, format "f  ;;  g").
UniMath/CategoryTheory/equivalences_lemmas.v:Local Notation "f ;; g" := (compose f g) (at level 50, format "f  ;;  g").
UniMath/CategoryTheory/functor_categories.v:Local Notation "f ;; g" := (compose f g) (at level 50, format "f  ;;  g").
UniMath/CategoryTheory/limits/cones.v:Local Notation "f ;; g" := (compose f g)(at level 50).
UniMath/CategoryTheory/limits/products_from_pullbacks.v:Local Notation "f ;; g" := (compose f g)(at level 50).
[...]

Issue #325, opened 2016-03-18
Officializing "→" ?


benediktahrens:
The following notation is declared in CT/UnicodeNotations:

Notation "x → y" := (x -> y)
  (at level 90, y at level 200, right associativity): type_scope.
(* written \to in Agda input method *)

It is currently used in files in CT, Ktheory, SubstitutionSystems, and Folds.

Should the notation be declared globally in Foundations/Basics/Preamble ?

vladimirias:
What is it the notation for?

benediktahrens:
On 03/18/2016 10:47 AM, Vladimir Voevodsky wrote:
> What is it the notation for?

It is the notation for the function type.

JasonGross:
Alternatively, you can `Require Export Coq.Unicode.Utf8.` to get that notation; it's in the Coq standard library.

benediktahrens:
On 03/25/2016 02:31 PM, Jason Gross wrote:
> Alternatively, you can `Require Export Coq.Unicode.Utf8.` to get that
> notation; it's in the Coq standard library.

We only compile very little of the Coq standard library, and this 
particular file is not among those we compile.

vladimirias:
I think we only compile or at least only import Coq.init

benediktahrens:
On 03/25/2016 03:03 PM, Vladimir Voevodsky wrote:
> I think we only compile or at least only import Coq.init

Yes, we only make the target "states", which depends on 
Coq.Init.Prelude, which itself depends on the files in Coq.Init.

 From what I understand, the module Prelude is loaded by default when 
Coq is started, and the flag -noinit would need to be used to avoid that.

benediktahrens:
Done in #330 , closing.

PR #330, opened 2016-03-23
fix notations

DanGrayson:
Resolves issue https://github.com/UniMath/UniMath/issues/322 .

We changed the level of Notation "x → y" from 90 to 99 to match the core
of Coq, and we removed the Notation "f ;; g" from the Ktheory package.

I don't understand exactly what provokes this incomptability -- there
are certainly some bugs in the way Coq takes care of such things.  In
particular, we aren't reserving any notations yet -- it's a good
suggestion, especially if notation reservation within a given scope
would work.

### DIFF
--- a/UniMath/Ktheory/AbelianGroup.v
+++ b/UniMath/Ktheory/AbelianGroup.v
@@ -538,3 +538,9 @@ Module Category.
   End DirectSum.
 
 End Category.
+
+(*
+Local Variables:
+compile-command: "make -C ../.. TAGS UniMath/Ktheory/AbelianGroup.vo"
+End:
+*)
--- a/UniMath/CategoryTheory/UnicodeNotations.v
+++ b/UniMath/CategoryTheory/UnicodeNotations.v
@@ -3,11 +3,6 @@ Require Export UniMath.Foundations.Basics.PartD
                UniMath.CategoryTheory.precategories
                UniMath.CategoryTheory.functor_categories.
 
-Notation "x → y" := (x -> y)
-  (at level 99, y at level 200, right associativity): type_scope.
-(* written \to in Agda input method *)
-(* the level comes from sub/coq/theories/Unicode/Utf8_core.v *)
-
 Notation "a --> b" := (precategory_morphisms a b)(at level 50).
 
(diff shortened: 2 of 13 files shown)


JasonGross:
Reserved notations are not specific to scopes.  Parsing rules are not specific to scopes; parsing happens before scope resolution happens, I believe, and it's only after parsing that notations are resolved.  For any given symbols/notation, it should always appear at the same level, with the same associativity, etc, regardless of scopes.

Note that reserving notations only impacts parsing, not interpretation of notations.

DanGrayson:
Thanks, we will have to face that eventually.  I'll reopen the issue so further discussion can happen there.

UniMath/Ktheory/AbelianGroup.v
```diff
@@ -538,3 +538,9 @@ Module Category.
   End DirectSum.
 
 End Category.
+
+(*
+Local Variables:
```
benediktahrens: Are those Emacs instructions here on purpose?

DanGrayson: Yes, I like being able to compile just the file I'm working on.  But it may be inconvenient 
for others to deal with, if they want to edit my files.

benediktahrens: On 03/24/2016 10:19 AM, Daniel R. Grayson wrote:
>> @@ -538,3 +538,9 @@ Module Category. End DirectSum.
>> 
>> End Category. + +(* +Local Variables:
> 
> Yes, I like being able to compile just the file I'm working on.  But
> it may be inconvenient for others to deal with, if they want to edit
> my files.

It does not cause any inconvenience for me, so I am fine with it.

benediktahrens:
Wouldn't this be the perfect occasion to put the arrow "→" in Basics.Preamble, as suggested in #325 ?

DanGrayson:
will do ...

DanGrayson:
done ...

benediktahrens:
On 03/25/2016 03:42 PM, Daniel R. Grayson wrote:
> done ...

lgtm

I hadn't noticed that you had used that arrow as a notation for
morphisms in a category.

Maybe we can find a Unicode arrow as notation for morphisms that is
easily distinguishable from → and ⇒ .
Anyway, that is a different issue.

PR #331, opened 2016-03-24
Links for latex documentation

benediktahrens:
- intra-document links to definitions
- color syntax highlighting
- bibliography at the end of the pdf instead of at end of Basics.Preamble


DanGrayson:
Body only -- good!

PR #339, opened 2016-04-01
remove accidential dependency on univalence axiom from yoneda

benediktahrens:


DanGrayson:
g

andrejbauer:
Actually, they installed smilies in the upper right corner of a message, in case `g` is too long a message for you...

benediktahrens:
On 04/02/2016 04:29 AM, Andrej Bauer wrote:
> Actually, they installed smilies in the upper right corner of a
> message, in case `g` is too long a message for you...

There do not seem to be email notifications for smilies, so an actual
message is better, imo.

PR #344, opened 2016-04-04
induced limit morphism is iso iff source is limit

benediktahrens:


benediktahrens:
We wait until Anders' future PR is merged in.

DanGrayson:
g

PR #348, opened 2016-04-08
The product bifunctor is omega cocontinuous

mortberg:


UniMath/CategoryTheory/polynomialfunctors.v
```diff
@@ -301,4 +299,281 @@ Defined.
...
 
+(* The functor "* : C^2 -> C" is omega cocont *)
+Section binprod_functor.
+
+Variables (C : precategory) (PC : Products C) (hsC : has_homsets C).
+Variables (hE : has_exponentials PC).
+
+Definition fun_lt (cAB : chain (product_precategory C C)) :
+  forall i j, i < j ->
+              C ⟦ ProductObject C (PC (pr1 (pr1 cAB i)) (pr2 (dob cAB j))),
```
benediktahrens: use access function dob instead of pr1
it might also be good to define obj_1 and obj_2 as projections for objects in a product precategory.

DanGrayson:
g

PR #349, opened 2016-04-18
Filters and Topology

cathlelay:
There is the first pull request about analysis.
I formalize the notion of "filter" which is a generalization of neighborhoods and topological spaces.
It is based on the Bourbaki's book "General Topology: Chapters 1-4".


andrejbauer:
Did you have to remove any uses of excluded middle in comparison to Boubraki? It looks like everything so far ought to work well without excluded middle, except possibly things that involve complements.

andrejbauer:
A small comment: you use `infinite_union` the general form of union that need not actually be infinite. It might be better to just call it `union` (or `general_union`, but I do not like that).

DanGrayson:
It might be better to organize our files by mathematical topic ("Foundations/Topology") rather than
by the name of the author of the textbook where we get the results ("Bourbaki").  After all, we may
not formalize the entire book, and we may get results from other books that are closely related to
these results.

vladimirias:
1. Maybe it would be better to call the package BourbakiTopology.

2. Please comply with the 100 characters per line restriction for new additions to the UniMath!

vladimirias:
union_all ?

vladimirias:
There has been a suggestion to take algebra and other such topics from the foundations leaving in foundations only the basics. I think that it makes sense. In any case maybe it is better not to add topology, filters, uniform spaces etc. that Cath is working on to foundations.

cathlelay:
Except in FilterBase, I didn't need yet hypotheses which can be qualified
of "classical".

2016-04-18 15:58 GMT-04:00 Andrej Bauer <notifications@github.com>:

vladimirias:
Can you point out where in FilterBase?

DanGrayson:
Better than BourbakiTopology would be Bourbaki/Topology

vladimirias:
Then probably Topology/Bourbaki

cathlelay:
As the Bourbaki will also contain other notions of analysis, such as Riemann integral, I rename it into Analysis.

vladimirias:
I would prefer to have topology separate from analysis.

andrejbauer:
@cathlelay could you point to the exact place where classical logic is used? The basic theory of filters is intuitionistic, there should not be any uses of classical logic, unless perhaps one of the definitions is too classical.

cathlelay:
Finally, there is no classical logic.
I thought than one of the hypotheses needed classical logic, but it is not the case.

vladimirias:
Great!

andrejbauer:
Very good.

benediktahrens:
What is the status of this PR? I have not read it, but it seems that people were happy with it in the end.

DanGrayson:
There seems to be quite a bit of new good-looking code about point set topology, which someone should read carefully.

cathlelay:
Due to my work on uniform spaces (in a future pull request), I needed to define a topology using the set of neighborhood. Then, many definitions of topologies changed for more useful versions.

vladimirias:
I think that "complements" should be renamed, for example in "miscellaneous". The word "complements" in the context of topology strongly associates with the complements to subsets.

vladimirias:
I think that "filtermap" need to be renamed. Into "filter_im"?

vladimirias:
Similarly with the FilterMap.

vladimirias:
Do you need FilterMap_incr? Does not it follow from PreFilterMap_incr through the coercion from filters to pre-filters?

cathlelay:
I renamed Topology.Complements into Topology.Miscellaneous, filtermap into filterim, FilterMap into FilterIm.
The theorem FilterMap_incr (now FilterIm_incr) is useful, despite the coercion, when we search theorems about Filter ou FilterIm.

cathlelay:
This PR is really old.
Should I close it ?

tpannila:
No, do not close. Nobody has given any reason why this PR should not be merged.

DanGrayson:
Vladimir, what about this pull request?  Does it look okay?

vladimirias:
Yes. I want to read new files in more detail, but I am unlikely to find time for it until the beginning of the term but I can also do it after it was merged.

PR #352, opened 2016-04-19
The lambda calculus as the initial algebra of Lam(F) = Id + F * F + F o option

mortberg:
This pull request finishes the construction of the lambda calculus. It contains:

- A sufficient condition for when a functor is a left adjoint (adjunction_from_partial)
- Covariant Yoneda functor
- Construction of exponentials in [C,HSET]
- A proof that the precomposition functor is cocontinuous. This is done by constructing a right adjoint to "A^K : [C,A] -> [M,A], F |-> F o K" which in turn uses the construction of right Kan extensions when A has limits.


DanGrayson:
g

vladimirias:
LGTM

PR #353, opened 2016-04-27
Towards linking omega-cocontinuous functors and substitution systems

mortberg:
This pull request contains various results towards linking Benedikt's and my work on omega-cocontinuous functors to the work on substitution systems. In particular it contains:

- Packaging and notations for omega-cocontinuous functors and a cleanup of the lists example.
- Another example implementing binary trees as the initial algebra of F(X) = 1 + A * X * X
- Lifting of initial objects to functor categories
- Instantiation of the necessary hypotheses in substitution systems using the results from the work on omega-cocontinuous functors. 

There are some issues that needs to be resolved related to compilation times of substitution systems. I had to change product_functor to product_of_functors and coproduct_functor to sum_of_functors in substitution systems in order to make it possible to use the results on omega-cocontinuous functors, however this seems to lead to very increased compilation times.

### DIFF
--- a/UniMath/SubstitutionSystems/Lam.v
+++ b/UniMath/SubstitutionSystems/Lam.v
@@ -53,7 +53,54 @@ Arguments θ_target {_ _} _ .
 Arguments θ_Strength1 {_ _ _} _ .
 Arguments θ_Strength2 {_ _ _} _ .
 
+Section set_instances.
 
+Require Import UniMath.CategoryTheory.limits.graphs.colimits.
+Require Import UniMath.CategoryTheory.exponentials.
+Require Import UniMath.CategoryTheory.category_hset.
+Require Import UniMath.CategoryTheory.category_hset_structures.
+Require Import UniMath.CategoryTheory.chains.
+Require Import UniMath.CategoryTheory.polynomialfunctors.
+
+Let Lam_S : Signature _ _ := Lam_Sig HSET has_homsets_HSET TerminalHSET CoproductsHSET ProductsHSET.
+
+Local Notation "'EndHSET'":= ([HSET, HSET, has_homsets_HSET]) .
+
+Let hsEndC : has_homsets EndHSET := functor_category_has_homsets _ _ has_homsets_HSET.
+
+Lemma Lam_Initial_HSET : Initial (precategory_FunctorAlg (Id_H _ _ CoproductsHSET Lam_S) hsEndC).
+Proof.
+use colimAlgInitial.
+- unfold Id_H, Const_plus_H.
+  apply is_omega_cocont_sum_of_functors.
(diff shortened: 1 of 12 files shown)


mortberg:
I should add that before the issues with compilation times has been resolved I don't think this PR should be merged, but by opening a PR it is easier to see what changes I've made and Benedikt can push patches to it.

vladimirias:
The file that only proves some properties of the omega cocontinuous functors should not be called “polynomial functors”.

There is a general notion of a polynomial functor in an lccc that we will need in the future and we will need this filename.

(BTW Functors of the form C0 + C1\times Id + … + Cn\times Id^n should probably be called… small-polynomial since even on hSet a general polynomial functor
I_f(X)=Hom_B(A, A\times X) for f:A -> B is not necessarily of this form)

UniMath/SubstitutionSystems/Lam.v
```diff
@@ -53,7 +53,54 @@ Arguments θ_target {_ _} _ .
 Arguments θ_Strength1 {_ _ _} _ .
 Arguments θ_Strength2 {_ _ _} _ .
 
+Section set_instances.
```
benediktahrens: Could this section go into a separate file, please? I would like to keep changes to existing files in SubstSys minimal.

benediktahrens:
From what I understand, there are two changes that affect Lam.v and hence could be responsible for the slowdown:
- replacement of `coproduct_functor` by `sum_of_functors`
- replacement of the manual definition of F \mapsto F o option to one using the precomposition functor
Can you confirm that it is the former that causes the slowdown?

benediktahrens:
In case it is indeed the switch to `sum_of_functors` that causes the slowdown, I suggest two things to try, in the given order:

1. replace `coproduct_functor` by `sum_of_functors` everywhere. it might be that with this change, Coq does not need to compute that those two functors agree judgmentally pointwise

2. prove that `coproduct_functor F G` is omega-cocontinuous whenever `F` and `G` are, by showing that it is equal to `sum_of_functors F G`. Those two being pointwise convertible, that should be easy to prove.

The second way messes up any computational behaviour that is still there.

benediktahrens:
I also suggest doing the experiments in a separate branch and in a separate PR - this one is full of noise, making it difficult to see the changes that are relevant to the problem.
I.e., please make a new branch where you only replace `coproduct_functor` by `sum_of_functors` in SumOfSignatures.v.

mortberg:
Thanks for the comments! I'll close this PR and split it into two now. One PR will contain everything not related to the linking of substsys and cocont (I'll also rename polynomialfunctors to cocontfunctors in this PR) and one containing the linking. This way it will be easier to localize where the slowdown in compilation time is coming from.

PR #355, opened 2016-05-03
Finish the linking of SubstSys and Cocont

mortberg:
This PR provides the connection between SubstSys and Cocont, in particular it contains:

- Proofs that more functors are omega cocont, in particular (co)product_functor

- Instantiation of the hypotheses of InitialHSS with Lam for HSET (LamHSET.v)

- Definition of the product of two Signatures, in particular proof of strength laws for the product (ProductOfSignatures.v)

- Construction of a Signature in the case when the functor is precomposition with a functor G. This is done by constructing a family of simpler distributive laws called delta (i.e. we construct something simpler than the theta in Signature for this special case). (currently at the top of SigToMonad.v, where should this be put?)

- Instantiations of the delta family when G = option and G = F^n

- Finally, by combining all of this, we get a function from lists of lists of natural numbers to a monad on Set obtained from the initial heterogeneous substitution system. See the end of SigToMonad.v for an example where this function is applied to the signature for the lambda calculus (i.e. [[0,0],[1]).

### DIFF
--- a/UniMath/CategoryTheory/cocontfunctors.v
+++ b/UniMath/CategoryTheory/cocontfunctors.v
@@ -6,6 +6,7 @@ This file contains proofs that the following functors are
 - Constant functor: F_x : C -> D, c |-> x
 - Identity functor
 - Composition of omega-cocontinuous functors
+- Iteration of omega-cocontinuous functors: F^n : C -> C
 - Pairing of omega-cocont functors (F,G) : A * B -> C * D, (x,y) |-> (F x,G y)
 - Delta functor: C -> C^2, x |-> (x,x)
 - Binary coproduct functor: C^2 -> C, (x,y) |-> x + y
@@ -105,6 +106,19 @@ Defined.
 Definition omega_cocont_functor_composite {C D E : precategory}
   (hsE : has_homsets E) (F : omega_cocont_functor C D) (G : omega_cocont_functor D E) :
   omega_cocont_functor C E := tpair _ _ (is_omega_cocont_functor_composite hsE _ _ (pr2 F) (pr2 G)).
+
+(* Functor iteration preserves omega cocontinuity *)
+Lemma is_omega_cocont_iter_functor {C : precategory} (hsC : has_homsets C)
+  (F : functor C C) (hF : is_omega_cocont F) n : is_omega_cocont (iter_functor F n).
+Proof.
+induction n as [|n IH]; simpl.
+- apply (is_omega_cocont_functor_identity _ hsC).
+- apply (is_omega_cocont_functor_composite hsC _ _ IH hF).
+Defined.
+
+Definition omega_cocont_iter_functor {C : precategory} (hsC : has_homsets C)
+  (F : omega_cocont_functor C C) n : omega_cocont_functor C C :=
+  tpair _ _ (is_omega_cocont_iter_functor hsC _ (pr2 F) n).
 
 (* A pair of functors (F,G) : A * B -> C * D is omega_cocont if F and G are *)
 Section pair_functor.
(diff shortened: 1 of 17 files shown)


UniMath/CategoryTheory/cocontfunctors.v
```diff
@@ -888,13 +923,19 @@ Notation "'Id'" := (omega_cocont_functor_identity _ has_homsets_HSET) :
...
     cocont_functor_hset_scope.
 
 Notation "F + G" :=
-  (omega_cocont_sum_of_functors _ _ ProductsHSET CoproductsHSET
-     has_homsets_HSET has_homsets_HSET _ _ (pr2 F) (pr2 G)) :
-    cocont_functor_hset_scope.
+  (omega_cocont_coproduct_of_functors _ _ ProductsHSET CoproductsHSET
+     has_homsets_HSET has_homsets_HSET F G) : cocont_functor_hset_scope.
+
+(* omega_cocont_coproduct_functor has worse computational behavior and
```
benediktahrens: Can you expand on that comment? Do you mean "worse than omega_cocont_coproduct_of_functors"?

UniMath/SubstitutionSystems/SigToMonad.v
```diff
@@ -0,0 +1,640 @@
+(**
+
+From signatures to monads
+
+Written by: Anders Mörtberg, 2016
```
benediktahrens: mention Ralph's notes here

benediktahrens:
In SubstitutionSystems/SigToMonad, from line 460 on, some of the lemmas and constructions are done for HSET, but seem to be doable just as well for any good base category C.
Could you generalize those without much work, and instantiate only later, for the test case of the lambda calculus? (Even that can be done for any good category, but that would be stretching it.)

benediktahrens:
As a test of correctness, can you derive the usual constructors and and (simple) elimination principle for the lambda calculus - just as you did for lists? 
This does not have anything to do with monads and SubstSys, but would go well in the file CT/lambdacalculus.

UniMath/SubstitutionSystems/SigToMonad.v
```diff
@@ -0,0 +1,640 @@
...
+
+Require Import UniMath.SubstitutionSystems.LamHSET.
+
+Let Lam_S : Signature HSET has_homsets_HSET :=
+  Lam_Sig HSET has_homsets_HSET TerminalHSET CoproductsHSET ProductsHSET.
+
+Goal (pr1 Lam_S = pr1 (SigToSignature LamSig)).
+now apply idpath.
+Abort.
+
```
benediktahrens: Also, here it would be a good sanity check to have the usual constructors and (simple) eliminator for the lambda calculus obtained from [0,0], [1].

UniMath/SubstitutionSystems/SigToMonad.v
```diff
@@ -0,0 +1,640 @@
...
+Require Import UniMath.CategoryTheory.ProductPrecategory.
+Require Import UniMath.SubstitutionSystems.Signatures.
+Require Import UniMath.CategoryTheory.limits.FunctorsPointwiseCoproduct.
+Require Import UniMath.CategoryTheory.limits.FunctorsPointwiseProduct.
+Require Import UniMath.CategoryTheory.EndofunctorsMonoidal.
+Require Import UniMath.CategoryTheory.Monads.
+Require Import UniMath.SubstitutionSystems.SumOfSignatures.
+Require Import UniMath.SubstitutionSystems.ProductOfSignatures.
+Require Import UniMath.SubstitutionSystems.SubstitutionSystems.
+Require Import UniMath.SubstitutionSystems.LamSignature.
```
benediktahrens: I'd prefer this file not to depend on files implementing examples. Could you put the example stuff in a separate file?

benediktahrens:
There are two things called `option_functor`:

- `CategoryTheory/lambdacalculus.v:Definition option_functor : [HSET,HSET,has_homsets_HSET].'
- `SubstitutionSystems/LamSignature.v:Definition option_functor: functor C C := coproduct_functor _ _ CC (constant_functor _ _  one) (functor_identity C).`

Can you unify the two and put it in a reasonable place? I suggest the file where coproduct_functor is defined.

mortberg:
This series of patches adresses all of Benedikt's comments.

benediktahrens:
Looks very good to me!

PR #356, opened 2016-06-01
Generalize the notion of signatures

mortberg:
This PR generalizes the signatures from lists of lists to sets of lists. In particular it contains:

- Generalization of binary (co)products to (co)products indexed by an arbitrary type. 
- Proofs that these (co)products lifts to functor categories
- Proof that generalized versions of some functors (delta, (co)product, pair_functor) are omega cocontinuous
- Generalization of sum of signatures
- Generalization of the notion of signature (GenSig) and the construction of a monad from a GenSig

### DIFF
--- a/UniMath/CategoryTheory/ArbitraryProductPrecategory.v
+++ b/UniMath/CategoryTheory/ArbitraryProductPrecategory.v
@@ -5,6 +5,7 @@ Anders Mörtberg
 
 2016
 
+Definition of the general product category
 
 ************************************************************)
 
@@ -60,23 +61,6 @@ Qed.
 
 End arbitrary_product_precategory.
 
-(** Objects and morphisms in the product precategory of two precategories *)
-(* Definition prodcatpair {C D : precategory} (X : C) (Y : D) : product_precategory C D. *)
-(* Proof. *)
-(*   exists X. *)
-(*   exact Y. *)
-(* Defined. *)
-
-(* Local Notation "A ⊗ B" := (prodcatpair A B) (at level 10). *)
-
-(* Definition prodcatmor {C D : precategory} {X X' : C} {Z Z' : D} (α : X --> X') (β : Z --> Z') *)
-(*   : X ⊗ Z --> X' ⊗ Z'. *)
-(* Proof. *)
-(*   exists α. *)
-(*   exact β. *)
-(* Defined. *)
-
-(* Define pairs of functors and functors from pr1 and pr2 *)
 Section functors.
 
 Definition arbitrary_pair_functor_data (I : UU) {A B : precategory}
--- a/UniMath/CategoryTheory/ArbitraryProductPrecategory.v
+++ b/UniMath/CategoryTheory/ArbitraryProductPrecategory.v
@@ -1,0 +1,131 @@
+
+(** **********************************************************
+
+Anders Mörtberg
+
+2016
+
+
+************************************************************)
+
+Require Import UniMath.Foundations.Basics.PartD.
+
+Require Import UniMath.CategoryTheory.precategories.
+Require Import UniMath.CategoryTheory.functor_categories.
+Require Import UniMath.CategoryTheory.UnicodeNotations.
+
+Local Notation "# F" := (functor_on_morphisms F)(at level 3).
+
+Section arbitrary_product_precategory.
+
+Variable I : UU.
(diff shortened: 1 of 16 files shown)


vladimirias:
Great! Can we now have, as an example, the signature of the Martin-Lof type theory MLTT79 (see https://en.wikipedia.org/wiki/Intuitionistic_type_theory <https://en.wikipedia.org/wiki/Intuitionistic_type_theory>)?

UniMath/CategoryTheory/EquivalencesExamples.v
```diff
@@ -1,18 +1,32 @@
+(**
+
+This file contains some adjunctions:
+
+- The delta_functor is left adjoint to binproduct_functor
+- The bincoproduct_functor left adjoint to delta_functor
```
benediktahrens: Please update the description of the contents in the header.

UniMath/CategoryTheory/ArbitraryProductPrecategory.v
```diff
@@ -0,0 +1,133 @@
...
+Definition has_homsets_arbitrary_product_precategory (hsC : has_homsets C) :
+  has_homsets arbitrary_product_precategory.
+Proof.
+intros a b; simpl.
+apply impred_isaset; intro i; apply hsC.
+Qed.
+
+End arbitrary_product_precategory.
+
+(** Objects and morphisms in the product precategory of two precategories *)
```
benediktahrens: Please delete those comments.

UniMath/CategoryTheory/cocontfunctors.v
```diff
@@ -259,6 +268,115 @@ Defined.
 
 End pair_functor.
 
+(* A family of functor F^I : A^I -> B^I is omega_cocont if each F_i is *)
```
benediktahrens: Please use  "(**" so that the comment shows up in HTML.
Similar in other places.

UniMath/SubstitutionSystems/GenSigToMonad.v
```diff
@@ -0,0 +1,114 @@
+(**
+
+From general signatures to monads
```
benediktahrens: Could you add the definition of "general signature" in the header, plz?

benediktahrens:
LGTM.
I suggest the following renaming:
Product -> BinProduct
ArbitraryProduct -> Product
That mis-naming is entirely my fault. Due to its size (it'll touch a lot of code), I suggest doing this in a separate PR once the present PR is merged.

mortberg:
Yes! I'll do it in a separate PR as soon as this one has been merged.

--
Anders

On Thu, Jun 2, 2016 at 5:30 AM, Vladimir Voevodsky <notifications@github.com

mortberg:
This patch adresses Benedikt's comments

UniMath/SubstitutionSystems/GenSigToMonad.v
```diff
@@ -0,0 +1,119 @@
...
+(* Initial (Id + functor(S)) *)
+(* |-> *)
+(* I:= Initial (HSS(func(S), \theta) *)
+(* |-> *)
+(* M := Monad_from_HSS(I)    # *)
+Section GenSigToMonad.
+
+Variables (I : UU) (HI : isdeceq I).
+
+Definition GenSig : UU := I -> list nat.
```
benediktahrens: As per my comment on PR #357 , GenSig should have the indexing type and the decidability as a field, not as a parameter (and should be called BindingSignature).
This is because we want to have a map of types
BindingSignature -> HSS-Signature 
(where HSS-Signature is what is simply called Signature at the moment).
We also want a map
BindingSignature -> Sigma(C : category)(Initial(C))
which maps a binding signature to its category of representations and an initial object therein.

UniMath/SubstitutionSystems/GenSigToMonad.v
```diff
@@ -0,0 +1,119 @@
...
+
+(* [[nat]] to Signature *)
```
benediktahrens: This comment is misleading, you are not discussing [[nat]] here.

UniMath/SubstitutionSystems/GenSigToMonad.v
vladimirias: Sounds good.

PR #357, opened 2016-06-03
MLTT79

mortberg:
Definition of the signature of MLTT79 from which we then construct a monad


UniMath/SubstitutionSystems/MLTT79.v
```diff
@@ -0,0 +1,388 @@
...
+Defined.
+
+Definition has_homsets_HSET2 : has_homsets HSET2.
+Proof.
+apply functor_category_has_homsets.
+Defined.
+
+(* Convenient function for combining decidability proofs and
+   signatures *)
+Definition SumGenSig {A B : UU} (p1 : isdeceq A × GenSig A)
```
benediktahrens: I suggest defining "BindingSignature" to be a triple:

- a type A
- decidable equality on that type
- a map A -> list(nat)

SumSig then is a map of type
BindingSignature -> BindingSignature -> BindingSignature
as a (binary) sum should be.

UniMath/CategoryTheory/lists.v
```diff
@@ -280,6 +280,12 @@ Defined.
 Definition foldr_list {A B : UU} (f : A -> B -> B) (b : B) : list A -> B :=
   list_ind A (fun _ => B) b (fun a _ b' => f a b').
 
+Definition length_list {A : UU} : list A -> nat :=
+  foldr_list (fun _ x => S x) 0.
+
```
andrejbauer: Out of curiosity, is there a reason for have `S` η-expanded to `fun _ => S x`?

mortberg: Nope, good catch!

UniMath/SubstitutionSystems/MLTT79.v
mortberg: Done!

UniMath/CategoryTheory/lists.v
DanGrayson: `(fun _ x => S x)` is not the eta expansion of S.

mortberg: True, but I had eta expanded S under the binder, so it is now simply ``fun _ => S``.

DanGrayson: Ah, right!

benediktahrens: There is nothing wrong with expanding functions. It sometimes increases readability. Before Coq got eta conversion for functions, artificially expanding functions increased chances that things became convertible.

UniMath/SubstitutionSystems/GenSigToMonad.v
```diff
@@ -52,6 +52,25 @@ Require Import UniMath.CategoryTheory.HorizontalComposition.
 Local Notation "[ C , D , hs ]" := (functor_precategory C D hs).
 Local Notation "'HSET2'":= ([HSET, HSET, has_homsets_HSET]) .
 
+Section move_upstream.
```
benediktahrens: The right moment to move this upstream is probably now.

vladimirias: As well as lemmas about the deceq of coproduct and unit. In fact, deceq of the unit may exist in basics (deceq of nat certainly does).

vladimirias:
I don't understand FinSigConstructors . Shouldn't it be just the constant function corresponding to [ ]?

vladimirias:
It would be good at the top of the file to refer specifically to p.158 of the "Constructive mathematics ...".

The table describing the syntax in the comment is correct in the sense that it corresponds to the syntax described by Martin-Lof on p.158.

We should also say that this syntax contains mistakes in the sense that in order to obtain the expected behavior of the type theory more arguments need to be added to some of the operations e.g. J should contain the argument corresponding to the type family. 

I think we should provide the signature for the correct syntax with a note describing the syntax originally chosen by Martin-Lof and the necessary corrections in the comment. The correct syntax can be read from the inference rules given on pp. 166 - 173.

mortberg:
I fixed all comments now except for Vladimir's comment about adding more arguments to some operations. The reason is that I'm not sure how many arguments to add and where. For instance, do we need to add the types to the application as in Streicher's book?

PR #358, opened 2016-06-06
Remove nested match from `onefiber`

peterlefanulumsdaine:
(as per email discussion with VV)

Alternative version, expunging match, induction, etc entirely from `onefiber`: https://github.com/UniMath/UniMath/compare/master...peterlefanulumsdaine:onefiber-reduced


benediktahrens:
LGTM

benediktahrens:
Is there a reason not to use the "induction" tactic, which also produces subterms of the form "coprod_rect...."?
Alternatively, if one doesn't like the fact that the induction tactic does more than merely applying the eliminator, one could introduce a notation, say, "case ..." for "refine coprod_rect...".
 
I think that a case distinction should look a bit better/simpler than "refine coprod_rect...".

vladimirias:
Let's merge the alternative variant (without "match"es) and then think about Benedikt's suggestion.

PR #361, opened 2016-06-12
Added zero objects

tpannila:
Hi.

I added zero objects in two ways, using limits and colimits, and directly, just
like initial and terminal are done.

In "CategoryTheory/limits/graphs/zero.v" I used induction on empty type to be
able to construct Zero from Initial and Terminal with InitialObject =
TerminalObject. This approach doesn't look "correct" but I was unable to give
a better proof. See lines 288 and 290 of the file for the problem
I solved by using the lemmas zeroCocone_isaprop (line 234) and
zeroCone_isaprop (line 253). Is there a better solution?


benediktahrens:
Hi,
thanks a lot for your PR. 
Indeed, your definition of Zero in terms of graphs looks unsatisfactory. 
Instead of defining Zero as consisting of an initial and a terminal such that the underlying objects are equal, you should say that Zero consists of 

- an object
- a cone on that object
- a cocone on that object
- such that both (co)cones are (co)limiting.

This is the analogue of your correct direct definition of Zero in limits/zeros.

Using already existing definitions from limits/graphs/initial and limits/graphs/terminal, you can define
`isZero(c) := isInitial(c) x isTerminal(c)`
and
`Zero := total2 (fun c => isZero c)`

Does this answer your question?

tpannila:
Yes. Your definition is the correct one and using it I was able to solve the problem. I rewrote the file
and pushed changes to my fork. Is "graphs/zero.v" now ok?

UniMath/CategoryTheory/limits/graphs/zero.v
```diff
@@ -0,0 +1,234 @@
...
+
+  (** Definition of Zero. *)
+  Definition Zero : UU := Σ c : C, isZero c.
+  Definition mk_Zero (c : C) (H : isZero c) : Zero := tpair _ c H.
+  Definition ZeroObject (Z : Zero) : C := pr1 Z.
+
+  (** Construction of Initial and Terminal from Zero. *)
+  Definition Zero_to_Initial (Z : Zero) : Initial C.
+  Proof.
+    induction Z.
```
benediktahrens: Instead of using induction on Z here, you should use the projections to get the data you need to build the initial object from Z. This will make some proofs work by "apply idpath" later, without the need to again use induction on Z.

UniMath/CategoryTheory/limits/graphs/zero.v
```diff
@@ -0,0 +1,234 @@
...
+    exact (mk_Terminal t (dirprod_pr2 p)).
+  Defined.
+
+  (** The following lemmas show that the underlying objects of Initial
+    and Terminal, constructed above, are equal to ZeroObject. *)
+  Lemma ZeroObject_equals_InitialObject (Z : Zero) :
+    ZeroObject Z = InitialObject (Zero_to_Initial Z).
+  Proof.
+    induction Z; unfold Zero_to_Initial; simpl.
```
benediktahrens: This induction becomes unnecessary if in the definition above you use the projections instead of inducting on Z.

UniMath/CategoryTheory/limits/graphs/zero.v
```diff
@@ -0,0 +1,234 @@
...
+    apply idpath.
+  Defined.
+
+  Lemma ZeroObject_equals_TerminalObject (Z : Zero) :
+    ZeroObject Z = TerminalObject (Zero_to_Terminal Z).
+  Proof.
+    induction Z; unfold Zero_to_Terminal; simpl.
```
benediktahrens: This induction becomes unnecessary if in the definition above you use the projections instead of inducting on Z.

UniMath/CategoryTheory/limits/graphs/zero.v
```diff
@@ -0,0 +1,234 @@
...
+  (** Using the above lemma we can construct an isomorphisms between any two
+    ZeroObjects. *)
+  Definition iso_Zeros (Z Z' : Zero) : iso (ZeroObject Z) (ZeroObject Z') :=
+    tpair _ (ZeroArrowFrom Z (ZeroObject Z')) (isiso_from_Zero_to_Zero Z Z').
+
+  Definition hasZero := ishinh Zero.
+
+  (** Construct Zero from Initial and Terminal for which the underlying objects
+    are equal. *)
+  Definition Initial_and_Terminal_to_Zero
```
benediktahrens: This lemma would be slightly more interesting if you only assumed an isomorphism between the initial and the terminal object.

UniMath/CategoryTheory/limits/graphs/zero.v
```diff
@@ -0,0 +1,236 @@
...
+
+  Lemma ZeroObject_equals_TerminalObject (Z : Zero) :
+    ZeroObject Z = TerminalObject (Zero_to_Terminal Z).
+  Proof.
+    apply idpath.
+  Defined.
+
+  (** We construct morphisms from ZeroObject to any other object c and from any
+    other object c to the ZeroObject. *)
+  Definition ZeroArrowFrom (Z : Zero) (c : C) : C⟦ZeroObject Z, c⟧.
```
benediktahrens: This definition is slightly odd. There is a "canonical" choice for that arrow, given by the center of contraction of the type C[ZeroObject Z,c].
There is no need to fiddle around with equalities of objects.
Of course, the arrow you define is propositionally equal to the canonical one.

UniMath/CategoryTheory/limits/graphs/zero.v
```diff
@@ -0,0 +1,236 @@
...
+  Proof.
+    set (f := InitialArrow (Zero_to_Initial Z) c).
+    set (g := idtomor _ _ (ZeroObject_equals_InitialObject Z)).
+    exact (g ;; f).
+  Defined.
+
+  Definition ZeroArrowTo (Z : Zero) (c : C) : C⟦c, ZeroObject Z⟧.
```
benediktahrens: Same applies here: this arrow can be projected out of Z directly.

benediktahrens:
Looks good to me.

tpannila:
I fixed the definition of morphisms from ZeroObject and to ZeroObject to use
directly the definitions InitialArrow and TerminalArrow by first taking
Initial and Terminal from Zero. This simplified proofs of uniqueness of such
morphisms.  

What do you think about these changes?

benediktahrens:
On 06/14/2016 04:09 AM, tpannila wrote:
> I fixed the definition of morphisms from ZeroObject and to ZeroObject to use
> directly the definitions InitialArrow and TerminalArrow by first taking
> Initial and Terminal from Zero. This simplified proofs of uniqueness of such
> morphisms.  
> 
> What do you think about these changes?

I think they are good!

PR #364, opened 2016-06-18
Added some results on arbitrary (co)products

tpannila:
Hi Benedikt and others.

I managed to formalize the following results about arbitrary (co)products:
- Construction of an arbitrary (co)product from two arbitrary (co)products and
  the assumption that we have a (co)product of the two arbitrary (co)products.
- Construction of a (co)product from an arbitrary (co)product of two elements
  and vice versa.
  
* In the proofs I have used eapply and Unshelve. Should I try to remove
  these?
* Are there other things I should do to the proofs?

To prove these, I needed some general results which should probably be moved
into some other files.
- The lemmas and definitions with name "coprod_families*" are both in the
  files coproduct.v and product.v. Are these kind of definitions and lemmas
  already done somewhere? If not, where one should move these? Should these
  be renamed?
- In product.v and coproduct.c I have the "tworec", "stn2ind", and
  "pair_to_stn2*". Where one should move these? Should these be renamed?

### DIFF
--- a/UniMath/CategoryTheory/limits/coproducts.v
+++ b/UniMath/CategoryTheory/limits/coproducts.v
@@ -13,6 +13,8 @@ Require Import UniMath.CategoryTheory.total2_paths.
 Require Import UniMath.CategoryTheory.precategories.
 Require Import UniMath.CategoryTheory.UnicodeNotations.
 Require Import UniMath.CategoryTheory.ProductPrecategory.
+Require Import UniMath.CategoryTheory.limits.arbitrary_coproducts.
+Require Import UniMath.Foundations.Combinatorics.StandardFiniteSets.
 
 (** * Definition of binary coproduct of objects in a precategory *)
 Section coproduct_def.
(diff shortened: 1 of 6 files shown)


UniMath/CategoryTheory/limits/coproducts.v
```diff
@@ -526,6 +528,237 @@ Qed.
...
+
+(** In this section we construct an arbitrary coproduct from two arbitrary
+  coproducts by taking the disjoint union of the objects. We need to assume that
+  the coproduct of the given two arbitrary coproducts exists. *)
+Section coproduct_from_coproducts.
+  Variable C : precategory.
+  Hypothesis hs : has_homsets C.
+
+  (** Disjoint union of the objects a1 and a2. *)
+  Definition coprod_families {I1 I2 : UU} (a1 : I1 -> C) (a2 : I2 -> C):
```
benediktahrens: This is `sumofmaps` in Foundations.Basics.PartA. The lemmas about the behaviour of `sumofmaps` (or `coprod_families`) are not really necessary. If you want to record them anyway, the right place would be directly after the definition of `sumofmaps`.

UniMath/CategoryTheory/limits/coproducts.v
```diff
@@ -526,6 +528,237 @@ Qed.
...
+    set (f := pair_to_stn2_mors ConeOb a in1 in2).
+    refine (mk_ArbitraryCoproductCocone _ _ a ConeOb f _ ).
+    refine (mk_isArbitraryCoproductCocone _ _ hs _ _ _ _).
+    intros c g.
+    set (f1 := g stn0).
+    set (f2 := g stn1).
+    set (ar := CoproductArrow _ Cone f1 f2).
+    set (com1 := CoproductIn1Commutes _ _ _ Cone _ f1 f2).
+    set (com2 := CoproductIn2Commutes _ _ _ Cone _ f1 f2).
+    refine (tpair _ (tpair _ ar _ ) _).
```
benediktahrens: In this place it seems preferable to use `simple refine (...)` or, better, use `use iscontrpair`.
This will present the goals in a more natural order and avoid the `Unshelve` below.

Actually, it might be useful to give a construction of a term of type ` ∃! x, P x` from hypotheses 
`(x : A), (p : P x), (h : forall y, isaprop (P y)) (H : forall y, P y -> y = x)`. This thing should go in Foundations, as it would be useful in many places.

benediktahrens:
Hi,

I am unsure about what to do with the finite types you are using. I find it a bit annoying that one has to rewrite, as in `pair_to_stn2_mors`, so I am wondering if there are no better finite types to work with.
For instance, would working with
`
stn_alt (n : nat) :=
match n with 
| 0 => empty
| S n' => coprod (stn_alt n') unit
end
`
work better?

This question does not only apply to your code, but to many other places as well. @DanGrayson is an expert on finite sets in UniMath, he might have an opinion.

tpannila:
Hi,

I added subtypeEquality'' into foundations, defined finite (co)products, and
constructed finite (co)products from (co)products and terminal (resp. initial).
I was able to remove eapplys and Unshelves from my proofs by using the tactic
use as you suggested. subtypeEquality'' is the general construction you
suggested.

I have not yet tried to work with the definition stn_alt.

The proofs on finite (co)products are not very elegant in my mind. Is there
something one can do to make these proofs more elegant?

vladimirias:
Could someone please do the renaming that Benedikt has suggested ASAP? I can not read the PR files with these names it is too confusing.

UniMath/Foundations/Basics/PartB.v
```diff
@@ -469,6 +469,20 @@ Corollary subtypeEquality' {A : UU} {B : A -> UU}
 (* This variant of subtypeEquality is not often needed. *)
 Proof. intros ? ? ? ? e is. apply (total2_paths e). apply is. Defined.
 
+(* This variant is used for categories. *)
+Corollary subtypeEquality'' {A : UU} {B : A -> UU} (x : A) (b : B x)
```
benediktahrens: That's nice, except for the name. Since this lemma is for proving unique existence, I suggest calling it `exists_unique` or something similar.

UniMath/CategoryTheory/limits/finite_coproducts.v
```diff
@@ -0,0 +1,103 @@
...
+End finite_coproduct_def.
+
+(** In the following section we prove that initial object and coproducts imply
+  finite coproducts. *)
+Section finite_coproduct_criteria.
+  Variable C : precategory.
+  Hypothesis hs : has_homsets C.
+
+  (* Couldn't find the following elsewhere. *)
+  Definition unit_to_ob (c : C) : unit -> C := fun tt : unit => c.
```
benediktahrens: This is `termfun` in Foundations.Basics.PartA.

UniMath/CategoryTheory/limits/finite_coproducts.v
```diff
@@ -0,0 +1,103 @@
...
+    initial and coproducts. The result is proved by induction on the number of
+    objects in the finite coproduct. *)
+  Theorem fin_coproduct_from_initial_and_coproducts :
+    Initial C -> Coproducts C -> FinCoproducts C.
+  Proof.
+    intros I Coprods. unfold FinCoproducts. intros n. induction n.
+
+    (* Case n = 0 follows from the fact that empty coproduct can be
+       constructed from initial. *)
+    rewrite (UnivalenceAxiom.weqtopaths (weqstn0toempty)).
```
benediktahrens: Appealing to the univalence axiom here is a bit heavy and unnecessary. 
Another option is to construct the cocone from scratch, doing essentially the same thing you did for constructing `empty_coproduct_from_initial`. While that is not very satisfactory either, it will at least behave well computationally.

UniMath/CategoryTheory/limits/finite_coproducts.v
```diff
@@ -0,0 +1,103 @@
...
+Require Import UniMath.CategoryTheory.limits.arbitrary_coproducts.
+Require Import UniMath.CategoryTheory.limits.coproducts.
+Require Import UniMath.CategoryTheory.limits.initial.
+
+(** Definition of finite coproduct. *)
+Section finite_coproduct_def.
+
+  Variable C : precategory.
+
+  Definition FinCoproducts :=
```
DanGrayson: I would expect a name like "FinCoproducts" to refer to coproducts indexed by finite sets, not by standard finite sets.  Maybe "IteratedCoproducts" instead, since finite coproducts might also be worth implementing (using univalence).

vladimirias: I agree completely. I wanted to write a similar comment myself.

In fact, initial object and binary coproducts do not imply the existence of finite coproducts only finite *ordered* coproducts (same for products). For example, the category, which Fiore denoted F, of natural numbers and finite set morphisms between them, has finite *ordered* coproducts but does not have finite coproducts.

Indeed, the loop stn(2)=stn(2) corresponding to the transposition will translate, through maponpaths, into a loop of any stn(2)-indexed coproduct. So the coproducts being considered are not even coproducts indexed by the standard finite sets considered as finite sets.

Binary coproducts X_0\coprod X_1 and coproducts \coprod_{i\in stn(2)} X_i where X_i is a functor from the category that "consists of” the finite set stn(2) are two different things.

UniMath/CategoryTheory/limits/products.v
```diff
@@ -194,6 +196,212 @@ Qed.
...
+
+  (** Construct a family of 2 objects from a pair of objects. *)
+  Definition pair_to_stn2 (c1 c2 : C) : (stn 2) -> C.
+  Proof.
+    intros X.
+    induction (stn2ind X).
+    exact c1.
+    exact c2.
+  Defined.
+
```
mortberg: I recently had to define similar functions for `stn 3` and `stn 4`. I remember asking @DanGrayson about it and he said he had defined a generic elimination function for `stn n` taking a vector of length `n` with the arguments. I can't remember where to find this function though...?

DanGrayson: Here's the notation I like to use for making lists, where a list is a pair (n,f), and f:stn n -> X :

https://github.com/UniMath/UniMath/blob/a090d5de90843f97369970ffbbd8a6e829a47b50/UniMath/Foundations/Algebra/IteratedBinaryOperations.v#L132

tpannila:
Hi,

I am unsure what I should do with this pull request. On my other pull request
I tried to do rebase and push after that. This did not work and github gave
an error.

Should I cancel this pull request, remove this branch from my fork, rebase this
branch locally, do the renaming and fixes people have suggested, and file a
new pull request after that?

DanGrayson:
Dear @tpannila ,

Your contributions have sparked my curiosity -- what is your name?  Are you willing
to put it and your picture in your profile?

All the best,

Dan

tpannila:
Hi @DanGrayson ,

my name is Tomi Pannila. I added my name to my profile. I don't have a picture
of myself at the moment.

benediktahrens:
On 06/21/2016 01:27 AM, tpannila wrote:
> Hi,
> 
> I am unsure what I should do with this pull request. On my other pull request
> I tried to do rebase and push after that. This did not work and github gave
> an error.
> 
> Should I cancel this pull request, remove this branch from my fork, rebase this
> branch locally, do the renaming and fixes people have suggested, and file a
> new pull request after that?

What do you mean when you say "github gave an error"?
Whenever you rebase, you change the history of your branch. In that
case, when you try to push to github, you are actually overwriting the
old branch with the new one with modified history. Github protects you
from doing that by default.

You can still push your rebased branch by using
`git push -f <name_of_branch>`
Here, the `-f` flag removes that protection.


Alternatively, you can push to a different branch on github using
`git push <name_of_branch>:<name_of_branch_on_github>`.
This will create a new branch on github with the name you have specified.
You can then close the old PR and open a new one from the new branch.

Issue #366, opened 2016-06-18
Renaming of (co)products


benediktahrens:
In the near future, the following renaming will happen:

- `Coproduct` and `Product` will be renamed into `BinCoproduct` and `BinProduct`, and similar for access functions
- `ArbitraryCoproduct` will be renamed into `Coproduct`, and analogously for products. Similar for access functions.

I document this here, because the renaming is likely to break some code, e.g., in pull requests. In particular, @tpannila might be affected.

tpannila:
Good. The new names are much better. I will fix my pull requests to match the
new names when the renaming happens.

mortberg:
I'll do the renaming now. Should have it finished before the end of the day.

benediktahrens:
Done in #367 , closing.

PR #367, opened 2016-06-19
Renaming of (co)products

mortberg:
This PR addresses https://github.com/UniMath/UniMath/issues/366 by:

- Renaming (co)product to bin(co)product

- Renaming arbitrary_(co)product to (co)product

- Renaming (co)product_functor defined in FunctorsPointwise(Co)Product to (co)product_of_functors (and similarily for binary versions). The less direct version (defined by composing delta, pair and product functors) is now called product_of_functors_alt.


UniMath/CategoryTheory/.package/files
```diff
@@ -18,13 +18,13 @@ limits/initial.v
 limits/graphs/initial.v
 limits/zero.v
 limits/graphs/zero.v
+limits/bincoproducts.v
 limits/coproducts.v
-limits/arbitrary_coproducts.v
-limits/graphs/coproducts.v
+limits/graphs/bincoproducts.v
```
benediktahrens: Would a camelcase `BinCoproducts.v` be easier to read?

UniMath/CategoryTheory/limits/FunctorsPointwiseBinCoproduct.v
```diff
@@ -0,0 +1,299 @@
...
+Variable HD : BinCoproducts D.
+Variable hsD : has_homsets D.
+
+Section bincoproduct_of_functors.
+
+Variables F G : functor C D.
+
+Local Notation "c ⊗ d" := (BinCoproductObject _ (HD c d))(at level 45).
+
+Definition bincoproduct_of_functors_ob (c : C) : D := F c ⊗ G c.
```
benediktahrens: The spelling/capitalization should be unified. Since it is `BinCoproduct`, it should probably be `BinCoproduct_of_functors`?

benediktahrens:
I have taken a superficial look, and am ready to trust in Travis.

UniMath/CategoryTheory/.package/files
mortberg: I prefer all lowercase letters in filenames, but I'm not too fanatic about it (it is not like trailing whitespace... :)). However there seems to be no standard for this in UniMath at the moment. Maybe we could decide on a standard for the CategoryTheory library and make another PR where we make the file names more uniform?

UniMath/CategoryTheory/limits/FunctorsPointwiseBinCoproduct.v
mortberg: Maybe it should also be camelcase? I've noticed we are not at all consequent about this... However making identifier names more standars should probably be done in another PR. I'll only make the change you suggest for now.

UniMath/CategoryTheory/.package/files
mortberg: I take that back, the standard seems to be to use camelcase, so let's make another PR where we make the CT library filenames camelcase as well.

benediktahrens: Let's say that @DanGrayson imposed his standard on us.

DanGrayson: :-)

I was surprised to see "the standard" in a comment, as though my proposal had been approved or even acclaimed.

vladimirias: What is camelcase? How does one know how to write in it when one need to import bincoproducts.v?

DanGrayson: I didn't know, either: https://en.wikipedia.org/wiki/CamelCase

PR #368, opened 2016-06-20
Monics, epis, (co)equalizers, and (co)kernels

tpannila:
Hi,

due to renaming of (co)products, I did not try to fix my other pull request
today, but tried to do something which would not conflict with renaming. This
pull request contains the following:

- Direct implementation of equalizers and coequalizers, the files equalizers.v
  and coequalizers.v
- Direct implementation of kernels and cokernels, the files kernels.v and
  cokernels.v
- Definition of monics and epis, the files Monics.v and Epis.v
- Some changes to zero.v. These I used for cokernels and kernels.

The files I have added should prove the following:

- Equalizer arrows are monics and Coequalizer arrows are epis.
- Kernel arrows are monics and Cokernel arrows are epis.
- Composition of monics and epis is monic and epi, respectively.
- If composition of morphisms is a monic, then the first morphism is a monic.
  If composition of morphisms is an epi, then the second morphism is an epi.

When writing these files, I tried to mimic the file Pullbacks.v.
All comments are welcome.

### DIFF
--- a/UniMath/CategoryTheory/Epis.v
+++ b/UniMath/CategoryTheory/Epis.v
@@ -1,0 +1,84 @@
+(* Definition of epi *)
+Require Import UniMath.Foundations.Basics.PartD.
+Require Import UniMath.Foundations.Basics.Propositions.
+Require Import UniMath.Foundations.Basics.Sets.
+
+Require Import UniMath.CategoryTheory.precategories.
+Require Import UniMath.CategoryTheory.UnicodeNotations.
+
+Section def_epi.
+
+  Variable C : precategory.
+
+  (** Definition and construction of isEpi. *)
+  Definition isEpi {x y : C} (f : x --> y) :=
+    forall (z : C) (g h : y --> z), f ;; g = f ;; h -> g = h.
+  Definition mk_isEpi {x y : C} (f : x --> y) :
+    (forall (z : C) (g h : y --> z), f ;; g = f ;; h -> g = h) -> isEpi f.
+  Proof. intros X. unfold isEpi. apply X.  Defined.
+
+  (** Definition and construction of Epi. *)
+  Definition Epi (x y : C) := Σ f : x --> y, isEpi f.
+  Definition mk_Epi {x y : C} (f : x --> y) (H : isEpi f) : Epi x y.
+  Proof. apply (tpair _ f H). Defined.
+
+  (** Gets the arrow out of Epi. *)
+  Definition EpiArrow {x y : C} (M : Epi x y) := pr1 M.
+
+  (** Identity to isEpi and Epi. *)
+  Lemma id_isEpi {x : C} : isEpi (identity x).
+  Proof.
+    apply mk_isEpi.
+    intros z g h H.
+    rewrite <- id_left; apply pathsinv0; rewrite <- id_left; apply pathsinv0.
+    exact H.
+  Defined.
+
+  Lemma id_Epi {x : C} : Epi x x.
+  Proof. exact (tpair _ (identity x) (id_isEpi)). Defined.
+
+  (** Isomorphism to isEpi and Epi. *)
+  Lemma iso_isEpi {x y : C} (f : x --> y) (H : is_iso f) : isEpi f.
+  Proof.
+    apply mk_isEpi.
+    intros z g h X.
+    apply (pre_comp_with_iso_is_inj _ x _ _ f H).
+    exact X.
+  Defined.
+
+  Lemma iso_Epi {x y : C} (f : x --> y) (H : is_iso f) : Epi x y.
+  Proof. apply (mk_Epi f (iso_isEpi f H)). Defined.
+
+  (** Composition of isEpis and Epis. *)
+  Definition isEpi_comp {x y z : C} (f : x --> y) (g : y --> z) :
+    isEpi f -> isEpi g -> isEpi (f ;; g).
+  Proof.
+    intros X X0. unfold isEpi. intros z0 g0 h X1.
+    repeat rewrite <- assoc in X1. apply X in X1. apply X0 in X1. apply X1.
+  Defined.
+
+  Definition Epi_comp {x y z : C} (M1 : Epi x y) (M2 : Epi y z) :
+    Epi x z := tpair _ (EpiArrow M1 ;; EpiArrow M2)
+                       (isEpi_comp (EpiArrow M1) (EpiArrow M2)
+                                     (pr2 M1) (pr2 M2)).
+
+  (* This general result should be moved somewhere? *)
+  Lemma precomp_with_eq {x y z : C} {f g : y --> z} (h : x --> y) :
+    f = g -> h ;; f = h ;; g.
+  Proof.
+    intros X.
+    rewrite X.
+    apply idpath.
+  Defined.
+
+  (** If precomposition of g with f is an epi, then g is an epi. *)
+  Definition isEpi_precomp {x y z : C} (f : x --> y) (g : y --> z) :
+    isEpi (f ;; g) -> isEpi g.
+  Proof.
+    intros X. intros w φ ψ H.
+    apply (precomp_with_eq f) in H.
+    repeat rewrite assoc in H.
+    apply (X w _ _ H).
+  Defined.
+
+End def_epi.
--- a/UniMath/CategoryTheory/Epis.v
+++ b/UniMath/CategoryTheory/Epis.v
@@ -18,24 +18,13 @@ Section def_epi.
   Proof. intros X. unfold isEpi. apply X.  Defined.
 
   (** Definition and construction of Epi. *)
-  Definition Epi (x y : C) := Σ f : x --> y, isEpi f.
-  Definition mk_Epi {x y : C} (f : x --> y) (H : isEpi f) : Epi x y.
-  Proof. apply (tpair _ f H). Defined.
+  Definition Epi (x y : C) : UU := Σ f : x --> y, isEpi f.
+  Definition mk_Epi {x y : C} (f : x --> y) (H : isEpi f) :
+    Epi x y := tpair _ f H.
 
   (** Gets the arrow out of Epi. *)
-  Definition EpiArrow {x y : C} (M : Epi x y) := pr1 M.
-
-  (** Identity to isEpi and Epi. *)
-  Lemma id_isEpi {x : C} : isEpi (identity x).
-  Proof.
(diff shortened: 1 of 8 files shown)


UniMath/CategoryTheory/Epis.v
```diff
@@ -0,0 +1,84 @@
...
+    (forall (z : C) (g h : y --> z), f ;; g = f ;; h -> g = h) -> isEpi f.
+  Proof. intros X. unfold isEpi. apply X.  Defined.
+
+  (** Definition and construction of Epi. *)
+  Definition Epi (x y : C) := Σ f : x --> y, isEpi f.
+  Definition mk_Epi {x y : C} (f : x --> y) (H : isEpi f) : Epi x y.
+  Proof. apply (tpair _ f H). Defined.
+
+  (** Gets the arrow out of Epi. *)
+  Definition EpiArrow {x y : C} (M : Epi x y) := pr1 M.
```
benediktahrens: You can make this projection into a "coercion", such that Coq inserts `EpiArrow` automatically for you whenever it expects a morphism instead of an epi.
For an example of what that looks like, see the definition of `iso a b` in CategoryTheory.precategories.

UniMath/CategoryTheory/Epis.v
```diff
@@ -0,0 +1,84 @@
...
+
+  (** Identity to isEpi and Epi. *)
+  Lemma id_isEpi {x : C} : isEpi (identity x).
```
benediktahrens: A nicer way to prove this is `pre_comp_with_iso_is_inj` (from precategories.v).

benediktahrens: Or using your `iso_isEpi` below, of course.

UniMath/CategoryTheory/Epis.v
```diff
@@ -0,0 +1,84 @@
...
+    repeat rewrite <- assoc in X1. apply X in X1. apply X0 in X1. apply X1.
+  Defined.
+
+  Definition Epi_comp {x y z : C} (M1 : Epi x y) (M2 : Epi y z) :
+    Epi x z := tpair _ (EpiArrow M1 ;; EpiArrow M2)
+                       (isEpi_comp (EpiArrow M1) (EpiArrow M2)
+                                     (pr2 M1) (pr2 M2)).
+
+  (* This general result should be moved somewhere? *)
+  Lemma precomp_with_eq {x y z : C} {f g : y --> z} (h : x --> y) :
```
benediktahrens: This is an instance of the very general `maponpaths`. There is usually no need to have this specialized instance, `apply maponpaths` usually works.
If you want to keep this lemma anyway, you should prove it using `apply maponpaths`. You could move it to precategories.v.

UniMath/CategoryTheory/Monics.v
```diff
@@ -0,0 +1,84 @@
...
+    repeat rewrite assoc in X1. apply X0 in X1. apply X in X1. apply X1.
+  Defined.
+
+  Definition Monic_comp {x y z : C} (M1 : Monic x y) (M2 : Monic y z) :
+    Monic x z := tpair _ (MonicArrow M1 ;; MonicArrow M2)
+                       (isMonic_comp (MonicArrow M1) (MonicArrow M2)
+                                     (pr2 M1) (pr2 M2)).
+
+  (* This general result should be moved somewhere? *)
+  Lemma postcomp_with_eq {x y z : C} {f g : x --> y} (h : y --> z) :
```
benediktahrens: This is `cancel_postcomposition` from precategories.v.

UniMath/CategoryTheory/limits/cokernels.v
```diff
@@ -0,0 +1,137 @@
...
+  Hypothesis Z : Zero C.
+
+  (** This rewrite is used to rewrite an equality f ;; g = ZeroArrow to
+     f ;; g = ZeroArrow ;; g. This is because Coequalizers need the latter
+     equality. *)
+  Lemma CokernelEqRw {x y z : C} {g : x --> y} {f : y --> z}
+        (H : g ;; f = ZeroArrow _ Z x z) :
+    g ;; f = ZeroArrow _ Z x y ;; f.
+  Proof.
+    rewrite <- (ZeroArrow_comp_left _ Z x y z f) in H.
```
benediktahrens: We often do not care how we prove equalities between arrows in a category. If, for some reason, you would like to get a "nice" proof, you could use `pathvia (ZeroArrow _ Z x z)` here to get your equality via transitivity of the equality in the hypothesis and (the inverse of) a canoncial equality.

UniMath/CategoryTheory/limits/cokernels.v
```diff
@@ -0,0 +1,137 @@
...
+             (H : g ;; f = (ZeroArrow _ Z x z))
+             (isE : isCoequalizer g (ZeroArrow _ Z x y) f (CokernelEqRw H))
+    : Cokernel g.
+  Proof.
+    use (mk_Coequalizer g (ZeroArrow _ Z x y) f (CokernelEqRw H)).
+    apply isE.
+  Defined.
+  Definition Cokernels := forall (y z : C) (g : y --> z), Cokernel g.
+  Definition hasCokernels := forall (y z : C) (g : y --> z), ishinh (Cokernel g).
+  Definition CokernelOb {y z : C} {g : y --> z} (CK : Cokernel g)
```
benediktahrens: Here and elsewhere, it would be good practice to also write the output type, such as 
`Definition CokernelOb {y z : C} ... : C :=...`

UniMath/CategoryTheory/limits/cokernels.v
```diff
@@ -0,0 +1,137 @@
...
+    := CoequalizerObject CK.
+  Coercion CokernelOb : Cokernel >-> ob.
+  Definition CokernelArrow {y z : C} {g : y --> z} (CK : Cokernel g) :=
+    CoequalizerArrow CK.
+  Definition CokernelEqAr {y z : C} {g : y --> z} (CK : Cokernel g) :=
+    CoequalizerEqAr CK.
+  Definition CokernelOut {y z : C} {g : y --> z} (CK : Cokernel g)
+             (w : C) (h : z --> w) (H : g ;; h = ZeroArrow _ Z y w) : C⟦CK, w⟧.
+  Proof.
+    exact (pr1 (pr1 (isCoequalizer_Coequalizer CK w h (CokernelEqRw H)))).
```
benediktahrens: It will probably be beneficial if you define CokernelOut as a `CoequalizerOut`, e.g., by using 
`use CoequalizerOut.` 
here.

benediktahrens:
LGTM modulo inline comments.

mortberg:
This is very nice. I guess you are working your way to abelian categories? I know there are some issues in proving that abelian groups (and more generally R-modules) are abelian categories in standard type theory (with the setoid definition of categories), so it would be interesting to investigate the status of this in UniMath.  For details see [this](http://assert-false.net/arnaud/papers/Towards%20constructive%20homological%20algebra%20in%20type%20theory.pdf) note.

tpannila:
Hi,

I fixed the things Benedikt pointed out. Mortberg, I am working towards
abelian categories. Thank you for the note you linked.

While checking my changes, I changed "id_isEpi" to "identity_isEpi". I forgot
that emacs has preloaded the compiled files, and thus "id_isEpi" was fine in
emacs. I forgot to recompile the stuff using make. The commit "Fixed monics,
epis, (co)equalizers, and (co)kernels" does _not_ compile, because Epi.v does
not to compile. This is fixed in the commit "Fixed a typo".

My bad.

benediktahrens:
@tpannila don't worry about PRs that don't compile. We are using Travis CI, which checks whether a given PR compiles. If you look at https://github.com/UniMath/UniMath/pull/368 , you'll see either a red "x" or a green checkmark next to each commit number, signifying either failure or success of Travis in compiling the corresponding commit. Clicking on the red "x" or green checkmark will give you more info.
In short, we will notice when a commit does not compile and not merge it.

vladimirias:
Will we notice an “Admitted” or an extra axiom?

benediktahrens:
On 06/21/2016 03:30 PM, Vladimir Voevodsky wrote:
> Will we notice an “Admitted” or an extra axiom?

No, not at the moment.

We could add such a test to the Travis script, but I am afraid it would
be quite fragile.
The difficulty is that in Travis we cannot examine only the new content,
but only the whole of UniMath with the PR merged in.
That means that when checking for Axioms, we have to have a way to
distinguish between those we like and those we don't.
In other words, we need to have a whitelist for those axioms that are
currently in there (which should not contain @DanGrayson 's
`PostponedProof` :-) ).

A check for `Admitted.` would be easier, since we don't have any of
those, and would simply reject a PR where Travis finds an occurrence.

benediktahrens:
On 06/21/2016 03:48 PM, Benedikt Ahrens wrote:
> 
> 
> On 06/21/2016 03:30 PM, Vladimir Voevodsky wrote:
>> Will we notice an “Admitted” or an extra axiom?
> 
> No, not at the moment.
> 
> We could add such a test to the Travis script, but I am afraid it would
> be quite fragile.
> The difficulty is that in Travis we cannot examine only the new content,
> but only the whole of UniMath with the PR merged in.
> That means that when checking for Axioms, we have to have a way to
> distinguish between those we like and those we don't.
> In other words, we need to have a whitelist for those axioms that are
> currently in there (which should not contain @DanGrayson 's
> `PostponedProof` :-) ).
> 
> A check for `Admitted.` would be easier, since we don't have any of
> those, and would simply reject a PR where Travis finds an occurrence.


I was wrong in several respects.

a) It is not quite easy even to check for `Admitted.`, because they
could appear in commented code - where we do not necessarily want to
forbid them.

b) We do have some `Admitted.` in the code. I'll file issues for them.

vladimirias:
Such a whitelist is easy. We know what it contains: several types of function extensionality and the univalence axiom. That’s it.

vladimirias:
If we can have a script that will show, by a request, right at the github, all axioms and all admitted  it would be good.

benediktahrens:
On 06/21/2016 04:25 PM, Vladimir Voevodsky wrote:
> If we can have a script that will show, by a request, right at the
> github, all axioms and all admitted  it would be good.

I don't think that there is a way Github can do that.

The only way to achieve that I can think of is to have Travis show the
Axioms and Admitted, via something similar to what I did in the Makefile
in PR #375.
That would require one to click on the green checkmark (or the red "x")
next to a commit number, and to take a look at the Travis logs. Not very
convenient...

PR #375, opened 2016-06-21
added Makefile targets to show axioms and admitted

benediktahrens:
poor man's checks for bad stuff in the Coq code:

- `$ make admitted` shows `Admitted`.
- `$ make axiom` shows uses of the word "Axiom".

There is a lot of noise in the output, due to `coq_makefile` being called, and there are false positives since "Axiom" sometimes occurs in comments.

I'd be happy to see something better than this.


vladimirias:
Dan, what do you think? The makefile is your area of expertise.

PR #376, opened 2016-06-21
product precategory only was cartesian power, now more general

rmatthes:
ProductPrecategory.v now with precategory indexed over I
power_precategory corresponds to the previous definitions
pr_functor not instantiated for this special situation
other three files only minimally changed so that they compile again

Remark: "Definition of the general product category" in the header information of UniMath/CategoryTheory/ProductPrecategory.v
seemed to me too generally-sounding. However, the current generalization was not dictated by a need for the generality elsewhere but rather by the fact that the proofs did not even have to be changed.
If in doubt, consider this as my test of the infrastructure set up for contributing to UniMath.


benediktahrens:
LGTM.
I tested compilation on my machine, no clue why Travis CI is not triggered on this PR.

PR #377, opened 2016-06-21
disallow Admitted

DanGrayson:


vladimirias:
If possible I would have this line emphasized by a different font (bold?).

Also, we should add that one is forbidden from redefining existing identifiers.

Otherwise things such as the Andrej’s stunt become possible by re-defining nat or bool or empty etc.

Finally, should we mention Abort. separately?

DanGrayson:
Bold face would make it too "busy".

Is there anything to say about "Abort"?  Its use is certainly okay.

As for stunts like redefining bool, is anyone going to do that?  I don't see the danger.

vladimirias:
What do you mean by “too busy”? It is a very important rule that should stand out from the rest of the text..

I am actually not sure what “Abort.” does never having used it myself. What does it do?

I think that the possibility of redefining an existing identifier is a very real thing. Whether for a stunt or simply by ignorance. There is also danger of there appearing PRs that are intended to play a joke. As more people contribute it will be harder and harder for us to check all of the PRs.

What can we do to check that whether a re-definition of an identifier has not occurred? I cc-ed Jason maybe he has some clever idea.

benediktahrens:
On 06/22/2016 03:38 PM, Vladimir Voevodsky wrote:
> What do you mean by “too busy”? It is a very important rule that
> should stand out from the rest of the text..
> 
> I am actually not sure what “Abort.” does never having used it
> myself. What does it do?

Whenever you write `Definition foo : bar.`, Coq enters proof mode
(independently of you writing `Proof.` or not).

`Abort` leaves that proof mode.

In contrast to `Admitted`, `Abort` does not lead to `foo : bar` to be
taken as an axiom. Instead, nothing happens. In particular, `foo` will
not be a defined identifier after `Abort.`.

A Coq file/module containing
```
Lemma foo : False.
Abort.
```
is a valid module, with no content.
The file
```
Lemma foo : False.
Admitted.
```
is also a valid module, with an axiom as contents.

> I think that the possibility of redefining an existing identifier is
> a very real thing. Whether for a stunt or simply by ignorance. There
> is also danger of there appearing PRs that are intended to play a
> joke. As more people contribute it will be harder and harder for us
> to check all of the PRs.
> 
> What can we do to check that whether a re-definition of an identifier
> has not occurred? I cc-ed Jason maybe he has some clever idea.

There is no protection in Coq against redefining an existing identifier.
If the same identifier `ident` is defined in two different files `a.v`
and `b.v`, in a file `c.v` importing both `a` and `b` one distinguishes
between them by typing `a.ident` and `b.ident`, respectively.

Also, it would be possible to introduce a notation `ident`, I believe. I
am not sure what the behaviour would be when having both an identifier
and a notation `ident`.


The only thing we can do to discover this kind of thing is to read the PRs.
If we don't have time to proofread a PR, we should not merge it.

vladimirias:
1. I think Abort has no place in the main library.

2. For the automated search for re-definition of identifiers I again cc Jason. Reading the PR is not sufficient - none of us knows all of the identifiers of all of the branches of the UniMath.

DanGrayson:
I have used several duplicate identifiers in my modules.  For example, I
have Group.zero as the zero group and Monoid.zero as the zero monoid.
Similarly, Group.ZZ is the free group on one generator and AbelianGroup.ZZ
is the free abelian group on one generator.  Since modules can be required
without being imported, there is no reason to avoid collisions, and the
qualified notation with dots has a certain beauty to it.

benediktahrens:
Similarly, we use the same identifiers for, e.g., the direct
formalization of pullbacks and the formalization of pullbacks as
specific limits.

This means that one can use one definition or the other simply by
changing what module one requires.

JasonGross:
If you have a small number of reserved identifiers, you can issue `Locate bool` (or possibly `Locate "bool"` for notations) in a file that `Require`s the entire library, and check that it only names one thing.

Aliases (`Notation bool := ...`) are similar to definitions; they get namespaced, and have kernel names (qualified names).  Notations (`Notation "'bool'" := ...`) reserve the symbol as a keyword, and make it impossible to add another definition of the symbol.  They don't have kernel names.

vladimirias:
Group_ZZ and AbelianGroup_ZZ have the same beauty as the ones with the . and it is easy to make a local notation in a module denoting Group_ZZ by ZZ.

If we start allowing the same identifier to denote different things, things can get very messy very fast especially as we move into the more complex concepts.

DanGrayson:
Re: "If possible I would have this line emphasized by a different font (bold?)."

I would prefer adding a new rule and putting it in bold.  It would say "read all the rules".  But I think of that as being implicitly present and implicitly in bold face.

vladimirias:
Then maybe ordering the rules such that the one about Admitted is on the top will do the job.

DanGrayson:
Thanks, done.

DanGrayson:
Re: " I think Abort has no place in the main library. "

It's better than putting the code in a comment, since it will get checked, and we don't forbid that.

PR #378, opened 2016-06-22
Iterated binary (co)products

tpannila:
Hi,

this pull request replaces my older pull request on the subject. This pull
request contains:
- Construction of binary (co)products from (co)products and vice versa.
- Direct definition of iterated binary (co)products.
- Construction of iterated binary coproduct (resp. iterated binary product)
  from initial and binary coproducts (resp. terminal and binary products).

What has changed
- Proofs no longer use Univalence. (I use funextfun, and this is defined in
  PartD.v)
- The proofs use the definition stn_alt. This approach was suggested by
  Benedikt. For some reason I called this iter. See Iter.v . I don't know if
  this is a good name. Suggestions for a better one?
- What I mistakenly called finite (co)products last time are now called
  iterated binary (co)products. This was suggested by Dan.
- Renamed subtypeEquality'' to exists_unique. Suggested by Benedikt.
- Use termfun instead of unit_to_ob. Suggested by Benedikt.

The last two definitions and lemma in Iter.v do not really seem to belong
there. Where one should put these?

All comments are welcome.

### DIFF
--- a/UniMath/CategoryTheory/.package/files
+++ b/UniMath/CategoryTheory/.package/files
@@ -67,3 +67,5 @@ PrecategoriesWithBinOps.v
 PrecategoriesWithAbgrops.v
 PreAdditive.v
 limits/BinDirectSums.v
+limits/FinOrdProducts.v
+limits/FinOrdCoproducts.v
--- a/UniMath/CategoryTheory/limits/FinOrdCoproducts.v
+++ b/UniMath/CategoryTheory/limits/FinOrdCoproducts.v
@@ -1,0 +1,206 @@
+(** A direct definition of finite ordered coproducts by using coproducts *)
+Require Import UniMath.Foundations.Basics.PartD.
+Require Import UniMath.Foundations.Basics.Propositions.
+Require Import UniMath.Foundations.Basics.Sets.
+
+Require Import UniMath.Foundations.Combinatorics.StandardFiniteSets.
+
+Require Import UniMath.CategoryTheory.total2_paths.
+Require Import UniMath.CategoryTheory.precategories.
+Require Import UniMath.CategoryTheory.UnicodeNotations.
+Require Import UniMath.CategoryTheory.ProductPrecategory.
+
+Require Import UniMath.CategoryTheory.limits.coproducts.
+Require Import UniMath.CategoryTheory.limits.bincoproducts.
+Require Import UniMath.CategoryTheory.limits.initial.
+
+
+(** Definition of finite ordered coproducts. *)
+Section def_FinOrdCoproducts.
+
+  Variable C : precategory.
+
+  Definition FinOrdCoproducts : UU :=
+    forall (n : nat) (a : stn n -> C), CoproductCocone (stn n) C a.
+  Definition hasFinOrdCoproducts := ishinh FinOrdCoproducts.
+
+End def_FinOrdCoproducts.
+
+
+(** Construction of FinOrdCoproducts from Initial and BinCoproducts. *)
+Section FinOrdCoproduct_criteria.
+
+  Variable C : precategory.
+  Hypothesis hs : has_homsets C.
+
+  (** Case n = 0 of the theorem. *)
+  Lemma InitialToCoproduct (I : Initial C):
+    ∀ (a : stn 0 -> C), CoproductCocone (stn 0) C a.
+  Proof.
+    intros a.
+    use (mk_CoproductCocone _ _ _ I
+                            (fun i : stn 0 => fromempty (weqstn0toempty i))).
+    use (mk_isCoproductCocone _ _ hs).
+    intros c g. use unique_exists.
+
+    apply (InitialArrow I c).
+    intros i. apply (fromempty (weqstn0toempty i)).
+    intros y. apply impred_isaprop. intros t. apply hs.
+    intros y X. apply InitialArrowUnique.
+  Defined.
+
+  (** Case n = 1 of the theorem. *)
+  Lemma ObjectToCoproduct:
+    forall (a : stn 1 -> C), CoproductCocone (stn 1) C a.
+  Proof.
+    intros a.
+    set (stn1ob := invweq(weqstn1tounit) tt).
+    use (mk_CoproductCocone _ _ _ (a stn1ob)).
+    intros i. exact (idtoiso (! (maponpaths a (isconnectedstn1 stn1ob i)))).
+
+    (* isCoproductcocone *)
+    use (mk_isCoproductCocone _ _ hs).
+    intros c g.
+    use (unique_exists (g stn1ob)).
+
+    (* Commutativity. *)
+    intros i. rewrite <- (isconnectedstn1 stn1ob i). apply id_left.
+
+    (* Equality of equalities of morphisms. *)
+    intros y. apply impred_isaprop. intros t. apply hs.
+
+    (* Uniqueness. *)
+    intros y X. rewrite <- (X stn1ob). apply pathsinv0. apply id_left.
+  Defined.
+
+  (** Finite ordered coproducts from initial and binary coproducts. *)
+  Theorem FinOrdCoproducts_from_Initial_and_BinCoproducts :
+    Initial C -> BinCoproducts C -> FinOrdCoproducts C.
+  Proof.
+    intros I BinCoprods. unfold FinOrdCoproducts. intros n. induction n.
+
+    (* Case n = 0 *)
+    apply (InitialToCoproduct I).
+
+    (* General case. *)
+    intros a.
+    set (a1 := fun (i : stn n) => a (dni_lastelement i)).
+    set (Cone1 := IHn a1).
+    set (a2 := (fun _ : stn 1 => a (lastelement n))).
+    set (Cone2 := ObjectToCoproduct a2). (* Case n = 1 *)
+    set (Cone1In := CoproductIn _ _ Cone1).
+    set (Cone2In := CoproductIn _ _ Cone2).
+    set (BinCone := BinCoprods (CoproductObject (stn n) C Cone1)
+                               (CoproductObject (stn 1) C Cone2)).
+    set (in1 := BinCoproductIn1 _ BinCone).
+    set (in2 := BinCoproductIn2 _ BinCone).
+    set (m1 := fun i1 : stn n => (Cone1In i1) ;; in1).
+    set (m2 := fun i2 : stn 1 => (Cone2In i2) ;; in2).
+
+    use (mk_CoproductCocone (stn (S n)) C a (BinCoproductObject _ BinCone) _).
+
+    (* Construction of the arrows from a i to BinCone *)
+    intros i. induction (natlehchoice4 (pr1 i) _ (pr2 i)).
+    exact (idtoiso (maponpaths a (dni_lastelement_eq n i a0))
+                   ;; m1 (stnpair n (pr1 i) a0)).
+    exact (idtoiso (maponpaths a (lastelement_eq n i b))
+                   ;;  m2 (invweq(weqstn1tounit) tt)).
+
+    (* Construction of isCoproductCocone. *)
+    use (mk_isCoproductCocone _ _ hs).
+
+    intros c g.
+    set (g1 := fun i : stn n => g(dni_lastelement i)).
+    set (ar1 := CoproductArrow _ _ Cone1 g1).
+    set (com1 := BinCoproductIn1Commutes  _ _ _ BinCone c ar1
+                                          (g(lastelement n))).
+    set (com2 := BinCoproductIn2Commutes _ _ _ BinCone c ar1
+                                         (g(lastelement n))).
+    set (com3 := CoproductInCommutes _ _ _ Cone1 _ g1).
+    set (com4 := CoproductInCommutes _ _ _ Cone2 _
+                                     (fun _ : stn 1 => g(lastelement n))).
+
+    use (unique_exists).
+
+    (* Construction of the unique arrow from BinCone to c. *)
+    use (BinCoproductArrow _ BinCone).
+    use (CoproductArrow _ _ Cone1). intros i. exact (g (dni_lastelement i)).
+    use (CoproductArrow _ _ Cone2). intros i. exact (g (lastelement n)).
+
+    (* First commutativity. *)
+    intros i. unfold coprod_rect. induction (natlehchoice4 (pr1 i) n (pr2 i)).
+    rewrite (dni_lastelement_eq n i a0). repeat rewrite <- assoc.
+    apply remove_id_left. apply idpath.
+
+    unfold m1. unfold in1. rewrite <- assoc. fold g1. fold ar1.
+    use (pathscomp0 (maponpaths (fun f : _ => Cone1In (stnpair n (pr1 i) a0) ;; f)
+                                com1)).
+    fold ar1 in com3. rewrite com3. unfold g1. apply idpath.
+
+
+    (* Second commutativity. *)
+    rewrite (lastelement_eq n i b). repeat rewrite <- assoc.
+    apply remove_id_left. apply idpath.
+
+    unfold m2. unfold in2. rewrite <- assoc. fold g1. fold ar1.
+    use (pathscomp0 (maponpaths (fun f : _ => Cone2In (lastelement 0) ;; f) com2)).
+    rewrite com4. apply idpath.
+
+
+    (* Equality on equalities of morphisms. *)
+    intros y. apply impred_isaprop. intros t. apply hs.
+
+    (* Uniqueness *)
+    unfold coprod_rect. intros k X.
+    apply BinCoproductArrowUnique.
+
+
+    (* From stn n *)
+    apply CoproductArrowUnique.
+    intros i. rewrite <- (X (dni_lastelement i)). rewrite assoc.
+    apply cancel_postcomposition.
+    induction (natlehchoice4 (pr1 (dni_lastelement i)) n
+                             (pr2 (dni_lastelement i))).
+    unfold m1. rewrite assoc. unfold in1.
+    apply cancel_postcomposition.
+    unfold Cone1In. apply pathsinv0.
+
+    set (e := dni_lastelement_is_inj (dni_lastelement_eq n (dni_lastelement i)
+                                                         a0)).
+    use (pathscomp0 _ (CoproductIn_idtoiso (stn n) C (a ∘ dni_lastelement) Cone1
+                                           e)).
+    apply cancel_postcomposition.
+    apply maponpaths. apply maponpaths. (* Why we need maponpaths twice? *)
+    rewrite <- (maponpathscomp _ a).
+    apply maponpaths. apply isasetstn.
+
+    (* This is false because of b *)
+    apply fromempty. induction i as [i i'].
+    cbn in b. induction (!b).
+    apply (isirreflnatlth _ i').
+
+
+    (* From stn 1 *)
+    apply CoproductArrowUnique.
+    intros i. rewrite <- (X (lastelement n)). rewrite assoc.
+    apply cancel_postcomposition.
+    induction (natlehchoice4 (pr1 (lastelement n)) n (pr2 (lastelement n))).
+
+    (* This case is false because of a0 *)
+    apply fromempty. cbn in a0. apply (isirreflnatlth _ a0).
+
+    apply pathsinv0. unfold m2. rewrite assoc. unfold in2.
+    apply cancel_postcomposition. unfold Cone2In.
+
+    (* This case makes sense *)
+    set (e := isconnectedstn1 i (invweq(weqstn1tounit) tt)).
+    use (pathscomp0 _ (CoproductIn_idtoiso (stn 1) C
+                                           (fun _ : _ => a(lastelement n))
+                                           Cone2 e)).
+    apply cancel_postcomposition.
+    apply maponpaths. apply maponpaths. (* Why we need maponpaths twice? *)
+    fold (@funcomp (stn 1) _ _ (fun _ : stn 1 => lastelement n) a).
+    rewrite <- (maponpathscomp (fun _ : stn 1 => lastelement n) a).
+    apply maponpaths. apply isasetstn.
+  Defined.
+End FinOrdCoproduct_criteria.
--- a/UniMath/CategoryTheory/limits/FinOrdProducts.v
+++ b/UniMath/CategoryTheory/limits/FinOrdProducts.v
@@ -169,7 +169,8 @@ Section FinOrdProduct_criteria.
     set (e := dni_lastelement_is_inj (dni_lastelement_eq n (dni_lastelement i)
                                                          a0)).
     use (pathscomp0 _ (ProductPr_idtoiso (stn n) C (a ∘ dni_lastelement) Cone1
-                                         e)).
+                                         (!e))).
+    rewrite maponpathsinv0.
     apply cancel_precomposition.
     apply maponpaths. apply maponpaths. (* Why we need maponpaths twice? *)
     apply maponpaths.
@@ -197,7 +198,8 @@ Section FinOrdProduct_criteria.
 
     set (e := isconnectedstn1 i (invweq(weqstn1tounit) tt)).
     use (pathscomp0 _ (ProductPr_idtoiso (stn 1) C (fun _ : _ => a(lastelement n))
-                                         Cone2 e)).
+                                         Cone2 (!e))).
+    rewrite maponpathsinv0.
     apply cancel_precomposition.
     apply maponpaths. apply maponpaths. (* Why we need maponpaths twice? *)
     apply maponpaths.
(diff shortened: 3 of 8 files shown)


UniMath/CategoryTheory/limits/IteratedBinProducts.v
```diff
@@ -0,0 +1,90 @@
...
+    set (Cone := BinProds (ProductObject (iter n) C A1)
+                          (ProductObject unit C A2)).
+    set (p1 := BinProductPr1 _ Cone).
+    set (p2 := BinProductPr2 _ Cone).
+    set (m1 := fun i1 : iter n => p1 ;; (A1pr i1)).
+    set (m2 := fun i2 : unit => p2 ;; (A2pr i2)).
+    set (ConeOb := BinProductObject _ Cone).
+    set (homoteq := @iter_equality _ (S n) _ a (iterincleq _ n a)).
+
+    rewrite <- homoteq.
```
benediktahrens: This use of `rewrite` will mess up your definition in later use, in the sense that you won't recognize your iterated binary product as an inductively obtained binary product. (Well, maybe you will, but Coq won't.)
I believe it would be better here to use `use mk_CoproductCocone` to build the desired cocone from scratch, as well as prove its universal property.

Disclaimer: this is just a suspicion, I have not tried any of that seriously.

benediktahrens:
I see that the constructions are basically the same with `iter` than with `stn` - or am I missing something? Sorry for that, that change was hence not helpful.

vladimirias:
If there is no considerable advantage it is better not to introduce alternative standard finite sets.

tpannila:
Hi benediktahrens  and vladimirias  ,

I'm unsure about using standard finite sets. The reason is that the definition
of stn, `stn` n = \Sigma i : nat, i < n` , requires an inequality which I feel
is unnecessary for iterated binary (co)products. What would be the benefit of
requiring such inequality for iterated binary (co)products? Isn't it much easier
to deal with iterated binary (co)products by using `ii1` and `ii2` for Iter?

I would like to hear both of your opinion on this before I continue to work on
this pull request, because if I need to work with standard finite sets I need
to rewrite some stuff.

vladimirias:
Hi Tomi,

I am not sure what you mean by the inequality being unnecessary. I suppose an iterated binary coproduct is still an iterated binary coproduct defined by induction as emptyset, X and (\coprod_{i\in stn n} X_i)\coprod X_n for the \coprod_{i\in stn(n+1)} X_i (using math notation) independently on which finite sets you use. The standard functions between stn i are introduced in stnfinsets.v and in particular there is there the inclusion of the initial segment stn n -> snt n+1 that you will need for this definition. I do not remember its notation in stnfinsets.v, in my paper http://arxiv.org/pdf/1602.00352.pdf <http://arxiv.org/pdf/1602.00352.pdf> p.14, which is irrelevant here except for this notation, it is denoted \iota n 1 .

Vladimir.

DanGrayson:
Dear Tomi,

The inequality i<n is part of the definition of stn n, so indirectly
it plays a role if you use stn n later, but the main concern later should be the properties of
stn n that are needed.  Since stn n and iter n are equivalent, they'll have the same properties,
and one would be just as good as another.  Many properties of stn n have been already proven
by Vladimir.

There might be some advantage in defining a finite sequence of elements of a type by induction,
rather than as a function from stn n or from iter n, for then you might be able to avoid using 
functional extensionality to prove that two sequences are equal, but you aren't doing that, I think.

tpannila:
Hi Vladimir and Dan,

Thank you for your comments and arguments.

My argument mostly concerns about simplicity. Iter is very simple. It uses only
3 very elementary type theoretic notions: empty, unit, and coproducts.

@DanGrayson  , I think you are correct, stn n and Iter n are equivalent, and
thus have the same properties. Why would we use a more complex definition
to get the same behavior that we get from a simpler definition? 


For standard finite sets I have only found the following weak equivalences
          stn 0 ≃ empty ,  stn 1 ≃ unit ,  stn (S n) ≃ stn n ⨿ stn 1
and I would like these to be equalities, like they are under Univalence. But
if we want to avoid Univalence in the proof of
`IteratedBinaryCoproducts_from_Initial_and_BinCoproducts`, to get better
computational behavior, then I guess we don't have equalities. Is this true?
Thus, I cannot use `sumofmaps`, `termfun`, or `fromempty` directly and I
have to do duplicates of these for stn n?

I can try to do this pull request with stn if people think that using stn is
the correct way to do this.

vladimirias:
There are many more properties of stn n’s in the stnfinsets.v including, for example,
weq (weq (stn n) (stn n)) (stn (factorial n)).

One strong, at least for some, argument against iter n is that it uses fixpoint function with values in the universe. Proof-theoretically it is a much more complex gadget than the inequality in natural numbers that is defined in the UniMath using a fixpoint with values in bool.

vladimirias:
BTW having written the previous message I noticed that there are *notations* that we re-define compared to the “standard library”. However these are notations, not identifiers. As far as I know we do not redefine any of the identifiers in init part of the standard library and that is the only part that we ever use.

DanGrayson:
Re: "For standard finite sets I have only found the following weak equivalences
stn 0 ≃ empty , stn 1 ≃ unit , stn (S n) ≃ stn n ⨿ stn 1
and I would like these to be equalities, like they are under Univalence. "

Equivalences are just about as good as equalities for all practical purposes, without using univalence.  Is there some great advantage?

I don't see anything called `IteratedBinaryCoproducts_from_Initial_and_BinCoproducts`, so I can't comment on it.

DanGrayson:
The use of "fixpoint" can be removed by defining "iter" as follows:

```
  Definition iter (n:nat) : UU.
    induction n as [|n X].
    - exact empty.
    - exact (X ⨿ unit).
  Defined.
```

DanGrayson:
Tomi, notice that our coding style in `UniMath/README.md` forbids the use of `Fixpoint`.

DanGrayson:
@mortberg -- I notice your `iterprod` also uses the forbidden `Fixpoint`, which can be removed as described above.

vladimirias:
Yes, one can hide the fixpoint inside the eliminator, but even if Coq did not define eliminators through match and fixpoint (which it does), it would still involve using the induction principle to define a function with values in a universe and it is a complex operation.

It means that the induction principle itself should allow type families with types lying in the universe one step higher from the universe where the basic types such as nat or stn n are.

It is not even clear that iter n belongs to the same universe (say U0) that nat and stn n do because the only way to get it is to evaluate nat_rect on a number of arguments the first of which is the function ( fun n : nat => U0 ) whose type is ( nat -> U1 ) where U1 is the universe of which U0 is an element.

DanGrayson:
Yes, but the reason we like using eliminators instead of match or fix is so the code will likely work in a new proof checker that doesn't implement the eliminators that way.

The point about universes is a good one, but how hard do we have to try to keep things in the same universe, when basic operations like set-quotient and prop-quotient are not, we're using Type in Type in Coq, and we don't know whether the resizing axioms are okay even in theory?

vladimirias:
I was only explaining why stn n is proof-theoretically a much more simple construct than iter n .

As for the resizing rules, the weak resizing - having a function that puts any hprop to U0 such that on hprops from U0 it is definitionally equal to the  identity is validated by the standard simplicial model.

They are sufficient to have set-quotients and “prop-quotient” in the same universe but these quotients will probably not have a very good computational behavior. However in everything else they will be just as good as the ones we use now.

We don’t know if the strong resizing rules can be shown to be equiconsistent with the ZFC but it is an open problem that people work on.

DanGrayson:
Re: "As for the resizing rules, the weak resizing - having a function that puts any hprop to U0 such that on hprops from U0 it is definitionally equal to the  identity is validated by the standard simplicial model."

Could you state this resizing rule precisely and explain why it's validated?

tpannila:
Hi Vladimir and Dan,

thank you for your comments.

@vladimirias  , your proof-theoretical explanation of universes, stn, and iter
was very useful for me. It made me to understand why stn is better than iter.

Due to my insufficient knowledge of universes, equivalence, and univalence,
I have decided to do the following. First I will study these things from the
HoTT book, and see how these things are defined and used in the Foundations.
Then I will try to fix this pull request with standard finite sets. This will take
some time, so if someone needs the contents of this pull request sooner, please
fix this pull request and file a new pull request, and I will close my pull
request.

benediktahrens:
Re: "Could you state this resizing rule precisely and explain why it's validated? (I'd like to understand it.) (Perhaps it's a bit off topic.)"

Maybe that is a topic that would be better discussed on our mailing list. To remind you, the address is univalent-mathematics@googlegroups.com .

vladimirias:
Hi,

I suggest you simply fix it using your intuition. You should not need Univalence Axiom and I would strongly suggest that you try not to use it or any other axiom.

Axioms break computation.

What you are given is a function F: (stn n) -> C where C is your category. You want to define an object finordcoprod F and morphisms finordin F i : F i --> finordcoprod F for each i : stn such that together the data ( finordin F i , finordcoprod F ) is a coproduct of the finite ordered set of objets represented by F. You do it by induction on n setting for n=0, finordcoprod F = initial (where initial is the initial object of C), for n=1, finordcoprod F = F 0  and for (S (S n)),

finordcoprd F = bincoprod ( finordcoprod (funcomp (stn_left 1 (S n)) F) (F (S n)).

The function stn_left is defined in the file StandardFiniteSets.

The construction of the morphisms finordin and the proof of the coproduct axiom should be straightforward. This is it. There is no need for the univalence or any axioms at all and no need to study anything fancy.

You are doing a great job and please don’t get distracted. The main argument towards (stn n) versus (iter n) is that we should not introduce two versions of the same mathematical object.

Vladimir.

tpannila:
Hi @vladimirias  , @benediktahrens  , @DanGrayson  and others,

I'm unable to do the proof. I have included the part of the proof I was able
 to do.  I tried to use `stn_left` to get a map  stn n -> stn (S n), but since
coq cannot unify n + 1 with S n, I used `dni_lastelement`. 

For me the main difficulty in the proof is exactly the construction of the
morphisms and verification of the coproduct axioms.

I will start to work on my other pull request.

```
(** A direct definition of finite ordered coproducts by using coproducts *)

Require Import UniMath.Foundations.Basics.PartD.
Require Import UniMath.Foundations.Basics.Propositions.
Require Import UniMath.Foundations.Basics.Sets.

Require Import UniMath.Foundations.Combinatorics.StandardFiniteSets.

Require Import UniMath.CategoryTheory.total2_paths.
Require Import UniMath.CategoryTheory.UnicodeNotations.
Require Import UniMath.CategoryTheory.ProductPrecategory.

Require Import UniMath.CategoryTheory.limits.coproducts.
Require Import UniMath.CategoryTheory.limits.bincoproducts.
Require Import UniMath.CategoryTheory.limits.initial.


(** Definition of finite ordered coproducts. *)
Section def_FinOrdCoproducts.

  Variable C : precategory.

  Definition FinOrdCoproducts : UU :=
    forall (n : nat) (a : stn n -> C), CoproductCocone (stn n) C a.
  Definition hasFinOrdCoproducts := ishinh FinOrdCoproducts.

End def_FinOrdCoproducts.


(** Construction of FinOrdCoproducts from Initial and BinCoproducts. *)
Section FinOrdCoproduct_criteria.

  Variable C : precategory.
  Hypothesis hs : has_homsets C.

  Lemma stn1isconnected : forall i : stn 1, invweq(weqstn1tounit) tt = i.
  Proof.
    intros i.
    apply (invmaponpathsweq weqstn1tounit).
    apply isconnectedunit.
  Defined.

  (** Case n = 1. *)
  Lemma identity_to_coproduct:
    forall (a : stn 1 -> C), CoproductCocone (stn 1) C a.
  Proof.
    intros a.
    set (stn1ob := invweq(weqstn1tounit) tt).

    use (mk_CoproductCocone _ _ _ (a stn1ob)).
    intros i. exact (idtoiso (! (maponpaths a (stn1isconnected i)))).

    use (mk_isCoproductCocone _ _ hs).
    intros c g.
    use (unique_exists).
    exact (g stn1ob).

    (* Commutativity. *)
    intros i. rewrite <- (stn1isconnected i). apply id_left.

    (* Equality of equalities of morphisms. *)
    intros y. apply impred_isaprop. intros t. apply hs.

    (* Uniqueness. *)
    intros y X. rewrite <- (X stn1ob). apply pathsinv0. apply id_left.
  Defined.

  Lemma stnneq : forall (n : nat) (i : stn (S n)) (a0 : pr1 i < n),
      i = dni_lastelement (stnpair n (pr1 i) a0).
  Proof.
    intros n i a0.
    apply isinjstntonat.
    apply idpath.
  Defined.

  Lemma stneq1 : forall (n : nat) (i : stn (S n)) (a0 : pr1 i = n),
      i = lastelement n.
  Proof.
    intros n i a0.
    unfold lastelement.
    apply isinjstntonat.
    apply a0.
  Defined.

  Theorem FinOrdCoproducts_from_Initial_and_BinCoproducts :
    Initial C -> BinCoproducts C -> FinOrdCoproducts C.
  Proof.
    intros I Coprods. unfold FinOrdCoproducts. intros n. induction n.

    (** Case n = 0 follows from the fact that empty coproducts can be
      constructed from Initial. *)
    intros a.
    use (mk_CoproductCocone _ _ _ I (fun i : stn 0 => fromempty (weqstn0toempty i))).
    use (mk_isCoproductCocone _ _ hs).
    intros c g. use unique_exists; cbn.

    apply (InitialArrow I c).
    intros i. apply (fromempty (weqstn0toempty i)).
    intros y. apply impred_isaprop. intros t. apply hs.
    intros y. intros X. apply InitialArrowUnique.

    (** General case. *)
    intros a.
    set (a1 := fun (i : stn n) => a (dni_lastelement i)).
    set (A1 := IHn a1).
    set (a2 := (fun _ : stn 1 => a (lastelement n))).
    set (A2 := identity_to_coproduct a2).
    set (A1in := CoproductIn _ _ A1).
    set (A2in := CoproductIn _ _ A2).
    set (Cone := Coprods (CoproductObject (stn n) C A1)
                         (CoproductObject (stn 1) C A2)).
    set (in1 := BinCoproductIn1 _ Cone).
    set (in2 := BinCoproductIn2 _ Cone).
    set (m1 := fun i1 : stn n => (A1in i1) ;; in1).
    set (m2 := fun i2 : stn 1 => (A2in i2) ;; in2).
    set (ConeOb := BinCoproductObject _ Cone).
    fold ConeOb in in1, in2, m1, m2.

    use (mk_CoproductCocone (stn (S n)) C a ConeOb _).

    (* Construction of the arrows from a i to ConeOb *)
    intros i. induction (natlehchoice4 (pr1 i) _ (pr2 i)).
    exact (idtoiso (maponpaths a (stnneq n i a0)) ;; m1 (stnpair n (pr1 i) a0)).
    exact (idtoiso (maponpaths a (stneq1 n i b))
                   ;;  m2 (invweq(weqstn1tounit) tt)).

    (* Construction of isCoproductCocone. *)
    use (mk_isCoproductCocone _ _ hs).
    intros c g.

    set (g1 := fun i : stn n => g(dni_lastelement i)).
    set (ar1 := CoproductArrow _ _ A1 g1). fold ar1.
    set (com1 := BinCoproductIn1Commutes  _ _ _ Cone c ar1 (g(lastelement n))).
    set (com2 := BinCoproductIn2Commutes _ _ _ Cone c ar1 (g(lastelement n))).
    set (com3 := CoproductInCommutes _ _ _ A1 _ g1).
    set (com4 := CoproductInCommutes _ _ _ A2 _ (fun _ : stn 1 => g(lastelement n))).

    use (unique_exists).

    (* Construction of the unique arrow from ConeOb to c. *)
    use (BinCoproductArrow _ Cone).
    use (CoproductArrow _ _ A1). intros i. exact (g(dni_lastelement i)).
    use (CoproductArrow _ _ A2). intros i. exact (g (lastelement n)).

    (* First commutativity. *)
    intros i. unfold coprod_rect. induction (natlehchoice4 (pr1 i) n (pr2 i)).
    rewrite (stnneq n i a0). cbn. repeat rewrite <- assoc. apply remove_id_left.
    apply idpath.

    unfold m1. unfold in1. rewrite <- assoc. fold g1. fold ar1.
    set (XX := maponpaths (fun f : _ => A1in (stnpair n (pr1 i) a0) ;; f) com1).
    cbn in XX.
    eapply pathscomp0. apply XX.
    fold ar1 in com3.
    rewrite com3.
    unfold g1.
    apply idpath.


    (* Second commutativity. *)
    rewrite (stneq1 n i b). cbn. repeat rewrite <- assoc. apply remove_id_left.
    apply idpath.

    unfold m2. unfold in2. rewrite <- assoc. fold g1. fold ar1.
    set (XX := maponpaths (fun f : _ => A2in (lastelement 0) ;; f) com2).
    eapply pathscomp0. apply XX.
    rewrite com4.
    apply idpath.


    (* Equality on equalities of morphisms. *)
    intros y. apply impred_isaprop. intros t. apply hs.

    (* Uniqueness *)
    unfold coprod_rect. intros k X.
    apply BinCoproductArrowUnique. apply CoproductArrowUnique.

    intros i.

    rewrite <- (X (dni_lastelement i)). rewrite assoc.
    apply cancel_postcomposition.
    destruct (natlehchoice4 (pr1 (dni_lastelement i)) n (pr2 (dni_lastelement i))).
    unfold m1. rewrite assoc. unfold in1.
    apply cancel_postcomposition.
    unfold A1in.
    apply pathsinv0.

    (* Can't do more than this. *)
  Abort.
End FinOrdCoproduct_criteria.
```

benediktahrens:
Please commit your partial work and push it to this branch. I'll then take a look.
Please don't paste big chunks of code into an issue comment.
Thanks a lot!

tpannila:
Hi Benedikt,

sorry for pasting the code. I did put the code to limits/IteratedBinCoproducts.v
and pushed. Please note that doing simpl in the proofs might not finish and
cbv delta may take so much memory that your operating system may crash.
Doing cbn works, but takes some time, 20-30s on my system. (Possibly
because coq unfolds things which are not meant to be unfolded).

benediktahrens:
Hi Tomi,

the situation does not look bad at all. You need a kind of lemma that you'll encounter very frequently. The general form should look somehow like this (I omit a few arguments)
```
Lemma foo  (a : diagram J C) (CC : Cocone a) (i j : J) (e : i = j) : 
   idtoiso (maponpaths a e) ;; CoconeIn a i = CoconeIn a j.
Proof.
induction e; apply id_left.
Qed.
```
Actually, the work consists in stating the lemma, the proof will then be very easy.
It is important to state the lemma with free variables `i` and `j`, so that you can induct on `e`.
I let you try this without digging deeper into it myself, but don't hesitate to ask again if you are encountering difficulties.

Also, you'll need to state the lemma in terms of Coproducts, not general Colimits.

benediktahrens:
Sorry, another remark: the situtation in your case is slightly more complicated, since the identity you want to induct on is an identity between elements of `stn (S n)`, where both are of the form `dni_lastelement ??`..
You can try with two additional ingredients:

- an identity of two elements `dni_lastelement x` and `dni_lastelement y` should be of the form `maponpaths dni_lastelement p` for `p : x = y`.
- might or might not be useful: `maponpaths (g o f) e = maponpaths g (maponpaths f e)`.

Let me know how it goes, please.

tpannila:
Hi Benedikt,

thank you for your help. I was able to complete the proof with these
suggestions. It will take me a few days to clean up the proof and fix this
pull request.

tpannila:
Hi,

I fixed the finite ordered (co)products. I added some results I needed to
`StandardFiniteSets.v` and to `NaturalNumbers.v` . I hope these are not
duplicates of any existing results.

UniMath/Foundations/NumberSystems/NaturalNumbers.v
```diff
@@ -222,6 +222,15 @@ Proof. intros n m is is' . apply ( isirreflnatgth n ( istransnatgth _ _ _ is is'
 Lemma isantisymmnegnatgth ( n m : nat ) : ¬ ( n > m ) -> ¬ ( m > n ) -> n = m .
 Proof . intro n . induction n as [ | n IHn ] . intros m ng0m ngm0  .  apply ( pathsinv0 ( negnatgth0tois0 _ ngm0 ) ) . intro m . destruct m as [ | m ] . intros ngsn0 ng0sn . destruct ( ngsn0 ( natgthsn0 _ ) ) .  intros ng1 ng2 .   apply ( maponpaths S ( IHn m ng1 ng2 ) ) .  Defined .
 
+Lemma iseqnatgth {i n : nat} (e1 : i = n) (e2 : natgth n i) : empty.
```
DanGrayson: It is better to not have lemmas like this with equalities like `i=n` as hypotheses, as it results in a profusion of lemmas.
Instead, use induction in your code, achieving `n>n`, and use the existing lemma that says that doesn't happen.

tpannila: This result is used on line 178 in FinOrdCoproducts.v . If I do induction on
the equality `b` , or on the equality obtained from it by rewriting with
`pr1_dni_lastelement` , the equality disappears and the other variable is not
replaced by the other in variables in the context. In particular, the inequality
`pr2 i` stays `n < pr1 i`. Thus, I don't know how to obtain `pr1 i < pr1 i` or
`n < n` by induction.

DanGrayson: Try this instead:
```
diff --git a/UniMath/CategoryTheory/limits/FinOrdCoproducts.v b/UniMath/CategoryTheory/limits/FinOrdCoproducts.v
index a3850de..c305355 100644
--- a/UniMath/CategoryTheory/limits/FinOrdCoproducts.v
+++ b/UniMath/CategoryTheory/limits/FinOrdCoproducts.v
@@ -175,8 +175,11 @@ Section FinOrdCoproduct_criteria.
     apply maponpaths. apply isasetstn.
 
     (* This is false because of b *)
-    apply fromempty. apply (iseqnatgth b (pr2 i)).
-
+    apply fromempty.
+    induction i as [i i'].
+    simpl in b.
+    induction (!b).
+    apply (isirreflnatlth _ i').
 
     (* From stn 1 *)
     apply CoproductArrowUnique.
```

tpannila: Very good! I did as you suggested. This allowed me to removed the Lemma iseqnatgth.

vladimirias:
Where is it? In which PR. I will take a look to check about possible duplication.

DanGrayson:
Vladimir, at the bottom of all these emails is this line:

"Reply to this email directly, view it on GitHub, or mute the thread."

Click on "view it onGitHub" to be taken to the PR.

vladimirias:
Thanks!

benediktahrens:
Hi Tomi,

`Print Assumptions FinOrdProducts_from_Terminal_and_BinProducts.` shows that you use the axiom of function extensionality somewhere. Do you know where exactly it is used? If so, we could think about whether it is possible to do without funext.

tpannila:
Hi Benedikt,

sorry I did not know that I use functional extensionality. I use the function
isinjstntonat, which uses funext according to coq.

See `Print Assumptions isinjstntonat`. I don't understand this because
`Print Assumptions invmaponpathsincl` does not show funext and neither does
`Print Assumptions isinclstntonat` . The proof of `isinjstntonat` is 

```
Proof. intros ? i j. apply (invmaponpathsincl pr1). apply isinclstntonat. Defined.
```

See line 920 of StandardFiniteSets.v . Does this help?

benediktahrens:
The problem is that the proof of `isInjectiveFunction` being a
proposition needs funext.

Unfortunately, `isInjectiveFunction` is defined as a map into hProp
instead of into `UU`. One thus always carries this axiom with, even
though one does not need it.

This makes `isInjectiveFunction` annoying to use. I suggest you don't
use it, and unfold its definition instead.
That means, you should change the statement of your lemma into:
```
∀ x x' : X, f x = f x' → x = x'
```
with `f` suitably replaced with the projection you are looking at.

Once this is fixed, could you please check if your result does still
depend on any axioms?


I also think that `isInjectiveFunction` should be fixed, or a big
warning should be printed next to it.

tpannila:
This is not enough. I found that `impred_isaprop` uses funext since `impred`
uses funext. And this uses funext at least because `funcontr` uses funext. See
the first comment in PartD.v for `funconr` .

I use `impred_isaprop` for example in InitialToCoproduct (FinOrdCoproducts.v ,
line 48) to prove one `isaprop (...)`. Are there alternative ways to prove
such things without using impred?

benediktahrens:
In `InitialToCoproduct`, you should not use `uniqe_exists`. Instead, you should provide the initial arrow, and when proving the property and uniqueness, you should use the fact that you have a hypothesis of `t < 0`, i.e., an element of `stn 0`.
More specifically, instead of `exists_unique`, use
```
    mkpair. 
    - mkpair. 
      + apply InitialArrow.
```
and solve the remaining goals using the hypothesis of `i : stn 0`.
Let me know if that works or not.

tpannila:
Hi Benedikt,

the initial arrow is provided for the `unique_exists` on line 46. If I replace
the unique_exists by the following
```
use (tpair _ _ _). use (tpair _ (InitialArrow I c) _).
intros i. apply (fromempty (weqstn0toempty i)).
intros i. use total2_paths. apply InitialArrowUnique. cbn.
apply proofirrelevance.
```
the goal is
```
isaprop (∀ i0 : stn 0, fromempty (negstn0 i0) ;; InitialArrow I c = g i0)
```
This I usually solve by `impred_isaprop`. Note that before impred_isaprop
there is no term of `stn 0` in the context. Thus I cannot use fromempty.

benediktahrens:
You are right, there does not seem to be a way to get around a use of funext here. 

I'll merge now. If we find a way to do without funext, we can always change the code later.

PR #379, opened 2016-06-23
Preadditive categories and binary biproducts

tpannila:
Hi,

- Definition of precategories such that each hom (not homset) has a binary
  operation. (PrecategoriesWithBinOps.v)
- Definition of precategories such that each homset has a structure of an
  abelian group. (PrecategoriesWithAbgrops.v)
- Definition of preadditive categories. (PreAdditive.v)
- Definition of binary biproducts. (BinBiproducts.v)

The main results of this pull request are the following:

- As a product in preadditive category, the unique arrow of a binbiproduct
  can be written as `p1 ;; f + p2 ;; g` for given morphisms f and g.
  (FromBinBiproductFormulaUnique)
- As a coproduct in preadditive category, the unique arrow of a binbiproduct
  can be written as `f ;; i1 + g ;; i2` for given morphisms f and g.
  (ToBinBiproductformulaUnique)
- We prove that in preadditive categories the morphisms `a \oplus b -->
  c \oplus d` , induced from `f : a --> c` and `g : b --> d` ,  by viewing
  the biproduct as a product and a coproduct are equal. This proof
  uses the idea that the morphism is given by `p1 ;; f ;; i1 + p2 ;; g ;; i2`.
  (BinBiProductIndArEq)

There are a few things I would like to do:

- I would like to remove the rewrites I use in the proofs that one can finds in
  BinBiproducts.v . I would like to use pathscomp0, but this makes the proofs
  harder to read. Is there a tactic which behaves like rewrite but uses only
  pathscomp0 ?
- I would like to use the notation \oplus for binbiproduct. What are the
  parameters I should pass to this when one writes `Notation "f '\oplus' g" =
  ... ` ? Does this conflict with other uses of \oplus ?
- Similarly, I would like to use `+` to denote the operation in each homset.
  This notation is also used for natural numbers. Is this a problem?

I did not find other files which put some algebra on homsets of a precategory,
so I don't know whether this approach to preadditive categories, by using
preadditive categories with abgrops and precategories with binops, is
satisfactory. I would like to get some feedback on this approach.

All comments are welcome.


DanGrayson:
`refine` can be useful for eliminating `rewrite`, as here:

```
--- a/UniMath/CategoryTheory/BinProductPrecategory.v
+++ b/UniMath/CategoryTheory/BinProductPrecategory.v
@@ -112,14 +112,14 @@ Proof.
     unfold functor_fix_snd_arg_mor; simpl.
     unfold functor_fix_snd_arg_ob; simpl.
     assert (functor_id_inst := functor_id F).
-    rewrite <- functor_id_inst.
+    simple refine (_ @ functor_id_inst _).
     apply maponpaths.
     apply idpath.
   + intros c c' c'' f g.
     unfold functor_fix_snd_arg_data; simpl.
     unfold functor_fix_snd_arg_mor; simpl.
     assert (functor_comp_inst := functor_comp F (dirprodpair c d) (dirprodpair c' d) (dirprodpair c'' d)).
-    rewrite <- functor_comp_inst.
+    simple refine (_ @ functor_comp_inst _ _).
     apply maponpaths.
     unfold compose at 2.
     unfold binproduct_precategory; simpl.
```

DanGrayson:
The name of `is_PrecategoryWithBinOps` beginning with `is` makes me think it defines a property, but instead it defines an element of added structure, so it should be renamed.

DanGrayson:
How is `mk_isPrecategoryWithBinOps` ever going to be useful?  It isn't used, and even if it were used, it probably could be deleted or replaced with a cast, since it's implemented as an identity function.

DanGrayson:
Similarly for `mk_isPrecategoryWithAbgrosp`, which is misspelled.

DanGrayson:
About ⊕, you could write it as in `SubstitutionSystems/SubstitutionSystems.v`, where it
is made a local notation, and thus won't conflict with anything.  On the other hand, that doesn't make
it available for others to use, and we don't want to encourage users to make many copies, even of one
line of code, so a good way is to have a submodule containing non-local notations, which anyone can import.  I don't know whether that's best.

It's important for all uses of the same notation to have the same syntax parts, so conflicts don't arise.

DanGrayson:
Perhaps a better way to limit the applicability of notations is to restrict them to a scope, as in 

```
Notation "x + y" := ( op x y ) : addmonoid_scope .
```

, for example.

But notice that there is no reason for the name of a scope to contain the string "_scope".

vladimirias:
Yes, this name convention is very convenient for the understanding and we should avoid breaking it.

tpannila:
Hi Dan and Vladimir,

Thank you for your comments.

@DanGrayson , I removed `mk_isPrecategoryWithAbrops` and
`mk_isPrecategoryWithBinOps`, and fixed the `is_` stuff. I simplified the file
BinBiproducts.v and added the following result.

- If a precategory has a zero object, then in a homset the unit element is
  given by the zero arrow (PreAdditive.v , PreAdditive_unel_zero).

I played with notations and found out that adding notation does not
significantly improve readability of the proofs in BinBiproducts.v . Thus I have
decided not to introduce any new notation for now. 

Also, I noticed that the proofs where I use rewrite only prove equality
between morphisms, and one does not need to unfold these proofs. Thus, it is
probably not so important to remove these rewrites.


If there is no further mistakes in this pull request it is ready to be merged.

UniMath/CategoryTheory/limits/BinBiproducts.v
```diff
@@ -0,0 +1,379 @@
...
+    rewrite <- assoc. rewrite <- assoc.
+    rewrite (BinBiproductIdIn2 B).
+    rewrite (BinBiproductUnit1 B).
+    rewrite id_right.
+    set (XX := PreAdditive_premor_0 A c a b f).
+    unfold PrecategoryWithAbgrops_premor in XX.
+    unfold PrecategoryWithAbgrops_unel.
+    rewrite XX.
+    apply (PrecategoryWithAbgrops_lunax A c b).
+  Defined.
```
benediktahrens: This is just a remark, not necessarily something to correct right now: proofs that are like this one are not meant to be reasoned about - as you say, you won't ever unfold those proofs. It is hence not important to remove the rewrites. 
But it also means that you can close the proof using `Qed.` instead of `Defined.`. That sometimes helps speed type-checking up when you use the lemma later.

vladimirias:
Why is it “binary biproducts”. Why “bi” twice?

UniMath/CategoryTheory/limits/BinBiproducts.v
vladimirias: I suggest, as a matter of style, not to use Qed, One may complete the proof of a lemma with Defined. and then, in the next line write

Opaque name_of_the_lemma .

It is longer, but has an important advantage of it being possible to use ‘Transparent name_of_the_lemm.” even in the middle of a later proof where the content of the proof term of name-Of_the_lemma is needed and then to lock it again  by “Opaque the name_of_the_lemma.” after it has been used as needed.

It is unfortunate that “Transparent” does not allow to access the content of proof terms that are locked with Qed.

Maybe we should file an issue with Coq requesting this to be made possible. This would save a lot of typing for us.

For now I suggest Defined. followed by Opaque …. .

DanGrayson:
A "biproduct" is what others call a "direct sum", i.e., a situation where
the coproduct and
the product are isomorphic (by the natural map).

vladimirias:
Maybe direct_sum is better? This is the name that I know mathematicians know. I am doubtful about the biproduct.

DanGrayson:
I think that (or directSum) would be better, too.  I notice that Mac Lane says this in his book:

"In special categories, such as Ab and R-Mod, the biproduct is often called a direct sum. "

But I think that way of speaking didn't catch on, and "direct sum" is the preferred term.

tpannila:
Hi Benedikt, Vladimir, and Dan,

thank you for your comments. I added Opaque and renamed Biproducts to
DirectSums.

UniMath/CategoryTheory/limits/BinBiproducts.v
mortberg: Just so you know, Opaque is not the same as Qed. When you write Qed the term is really hidden and will never get unfolded, but Opaque is just a hint to the kernel and some tactics. See the comment by Robbert Krebbers in: 

http://coq-club.inria.narkive.com/1gCV6IBa/feature-request-sealed-definitions

So I don't think we can replace all Qed's by Defined followed by Opaque (but it would be interesting if someone wanted to test this). Benedikt and I had to be very careful about putting Qed/abstract while implementing the computation of initial algebras so that things don't get unfolded too much during either unification or Qed in order not to run out of memory or Coq running forever. I sometimes also had to put an occasional Opaque on certain constants in order for some Qed's not taking forever.

benediktahrens:
This PR now has a merge conflict. Could you fix it, please? I think it is ready to be merged otherwise.

tpannila:
Hi Benedikt,

I think I rebased the branch successfully. The only conflict was in
`UniMath/CategoryTheory/.package/files`

UniMath/CategoryTheory/limits/BinBiproducts.v
vladimirias: I know they are not the same so my previous suggestion remains.

Originally there were no Qed’s in the Foundations. Only Defined with an occasional Opaque. It worked quite well.

PR #380, opened 2016-06-23
Better doc for limits

rmatthes:
update/creation of metadata files for limits and its subdirectories

### DIFF
--- a/UniMath/CategoryTheory/limits/README.md
+++ b/UniMath/CategoryTheory/limits/README.md
@@ -1,16 +1,43 @@
-Rezk Completion
+limits
 ===============
 
-This subdirectory contains the definition of some limits for the notion of precategory defined in the package *RezkCompletion*.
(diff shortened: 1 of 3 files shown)


DanGrayson:
Dear Ralph,

Could you associate the email address used on your commits with your github account and add your name to your github profile?

vladimirias:
This is a great PR. We need similar ones for all our directories. I also proposed and propose again that we need to have a bibliography file and to provide, where available, references to papers and preprints that formalizations are based on. Where it is new material we should say that this or that file contains what is to the best of our knowledge, new research. 
Even where the formalization based on textbook material such as on Bourbaki or MacLane book on categories it makes sense to give explicit references to them.
This way at least some bits of the text will be immediately familiar to someone who without the knowledge of Coq or UniMath decides to have a look at what we are doing.

PR #385, opened 2016-06-27
Pullback of a monic is a monic

tpannila:
Hi,

- A proof that a pullback of a monomorphism is a monomorphism. (pullbacks.v)
- A proof that a pushout of an epimorphism is an epimorphism. (pushouts.v)
The file pushouts.v is similar to the file pullbacks.v . There was a section
about gluing pullbacks in pullbacks.v . I have not translated this to pushouts
since my main goal was to verify the above two results.

The function `transportf_isotoid'` added to precategories.v is used on line 270
in pushouts.v . It differs from `transportf_isotoid` by postcomposing with the
isomorphism, not precomposing.

### DIFF
--- a/UniMath/CategoryTheory/limits/pushouts.v
+++ b/UniMath/CategoryTheory/limits/pushouts.v
@@ -1,0 +1,330 @@
+(* Direct implementation of pushouts *)
+Require Import UniMath.Foundations.Basics.PartD.
+Require Import UniMath.Foundations.Basics.Propositions.
+Require Import UniMath.Foundations.Basics.Sets.
+
+Require Import UniMath.CategoryTheory.precategories.
+Require Import UniMath.CategoryTheory.UnicodeNotations.
+Require Import UniMath.CategoryTheory.limits.initial.
+Require Import UniMath.CategoryTheory.Epis.
+
+Local Notation "a --> b" := (precategory_morphisms a b)(at level 50).
+
+
+Section def_po.
+
+  Context {C : precategory}.
+  Variable hs: has_homsets C.
+
+  Definition isPushout {a b c d : C} (p1 : a --> b) (p2 : a --> c)
+             (f : b --> d) (g : c --> d) (H : p1 ;; f = p2 ;; g) : UU :=
+    forall e (h : b --> e) (k : c --> e)(H : p1 ;; h = p2 ;; k),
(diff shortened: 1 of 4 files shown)


UniMath/CategoryTheory/limits/pushouts.v
```diff
@@ -0,0 +1,330 @@
...
+    (forall e (h : C ⟦b, e⟧) (k : C⟦c,e⟧)(Hk : p1 ;; h = p2 ;; k),
+        iscontr (total2 (fun hk : C⟦d,e⟧ => dirprod (f ;; hk = h)(g ;; hk = k))))
+    →
+    isPushout p1 p2 f g H.
+  Proof.
+    intros H' x cx k sqr.
+    apply H'. assumption.
+  Defined.
+
+  Lemma MorphismsIntoPushoutEqual {a b c d : C} {p1 : a --> b} {p2 : a --> c}
```
benediktahrens: This lemma would better be called `MorphismsOutofPushoutEqual`.

benediktahrens:
lgtm modulo remark on naming

tpannila:
Hi Benedikt,

thank you for your comment. I fixed the name as you suggested.

UniMath/CategoryTheory/limits/pushouts.v
```diff
@@ -0,0 +1,330 @@
...
+
+  Definition hasPushouts := forall (a b c : C) (p1 : a --> b) (p2 : a --> c),
+      ishinh (Pushout p1 p2).
+
+
+  Definition PushoutObject {a b c : C} {p1 : a --> b} {p2 : a --> c}:
+    Pushout p1 p2 -> C := fun H => pr1 (pr1 H).
+  Coercion PushoutObject : Pushout >-> ob.
+
+  Definition PushoutPr1 {a b c : C} {p1 : a --> b} {p2 : a --> c}
```
benediktahrens: What do you think about renaming `PushoutPr1` to `PushoutIn1`, in analogy with coproducts? 
If those get renamed, the names of the lemmas below should be changed accordingly.

tpannila:
Hi Benedikt,

sorry for the naming. I guess I was not thinking at all when I wrote the file.
I renamed the lemmas you suggested and renamed some variables in other
definitions and lemmas to match in1 and in2.

Issue #381, opened 2016-06-23
Both PartD.v and UnivalenceAxiom.v contain a 'funextfun'


mvr:
This leads to some very confusing error messages if some files import UnivalenceAxiom and some don't. One probably only wants to use the funextfun in PartD, for the nearby proof that it is a weak equivalence.

vladimirias:
I think we should have only one axiom in the library - the univalence axiom. Then we should deduce a number of its standard corollaries such as UAP and funextfun and funextempty and use the UA only through the weakest corollary necessary.

The file with the UA and the deduction of these corollaries from it should go between BasicsB which is axiom free and BasicsC where funextempty is used.

benediktahrens:
On 06/23/2016 09:10 PM, Vladimir Voevodsky wrote:
> I think we should have only one axiom in the library - the univalence
> axiom. Then we should deduce a number of its standard corollaries
> such as UAP and funextfun and funextempty and use the UA only through
> the weakest corollary necessary.

I disagree, for the following reason:
if the proof of a result `foo` only uses funext, then using the axiom
funext will make `Print Assumptions foo` show funext.
If instead one uses the corollary fuext, derived from univalence, then
`Print Assumptions foo` will show univalence.

I think that being able to get more fine-grained information from `Print
Assumptions` is worth having several axioms, even if some could (and
should somewhere) be derived from others.


I agree with the original request, however, that the axioms could be
organized a bit better.

vladimirias:
We can then have the structure I suggested with the UA and all its corollaries being in a file between PartB and PartC and to have, in this file, first funext_lemma proved on the basis of UA and then right after it funext as an axiom. At the header of the file we will explain the reason for this structure.

Then it seems we will have both a better organization, clear statement that all of these axioms are corollaries of the UA and the axioms themselves that can be distinguished with "Print assumptions”.

DanGrayson:
That sounds good to me.  Perhaps one could define funext_statement, as well, so that one
could have
```
Definition funext_statement := ... .

Theorem funext_theorem: funext_statement.
Proof. ... Defined.

Axiom funext_axiom: funext_statement.
```
Then the reader's eye would not have to scan the two statements to check them for equality.

vladimirias:
LGTM

andrejbauer:
The HoTT library has a similar arrangement. We also use type classes (which I think UniMath avoids) so that we automatically derive `funext` from `univalence` when univalence was assumed but `funext` is required.

[...]

PR #387, opened 2016-06-29
Axiom reorganization

DanGrayson:
This addresses #381.

It also removes the obsolete lemmas on eta-correction for functions.

Another option would be to preserve uahp as an axiom.

### DIFF
--- a/UniMath/CategoryTheory/HLevel_n_is_of_hlevel_Sn.v
+++ b/UniMath/CategoryTheory/HLevel_n_is_of_hlevel_Sn.v
@@ -118,7 +118,7 @@ Lemma isofhlevel0pathspace (X Y : UU) :
     iscontr X -> iscontr Y -> iscontr (X = Y).
 Proof.
   intros pX pY.
-  set (H := isofhlevelweqb 0 (tpair _ _ (univalenceaxiom X Y))).
+  set (H := isofhlevelweqb 0 (tpair _ _ (univalenceAxiom X Y))).
   apply H.
   exists (isofhlevel0weq _ _ pX pY ).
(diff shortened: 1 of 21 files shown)


benediktahrens:
Looks great, thanks!

vladimirias:
Sorry Dan, but I don’t like it. We use funext and funextfun a lot and writing funextAxiom instead is a lot of extra typing.

I suggest we have funextTheorem and funext and similarly for others.

DanGrayson:
Hmm, I don't understand your remark, since `funextfun` and `funextsec` are still there, and there is no need to type `funextAxiom`.  There never was a `funext`.

DanGrayson:
Oops, wait a minute, I forgot to even use `funextAxiom` anywhere at all.  It should have been used in the proof of `funextsec`, but instead, it the proof of `funextsec` used univalence.  I'll fix that.

vladimirias:
I think this is what caught my eye:

> Theorem isapropneg (X:UU): isaprop (neg X).
> -Proof. intro.  apply invproofirrelevance . intros x x' .   apply ( funextempty X x x' ) . Defined .
> +Proof. intro.  apply invproofirrelevance . intros x x' .   apply ( funextemptyAxiom X x x' ) . Defined .

DanGrayson:
Yes, I like having axiom names ending with "Axiom".  That particular axiom is used only once, in `isapropneg`.

DanGrayson:
Okay, I fixed that problem and isolated one more axiom from Vladimir's code.  I think it's all
okay now.

vladimirias:
LGTM

Issue #386, opened 2016-06-27
Make rngaddabgr into a coercion?


vladimirias:
I can not think of a reason not to make this function into a coercion and it seems that doing so will allow to simplify a lot of code.

DanGrayson:
It's a good idea.

benediktahrens:
Done in #388 , closing

PR #388, opened 2016-06-29
make rngaddabgr into a coercion

DanGrayson:
and eliminate all but two of its uses.

This addresses #386.


benediktahrens:
Can you say something about the two uses that you have not removed, please? Does the coercion not trigger there? Or was it your choice to retain the maps, as to make things more explicit?

DanGrayson:
They were where `ringaddabr` was used to make the type that a variable was declared to be, and removing it resulted in an error a little later.

Issue #373, opened 2016-06-21
Remove `Admitted.``` in Foundations/Combinatorics/SimplicialSets.v


benediktahrens:
Remove

```
Definition precatDelta : precategory .
Proof.
...
Admitted.
```

DanGrayson:
This file is commented out in the list of files, so it doesn't ever get compiled.  It was an experiment of Vladimir.  I'll re-enable it and remove the "Admitted".

But it is a file in Foundations that requires files from CategoryTheory, breaking our convention that the list of "packages" can be linearly ordered.  Should it be put somewhere else?

benediktahrens:
On 06/29/2016 10:31 PM, Daniel R. Grayson wrote:
> This file is commented out in the list of files, so it doesn't ever
> get compiled.  It was an experiment of Vladimir.  I'll re-enable it
> and remove the "Admitted".
> 
> But it is a file in Foundations that requires files from
> CategoryTheory, breaking our convention that the list of "packages"
> can be linearly ordered.  Should it be put somewhere else?

Yes, I think it should go into CategoryTheory, if Vladimir agrees.
Would you amend your PR accordingly?

DanGrayson:
Sure.

vladimirias:
It certainly should be taken out of foundations.

There are also two (I think) mistakes there. One is that simplicial sets are defined as covariant functors instead of contravariant (if I remember right) and there was another one which I can not recall right now.

vladimirias:
Not to category theory. We may start a new large folder called “HomotopyTheory” and put it there.

DanGrayson:
Sounds good to me.

benediktahrens:
This `Admitted.` is still present.

PR #389, opened 2016-06-29
fix file SimplicialSets

DanGrayson:
resolves https://github.com/UniMath/UniMath/issues/373


DanGrayson:
See discussion in https://github.com/UniMath/UniMath/issues/373 , further changes contemplated...

PR #390, opened 2016-06-29
change notation for "forall"

### LINKED ISSUE #369: Introducing "forall" for hProp
I suggest that we renaim the current "forall" notation into the \Pi as it should be and introduce \forall for the dependent product of a family of hProp's with values in hProp - a generalization of the current himpl. 

It will then be similar to how we have two notations one for hexists and one  for total2.

DanGrayson:
We rename our notation for ∀ to Π.

We make a global definition for ∀ that applies just to families of
propositions and yields a proposition (hProp).

We change all existing uses of "forall" and ∀ to Π.  Later on, those
that applied to hProps could be switched back to ∀.

The reason we change all uses of "forall" is that ∀ is normally read as
"for all".

Here is the definition of ∀ from Propositions.v:
```
    Lemma isaprop_forall_hProp (X:UU) (Y:X->hProp) : isaprop (Π x, Y x).
    Proof. intros. apply impred_isaprop. intro x. apply propproperty. Defined.

    Definition forall_hProp {X:UU} (Y:X->hProp) : hProp := hProppair (Π x, Y x) (isaprop_forall_hProp X Y).

    Notation "∀  x .. y , P" := (forall_hProp (fun x => .. (forall_hProp (fun y => P)) ..))
      (at level 200, x binder, y binder, right associativity) : type_scope.
```

This resolves #369 .

### DIFF
--- a/UniMath/CategoryTheory/Epis.v
+++ b/UniMath/CategoryTheory/Epis.v
@@ -11,16 +11,8 @@ Section def_epi.
   Variable C : precategory.
 
   (** Definition and construction of isEpi. *)
-<<<<<<< HEAD
-  Definition isEpi {x y : C} (f : x --> y) :=
-    Π (z : C) (g h : y --> z), f ;; g = f ;; h -> g = h.
-||||||| merged common ancestors
-  Definition isEpi {x y : C} (f : x --> y) :=
-    forall (z : C) (g h : y --> z), f ;; g = f ;; h -> g = h.
-=======
   Definition isEpi {x y : C} (f : x --> y) : UU :=
-    forall (z : C) (g h : y --> z), f ;; g = f ;; h -> g = h.
->>>>>>> 2c0c40deb560dc855144f924c8a63184d7e0462d
+    Π (z : C) (g h : y --> z), f ;; g = f ;; h -> g = h.
   Definition mk_isEpi {x y : C} (f : x --> y) :
     (Π (z : C) (g h : y --> z), f ;; g = f ;; h -> g = h) -> isEpi f.
   Proof. intros X. unfold isEpi. apply X.  Defined.
--- a/UniMath/CategoryTheory/Epis.v
+++ b/UniMath/CategoryTheory/Epis.v
@@ -12,9 +12,9 @@ Section def_epi.
 
   (** Definition and construction of isEpi. *)
   Definition isEpi {x y : C} (f : x --> y) :=
-    forall (z : C) (g h : y --> z), f ;; g = f ;; h -> g = h.
+    Π (z : C) (g h : y --> z), f ;; g = f ;; h -> g = h.
   Definition mk_isEpi {x y : C} (f : x --> y) :
-    (forall (z : C) (g h : y --> z), f ;; g = f ;; h -> g = h) -> isEpi f.
+    (Π (z : C) (g h : y --> z), f ;; g = f ;; h -> g = h) -> isEpi f.
   Proof. intros X. unfold isEpi. apply X.  Defined.
 
   (** Definition and construction of Epi. *)
--- a/UniMath/CategoryTheory/HLevel_n_is_of_hlevel_Sn.v
+++ b/UniMath/CategoryTheory/HLevel_n_is_of_hlevel_Sn.v
@@ -46,7 +46,7 @@ Defined.
     the hProposition [P] with this fibration.
 *)
 
-Lemma ident_is_prop : forall (P : UU -> hProp) (X X' : UU)
+Lemma ident_is_prop : Π (P : UU -> hProp) (X X' : UU)
       (pX : P X) (pX' : P X') (w : X = X'),
    isaprop (transportf (fun X => P X) w pX = pX').
 Proof.
@@ -135,7 +135,7 @@ Defined.
 
 (** *** The case [n = S n'] *)
 
-Lemma isofhlevelSnpathspace : forall n : nat, forall X Y : UU,
+Lemma isofhlevelSnpathspace : Π n : nat, Π X Y : UU,
       isofhlevel (S n) Y -> isofhlevel (S n) (X = Y).
 Proof.
   intros n X Y pY.
@@ -154,7 +154,7 @@ Defined.
 
 (** ** The lemma itself *)
 
-Lemma isofhlevelpathspace : forall n : nat, forall X Y : UU,
+Lemma isofhlevelpathspace : Π n : nat, Π X Y : UU,
       isofhlevel n X -> isofhlevel n Y -> isofhlevel n (X = Y).
 Proof.
   intros n.
@@ -178,7 +178,7 @@ Definition HLevel n := total2 (fun X : UU => isofhlevel n X).
 
 (** * Main theorem: [HLevel n] is of hlevel [S n] *)
 
-Lemma hlevel_of_hlevels : forall n,
+Lemma hlevel_of_hlevels : Π n,
       isofhlevel (S n) (HLevel n).
 Proof.
   intro n.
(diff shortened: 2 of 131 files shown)


vladimirias:
I think that the use of “forall” should be permitted. It is a Coq identifier and someone who starts to write in UniMath may prefer to use it rather than to go through an extra hurdle of learning how to use unicode.

DanGrayson:
Okay, permit "forall".  I didn't add a rule forbidding it in this pull request, so I guess no change is required.

benediktahrens:
I don't like the bundling into an hProp that is done by the unicode forall. 
We have seen that this makes *definitions* depend on axioms, e.g., in the case of `isInjectiveFunction` in a PR by @tpannila , even when one does not use the fact that the defined type is a proposition.

DanGrayson:
Is there no feature of Coq that allows us to use pr1 of a pair in a proof
without also using pr2?  There should be.  If not, maybe we need to replace
isInjectiveFunction by isInjectiveFunction and isInjectiveFunction_hProp.
Other examples of such pairs are (ishinh_UU, ishinh), (unit,htrue),
(empty,hfalse).

Users can use Π instead of ∀ if they prefer to get a type.  Or maybe what you really want is
to not encourage further use of hProp by making new easy notation for it.  (By
the way, it already exists in the master branch in the scope "prop".)

We could limit it once again to a scope with name "prop" to make it harder to
do.

By the way, there is a Π for hSet's in the scope with the name "set".

By the way, also, Print Assumptions is broken on the trunk, and prints out way
too much.

benediktahrens:
On 07/02/2016 11:29 PM, Daniel R. Grayson wrote:
> Is there no feature of Coq that allows us to use pr1 of a pair in a proof
> without also using pr2?  

Well, the types `A` and `fun x : A => B x` are arguments to `pr1 :
total2 B -> A`, so even if you don't use `pr2`, the family `B` will
still be mentioned.
I don't know if primitive projections are different in that respect.

> There should be.  If not, maybe we need to replace
> isInjectiveFunction by isInjectiveFunction and isInjectiveFunction_hProp.
> Other examples of such pairs are (ishinh_UU, ishinh), (unit,htrue),
> (empty,hfalse).
> 
> Users can use Π instead of ∀ if they prefer to get a type.  Or maybe what you really want is
> to not encourage further use of hProp by making new easy notation for it.  (By
> the way, it already exists in the master branch in the scope "prop".)

I am not sure what should be done. One problem is that the issue - an
"invisible" use of an axiom" - is difficult to identify, in particular
when you have never seen such an issue before.
So, yes, I would like to discourage use of the bundled hProps, at least
when the proof of `isaprop` needs an axiom.


However, since the axiom is only used in the proof of `isaprop`, it
probably does not get in the way of computation - the first projection
should still compute well, even when computing a proposition with
computational content (e.g., `isweq`). The issue is only in that it
creates a formal dependency on funext, reported by `Print Assumptions`.

DanGrayson:
On Sun, Jul 3, 2016 at 6:41 AM, Benedikt Ahrens <notifications@github.com>
wrote:

> On 07/02/2016 11:29 PM, Daniel R. Grayson wrote:
> > Is there no feature of Coq that allows us to use pr1 of a pair in a proof
> > without also using pr2?
>
> Well, the types `A` and `fun x : A => B x` are arguments to `pr1 :
> total2 B -> A`, so even if you don't use `pr2`, the family `B` will
> still be mentioned.
> I don't know if primitive projections are different in that respect.
>

Right.  But many programming languages have "macro facilities" that allow
code to run
at compile time.  One could imagine adding such a facility to Coq so that a
definition involving
`pr1 ab` could have that reference simplified to `a` if `ab` has been
defined explicitly as a pair `(a,b)`.  Then
the resulting definition would not refer to `ab`, and thus also not to
`b`.  Perhaps such a facility
could be designed for Coq.

DanGrayson:
I'd love it if we could accept this pull request before any others and postpone further discussion for another issue, since every other pull request accepted makes a conflict with this one that I have to resolve.

I'll repair the conflicts again ...

vladimirias:
I agree that in most cases one should use \Pi.

vladimirias:
> On Jul 2, 2016, at 5:29 PM, Daniel R. Grayson <notifications@github.com> wrote:
> 
> By the way, there is a Π for hSet's in the scope with the name "set".


This we should not use it would be confusing to have two uses of \Pi. We can have \Piset for \Pi in hSet.

DanGrayson:
Well, we already have many identical notations distinguished from each other only by scope, such as 0, 1, +, and *.  How is Π for hSet different from 0 for nat and additive monoids?

vladimirias:
This is true, but somehow I feel that there is a difference. Isn't \Piset a good notation?

DanGrayson:
It does make it more visually clear.  Okay, will change...

JasonGross:
Note that you can do something like
`Definition foo := Eval simpl @pr1 in pr1 (x, y).` or something, and sometimes this might give you what you want.

DanGrayson:
Re: "Isn't \Piset a good notation?"

Let me make one more attempt at convincing: in set theory, the standard notation for the product of sets is Π, so why not try to use that here?  Mathematicians will like it.

Coercions allow the set-product of sets to be used where a type is expected, and we don't mind keeping that conversion invisible.  So why bother making visible the reverse situation, where product of sets is used as a set?

DanGrayson:
Jason, how would you make it work in this context?:
```
Require Import UniMath.Foundations.Basics.Sets.
Definition i { X Y : hSet } (f:X -> Y) : UU. Proof. exact (isInjectiveFunction f). Defined.
Set Printing All.
Print i.
(*
    fun (X Y : hSet) (f : forall _ : pr1hSet X, pr1hSet Y) =>
    hProptoType (@isInjectiveFunction X Y f)
         : forall (X Y : hSet) (_ : forall _ : pr1hSet X, pr1hSet Y), UU
  *)
Definition i' { X Y : hSet } (f:X -> Y) : UU := Eval simpl hProptoType in (i f).
Print Assumptions i'.
```

vladimirias:
Sorry Dan, but I am unconvinced. In fact, it will be useful for explaining things to mathematicians to have both \Pi and \Piset. We may eventually also consider \Pilevel for \Pi of types of a given h-level.

DanGrayson:
I have similar notations in scope "set" for a=b, a!=b, and X×Y.  Do you object to those, too?

JasonGross:
@DanGrayson You can try one of these:
```coq
Arguments hProptoType !_ / .
Arguments i / .
Arguments isInjectiveFunction / .
Definition i' { X Y : hSet } (f:X -> Y) : UU := Eval simpl in (i f).
```
or
```coq
Definition i' { X Y : hSet } (f:X -> Y) : UU := Eval hnf in (i f).
```
or
```coq
Definition i' { X Y : hSet } (f:X -> Y) : UU := Eval cbv [i hProptoType isInjectiveFunction] in (i f).
```

Do any of these suit your needs?

vladimirias:
> On Jul 5, 2016, at 9:12 PM, Jason Gross <notifications@github.com> wrote:
> 
> Arguments hProptoType !_ / .
> Arguments i / .
> Arguments isInjectiveFunction / .
Sorry, what do these mean?

DanGrayson:
I don't understand the code, but it looks too special and cryptic to be useful.  Imagine that the use of `hProptotype` arising from a coercion is deep inside the body of the definition of i, or that there are multiple things like that that should be handled also.

vladimirias:
I suppose I don’t and I suppose that it makes little sense then to object to \Pi so I take back my objection.

DanGrayson:
Thank you for reconsidering.

We can think of it this way: a universe, such as hSet, is closed under all the interesting operations,
and a command like `Open Scope set` arranges for Coq to take care of that for us in a few cases.

PR #393, opened 2016-06-30
Primitive projections

DanGrayson:
Various changes are needed to our code to work with primitive projections.  I'm at an
impasse while waiting for a Coq bug fix, so I submit these changes here.  (I don't turn on
primitive projections in this request, I just make changes to the proofs that ensure they
will work both with and without primitive projections.

I also add a utility displaying which are the slowest steps in a proof.

And I also add something to the style guide about explicitly naming new variables introduced with 
`induction`.  (With a new definition of `total2`, the names invented for the parts by Coq
changed, annoyingly in many places.)

This is progress toward https://github.com/UniMath/UniMath/issues/111


benediktahrens:
LGTM

benediktahrens:
Sorry, more merge conflicts...

benediktahrens:
Is there a reason why you don't use the tactic `use foo` instead of `refine (foo _ _ _ _ )`? If `use` doesn't work, we should fix it.

DanGrayson:
No, it's just that using built-in Coq commands seems clearer to me, and I got into the habit of typing that before the "use" tactic was introduced.

DanGrayson:
The latest changes include fixing all those uses of nested `intros` tactics whose behavior has changed on the trunk of Coq.  The main difference is that `intros (a,b,c,...)` needs as many parts as the inductiive object has.

DanGrayson:
This one is ready.

PR #399, opened 2016-07-03
Additive and Abelian categories

tpannila:
Hi,

this pull request contains the following
- Definition of the precategory consisting of setswithbinops.
  (category_binops.v)
- Definition of the precategory consisting of abelian groups.
  (category_abgr.v)
- Definition of additive category.
  (Additive.v)
- A proof that the category consisting of abelian groups is additive.
  (ABGR_additive in category_abgr.v)

I have not been able to do the following
- Verify that binop_precategory and abgr_precategory are actually categories and
  not just precategories with has_homsets.
- Formalization of abelian categories and the fact that the category of abelian
  groups is an abelian category.

I guess the problem in the latter is that I don't know how to work with hProp.
In particular, I don't know how to make an equality of morphisms in a
precategory with has_homsets into an hProp, if this equality is the goal. This, is
needed(?) in proving that kernels exist in abgr_precategory. Similarly, in the
definition of abelian categories I have used \forall and \exists instead of
forall and exists, and because of these I don't know how to construct kernels
and cokernels in abelian category. Please, see the end of the files
category_abgr.v and Abelian.v .

Both category_binops.v and category_abgr.v use funextfun.

If you want that I break this pull request into smaller pull requests please
let me know. All comments are welcome.

### DIFF
--- a/UniMath/CategoryTheory/Abelian.v
+++ b/UniMath/CategoryTheory/Abelian.v
@@ -20,6 +20,8 @@ Require Import UniMath.CategoryTheory.limits.equalizers.
 Require Import UniMath.CategoryTheory.limits.coequalizers.
 Require Import UniMath.CategoryTheory.limits.kernels.
 Require Import UniMath.CategoryTheory.limits.cokernels.
+Require Import UniMath.CategoryTheory.limits.pushouts.
+Require Import UniMath.CategoryTheory.limits.pullbacks.
 Require Import UniMath.CategoryTheory.Monics.
 Require Import UniMath.CategoryTheory.Epis.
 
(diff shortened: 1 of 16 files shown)


benediktahrens:
You can use  `Search ( ∥ _ ∥ ) .` to find lemmas and constructions involving propositional truncation. Note that `∥` is typed `\||`.
In your case, you want to use the following tactic: `use (squash_to_prop p).`, where `p` is the hypothesis of the truncated type.

benediktahrens:
In Abelian, try this, same as in my previous comment:
```
  Definition Abelian_EpiToCokernel {y z : A} (E : Epi A y z) :
    Cokernel Abelian_Zero (KernelArrow _ (Abelian_Kernels _ _ E)).
  Proof.
    set (AEC := Abelian_isAbelianEpiCokernels y z E).
    use mk_Cokernel.
    exact z. exact (EpiArrow _ E).
    use (squash_to_prop AEC).
    - apply hs.
    - intros.
```

benediktahrens:
Proving that a precategory is a category is often non-trivial. In the case of `BINOP`, the situation gets more complicated by our use of a slightly strange definition of isomorphism in precategories.
Here are some hints, in a random order. Please take a look at them, and ask more questions afterwards:

- Instead of using `is_isomorphism`, you almost always want to use `is_z_isomorphism`. There is also a lemma `is_iso_qinv` that often comes in handy.
- In order to construct an equivalence between `a = b` and `iso a b` for binops a and b, you will probably construct some intermediate equivalences. 
Maybe something like this:
```
a = b
\simeq
(X,m) = (Y,n)
\simeq
Sigma (H : X = Y), transport H m = n
\simeq
Sigma (f : X \equiv Y), f o m o f^{-1}^2 = n
\simeq
Sigma (f : X -> Y)(g : Y -> X), (are_inverses f g) \times (f o m o f^{-1}^2 = n)
\simeq
Sigma (f : X -> Y)(g : Y -> X), (are_inverses f g) \times (f o m o f^{-1}^2 = n) \times (g o n o g^{-1}^2 = m)
\simeq
Sigma (f : X -> Y),(f o m o f^{-1}^2 = n) \times \Sigma (g : Y -> X), \times (g o n o g^{-1}^2 = m) \times (are_inverses f g)
\simeq
z_iso a b
\simeq
iso a b
```
- afterwards you can show that your constructed equivalence is pointwise equal (`homot`) to `idtoiso`. Combine this with `isweqhomot` to get that `idtoiso` is an equivalence.
- useful functions to construct equivalences of sigma types are `weqbandf` (equivalence in base and fibers). Maybe a look at the file slicecat.v can be helpful.
- Another way to prove that `BINOP` is a category is to directly define a function that goes from isos to identities, and show that they are mutually inverse, using `gradth`. Actually, from @peterlefanulumsdaine , I learned about the following lemma
```
Lemma eq_equiv_from_retraction {A} {P : A -> A -> UU} 
    (w : forall a b, a = b -> P a b)
    (v : forall a b, P a b -> a = b)
  : (forall a b (p : P a b), w _ _ (v _ _ p) = p)
  -> forall a b, isweq (w a b).
Proof.
  intros wv a.
  apply isweqtotaltofib. (* first of the two key steps *)
  use gradth.
  - intros bp. exists (pr1 bp). apply v, (pr2 bp).
  - intros be; apply connectedcoconusfromt. (* the second key step *)
  - intros bp. use total2_paths. apply idpath. apply wv.
Qed.
```
which shows that one only needs to provide one family of equalities.
This lemma is not yet in UniMath, but should be there.

benediktahrens:
I suggest you finish the proofs involving truncation, and then we merge this PR.
The proof that your precategories are categories can go into a subsequent PR.

tpannila:
Hi Benedikt,

thank you for your comments and advice. Using `squash_to_prop` , I have been
able to prove that kernels exist in the precategory of abelian groups. I have some
difficulties with cokernels, but I try to work those out. I will push an update on my
progress today or tomorrow.

I will leave the proofs that the precategories are categories to subsequent
pull requests.

tpannila:
I have now proved that the precategory of abelian groups has kernels and
cokernels. I think I will leave the proofs that monics are kernels and epis
are cokernels to my next pull request.

I have encountered a new problem in Abelian.v . In Abelian_monic_kernel, line
134, I construct a kernel such that the kernel arrow is given by a monic M.
In Abelian_subobjects_isPullback, line 226, I use this construction, but the
reference to monic M is lost when I try to apply it. Does this have something
to do with \ex ? I really need coq to remember the construction.

benediktahrens:
On 07/07/2016 10:07 AM, Tomi Pannila wrote:
> I have now proved that the precategory of abelian groups has kernels and
> cokernels. I think I will leave the proofs that monics are kernels and epis
> are cokernels to my next pull request.
> 
> I have encountered a new problem in Abelian.v . In Abelian_monic_kernel, line
> 134, I construct a kernel such that the kernel arrow is given by a monic M.
> In Abelian_subobjects_isPullback, line 226, I use this construction, but the
> reference to monic M is lost when I try to apply it. Does this have something
> to do with \ex ? I really need coq to remember the construction.

In UniMath (and in univalent foundations in general) we can distinguish
between structures and properties.
When you use \ex in the definition of something, then it'll be a
property, and the construction won't be remembered.
If you want a structure, use \Sigma instead.

DanGrayson:
At that point, you're proving `isPullback M1 M2 p1 p2 H`, which is a proposition, so you can unfold the squashed type to get the construction.  See `hinhuniv` and its corollaries.

tpannila:
Hi Benedikt and Dan,

I replaced the \ex stuff by \Sigma and it works perfectly. It will take me a day or two to do Abelian.v properly .

tpannila:
I fixed Abelian.v . The proofs in Abelian.v are unreadable because there
are no diagrams which explain what I do in the proofs. Thus one might want to
view the enhanced-html file of Abelian.v, generated by `make doc`, before
looking at Abelian.v . I have basically followed Borceux Handbook of
Categorical Algebra 2 and one can take a look at it if one wants to see some
diagrams. If one doesn't have this book, then my Master's thesis,
(The site which hosts it is down at the moment. I will add a link here
 when the site is working again), pages 23-26, contain the same stuff.

Some of the proofs contain more equations, proved by using assert, than necessary.
I will remove the unnecessary equations in my next pull request, or I can remove them
now if people want to. 

Here is a short summary of the results one finds in Abelian.v , in the order
they occur in the file:
- Definition of Abelian categories.
- A proof that every morphism which is monic and epi is an isomorphism.
- Pullbacks of subobjects and pushouts of quotient objects.
- Equalizers and Coequalizers.
- A proof that the canonical morphism CoIm f -> Im f is an isomorphism, and
  that every morphism can be factored as Epi ;; Monic in 2 canonical ways,
  by interpreting the isomorphism CoIm f -> Im f as an epi and as a monic.
- Construction of kernel (resp. cokernel) for every monic (resp. epi) so
  that the monic (resp. epi) is the KernelArrow of the kernel (resp.
  cokernelarrow of the cokernel).

Please let me know if there is something you would like to be fixed.

vladimirias:
Why not to put your thesis on your GitHub? I have several pdf’s on my GitHub.

vladimirias:
l. 189 in abelian.v “if” should be “is”

vladimirias:
l. 241 you say “pullbacks of subobjects” but you only construct pullbacks of monics w.r.t. monics, not monics w.r.t. any morphism. And similarly for epis. Why?

vladimirias:
In abelian.v I did not see where you state or prove that binary coproducts are also binary products, i.e., that A is additive. Do you state or prove it?

vladimirias:
In category_abgr l. 127 why is there “h” in the name, why not just abgr_iso_equiv?

vladimirias:
In l. 194 there also seems to be an extra h in the name.

vladimirias:
In the construction of cockernels in ABGR isn’t it easier to define images first and then cokernels as quotient groups by images?

tpannila:
Hi Vladimir,

thank you for your comments. Indeed, it is better to have my Master's thesis on
github.
- I uploaded my thesis to Github,
  https://github.com/tpannila/Masters_thesis . 
- I fixed the "if" to "is" and the names with extra h you found.
- I do not state or prove that abelian categories are additive. This
  requires some extra work.
- In category theory subobject of an object X is an equivalence class of monics
  to X. This result is needed to prove that equalizers exist in abelian
  categories. Similarly, pushouts of quotient objects are needed for
  coequalizers. After one has proved that equalizers exist, it should follow
  from general category theory that pullbacks exist. If category has terminal
  object, equalizers, and binary products, then it has all pullbacks.
  Similarly for pushouts.
- In category_abgr.v the image is constructed in `ABGR_image`, line 853, but
  is never used. The reason is that I form the cokernel by using `abgrquot`
  in `ABGR_cokernel_abgr`, line 1019, which needs binopeqrel on B. The definition 
  `abgrquot` is in Foundations/Algebra/Monoids_and_Groups.v, line 738.

benediktahrens:
- I won't have time to look at this PR for another 10 days or so.
- There are merge conflicts at the moment.

UniMath/CategoryTheory/Abelian.v
```diff
@@ -0,0 +1,1328 @@
...
+  (** Data which contains kernels, cokernels, the data that monics are kernels
+    of some morphisms, and the data that epis are cokernels of some
+    morphisms. *)
+  Definition AbelianData (C : precategory) (AD1 : AbelianData1 C) : UU :=
+    (AbelianData2 C AD1)
+      × (AbelianMonicKernelsData C AD1)
+      × (AbelianEpiCokernelsData C AD1).
+
+  (** Definition of abelian categories. *)
+  Definition Abelian : UU :=
```
benediktahrens: Could you call this `Abelian_precategory` or similar, please, and not just `Abelian`?

UniMath/CategoryTheory/Abelian.v
```diff
@@ -0,0 +1,1328 @@
...
+    Σ A : (Σ C : precategory, AbelianData1 C), AbelianData (pr1 A) (pr2 A).
+  Definition Abelian_precategory (A : Abelian) :
```
benediktahrens: This could be called `precategory_of_abelian_precat` or another unwieldy name, since you never use it explicitly.

UniMath/CategoryTheory/Abelian.v
vladimirias: Right.

tpannila:
I fixed the two names Benedikt suggested and rebased. Should I replace the
prefix Abelian by Abelian_precategory in Abelian.v ?

Also, I added some new results:
- Pushouts and Pullbacks in Abelian_precategory. See lines 829 and 836 of
  Abelian.v .
- A morphism is monic if its kernel is zero, line 922 Abelian.v , and a
  morphism is epi if its cokernel is zero, line 1032 Abelian.v.
- I started to formalize the fact that Abelian categories are additive in
  AbelianisAdditive.v . This is work in progress.

tpannila:
I proved that Abelian precategories are Additive, line 975 of AbelianisAdditive.v .

tpannila:
I proved that Monics are inclusions, line 1708 of category_abgr.v, and Epis are
surjections, line 1551 of category_abgr.v, in the caegory of abelian groups.
Also, I changed some Defined to Qed so that unfold and cbn work better.
There are a few duplicates in category_abgr.v . I try to remove these in a
few days.

tpannila:
I think I formalized that Monics are Kernels in ABGR, line 1986 of
category_abgr.v . In the [paper](http://assert-false.net/arnaud/papers/Towards%20constructive%20homological%20algebra%20in%20type%20theory.pdf) , that @mortberg linked in #368 , 
it is said on page 6, line 6,
```
(that is, a monomorphism for instance is not necessarily the kernel of a map).
```
Thus, I guess this shows that in UniMath we are able to solve this problem
(not sure yet, have to recheck my definitions and proofs).

I will try to show that ABGR, the category of abelian groups, is abelian. If
I manage to do this, then I will not add any new results after it and I start
to clean this pull request.

DanGrayson:
That paper seems to be about Bishop sets, not h-sets.

tpannila:
I was able to show that ABGR is an Abelian_precategory, line 2427 of
category_abgr.v . I hope this gives an answer for what is the status of 
the category of abelian groups in UniMath. It is an abelian category. I 
guess this is what @mortberg wished to know in his comment in #368 .

Now I start to clean the proofs in this pull request.

@DanGrayson : You are right. Please read @mortberg 's comment in #368 .

tpannila:
Added some comments and simplified some proofs. There are still a few proofs to
clean. For example, in Abelian.v , the proofs of the following results
- Abelian_subobjects_Pullback, line 288
- Abelian_quotobjects_Pushout, line 452
- Abelian_Equalizer, line 635
- Abelian_Coequalizer, line 747

are long and use asserts. I will try to do this tomorrow.

DanGrayson:
Using "assert" is not a problem.

mortberg:
Indeed, the paper is representing categories using Bishop sets/setoids. This used to be the way to do category theory in type theory prior to Univalent Foundations. The most common argument I've heard against this approach is that it gets very cumbersome to work with in a proof assistant (for example rewriting becomes slow). However the paper I linked to before shows that you cannot prove that the category of abelian groups is an abelian category with this definition in Coq (without additional axioms). @tpannila  has now showed that this is provable in UniMath. 

@tpannila: What axioms are you relying on? (You can use `Print Assumptions` to get this information). Also, are you using `hexists` anywhere or are all definitions expressed with `total2`?

DanGrayson:
Re: "@tpannila has now showed that this is provable in UniMath."

Not sure what you mean -- it's a different "this" when h-sets are used.

tpannila:
@mortberg , `Print Assumptions ABGR_Abelian_precategory` gives the
following output.
```
Axioms:
propositionalUnivalenceAxiom : propositionalUnivalenceStatement
isweqtoforallpathsAxiom : isweqtoforallpathsStatement
funcontrAxiom : funcontrStatement
Theory:
Type hierarchy is collapsed (logic is inconsistent)
```
I do not use hexists in definitions but I use it indirectly in proofs about
integers, because the definition of integers uses hexists, and I have used 
something  similar to hexists in  `ABGR_setquot_iso`  (I will fix this name), 
line 1559,  category_abgr.v .  This because `issurjective` by definition is 
something  similar to hexists. 

The fact that integers use hexists can be seen by using 
`Print All Dependencies hzaddabgr` . 

To see the files where hexists is used one can use `git grep --count hexists` .
This gives the following output.
```
UniMath/Foundations/Algebra/BinaryOperations.v:2
UniMath/Foundations/Algebra/Monoids_and_Groups.v:4
UniMath/Foundations/Basics/Propositions.v:8
UniMath/Foundations/Combinatorics/StandardFiniteSets.v:7
UniMath/PAdics/fps.v:2
UniMath/PAdics/frac.v:2
UniMath/PAdics/lemmas.v:12
UniMath/PAdics/padics.v:14
UniMath/PAdics/z_mod_p.v:2
UniMath/RealNumbers/NonnegativeReals.v:13
UniMath/RealNumbers/Sets.v:1
```
Is hexists problematic? 


@DanGrayson , Yes, you are right. Here is the reason I need to clean those
proofs. The proofs are closed with Defined. This is needed because the proofs
construct structures and we want to remember the construction. To make cbn and
unfold compute better, what we really want to remember is what objects and
morphisms we used in the construction, not why these objects and morphisms
satisfy the properties the structures require.

For example, consider the Abelian_subobjects_Pullback, line 288 Abelian.v . In 
this case we want to remember the pullback object and the projections from the 
pullback object. We do not need to remember the why this pullback object 
and projections form a pullback and thus we hide this behind Qed. When this is
done, if one unfolds Abelian_subobjects_Pullback one sees just the contruction, 
not why  it works. This should speed up type checking.

tpannila:
I cleaned the proofs in Abelian.v . Please let me know if there are any changes
you would like to be made.

mortberg:
@DanGrayson Yes, I guess my point is that UniMath is a good language for formalizing constructive algebra in type theory. We can now define notions in a "correct" way so that the properties we want to prove can be proved constructively.

mortberg:
@tpannila Interesting! The use of `hexists` or propositional truncation is not problematic, I was just curious where it was used.

benediktahrens:
I have checked this PR superficially and formally, but not contentwise (e.g., for minimality of hypotheses). I don't think I'll have time to do the latter for a while.
Since this PR has been around for a while, and is formally ok (no axioms etc.), I suggest merging it now. Improvements can still be suggested and done later.

PR #405, opened 2016-07-09
Updated doc for substitution systems

rmatthes:
just an overhaul of the README so that precisely the existing files are described
doc of files that are now elsewhere has been moved to the relevant README - see the commit log


vladimirias:
LGTM

PR #409, opened 2016-07-20
Bicategories Definitions

mvr:
We define prebicategories, internal equivalences and bicategories. We prove that Cat and PreCat are prebicategories (but not that Cat is a bicategory).

There is also some initial work on showing that paths between precategories are equivalent to isomorphisms of precategories, as defined in the univalent categories paper. There are maps (A = B) <-> (catiso A B), but proving that they are inverse turned out to be beyond me.


benediktahrens:
The file bicategories/bicategory.v is missing, but included in .package/files, and required in internal_equivalence.v.
Could you add the file to the git repository, please?

mvr:
Thanks, I couldn't tell what the error was!

UniMath/CategoryTheory/catiso.v
```diff
@@ -0,0 +1,499 @@
...
+Require Import UniMath.CategoryTheory.whiskering.
+
+Local Notation "a --> b" := (precategory_morphisms a b)(at level 50).
+Local Notation "f ;; g" := (compose f g) (at level 50, format "f  ;;  g").
+Local Notation "[ C , D , hs ]" := (functor_precategory C D hs).
+Local Notation "# F" := (functor_on_morphisms F)(at level 3).
+
+(* General lemmas that should probably go somewhere else *)
+
+Lemma eqweq_twice_is_eqweq_of_comp {A B C : UU}
```
benediktahrens: This should go into Basics/UnivalenceAxiom. A better name might be `eqweqmap_pathscomp0`.

UniMath/CategoryTheory/catiso.v
```diff
@@ -0,0 +1,499 @@
...
+Lemma eqweqfun_twice_is_eqweqfun_of_comp {A B C : UU}
+      (p : A = B) (q : B = C) (a : A)
+  : (eqweqmap q) ((eqweqmap p) a)
+    = eqweqmap (pathscomp0 p q) a.
+Proof.
+  induction p.
+  reflexivity.
+Defined.
+
+Lemma inv_idweq_is_idweq {A : UU} :
```
benediktahrens: Should go in Basics.PartA.

benediktahrens:
Could you put the contents of catiso.v into suitable files?
The definition of when a functor is an isomorphism should to into functor_categories.v.
Most other stuff should go somewhere in Basics - let me know if you don't find a suitable place.

vladimirias:
Let me make a somewhat global comment. 
I would define a prebicategory as a precategory with some extra structure - the precategory structure on the homtypes, functor structure on the composition, pentagons and triangles for the unity paths. 
Then many things that we have done for precategories that does not require homsets will apply directly. For example, a bifunctor between two prebicategories will be a functor between the underlying precategories plus again some extra structure.
To be able to define (pre-)bicategories in this way was the reason for developing as much as possible of the precategory theory without the homsets assumption. 
Your approach, it seems, does not let you to use a lot of stuff that would otherwise be useful.

mvr:
My proofs of those two lemmas uses isapropisweq, which only shows up in PartD. Any suggestions?

A more critical question is: Is there any hope for this pull request? Or would it need to be redone using the other structure to be accepted?

vladimirias:
It can certainly be accepted with this approach, but I am looking forward to the other approach tried as well. Maybe I am mistaken and the approach that I have suggested is not more efficient than yours except in the first few definitions.

I am also interested in how hard or easy is it to compare the two approaches. There might be a difference showing up in the possible concepts of partial univalence of a bi-category.

I prefer here the terminology of category/univalent category instead of precategory/category because we can say something like (0,1)-univalent bi-category for the one where Mor-categories are univalent and (1,0)-univalent for the one where Mor-categories are not univalent but 2-equivalence of objects is the same as their identity.

However, we can certainly accept this PR and discuss these issues afterwards as long as we think of a way to distinguish two approaches by having two different names for two different definitions of a bi-category.

vladimirias:
binproduct_precategory etc. should be precategory_.._binproduct.

It is like grquot not quotgr.

vladimirias:
Main definitions such as precategory_ob_mor_binproduct look much better when they are direct, without the use of “Proof”.

vladimirias:
Also, please make such definitions a little more verbose. I can not understand, for example, in precategory_ob_mor_binproduct, what are the types of cd and cd’.

BTW, if you don’t like to write total2 ( fun …. ) there is now a system of notation introduced by Benedikt where you can use \Sigma that is translated into total2 (see preamble.v in Basics).

vladimirias:
binprodcatpair should be precatbinprodpair (or precat_binprod_pair).

vladimirias:
There are a few lines that are longer than 100 characters.

vladimirias:
There is productprecategory.v in CategoryTheory with the definition of a product of a family of precategories. Is there somewhere the binary product in CategoryTheory?

vladimirias:
In Catlike_associator and maybe other similar complex definitions it makes a lot of sense to use notations that makes it possible to see such definitions as a whole as opposed to having to read it as a long text.

vladimirias:
Also, Cat should be bicat etc, shouldn’t it?

vladimirias:
Why Catlike_associator needs homsets? nat_trans should use only functor_data and precategory_data.

vladimirias:
There is binproduct_precategory in CategoryTheory.

It is btw, binproduct_precategory there not precategory_binproduct. Is it how we name the product of two groups or monoids?

vladimirias:
I also want to make a general comment, which I think is important.

While the difference between _ob_mor and _data is somewhat arbitrary, the difference between _data and all kinds of extra information that uses the identity types is very clear and should be kept such in the future.

Up to the level of _data we do not use identity types and this is what separates it from the next levels, which we do not call “data" even though, unless has_homsets is specified, the associativity is also data.

We should keep this distinction. In the future, for example in the type theory of Angiuli-Harper, there will be more than one identity type and using one or another in the structures above the _data will create different concepts. I can see even now how some of those concepts will be important.

This means, for example, that precategory can not be used in bicategory_ob_mor, only precategory_data can and similarly for many other concepts.

mvr:
Thank you for your comments! Some of my changes are to existing files: the binproduct_precategory code was there before I started, so my changes aren't adding a second copy. I will have a go at cleaning up that code.

The way I have it written, Cat and Precat are the prebicategories of categories and precategories. I need homsets there so that the functors between two (pre)categories themselves form a precategory.

It looks like I lost the coin flip twice on design choices! Commit 1327e75 is where I switched from a precategory with extra structure to the current design, and commit 3147a36 is where I switched to using precategory directly instead of keeping the data separate.

benediktahrens:
lgtm

DanGrayson:
I'm content to let you and Vladimir judge this one.

UniMath/CategoryTheory/BinProductPrecategory.v
```diff
@@ -37,76 +37,266 @@ Local Notation "G ∙ F" := (functor_composite _ _ _ F G) (at level 35).
...
+Proof.
+  use tpair. use tpair.
+  (* functor_on_objects *) intros; exact tt.
+  (* functor_on_morphisms *) intros F a b; apply identity.
+  split.
+  (* functor_id *) intros x; apply paths_refl.
+  (* functor_comp *) intros x y z w v; apply paths_refl.
+Defined.
+
+(* TODO: perhaps generalise to constant functors? *)
```
benediktahrens: Yes, please use `constant_functor` (from functor_categories.v) here.

UniMath/CategoryTheory/BinProductPrecategory.v
```diff
@@ -37,76 +37,266 @@ Local Notation "G ∙ F" := (functor_composite _ _ _ F G) (at level 35).
...
+Proof.
+  exists (ob_as_functor_data c).
+  split.
+  (* functor_id *) intros; constructor.
+  (* functor_comp *) intros x y z w v; simpl. apply pathsinv0, id_left.
+Defined.
+
+End unit_precategory.
+
+(** Fixing one argument of C × D -> E results in a functor *)
```
benediktahrens: The content of this section should be moved to CategoryTheory.BinProductPrecategory.v.

benediktahrens:
I would really love to see if the following lemma, which I learned about from @peterlefanulumsdaine , can be helpful in proving that the map from identities of categories to isomorphisms of categories is an equivalence:
```
Lemma eq_equiv_from_retraction {A} {P : A -> A -> UU} 
    (w : forall a b, a = b -> P a b)
    (v : forall a b, P a b -> a = b)
  : (forall a b (p : P a b), w _ _ (v _ _ p) = p)
  -> forall a b, isweq (w a b).
Proof.
  intros wv a.
  apply isweqtotaltofib. (* first of the two key steps *)
  use gradth.
  - intros bp. exists (pr1 bp). apply v, (pr2 bp).
  - intros be; apply connectedcoconusfromt. (* the second key step *)
  - intros bp. use total2_paths. apply idpath. apply wv.
Qed.
```
(I feel like I have said this before.)

UniMath/CategoryTheory/BinProductPrecategory.v
mvr: Do you mean define ob_as_functor using constant_functor, or remove ob_as_functor all together?

benediktahrens: On 09/07/2016 06:01 PM, Mitchell Riley wrote:
>> +  (* ob, mor *) exists unit. intros; exact unit.
>> +  (* identity, comp *) split; intros; constructor.
>> +  (* id_left *) simpl; split; try split; intros; apply isconnectedunit.
>> +Defined.
>> +
>> +Definition unit_functor C : functor C unit_precategory.
>> +Proof.
>> +  use tpair. use tpair.
>> +  (* functor_on_objects *) intros; exact tt.
>> +  (* functor_on_morphisms *) intros F a b; apply identity.
>> +  split.
>> +  (* functor_id *) intros x; apply paths_refl.
>> +  (* functor_comp *) intros x y z w v; apply paths_refl.
>> +Defined.
>> +
>> +(* TODO: perhaps generalise to constant functors? *)
> 
> Do you mean define ob_as_functor using constant_functor, or remove ob_as_functor all together?

It should be possible to replace every occurrence of `ob_as_functor c`
by `constant_functor unit_precategory _ c` directly, without giving an
extra definition.

If for some reason that does not work, or gets very unreadable, then
defining `ob_as_functor c` as `constant_functor ...` is the next-best
option.

benediktahrens:
I think we can merge this PR now.

vladimirias:
Since the concept is now called precategory_binroduct, the file should also be called PrecategoryBinProduct.

vladimirias:
Please let me have a further look. It might take some time - around a week, sorry, it is the start of work here at the IAS and many things need to be taken care of.

vladimirias:
I have a question: why precategory_binprod and in particular Catlike_associator appear in the definition of the pre-bicategory of precategories? The pentagon axiom for the pre-bicategory of precategories is just reflexivity that appears as a result of strict associativity of functor compositions. What am I missing?

mvr:
We need to define a natural transformation between F (G H) and (F G) H, this "Step 1", and the natural transformation is of course (fun a => identity _). But we also need to show that this natural transformation is itself natural in F, G and H, this is "Step 2".

The proof of the pentagon is further down the file in Catlike_pentagon, and it's indeed almost reflexivity, although we have to cancel some identity morphisms.

vladimirias:
What I do not understand is where the precategory_binproduct is from. It does not show up in the definition of pre-bicategory, does it?

mvr:
It does the way I have defined it, composition of 1-morphisms is given as a functor B(a,b) x B(b,c) -> B(a,c)

vladimirias:
LGTM. Sorry for the delay. It gave me a lot of thought. @mvr please resolve the conflicts that have accumulated and let's merge it.

vladimirias:
Mitchell, how did it happen that you have taken Peter Lumsdaine's code and submitted it to the UniMath as your own?

peterlefanulumsdaine:
Just to be clear — for my part, I do not think this is problematic.  The parts which were taken directly from my code were attributed as such at the time, in commit c61c283a6c640faa9aea052cbe131099f3155f71 — it looks like that attribution just got lost during the merge with UniMath, having ended up in a file that was redundant since @mortberg had provided the same constructions.  His definition of pre-bicategory looks to be based on mine in some respects (I shared my code with Mitchell exactly so that he could look at an existing development and re-use whatever aspects he felt worked well), but is not exactly the same.  And the subsequent parts of the development — the definitions of equivalences, and of (saturated) bicategories — are entirely Mitchell’s.  So I’m glad to have the acknowledgement reinstated, but I think the bulk of the code here is certainly Mitchell’s own.

vladimirias:
This is great that it is so. However, the similarities with your code and with your approach in the definition of pre-bicategories are very substantial and I think that the acknowledgement of your role in shaping this code must have been there from the start.

We also had some discussion with Mitchell about why he has chosen this approach to the definition and not the one where a pre-bicategory is a pre-category with additional structure on the Mor-types and he never mentioned in this discussion that his choice might have been affected by the availability of your code that used the first option.

And it appears that we would not have learned about your role in the development of this package if I have not noticed the /bicategories directory in TypeTheory and have not started asking questions about it.

I understand very well how you may share some of your ideas and work without caring whether you will be mentioned in connection with it or not. However, due to some very bad things that happen in our community not so long ago because of false attribution of credit we must be very pedantic in all things related to credit and authorship and we absolutely must teach our students to be very stringent in this discipline as well.

Vladimir.

mvr:
I am very sorry to both you and Peter, it was never my intention to claim credit for work that was not mine. At the beginning of my project, Peter was kind enough to send through a definition of bicategory that he had previously written up. I had little Coq experience, and did not understand his definition very clearly, but I expected that the lemmas he proved about product precategories would be useful so I copied them into my code verbatim with a line "(* Imported from Peter Lumsdaine's old bicategory code: *)". Some of that code still exists in the PrecategoryBinProduct file, but somehow along the way the attribution was lost.

While doing my project, the structure of my definition of a prebicategory changed a couple of times, as I mentioned earlier in this pull request. The initial change between the two possible structures was prompted by Chris, who suggested I follow the definition that appears in Leinster's "Basic Bicategories". Looking back at Peter's code now, this lead to definitions that are very similar to his, and it is likely that I was influenced by seeing his code.

Some of the contents of this pull request covers the same concepts as those in Peter's code: this is prebicategory.v and Cat.v. The remainder is new work, including the definition of a saturated bicategory, the lemmas in whiskering.v and catiso.v.

Again, I sincerely apologise; I should not have been so careless. I will add acknowledgements to the files that are similar to his work.

PR #410, opened 2016-07-24
Some special limits and colimits

tpannila:
Hi,

I formalized some special limits and colimits using graphs, following
graphs/pullbacks.v, and I showed these definitions coincide with direct
definitions. That is, we can construct the directly defined structures from
(co)limits structures and vice versa. 

I changed slightly the definition of diagram in graphs/colimits.v . This was 
needed(?) to have two edges, between same vertices, to map into different
morphisms in a precategory C. The new definition is needed for example in
graphs/equalizers.v .

### DIFF
--- a/UniMath/CategoryTheory/limits/graphs/equalizers.v
+++ b/UniMath/CategoryTheory/limits/graphs/equalizers.v
@@ -1,0 +1,359 @@
+(** * Equalizers defined in terms of limits *)
+(** ** Contents
+- Definition of equalizers
+- Coincides with the direct definition
+*)
+
+Require Import UniMath.Foundations.Basics.PartD.
+Require Import UniMath.Foundations.Basics.Propositions.
+Require Import UniMath.Foundations.Basics.Sets.
+
+Require Import UniMath.CategoryTheory.precategories.
+
+Require Import UniMath.CategoryTheory.limits.graphs.limits.
+Require Import UniMath.CategoryTheory.limits.graphs.colimits.
+Require Import UniMath.CategoryTheory.UnicodeNotations.
+Require Import UniMath.CategoryTheory.limits.equalizers.
+
+
+(** * Definition of equalizers in terms of limits *)
+Section def_equalizers.
+
+  Variable C : precategory.
+  Variable hs: has_homsets C.
+
+  Inductive two := One | Two .
+
+  Definition Equalizer_graph : graph.
+  Proof.
+    exists two.
+    exact (fun a b =>
+             match (a, b) with
+             | (One, Two) => unit ⨿ unit
+             | _ => empty
+             end).
+  Defined.
+
+  Definition Equalizer_diagram {a b : C} (f g : C⟦a, b⟧) :
+    diagram Equalizer_graph C.
+  Proof.
+    unfold diagram. cbn.
+    use (tpair _ (fun x => match x with
+                        | One => a
+                        | Two => b end) _).
+    intros x y e.
+    induction x; induction y; try induction e.
+    exact f. exact g.
+  Defined.
+
+  Definition Equalizer_cone {a b : C} (f g : C⟦a, b⟧)
+             (d : C) (h : C⟦d, a⟧) (H : h ;; f = h ;; g) :
+    cone (Equalizer_diagram f g) d.
+  Proof.
+    use mk_cone.
+    intros v. induction v.
+    exact h. exact (h ;; f).
+
+    intros u v e.
+    induction u; induction v; try induction e.
+    apply idpath. exact (!H).
+  Defined.
+
+  Definition isEqualizer {a b : C} (f g : C⟦a, b⟧)
+             (d : C) (h : C⟦d, a⟧) (H : h ;; f = h ;; g) : UU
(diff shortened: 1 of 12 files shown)


UniMath/CategoryTheory/limits/graphs/colimits.v
```diff
@@ -54,7 +54,7 @@ Definition vertex : graph -> UU := pr1.
 Definition edge {g : graph} : vertex g -> vertex g -> UU := pr2 g.
 
 Definition diagram (g : graph) (C : precategory) : UU :=
-  Σ (f : vertex g -> C), Π (a b : vertex g), edge a b -> C⟦f a, f b⟧.
+  Σ (f : vertex g -> C), Π (a b : vertex g), Π (e : edge a b), C⟦f a, f b⟧.
```
benediktahrens: This is literally the same as the old definition, since `C⟦f a, f b⟧` does not depend on `e`.
You can leave it as it is now, but the change is not necessary, as it is not a change.

tpannila: You are right. I had problems in defining equalizers and I changed this definition,
together with something else, which made some problem to disappear. This led
me to believe this definition was different. I rebased and switched back to the old
definition.

UniMath/CategoryTheory/limits/graphs/pushouts.v
```diff
@@ -0,0 +1,438 @@
...
+Require Import UniMath.CategoryTheory.UnicodeNotations.
+Require Import UniMath.CategoryTheory.limits.pushouts.
+
+(** * Definition of pushouts in terms of colimits *)
+Section def_po.
+
+  Variable C : precategory.
+  Variable hs: has_homsets C.
+
+  Inductive three := One | Two | Three.
```
benediktahrens: Note that the use of this inductive is against the rules of UniMath.
The same inductive type is used for pullbacks, so I guess that in the end, all of this is my fault.
Anyway, ideally this inductive would be replaced, in both pullbacks and pushouts, by another type of three elements, one that is built from the permitted type constructors.

UniMath/CategoryTheory/limits/graphs/equalizers.v
```diff
@@ -0,0 +1,359 @@
...
+  Variable C : precategory.
+  Variable hs: has_homsets C.
+
+  Inductive two := One | Two .
+
+  Definition Equalizer_graph : graph.
+  Proof.
+    exists two.
+    exact (fun a b =>
+             match (a, b) with
```
benediktahrens: We are trying to avoid the `match` construct, replacing it with use of eliminators.

UniMath/CategoryTheory/limits/graphs/equalizers.v
```diff
@@ -0,0 +1,359 @@
...
+Require Import UniMath.CategoryTheory.limits.equalizers.
+
+
+(** * Definition of equalizers in terms of limits *)
+Section def_equalizers.
+
+  Variable C : precategory.
+  Variable hs: has_homsets C.
+
+  Inductive two := One | Two .
```
benediktahrens: Instead of this type, you could/should use bool.

UniMath/CategoryTheory/limits/graphs/pushouts.v
vladimirias: The best choice is stn 3 - the standard set with 3 elements.

BTW I have a suggestion, let us define functions nattostn n : nat -> stn n for each n sending 0, …, n-1 to “itself” and numbers greater than n-1 to n-1. We can use this function as a coercion or introduce a short notation for it ( "fin i"  for “nattostn _ i” ?) making the use of stn n’s more convenient.

UniMath/CategoryTheory/limits/graphs/equalizers.v
vladimirias: “Inductive” in UniMath is *really* forbidden.

We may open another repository for UniMath extensions where the rules are weakened but in the UniMath itself the rules must be followed.

UniMath/CategoryTheory/limits/graphs/pushouts.v
benediktahrens: My PR #413 replaces the inductive type `three` by `stn 3`.
I suggest merging #413 first, and then adapting the present PR analogously to #413 .

tpannila:
I followed pullbacks.v to get rid of Inductive and match. Also, I cleaned some
of the proofs.

vladimirias:
Did you get a compilation time increase from removing the Inductive?

tpannila:
Yes. It takes a few seconds for coq to process when I use `two_rec_dep` two times, line 60 in graphs/equalizers.v , and `three_rec_dep` two times, line 68 in graphs/pushouts.v.

vladimirias:
One possible solution is to define stnn_rect in StandardFiniteSets. There is, maybe, even a way to construct it such that the computation rule holds definitionally (I am thinking about using nat_rect and stnntonat and nattostnn).

benediktahrens:
I am merging this PR now. Improvements to the induction principle on finite sets should be done in a subsequent PR, if necessary.

PR #413, opened 2016-07-28
remove Inductive three

benediktahrens:
replace the inductive type `three` by `stn 3`

- leads to some increase in compile time, in particular in nested applications of the dependent eliminator


DanGrayson:
g

PR #416, opened 2016-08-01
Added additive functors

tpannila:
This pull request contains definition of additive functors and the following 
results are formalized
- Additive functor preserves binary direct sums.
- A functor which preserves binary direct sums is additive.

Also, I added some smaller results on BinProducts, BinCoproducts, Zeros, 
and BinDirectSums.

### DIFF
--- a/UniMath/CategoryTheory/limits/binproducts.v
+++ b/UniMath/CategoryTheory/limits/binproducts.v
@@ -31,10 +31,18 @@ Section binproduct_def.
 
 Variable C : precategory.
 
-Definition isBinProductCone (c d p: C) (p1 : p --> c) (p2 : p --> d) :=
+Definition isBinProductCone (c d p : C) (p1 : p --> c) (p2 : p --> d) :=
   Π (a : C) (f : a --> c) (g : a --> d),
-    iscontr (total2 (fun fg : a --> p => dirprod (fg ;; p1 = f)
-                                                 (fg ;; p2 = g))).
+  iscontr (total2 (fun fg : a --> p => dirprod (fg ;; p1 = f) (fg ;; p2 = g))).
+
+Lemma isaprop_isBinProductCone (c d p : C) (p1 : p --> c) (p2 : p --> d) :
+  isaprop (isBinProductCone c d p p1 p2).
+Proof.
+  apply impred_isaprop. intros t.
+  apply impred_isaprop. intros t0.
+  apply impred_isaprop. intros t1.
+  apply isapropiscontr.
+Qed.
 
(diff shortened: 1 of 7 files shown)


benediktahrens:
According to [1], biproducts (or direct sums) can be defined for any category with a zero object, not just for (pre)abelian categories. 
Would it make sense to change your definition of `isBinDirectSumCone` (which has been added in a previous PR) to the more general setting? Or would that make things more complicated?

[1] https://en.wikipedia.org/wiki/Biproduct

DanGrayson:
Exercise 4 on p 198 of Mac Lane's "Categories for the Working Mathematician" shows that
a category with finite products, finite coproducts, such that the canonical map from the 
coproduct to the product is always an isomorphism, is also almost an additive category:
the Hom sets can all be made into commutative monoids.  The only other thing needed
to make an additive category is additive inverses for identity arrows.

I know of no useful examples of categories with biproducts that are not additive, so this suggests
that the added generality is probably not worth it.

On the other hand, I haven't looked at the definition you're referring to.

tpannila:
Rebased, fixed conflicts, and fixed coding style.

Benedikt, we can generalize the definition at some point, it should not make things much more
complicated.

benediktahrens:
lgtm

UniMath/CategoryTheory/limits/binproducts.v
```diff
@@ -589,3 +595,62 @@ Defined.
...
+    (* Arrow *)
+    - exact ((BinProductArrow C BP f g) ;; (iso_inv_from_iso i)).
+    (* Commutativity *)
+    - exact (iso_to_isBinProductCone_comm BP i w f g).
+    (* Equality of equalities of morphisms. *)
+    - intros y0. apply isapropdirprod. apply hs. apply hs.
+    (* Uniqueness *)
+    - intros y0 T. exact (iso_to_isBinProductCone_unique BP i w f g y0 T).
+  Defined.
+  Opaque iso_to_isBinProductCone.
```
mortberg: I'm curious why you have to make this Opaque here?

mortberg:
LGTM

UniMath/CategoryTheory/limits/binproducts.v
tpannila: I don't have to. `isBinProductCone` is an hProp and most of the time we don't need to unfold the lemma to get the arrow. If we want to get the arrow we can use `Transparent`.

mortberg: Ok, thanks for the explanation.

PR #419, opened 2016-08-05
Added Grothendieck toposes

tpannila:
Hi,

- Formalization of undercategories (dual to slice_precat) in
  UnderPrecategories.v.
- Formalization of Subobjects and Quotobjects in C of an object c.
  These are defined by first taking the subcategory consisting of
  all monics (resp. epis) of C and then taking slice_precat
  (resp. UnderPrecategory) of c in this category. Subobjects (resp.
  quotient objects) are then objects of this category.
- Formalization of Grothendieck toposes in GrothendieckTopos.v .
  This definition should be carefully checked. I followed the definition 
  found in "Sheaves in Geometry and Logic", page 110.

Changes in other files:
- In Abelian.v, changed "subobject" to "monic" and "quotobject" to "epi".
- In Monics.v and Epis.v I defined the subcategories consisting of all
  monics and epis, respectively, and gave method to construct monics and epis in
  functor categories.
- In more_on_pullbacks.v, showed that a functor category [D, C hsC] has
  pullbacks if C has pullbacks using the result on pointwise isPullback.
- In functor_precategories.v I added functions "mk_*" to construct things.
- In yoneda.v I added a construction of a natural transformation
  yoneda c -> yoneda c' from a morphism h : c --> c'.
- pushouts.v, fixed typos
- graphs/pullbacks.v , if pullbacks exist as limits, then pullbacks exist as
  direct definition, and vice versa.
- category_hset_structures.v, uncomment the proof about pullbacks as a
  special case of limits.

All comments are welcome.

### DIFF
--- a/UniMath/CategoryTheory/GrothendieckTopos.v
+++ b/UniMath/CategoryTheory/GrothendieckTopos.v
@@ -1,0 +1,194 @@
+(** * Grothendick toposes *)
+(** ** Contents
+- Definition of Grothendieck topology
+ - Grothendieck topologies
+ - Precategory of sheaves
+- Grothendieck toposes
+*)
+Require Import UniMath.Foundations.Basics.PartD.
+Require Import UniMath.Foundations.Basics.Propositions.
+Require Import UniMath.Foundations.Basics.Sets.
+
+Require Import UniMath.CategoryTheory.UnicodeNotations.
+Require Import UniMath.CategoryTheory.limits.pullbacks.
+Require Import UniMath.CategoryTheory.limits.equalizers.
+Require Import UniMath.CategoryTheory.limits.products.
+Require Import UniMath.CategoryTheory.precategories.
+Require Import UniMath.CategoryTheory.functor_categories.
+Require Import UniMath.CategoryTheory.limits.pullbacks.
+Require Import UniMath.CategoryTheory.slicecat.
+Require Import UniMath.CategoryTheory.Monics.
+Require Import UniMath.CategoryTheory.Subobjects.
+Require Import UniMath.CategoryTheory.yoneda.
+Require Import UniMath.CategoryTheory.category_hset.
+Require Import UniMath.CategoryTheory.opp_precat.
+Require Import UniMath.CategoryTheory.category_hset_structures.
+Require Import UniMath.CategoryTheory.limits.graphs.pullbacks.
+Require Import UniMath.CategoryTheory.limits.graphs.equalizers.
+Require Import UniMath.CategoryTheory.sub_precategories.
+Require Import UniMath.CategoryTheory.equivalences.
+
+(** HSET Pullbacks and Equalizers from limits to direct definition *)
+Section HSET_Structures.
+
+  Definition HSET_Pullbacks : @limits.pullbacks.Pullbacks HSET :=
+    equiv_Pullbacks_2 HSET has_homsets_HSET PullbacksHSET.
+
+  Definition HSET_Equalizers: @limits.equalizers.Equalizers HSET :=
+    equiv_Equalizers2 HSET has_homsets_HSET EqualizersHSET.
+
+End HSET_Structures.
+
+(** * Definiton of Grothendieck topology
+    The following definition is a formalization of the definition in Sheaves in Geometry and Logic,
+    Saunders Mac Lane and Ieke Moerdijk, pages 109 and 110.
+
+    Grothendieck topology is a collection J(c) of subobjects of the Yoneda functor, for every object
+    of C, such that:
+    - The Yoneda functor y(c) is in J(c).
+    - Pullback of a subobject in J(c) along any morphism h : c' --> c is in J(c')
+    - If S is a subobject of y(c) such that for all objects c' and all morphisms
+      h : c' --> c in C the pullback of S along h is in J(c'), then S is in J(c).
+  *)
+Section def_grothendiecktopology.
+
+  Variable C : precategory.
+  Hypothesis hs : has_homsets C.
+
+  (** A sieve on c is a subobject of the yoneda functor. *)
+  Definition sieve (c : C) : UU :=
+    SubobjectsPrecategory
+      (functor_precategory (opp_precat C) HSET has_homsets_HSET)
+      (functor_category_has_homsets (opp_precat C) HSET has_homsets_HSET)
+      (yoneda C hs c).
+
+  (* Coq does not automatically convert the following types *)
+  Definition FunctorPrecatObToFunctor
+             (c : functor_precategory (opp_precat C) HSET has_homsets_HSET) :
+    functor (opp_precat C) HSET := c.
+
+  Definition FunctorPrecatMorToNatTrans
+             {c c': functor_precategory (opp_precat C) HSET has_homsets_HSET} (h : c --> c') :
+    nat_trans (FunctorPrecatObToFunctor c) (FunctorPrecatObToFunctor c') := h.
+
+  Definition sieve_functor {c : C} (S : sieve c) : functor (opp_precat C) HSET :=
+    precategory_object_from_sub_precategory_object _ _ (slicecat_ob_object _ _ S).
+
+  Definition sieve_nat_trans {c : C} (S : sieve c) :
+    nat_trans (sieve_functor S) (FunctorPrecatObToFunctor (yoneda C hs c)) :=
+    precategory_morphism_from_sub_precategory_morphism _ _ _ _ (slicecat_ob_morphism _ _ S).
+
+
+  (** ** Grothendieck topology *)
+
+  Definition collection_of_sieves : UU := Π (c : C), hsubtypes (sieve c).
+
+  Definition isGrothendieckTopology_maximal_sieve (COS : collection_of_sieves) : UU :=
+    Π (c : C), COS c (SubobjectsPrecategory_ob
+                        (functor_precategory (opp_precat C) HSET has_homsets_HSET)
+                        (functor_category_has_homsets (opp_precat C) HSET has_homsets_HSET)
+                        (identity (yoneda C hs c)) (identity_isMonic _)).
+
+  Definition isGrothendieckTopology_stability (COS : collection_of_sieves) : UU :=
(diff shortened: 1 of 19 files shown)


UniMath/CategoryTheory/GrothendieckTopos.v
```diff
@@ -0,0 +1,101 @@
...
+
+  Definition HSET_Pullbacks : @limits.pullbacks.Pullbacks HSET
+    := equiv_Pullbacks_2 HSET has_homsets_HSET PullbacksHSET.
+
+End HSET_Pullbacks.
+
+(** * Definiton of Grothendieck topos
+  The following definition is a formalization of the definition in
+  Sheves in Geometry and Logic, Saunders Mac Lane and Ieke Moerdijk, pages 109
+  and 110.
```
DanGrayson: On those pages is defined "Grothendieck topology", not "Grothendieck topos".

UniMath/CategoryTheory/GrothendieckTopos.v
```diff
@@ -0,0 +1,101 @@
...
+
+  Grothendieck topos is a collection J(c) of subobjects of the Yoneda functor,
+  for every object of C, such that:
+  - The Yoneda functor y(c) is in J(c).
+  - Pullback of a subobject in J(c) along any morphism h : c' --> c is in J(c')
+  - If S is a subobject of y(c) such that for any object c' and morphism
+    h : c' --> c in C the pullback is in J(c'), then S is in J(c).
```
DanGrayson: This is a misstatement of the transitivity axiom.  The current statement is a tautology, for
one can take c=c' and h=id.

UniMath/CategoryTheory/GrothendieckTopos.v
mortberg: Minor detail: it should be "Sheaves" and not "Sheves"

tpannila:
Hi,

thank you @DanGrayson and @mortberg for your comments. I made a blunder by 
calling Grothendieck topologies Grothendieck toposes. I rebased and fixed the 
things you suggested.
- Added equalizers stuff for HSET. The changes in category_hset_structures.v and
  graphs/equalizers.v. 
- Added a few accessor functions in slicecat.v
- Fixed the file GrothendieckTopos.v. I hope the definition is correct this time.

DanGrayson:
How about renaming the file `GrothendieckTopos.v`, too?

tpannila:
@DanGrayson ,  `GrothendieckTopos.v` contains a definition of Grothendieck topos on line
193. I renamed the old topos to topology and added some code so that I was able to define Grothendieck topos. I hope the new definition is correct.

UniMath/CategoryTheory/yoneda.v
```diff
@@ -418,3 +418,30 @@ Proof.
...
+  Hypothesis hs : has_homsets C.
+
+  Lemma YonedaNatTrans_is_nat_trans {c c': C} (h : C⟦c, c'⟧) :
+      is_nat_trans (yoneda_ob_functor_data C hs c) (yoneda_ob_functor_data C hs c')
+                   (λ (x : C) (y : yoneda_objects_ob C c x), y ;; h).
+  Proof.
+    intros x y f. cbn. use funextfun. intros x0. rewrite assoc. apply idpath.
+  Qed.
+
+  Definition YonedasNatTrans {c c': C} (h : C⟦c, c'⟧) :
```
benediktahrens: It seems to me that this is already formalized in 
```
Definition yoneda (C : precategory) (hs: has_homsets C) :
  functor C [C^op, HSET, has_homsets_HSET] :=
   tpair _ _ (is_functor_yoneda C hs).
```
above, isn't it?

UniMath/CategoryTheory/limits/more_on_pullbacks.v
```diff
@@ -526,6 +526,224 @@ Proof.
...
+Qed.
 
-End pullbacks_pointwise.
\ No newline at end of file
+End pullbacks_pointwise.
+
+
+(** * Pullbacks in functor_precategory
+  We show that pullbacks exist in the functor category [D, C, hs] if pullbacks exist
+  in C. *)
```
benediktahrens: To be slightly nitpicky, the comment might be misleading: "exist" often refers to limits as a property. 
Here, pullbacks are chosen, hence a structure. "Specified pullbacks" is another expression used.

UniMath/CategoryTheory/limits/more_on_pullbacks.v
```diff
@@ -526,6 +526,224 @@ Proof.
...
+Section pullbacks_functor_category.
+
+  Variable D C : precategory.
+  Hypothesis hs : has_homsets C.
```
benediktahrens: Would it be possible to derive this construction from the general construction of limits in functor categories? Would it be easier? Shorter?

UniMath/CategoryTheory/UnderPrecategories.v
```diff
@@ -0,0 +1,189 @@
...
+    apply (isaset_Under_mor X Y).
+  Qed.
+
+  Definition Under_mor_iso_eq (X Y : Under_ob)
+             (f g : @iso UnderPrecategory X Y) :
+    pr1 f = pr1 g -> f = g.
+  Proof.
+    case g. case f. clear f g. cbn. intros t p t0 p0 X0.
+    use (subtypePairEquality _ X0).
+    intros f. apply isaprop_is_iso.
```
benediktahrens: You probably don't need this lemma, use the general `eq_iso` instead.

UniMath/CategoryTheory/limits/more_on_pullbacks.v
vladimirias: The best expression is “we construct pullbacks in the functor category”.

UniMath/CategoryTheory/limits/more_on_pullbacks.v
vladimirias: Don’t we have a formalization of specifically pullbacks in presheaves?

benediktahrens: On 09/08/2016 03:51 PM, Vladimir Voevodsky wrote:
>> \ No newline at end of file
>> +End pullbacks_pointwise.
>> +
>> +
>> +(** * Pullbacks in functor_precategory
>> +  We show that pullbacks exist in the functor category [D, C, hs] if pullbacks exist
>> +  in C. *)
>> +Section pullbacks_functor_category.
>> +
>> +  Variable D C : precategory.
>> +  Hypothesis hs : has_homsets C.
> 
> Don’t we have a formalization of specifically pullbacks in presheaves?

We have a few lemmas about specifically pullbacks in presheaves in the
ALV project. That code has not been upstreamed to UniMath yet.
However, there are no new definitions there, but only lemmas - in
particular, no formalization of specifically pullbacks in presheaves.

tpannila:
Rebased, fixed conflicts, fixed some coding style stuff, and fixed the stuff Benedikt suggested.

@benediktahrens , if the category C has limits, then we can derive the construction from the
general construction. Suppose C does not have limits. C can still have pullbacks and the current
proof works also in this case.

benediktahrens:
Concerning the lifting of pullbacks to the functor category, the lemma `LimFunctorCone` of the file `limits/graphs/limits` will help. For this lemma, it is not necessary to have all limits in the target category, but only those limits of the shape under consideration.
I agree that that is not completely automatic: the map `diagram_pointwise` occurring in a hypothesis of `LimFunctorCone` does not literally give a pullback diagram when applied to a pullback diagram in the functor category. It will be required to destruct and reassemble the pullbacks given in the target category by hypothesis, in order to instantiate this hypothesis of `LimFunctorCone`. But that might still be shorter than your direct proof.

vladimirias:
It might be important to have the construction of pullbacks in the functor category that are given on objects, by construction, by pullbacks in the target category and on morphisms by the construction of the morphism between pullbacks from a morphism between the corresponding three objects with the two equalities of the appropriate compositions. Such a construction on the level of objects will be useful in many other places, so we need to have it and deducing from it the construction of the pullback presheaves should be straightforward.

Such pullbacks of functors can be much easier to use later than the ones obtained through general constructions.

Have not we discussed it before?

benediktahrens:
Right.
We did discuss this at least for binary (co)products, where we
encountered some significant slowdown when instantiating the general
construction of (co)limits in functor categories compared to the direct
construction in the style you are mentioning.

vladimirias:
The advantage of the approach that I mention is also in that it requires constructing a number of objects related to the pullbacks in a general category, such as the morphism between pullbacks (given by any pair of pullback squares, no need to require general pullbacks in the category) given by three morphisms and two equalities, but also probably other objects all of which will be useful for many other purposes later.

UniMath/CategoryTheory/limits/pullbacks.v
```diff
@@ -994,3 +994,209 @@ Defined.
...
+            (λ d : D, hpb (F d) (G d) (H d) (α d) (β d))
+            (λ (a b : D) (f : D ⟦ a, b ⟧),
+             PullbackArrow
+               (hpb (F b) (G b) (H b) (α b) (β b))
+               (hpb (F a) (G a) (H a) (α a) (β a))
+               (PullbackPr1 (hpb (F a) (G a) (H a) (α a) (β a)) ;; # G f)
+               (PullbackPr2 (hpb (F a) (G a) (H a) (α a) (β a)) ;; # H f)
+               (FunctorPrecategoryPullbacks_eq F G H α β a b f))) H
+         (λ x : D, PullbackPr2 (hpb (F x) (G x) (H x) (α x) (β x)))
+         (FunctorPrecategoryPullbacks_is_nat_trans2 F G H α β)) β.
```
benediktahrens: Do you think it would be a good idea to define some of these functors separately, in order to make the statements of the subsequent lemmas more readable?

benediktahrens:
lgtm

vladimirias:
I want to have a look. There is a possibility of universe consistency issues in this subject.

vladimirias:
LGTM 

Tomi can you resolve the conflicts and merge?

tpannila:
Resolved the conflicts.

PR #424, opened 2016-08-10
Cleaned Monoids_and_Groups.v

tpannila:
Cleaned long lines in Monoids_and_Groups.v and added Contents comment.

### DIFF
--- a/UniMath/Foundations/Algebra/Monoids_and_Groups.v
+++ b/UniMath/Foundations/Algebra/Monoids_and_Groups.v
@@ -1,8 +1,6 @@
-(** * Algebra I. Part B.  Monoids, abelian monoids groups, abelian groups. Vladimir Voevodsky. Aug. 2011 - .
-
-*)
+(** * Algebra I. Part B.  Monoids, abelian monoids groups, abelian groups. Vladimir Voevodsky. Aug. 2011 - . *)
 (** ** Contents
-- Standard Algebraic Structure
+- Standard Algebraic Structures
  - Monoids
   - Basics definitions
   - Functions between monoids compatible with structures (homomorphisms)
@@ -45,18 +43,17 @@
 
 (** ** Preamble *)
 
+
 (** Settings *)
 
 (** The following line has to be removed for the file to compile with Coq8.2 *)
 Unset Automatic Introduction.
-
 
 Unset Kernel Term Sharing.
 
 (** Imports *)
 
 Require Export UniMath.Foundations.Algebra.BinaryOperations.
-
 
 (** To upstream files *)
 
@@ -69,26 +66,26 @@ Require Export UniMath.Foundations.Algebra.BinaryOperations.
 
 (** ****  Basic definitions *)
 
-
-
 Definition monoid : UU := total2 (fun X : setwithbinop => ismonoidop (@op X)).
+
 Definition monoidpair :
-  Π (t : setwithbinop),
-  (λ X : setwithbinop, ismonoidop op) t → Σ X : setwithbinop, ismonoidop op
-  := tpair (fun X : setwithbinop => ismonoidop (@op X)).
+  Π (t : setwithbinop), (λ X : setwithbinop, ismonoidop op) t → Σ X : setwithbinop, ismonoidop op :=
+  tpair (fun X : setwithbinop => ismonoidop (@op X)).
+
 Definition monoidconstr :
-  Π (t : setwithbinop),
-  (λ X : setwithbinop, ismonoidop op) t → Σ X : setwithbinop, ismonoidop op
-  := monoidpair.
+  Π (t : setwithbinop), (λ X : setwithbinop, ismonoidop op) t → Σ X : setwithbinop, ismonoidop op :=
+  monoidpair.
+
 Definition pr1monoid : monoid -> setwithbinop := @pr1 _ _.
 Coercion pr1monoid : monoid >-> setwithbinop.
 
 Definition assocax (X : monoid) : isassoc (@op X) := pr1 (pr2 X).
+
 Definition unel (X : monoid) : X := pr1 (pr2 (pr2 X)).
-Definition lunax (X : monoid) : islunit (@op X) (unel X)
-  := pr1 (pr2 (pr2 (pr2 X))).
-Definition runax (X : monoid) : isrunit (@op X) (unel X)
-  := pr2 (pr2 (pr2 (pr2 X))).
+
+Definition lunax (X : monoid) : islunit (@op X) (unel X) := pr1 (pr2 (pr2 (pr2 X))).
+
+Definition runax (X : monoid) : isrunit (@op X) (unel X) := pr2 (pr2 (pr2 (pr2 X))).
 
 Notation "x + y" := (op x y) : addmonoid_scope.
 Notation "0" := (unel _) : addmonoid_scope.
@@ -101,12 +98,10 @@ Notation "1" := (unel _) : multmonoid_scope.
 Delimit Scope multmonoid_scope with multmonoid.
 
 
-
 (** **** Functions between monoids compatible with structure (homomorphisms) and their properties *)
 
-
-Definition ismonoidfun {X Y : monoid} (f : X -> Y) : UU
-  := dirprod (isbinopfun f) (paths (f (unel X)) (unel Y)).
+Definition ismonoidfun {X Y : monoid} (f : X -> Y) : UU :=
+  dirprod (isbinopfun f) (paths (f (unel X)) (unel Y)).
 
 Lemma isapropismonoidfun {X Y : monoid} (f : X -> Y) : isaprop (ismonoidfun f).
 Proof.
@@ -115,16 +110,16 @@ Proof.
   - apply (setproperty Y).
 Defined.
 
-Definition monoidfun (X Y : monoid) : UU
-  := total2 (fun f : X -> Y => ismonoidfun f).
-Definition monoidfunconstr {X Y : monoid} {f : X -> Y} (is : ismonoidfun f) :
-  monoidfun X Y := tpair _ f is.
+Definition monoidfun (X Y : monoid) : UU := total2 (fun f : X -> Y => ismonoidfun f).
+
+Definition monoidfunconstr {X Y : monoid} {f : X -> Y} (is : ismonoidfun f) : monoidfun X Y :=
+  tpair _ f is.
+
 Definition pr1monoidfun (X Y : monoid) : monoidfun X Y -> (X -> Y) := @pr1 _ _.
 
-Definition monoidfuntobinopfun (X Y : monoid) : monoidfun X Y -> binopfun X Y
-  := fun f => binopfunpair (pr1 f) (pr1 (pr2 f)).
+Definition monoidfuntobinopfun (X Y : monoid) : monoidfun X Y -> binopfun X Y :=
+  fun f => binopfunpair (pr1 f) (pr1 (pr2 f)).
 Coercion monoidfuntobinopfun : monoidfun >-> binopfun.
-
 
 Lemma isasetmonoidfun (X Y : monoid) : isaset (monoidfun X Y).
 Proof.
@@ -136,7 +131,6 @@ Proof.
     apply isapropismonoidfun.
 Defined.
 
-
 Lemma ismonoidfuncomp {X Y Z : monoid} (f : monoidfun X Y) (g : monoidfun Y Z) :
   ismonoidfun (funcomp (pr1 f) (pr1 g)).
 Proof.
@@ -144,18 +138,16 @@ Proof.
   unfold funcomp. rewrite (pr2 (pr2 f)).
   apply (pr2 (pr2 g)).
 Defined.
-
 Opaque ismonoidfuncomp.
 
-Definition monoidfuncomp {X Y Z : monoid}
-           (f : monoidfun X Y) (g : monoidfun Y Z) : monoidfun X Z
-  := monoidfunconstr (ismonoidfuncomp f g).
+Definition monoidfuncomp {X Y Z : monoid} (f : monoidfun X Y) (g : monoidfun Y Z) :
+  monoidfun X Z := monoidfunconstr (ismonoidfuncomp f g).
 
+Definition monoidmono (X Y : monoid) : UU := total2 (fun f : incl X Y => ismonoidfun f).
 
-Definition monoidmono (X Y : monoid) : UU
-  := total2 (fun f : incl X Y => ismonoidfun f).
 Definition monoidmonopair {X Y : monoid} (f : incl X Y) (is : ismonoidfun f) :
   monoidmono X Y := tpair _  f is.
+
 Definition pr1monoidmono (X Y : monoid) : monoidmono X Y -> incl X Y := @pr1 _ _.
 Coercion pr1monoidmono : monoidmono >-> incl.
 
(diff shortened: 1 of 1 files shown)


benediktahrens:
lgtm

vladimirias:
Is there a reason to have some titles in the Contents comment on two lines? I checked that they are on one line in the text so it is Ok with coqdoc, but why have them on two lines in the Contents comment?

vladimirias:
Actually, I am not happy with how the lines are cut in this PR. I can not see any order in it. The general rule that I have used and I think we have used is that the line of a single sentence, before the full stop, is only cut if this sentence would otherwise be longer that 100 characters. 

In this case it is cut such that : and  := do not occur at the start of the lines, something that you also do not follow,  and such that the resulting lines are as long as possible, but there are no firm rules (yet?) on how to do it.

You have cut lots of sentences that are much shorter than 100 characters.

You also leave no blanc line between the units (e.g. definitions), something that we agreed to do in all cases except possibly the Definition/Coercion pair when an identifier is declared a coercion. 

I think these rules need to be formulated and put into the README along the 100 characters rule, after we discuss them and agree to them. At the moment some of them (no := at the start of the line) are hidden somewhere in our discussions and the rest are simply rules that I have followed and it seem other did too.

DanGrayson:
I often start my lines in Ktheory with := or with : .  I think it's easier to see what the line contains that way.  I offer an example:
```
Definition makeGH_verticalPath {Y} {y:Y} (l:y = y) {T:Torsor ℤ} (t:T)
           {y' y''} (h:y' = y) (p:y'' = y')
  : makeGH l T t (p@h) = makeGH l T t h.
Proof. intros. destruct p. reflexivity. Defined.
```

As for spacing between units, I think it remains unchanged in this pull request, e.g., between abmonoid and abmonoidpair there is still no space.  I am unaware of a new rule about this.  Shouldn't all new rules be submitted as pull requests that do nothing but modify the style guide in README.md?

I don't see lots of lines shorter than 100 that have been newly wrapped.  Can you offer examples?

vladimirias:
> I don't see lots of lines shorter than 100 that have been newly wrapped. Can you offer examples?
> 
The first ones I see are lunax and runax. Then monoidfun. Then monoidmono.

> I often start my lines in Ktheory with := or with : . I think it's easier to see what the line contains that way.
> 

I think we had a discussion about it and about uniformizing it throughout the UniMath. I think, it is one of those things that is better to have restricted by a rule.

Vladimir.

DanGrayson:
Okay, those four functions are indeed examples of rewrapping lines already shorter than 100.

About : and := , if it does become a rule and gets recorded and thus advertised in the style guide, then we'll have a lot of files to fix.  Perhaps don't hold up this PR on that issue.

benediktahrens:
On 09/05/2016 04:04 AM, Daniel R. Grayson wrote:
> Okay, those four functions are indeed examples of rewrapping lines
> already shorter than 100.
> 
> About : and := , if it does become a rule and gets recorded and thus
> advertised in the style guide, then we'll have a lot of files to fix.
> Perhaps don't hold up this PR on that issue.

Please don't introduce too many rules about the formatting of text.
It starts feeling like in french administration...

Also, the mass changing of files is quite disruptive, for what I
perceive to be little benefit.


The main benefit of Tomi's PR is, in my opinion, the introduction of
bullets in the proofs. This is a big improvement compared to the state
before.
Making a lot of noise just for line breaks, on the other hand, seems
overkill to me.

vladimirias:
> About : and := , if it does become a rule and gets recorded and thus advertised in the style guide, then we'll have a lot of files to fix. Perhaps don't hold up this PR on that issue.
> 
> 

Ok. I can fix it when I get to PartC in my working on the files.

vladimirias:
I agree with you and Dan concerning this particular PR, however I do think that following strict rules for text formating in the core part of the library is a good idea. It makes the text to have the same feeling throughout the library which helps both the newcomers and us to read through it faster and more easily.

BTW it looks in this PR like Tomi was experimenting with different ways to introduce the line breaks. f it is so then maybe he can tell us his ideas?

I do not insist on my particular choice of rules, I just suggest that having such rules, not binding ones, is convenient both for the writer and for the reader.

DanGrayson:
Yes, uniform formatting of a file or even of a package would make things look nicer.  That's why I hope no one edits my files for style, since I think I already have made good choices based on style.

I wouldn't much mind seeing a few rules for the "Foundations" package posted in  file `UniMath/Foundations/Style.md`.  I wouldn't mind seeing the thing about : and := and blank lines between units there. 

I personally like to occasionally put no blank lines between things that are closely related:
```
Definition right_mult {G:gr} {X:Action G} (x:X) := fun g => g*x.
Definition left_mult {G:gr} {X:Action G} (g:G) := fun x:X => g*x.
```

I think the rule for indentation, by the way, should be that whatever coq-mode produces automatically is okay.  You may want to make coq-mode indentation mandatory within Foundations, to promote uniformity of look.  I'm pretty sure that's what Tomi has been using for his indentation, anyway, as he cleans the files.

DanGrayson:
Here's a rule for formatting proofs I would propose: always use Coq's structure syntax (using `{ } + - *`) to focus on a single goal as soon as a tactic makes additional goals.  Hence
```
Definition weqfp {X Y : UU} (w : X ≃ Y) (P : Y -> UU) :
  (Σ x : X, P (w x)) ≃ (Σ y, P y).
Proof.
  intros.
  exists (weqfp_map w P).
  refine (gradth _ (weqfp_invmap w P) _ _).
  { intros xp. use total2_paths.
    { simpl. apply homotinvweqweq. }
    simpl. rewrite <- weq_transportf_adjointness.
    rewrite transport_f_f. rewrite pathsinv0l.
    reflexivity. }
  { intros yp. simple refine (total2_paths _ _).
    { simpl. apply homotweqinvweq. }
    simpl. rewrite transport_f_f. rewrite pathsinv0l. reflexivity. }
Defined.
```
is allowed and
```
Definition weqfp {X Y : UU} (w : X ≃ Y) (P : Y -> UU) :
  (Σ x : X, P (w x)) ≃ (Σ y, P y).
Proof.
  intros.
  exists (weqfp_map w P).
  refine (gradth _ (weqfp_invmap w P) _ _).
  intros xp. use total2_paths.
  simpl. apply homotinvweqweq. 
  simpl. rewrite <- weq_transportf_adjointness.
  rewrite transport_f_f. rewrite pathsinv0l.
  reflexivity. 
  intros yp. simple refine (total2_paths _ _).
  simpl. apply homotweqinvweq. 
  simpl. rewrite transport_f_f. rewrite pathsinv0l. reflexivity. 
Defined.
```
is not allowed.  This rule should be absolutely mandatory, because it helps make our code less fragile.  If a change happens upstream, we want the downstream error to occur as soon as possible.

tpannila:
Hi all,

I've been inactive due to bad internet connection, but now that is solved.

@vladimirias , I tried to make most of the lines max. 80 characters.
This is some kind of standard in programming. I can fix this PR to max.
100 characters.

DanGrayson:
Re: "Making a lot of noise just for line breaks, on the other hand, seems
overkill to me."

I disagree.  We still have many overlong lines, like this wonderful one 2245 characters wide:
```
intros xa1 xa2 . change ( ( abgrfracrelint ( rigaddabmonoid X ) R ) xa1 ( dirprodpair ( @rigunel1 X ) ( @rigunel1 X ) ) -> ( abgrfracrelint ( rigaddabmonoid X ) R ) xa2 ( dirprodpair ( @rigunel1 X ) ( @rigunel1 X ) ) -> ( abgrfracrelint ( rigaddabmonoid X ) R ( @rigtorngop2int X xa1 xa2 ) ( dirprodpair ( @rigunel1 X ) ( @rigunel1 X ) ) ) ) .  unfold abgrfracrelint . simpl . apply hinhfun2 . intros t22 t21 .   set ( c2 := pr1 t21 ) . set ( c1 := pr1 t22 ) . set ( r1 := pr2 t21 ) . set ( r2 := pr2 t22 ) . set ( x1 := pr1 xa1 ) . set ( a1 := pr2 xa1 ) .  set ( x2 := pr1 xa2 ) . set ( a2 := pr2 xa2 ) . split with ( ( x1 * c2 + a1 * c2 ) + ( ( c1 * x2 + c1 * c2 ) + ( c1 * a2 + c1 * c2 ) ) ) . change ( pr1 ( R ( x1 * x2 + a1 * a2 + 0 + ( ( x1 * c2 + a1 * c2 ) + ( ( c1 * x2 + c1 * c2 ) + ( c1 * a2 + c1 * c2 ) ) ) )  ( 0 + ( x1 * a2 + a1 * x2 ) + ( x1 * c2 + a1 * c2 +  ( ( c1 * x2 + c1 * c2 ) + ( c1 * a2 + c1 * c2 ) ) ) ) ) ) . rewrite ( riglunax1 X _ ) .  rewrite ( rigrunax1 X _ ) .   rewrite ( assoc ( x1 * c2 ) _ _ ) .  rewrite ( rer _ ( a1 * a2 ) _ _ ) .  rewrite ( rer _ ( a1 * x2 ) _ _ ) . rewrite ( pathsinv0 ( assoc ( a1 * a2 ) _ _  ) ) . rewrite ( pathsinv0 ( assoc ( a1 * x2 ) _ _  ) ) . rewrite ( pathsinv0 ( assoc ( x1 * x2 + _ ) _ _ ) ) . rewrite ( pathsinv0 ( assoc ( x1 * a2 + _ ) _ _ ) ) .  rewrite ( rer _ ( a1 * a2 + _ ) _ _ ) .  rewrite ( rer _ ( a1 * x2 + _ ) _ _ ) .  rewrite ( pathsinv0 ( ld _ _ x1 ) ) . rewrite ( pathsinv0 ( ld _ _ x1 ) ) . rewrite ( pathsinv0 ( ld _ _ c1 ) ) . rewrite ( pathsinv0 ( ld _ _ c1 ) ) . rewrite ( pathsinv0 ( ld _ _ a1 ) ) .  rewrite ( pathsinv0 ( ld _ _ a1 ) ) .  rewrite ( pathsinv0 ( rd _ _ ( x2 + c2 ) ) ) .  rewrite ( pathsinv0 ( rd _ _ ( a2 + c2 ) ) ) . rewrite ( comm ( a1 * _ ) _ ) .  rewrite ( rer _ ( c1 * _ ) _ _ ) . rewrite ( pathsinv0 ( rd _ _ ( x2 + c2 ) ) ) .  rewrite ( pathsinv0 ( rd _ _ ( a2 + c2 ) ) ) .  clearbody r1 . clearbody r2 . change ( pr1 ( R ( x2 + 0 + c2 ) ( 0 + a2 + c2 ) ) ) in r1 . change ( pr1 ( R ( x1 + 0 + c1 ) ( 0 + a1 + c1 ) ) ) in r2 .  rewrite ( rigrunax1 X _ ) in r1 .   rewrite ( riglunax1 X _ ) in r1 .   rewrite ( rigrunax1 X _ ) in r2 .   rewrite ( riglunax1 X _ ) in r2 . rewrite ( comm c1 a1 ) . apply ( is _ _ _ _ r2 r1 ) .   Defined .
```

DanGrayson:
I'm also happy that Tomi is removing all those extra spaces, as in `intros xa1 xa2 . change ( ( abgrf...`

DanGrayson:
@tpannila -- our local variables file now sets the fill column to 100 for you, so it should be easy to re-wrap your PR.

The 80 character standard is one I know well, since my summer job in 1968 involved handling tens of thousands of IBM punched cards per day -- each one could hold just 80 characters.  Then eventually we had video terminals that adopted the same limit.  But since the laptop arrived on the scene 25 years ago we have essentially not been limited to 80 characters for any practical purpose.

vladimirias:
The coq-mode does not wrap long lines, at least in my somewhat outdated Proof General.

Does it in yours?

vladimirias:
This sounds Ok. It will take a lot of work to format proofs in the existing files. Also it seems that { } + - *) provide many notations for the same function. As always when it happens it leads to a possible… variegated (the only word that I have found to translate a Russian word “pestryj”) text.

DanGrayson:
Re: "The coq-mode does not wrap long lines, at least in my somewhat outdated Proof General."

Not automatically, no -- that would be consider bad emacs style.  It sets just the column to which wrapping would be done if you would do it.

You could use `C-x q`, which  runs the command `auto-fill-mode`.  Alternatively, there are ways to highlight long lines.  I just prefer to use this:

```
M-q runs the command fill-paragraph (found in global-map), which is an
interactive compiled Lisp function in ‘fill.el’.

It is bound to M-q.

(fill-paragraph &optional JUSTIFY REGION)

Fill paragraph at or after point.

If JUSTIFY is non-nil (interactively, with prefix argument), justify as well.
If ‘sentence-end-double-space’ is non-nil, then period followed by one
space does not end a sentence, so don’t break a line there.
The variable ‘fill-column’ controls the width for filling.

If ‘fill-paragraph-function’ is non-nil, we call it (passing our
argument to it), and if it returns non-nil, we simply return its value.

If ‘fill-paragraph-function’ is nil, return the ‘fill-prefix’ used for filling.

The REGION argument is non-nil if called interactively; in that
case, if Transient Mark mode is enabled and the mark is active,
call ‘fill-region’ to fill each of the paragraphs in the active
region, instead of just filling the current paragraph.
```

DanGrayson:
PS: line wrapping is up to emacs, not to proof general

tpannila:
@vladimirias , I rebased and fixed lines to 100 characters. Please check
if the coding style is what you want and comment if there are things you want
to be fixed.

vladimirias:
Thank you, this is much better. 
LGTM

DanGrayson:
@vladimirias : in addition to my comment https://github.com/UniMath/UniMath/pull/424#issuecomment-244743151 : 

Another way to wrap lines is to have it happen automatically, after each word.  You do that by turning automatic fill mode on with this function:
```
auto-fill-mode is an interactive compiled Lisp function in
‘simple.el’.

(auto-fill-mode &optional ARG)

Toggle automatic line breaking (Auto Fill mode).
With a prefix argument ARG, enable Auto Fill mode if ARG is
positive, and disable it otherwise.  If called from Lisp, enable
the mode if ARG is omitted or nil.

When Auto Fill mode is enabled, inserting a space at a column
beyond ‘current-fill-column’ automatically breaks the line at a
previous space.

When ‘auto-fill-mode’ is on, the ‘auto-fill-function’ variable is
non-nil.

The value of ‘normal-auto-fill-function’ specifies the function to use
for ‘auto-fill-function’ when turning Auto Fill mode on.

[back]
```

vladimirias:
@DanGrayson : could you give an example of how these two methods work on something that is a little complicated , for example on [ rigpair ] from Rigs_and_Rings (first make one long line out of it as it was before Tomi’s work)?

DanGrayson:
Well, those automated methods are mainly for prose, where it is okay to break at the rightmost space occurring to the left of column 100, thus trying to get lines as full as possible.  But here goes.

You have this long line:
```
Definition rigpair { X : setwith2binop } ( is : isrigops ( @op1 X ) ( @op2 X ) ) : rig  := tpair ( fun X : setwith2binop =>  isrigops ( @op1 X ) ( @op2 X ) ) X is .
```
On this one, `fill-paragraph` does nothing, probably because it knows it is code rather than text.

If your turn on auto-fill mode and type a space at the end of the line, you get this:
```
Definition rigpair { X : setwith2binop } ( is : isrigops ( @op1 X ) ( @op2 X ) ) : rig
  := tpair ( fun X : setwith2binop =>  isrigops ( @op1 X ) ( @op2 X ) ) X is . 
```
I don't know why it broke the line there -- it had more space.

If you try typing that line in from scratch, word by word, you will eventually reach this state at a width of 97:
```
Definition rigpair { X : setwith2binop } ( is : isrigops ( @op1 X ) ( @op2 X ) ) : rig := tpair (
```
and then after pressing  "fun " the auto-indentation happens and you get
```
Definition rigpair { X : setwith2binop } ( is : isrigops ( @op1 X ) ( @op2 X ) ) : rig
  := tpair ( fun 
```
, from which point you can type the rest of the line.  I don't know why it chose to break before the :=, but I like it.

Here's how it works on a comment.  Suppose we have this:
```
(** Dependent sums.

One can not use a new record each time one needs it because the general theorems about this construction would not apply to new instances of "Record" due to the "generativity" of inductive definitions in Coq.

We use "Inductive" instead of "Record" here.

Using "Record" which is equivalent to "Structure" would allow us later to use the mechanism of canonical structures with total2.  By using "Structure", we could also get eta for dependent pairs, by adding the option "Set Primitive Projections.".

However, the use of "Inductive" allows us to obtain proof terms that are expressed in terms of the eliminator total2_rect that, unlike the "match" construct that would appear in the proof terms if we used "Record", has a known interpretation in the framework of the univalent model.

*)
```
All the lines are long.  Then if I use `fill-paragraph` I get this:
```
(** Dependent sums.

One can not use a new record each time one needs it because the general theorems about this
construction would not apply to new instances of "Record" due to the "generativity" of inductive
definitions in Coq.

We use "Inductive" instead of "Record" here.

Using "Record" which is equivalent to "Structure" would allow us later to use the mechanism of
canonical structures with total2.  By using "Structure", we could also get eta for dependent pairs,
by adding the option "Set Primitive Projections.".

However, the use of "Inductive" allows us to obtain proof terms that are expressed in terms of the
eliminator total2_rect that, unlike the "match" construct that would appear in the proof terms if we
used "Record", has a known interpretation in the framework of the univalent model.

*)
```
The longest line is 100.

DanGrayson:
Here is another example, with auto fill mode on.  You have
```
Definition lam_map : HSET2⟦precomp_option LC,LC⟧ :=
  CoproductIn bool HSET2 (Coproducts_functor_precat _ _ _ _ _ _) false ;; BinCoproductIn2 HSET2 (BinCoproducts_functor_precat _ _ _ _ _ _) ;; LC_mor.
```
and you type a space at the end of the second line.  Then you get
```
Definition lam_map : HSET2⟦precomp_option LC,LC⟧ :=
  CoproductIn bool HSET2 (Coproducts_functor_precat _ _ _ _ _ _) false ;; BinCoproductIn2
              HSET2 (BinCoproducts_functor_precat _ _ _ _ _ _) ;; LC_mor. 
```

PR #426, opened 2016-08-11
Cleaned Domains_and_Fields.v

tpannila:
Cleaned long lines in Domains_and_Fields.v and added Contents comment.


benediktahrens:
lgtm

vladimirias:
LGTM and merging.

DanGrayson:
I notice that in a few places "assert" has been changed to "set", Tomi, such as in the second command of the proof of lcanfromlinv.  That may not be a beneficial change, since it puts more expressions into the context that have to be acted on by tactics such as "simpl".  It's too late for this PR, but perhaps keep that in mind for the future.

PR #427, opened 2016-08-12
Cleaned NaturalNumbers.v

tpannila:
Cleaned long lines in NaturalNumbers.v and added Contents comment.

### DIFF
--- a/UniMath/Foundations/NumberSystems/NaturalNumbers.v
+++ b/UniMath/Foundations/NumberSystems/NaturalNumbers.v
@@ -1,9 +1,49 @@
-(** * Natural numbers and their properties. Vladimir Voevodsky . Apr. - Sep. 2011
+(** * Natural numbers and their properties. Vladimir Voevodsky. Apr. - Sep. 2011
 
-This file contains the formulations and proofs of general properties of natural numbers from the univalent perspecive. *)
-
-
-
+This file contains the formulations and proofs of general properties of natural
+numbers from the univalent perspecive. *)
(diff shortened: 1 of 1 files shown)


benediktahrens:
Starting from `natgehandminuslinv`, the proof code is not indented any more. Before that, you had used the convention of indenting proof code by 2 spaces.

Starting from `isplushrelnatgth`, some proofs have remained in one line, with `Proof.` and `Defined.` in the same line.

benediktahrens:
lgtm modulo the small incompletenesses mentioned above

tpannila:
@benediktahrens , there is a block of code in a comment starting at `natgehandminuslinv` .
The block also contains an unfinished definition `natlehandminusr` and a proof without 
statement. Thus I left the comment untouched.  Should I treat code inside a comment the
same as code outside a comment?

benediktahrens:
On 09/07/2016 12:23 AM, Tomi Pannila wrote:
> @benediktahrens , there is a block of code in a comment starting at `natgehandminuslinv` .
> The block also contains an unfinished definition `natlehandminusr` and a proof without 
> statement. Thus I left the comment untouched.  Should I treat code inside a comment the
> same as code outside a comment?

Sorry, I only looked at the diff and hence did not see the context.
No, don't work on commented code, that is likely to be a wasted effort.

vladimirias:
It is possible that there will be a major reworking of the part of hnat that concerns minus.

Vladimir.

vladimirias:
LGTM and merging.

PR #445, opened 2016-09-06
Primitive projections 2 pull request 1

DanGrayson:
As last done in #393 and addressing issue #111, here I provide some more changes that are needed to make our code work with primitive projections enabled, although we don't enable them here.  I've spent a few days working on it and have reported 3 Coq bugs, one of which was judged not to be a bug.

Many of the changes just involve replacing `apply` by `refine`, as before, or involve elaborating certain parameters that were left blank before.

One new aspect is that index numbers of subexpressions can change, depending on whether primitive projections are enabled, because the parameters to `pr1` and `pr2` may not actually be there any longer.  This affects commands that specify the index explicitly with `at`, as occurs especially in Catherine's code, e.g., in NonnegativeReals.v.  So now we have a command that selects between one of two alternatives based on the mode we are in.

It's worthwhile submitting these changes, because I want to work on other things for a while and don't want them to get lost.  (I'm stuck on my own file, Ktheory/AffineLine.v, and don't know what to do about it.  Coq seems to run forever on a `refine` statement that formerly was quick.


benediktahrens:
I have asked this question in the past, I think:

Is there a reason why you don't use the `use` tactic?
This tactic allows to omit the underscores (when it works) that `simple
refine` requires.

DanGrayson:
Oh, right, you've drawn my attention to it once before, and I could have used it here.

By the way, one thing I've noticed in our code a lot of: using `apply x` when `exact x` would have worked.  What's that all about?

vladimirias:
Is exactly lime that? For example apply foo and exact (foo _) are different in that that if the number of explicit arguments of foo changes than the first variant remains Ok and the second one breaks.

DanGrayson:
Yes, with "exact" one must put in the correct number of arguments, explicitly.  But it seems that "apply x" always will work whenever "exact x" works, even though "exact x" will not always work when "apply x" works.  It's the case where "x" takes 0 arguments.  To me it seems strange to use "apply x" when the number of arguments is 0.

benediktahrens:
lgtm. the case distinction on whether we have eta for pairs or not is both cute and terrifying.

DanGrayson:
Yes, but it's better to have a case distinction than to try both ways and take the one that works.

DanGrayson:
Oh, and in 8.6 there will be an official facility for telling whether a projection is primitive.

vladimirias:
In your correction in PartA, what would have not worked with the original code?

DanGrayson:
In PartA, if I recall correctly, that's one case where the change wasn't necessary, but it simplifies the proof.  I should have mentioned that.

benediktahrens:
lgtm modulo the merge conflict

DanGrayson:
Vladimir, and comments on this one?

DanGrayson:
*any*

vladimirias:
Sorry, running late on the UniMath things again.

mortberg:
LGTM modulo conflicts. 

A general comment: selecting subexpressions by their index is not so good as the resulting proofs are fragile and might break if Coq changes as illustrated by this PR. Maybe the proofs can be rewritten so that they don't use `pattern` with specific indices? However I never used `pattern` so I don't have a specific solution in mind...

DanGrayson:
Perhaps Catherine has an opinion about rewriting those uses of `pattern`, which I also have never used.  That could be a future project.

DanGrayson:
I'll resolve the conflicts and merge this one ...

vladimirias:
LGTM

benediktahrens:
Is there a reason why you want to update Coq now instead of waiting for the next patch level?

Issue #397, opened 2016-07-02
abstract subproofs


DanGrayson:
To make sure our code compiles on the trunk, I've had to make this change:
```
--- a/UniMath/CategoryTheory/lists.v
+++ b/UniMath/CategoryTheory/lists.v
@@ -409,7 +409,7 @@ simple refine (tpair _ _ _).
     apply funextfun; intro p; rewrite (paireta p);
     generalize (colimArrowCommutes (mk_ColimCocone hF c L ccL) _
                  (mk_cocone _ (omega_cocontConstProdFunctor_subproof
-                               hF c L ccL HcL (f,,hf))) n);
+                               hF HcL (f,,hf))) n);
     unfold flip, curry, colimIn; simpl; intro H;
     now rewrite <- (toforallpaths _ _ _ (toforallpaths _ _ _ H (pr2 p)) (pr1 p))).
 - abstract (
```

In 8.5pl1 the subproof had this type:
```
Import UniMath_DOT_CategoryTheory_DOT_covyonedaomega_cocontConstProdFunctor_subproof
     : ∀ (hF : chain HSET) (HcL : HSET)
       (cc : cocone (mapdiagram constprod_functor hF) HcL) 
       (u v : vertex nat_graph) (e : edge u v),
       dmor hF e ;; flip (curry (pr1 cc v)) = flip (curry (pr1 cc u))
```
, but on the trunk it has this type:
```
omega_cocontConstProdFunctor_subproof
     : ∀ (hF : chain HSET) (c : HSET) (L : cocone hF c),
       isColimCocone hF c L
       → ∀ (HcL : HSET)
       	 (cc : cocone (mapdiagram constprod_functor hF) HcL)
         (u v : vertex nat_graph) (e : edge u v),
         dmor hF e ;; flip (curry (pr1 cc v)) = flip (curry (pr1 cc u))

```

It is unsound to use abstracted subproofs by name, since the precise form of the abstraction
has turned out to be Coq-implementation dependent, and highly unlikely to be understood by a future proof assistant.  Can someone find a proof avoiding citing the subproof?

Shall we add something to the style guide saying not to use abstracted subproofs by name?

DanGrayson:
I'll add something to the style guide ...

mortberg:
Good catch! I don't know what I was thinking while writing this proof... I don't think one should ever rely on automatically generated names for arguments or lemmas. For example one should only use `intros` without giving names to the introduced arguments when the names are not used later.

[...]

PR #446, opened 2016-09-06
Add a rule

DanGrayson:
this resolves issue #397

### DIFF
--- a/UniMath/README.md
+++ b/UniMath/README.md
@@ -62,9 +62,10 @@ less fragile and to make the files have a more uniform and pleasing appearance.
   * Do not start lines with `:` or with `:=`.
   * One should normally put an extra blank line between units.  Exceptions may
     be made for closely related items.
-
-Our files don't adhere yet to all of these conventions, but it's a goal we
-strive for.
+* When using `abstract` in a proof, it is unsound to refer later by name to the
+  abstracted lemma (whose name typically ends with `_subproof`), because
+  its type may vary from one version of Coq to another.  Coq's current behavior is also
+  unlikely to be duplicated precisely by a future proof assistant. 
 
 Another advantage of coding in this style is that the proofs should be easier
 to transport to another proof assistant.
--- a/UniMath/README.md
+++ b/UniMath/README.md
@@ -67,5 +67,8 @@ less fragile and to make the files have a more uniform and pleasing appearance.
   its type may vary from one version of Coq to another.  Coq's current behavior is also
   unlikely to be duplicated precisely by a future proof assistant. 
 
+Our files don't adhere yet to all of these conventions, but it's a goal we
+strive for.
+
 Another advantage of coding in this style is that the proofs should be easier
 to transport to another proof assistant.


benediktahrens:
lgtm

vladimirias:
What does "abstract" do? I don't see it in the manual.

DanGrayson:
It's in the Global Index, and is on p 262 of my edition of the manual.  You wrap it around a part of your proof that satisfies the current goal, and it puts all that stuff into a separate lemma that ends with Qed.  It's a way to make parts of your proof opaque without having to create separate lemmas yourself.

DanGrayson:
... Here is an example:
```
Definition dirprod_hSet (X Y : hSet) : hSet.
Proof.
  intros X Y. exists (X × Y).
  abstract (exact (isasetdirprod _ _ (setproperty X) (setproperty Y))).
Defined.
```
Now print out the proof and you get this:
```
dirprod_hSet = λ X Y : hSet, (X × Y),, dirprod_hSet_subproof X Y
     : hSet → hSet → hSet
```
Print out the subproof and you get this:
```
Fetching opaque proofs from disk for UniMath.Foundations.Basics.Sets

dirprod_hSet_subproof = 
λ X Y : hSet, isasetdirprod (pr1 X) (pr1 Y) (setproperty X) (setproperty Y)
     : Π X Y : hSet, isaset (X × Y)
```

vladimirias:
Is there a way to give a name to that lemma? E.g. to make it to be called dirprod_hSet_l1 instead?

benediktahrens:
Yes, apparently you could do
```
abstract (...) using dirprod_hSet_l1.
```
However, it is considered bad practice to refer to that lemma
`dirprod_hsSet_l1` outside the enclosing `Defined.` and, as Dan
discovered, the type of the produced opaque proof can change in between
versions of Coq.

We should not use that feature.

DanGrayson:
I use "using" a lot with "abstract", but just so the proof terms look smaller on the screen.  I use 1 letter names.

benediktahrens:
On 09/07/2016 07:25 PM, Daniel R. Grayson wrote:
> I use "using" a lot with "abstract", but just so the proof terms look
> smaller on the screen.  I use 1 letter names.

Does that mean that you create a lot of 1 letter identifiers that get
imported when one imports your files?

DanGrayson:
Yes, you're right:
````
Constant UniMath.Ktheory.Precategories.L
Constant UniMath.Ktheory.Bifunctor.L
  (shorter name to refer to it in current context is Bifunctor.L)
Constant UniMath.Ktheory.Representation.L
  (shorter name to refer to it in current context is Representation.L)
```

benediktahrens:
On 09/07/2016 08:02 PM, Daniel R. Grayson wrote:
> Yes, you're right:
> ````
> Constant UniMath.Ktheory.Precategories.L
> Constant UniMath.Ktheory.Bifunctor.L
>   (shorter name to refer to it in current context is Bifunctor.L)
> Constant UniMath.Ktheory.Representation.L
>   (shorter name to refer to it in current context is Representation.L)
> ```

Sorry, but that is awful.

DanGrayson:
No, lne never sees those things.

benediktahrens:
On 09/07/2016 09:56 PM, Daniel R. Grayson wrote:
> No, lne never sees those things.

Well, they behave like normal definitions, right? So they get imported
and are accessible by their short name, e.g., by "L", in any file that
imports UniMath.Ktheory.Precategories.

I don't think one can say "one never sees those things" about such a thing.

benediktahrens:
On 09/07/2016 08:02 PM, Daniel R. Grayson wrote:
> Yes, you're right:
> ````
> Constant UniMath.Ktheory.Precategories.L
> Constant UniMath.Ktheory.Bifunctor.L
>   (shorter name to refer to it in current context is Bifunctor.L)
> Constant UniMath.Ktheory.Representation.L
>   (shorter name to refer to it in current context is Representation.L)
> ```

Since everybody is in a rule-making mood right now, let me suggest
another one:

One-letter strings should not be used as identifiers. Instead, they are
reserved for naming variables.

DanGrayson:
Well, of course, if you type L in your file you'll get one of those things.  But I never do that.  And even if I did, the error would be immediately apparent, so it's not a problem.  And it doesn't prevent me from using L as the name of one of my variables.

I'm happy to change all my L's to _L's, if you like that rule a lot and are in a French bureaucratic mood.

benediktahrens:
On 09/07/2016 10:07 PM, Daniel R. Grayson wrote:
> Well, of course, if you type L in your file you'll get one of those
> things.  But I never do that.  And even if I did, the error would be
> immediately apparent, so it's not a problem.  

Well, it would be *to you*, because you know about those subproofs that
you gave the name "L".

For someone like me who just wants to use your files in a naive way
without reading through the proofs to discover your use of the obscure
"abstract (...) using L", an error message could be very confusing.
In particular, a superficial browsing of your file does probably not
reveal that bad habit of yours to the user, because one only looks for
identifiers after a `Lemma` or `Definition`.

> And it doesn't prevent
> me from using L as the name of one of my variables.
> 
> I'm happy to change all my L's to _L's, if you like that rule a lot
> and are in a French bureaucratic mood.

How is anything more bureaucratic than having a rule for the length of
lines? Compared to that, the issue of keeping variable names and
identifiers disjoint seems like one of the great issues mankind is facing.

I would appreciate it a lot if you'd rename those "L" to something else,
similar to what you did for "__t__" and "__p__" - which I find a great
idea, btw.

DanGrayson:
Agreed, see https://github.com/UniMath/UniMath/issues/450

But the rule about 100 as line length -- hmm, was that mine, do you think?

benediktahrens:
On 09/07/2016 10:32 PM, Daniel R. Grayson wrote:
> Agreed, see https://github.com/UniMath/UniMath/issues/450
> 
> But the rule about 100 as line length -- hmm, was that mine, do you
> think?

Maybe it wasn't. Let me mention the removal of whitespaces instead,
which set back the whole UniMath development by several days during
which people (me at least) were busy fixing merge conflicts over and
over again.
You might argue that Anders brought up the subject, but it was your
enthusiasm that made the idea go through :-)

DanGrayson:
:-)

vladimirias:
LGTM for the current rule and merging. For the other issue maybe we should raise an issue?

DanGrayson:
yes, the other issue (short id names) has been dealt with

mortberg:
Sorry about the whitespace, that was not very important and caused a lot of trouble for very little gain. Having a limit for line length is very important though :-)

DanGrayson:
Well, somehow I screwed up with the initial whitespace elimination, and only a portion of it was eliminated in the pull request.  I never did figure out why my method didn't really work, and why I didn't detect it.  But all those spaces are gone now, I just checked.

vladimirias:
I would even say one and two letter strings.

mortberg:
This rule sounds very good to me, as long as it is not covering local definitions. When doing a complicated proof with many hypotheses it is nice to define things outside of the proof and then one might want to use short identifier names.

vladimirias:
The number 100 was probably suggested by me. For the rule in general we had a discussion started, I think, by the fact that very long lines are hard to read in GitHub.

DanGrayson:
Re: "I would even say one and two letter strings."

Except we just suggested to Tomi that "Im" would be good; I'd prefer "image", though.

vladimirias:
Ab_Im?

DanGrayson:
`Ab_Im` is not as human-friendly as `image` is.

mortberg:
For this kind of thing I think a notation is perfect. You have a long (unique) identifier name (like `AbelianIm`) that doesn't clash with anything else and then a good notation put in a scope.

DanGrayson:
No, we don't need two names for things.  The long name that doesn't clash is the qualified one: `Abelian.image`.

vladimirias:
Yes, but it is short and confirms better to having different prefixes to distinguish constructions with similar names and different meaning such as image in an abelian category and image of a function between types.

vladimirias:
This is a good idea.

vladimirias:
No, this is a wrong approach. One can still use “image” outside of Abelian and it will clash with other “image”.

What is wrong with having a notation?

Or use Ab_Im or, better yet ab_im.

DanGrayson:
On Thu, Sep 8, 2016 at 8:24 PM, Vladimir Voevodsky <notifications@github.com
> wrote:

> No, this is a wrong approach. One can still use “image” outside of Abelian
> and it will clash with other “image”.
>

Why would it clash with anything?  Suppose I have a file foo1.v with
```
Definition n := 1.
```
in it and I have a file foo2.v with
```
Definition n := 2.
```
in it.

Then another file like this works fine:
```
Require Export UniMath.foo1.
Print n.                        (* shows 1 *)
Require Export UniMath.foo2.
Print n.                        (* shows 2 *)
Definition n := 3.
Print n.                        (* shows 3 *)
Print foo1.n.                   (* shows 1 *)
Print foo2.n.                   (* shows 2 *)
```


>
> What is wrong with having a notation?
>

Then you have two different names for one thing, and you have to remember
both of them.


>
> Or use Ab_Im or, better yet ab_im.
>

I don't like the idea of requiring everyone to use long names made ugly by
appending type information
to the name.  People who like less ambiguity could type Abelian.Im if they
want to.

DanGrayson:
Here's the real problem with notations: they preempt all other uses of the symbol.  For
example, this does not work:
```
Notation "'m'" := 3.
Definition f (m:nat) := m.
```

JasonGross:
The solution to this is to say
```coq
Notation m := 3.
```
Note the lack of quotes.  See the section in the manual on aliases for more details.

mortberg:
Also I think notations should always either be introduced locally (using `Local Notation`) or put in a scope. This way Dan's `m` can be perfectly fine if introduced locally in a section where `m` is supposed to mean 3. Global notations are bad because they pollute the namespace. 

The beauty with scopes is that the user can easily control what symbols mean what. For example if I want to do proofs about abelian categories I open the abelian categories scope and then things like `im`, `ker`, `mono`, etc. mean what I want. Later when I want to do proofs about R-modules I open that scope and the same symbols means what I expect them to. This way we get the same kind of overloading of symbols as in a mathematical textbook without having to fiddle with qualified names. Also I think another problem with using files/Modules for this is that it is more difficult to know which symbols are in scope when you have imported a lot of different files/Modules (which in turn might be exporting other files/Modules...). Maybe it is easy for me to know what things I get when importing files I've written myself, but for users it is not so easy (of course the user can easily figure out where things come from, but if you have to do this all the time and then figure out which qualifier to put in front of your identifier I would get tired rather quickly...). 

However I'm not suggesting adding a rule about these kind of things. Maybe a guideline could be to use notations and scopes for very commonly used mathematical notions (like `image`) while for less commonly used mathematical notions it might be better to use qualified names to disambiguate.

DanGrayson:
Re: `Notation m := 3.`

Oops, yes, that works, thanks.  (I thought I tried that.)

DanGrayson:
Re: `Notation m := 3.` again.

How do we put that in a scope?  I tried`Notation m := n : A.` but it didn't work.  (I was right
when I thought I had tried it.)

Section 12.3 of the manual on "Abbreviations" covers this, and it says the general form is
```
[Local]Notationident[identident...identident]:=term[(only parsing)].
```
So there seems to be possibility for putting abbreviations into a scope, as Anders desires.

Meanwhile, modules offer all the desired functionality.  Instead of opening a scope, you import the module, and then you have desired short names available.  (Yes, I know all of our modules are supposed to be entire files, according to our rules.)

mortberg:
Ok, fair enough. Then I would suggest that abbreviation should be local while notations should either be local or in a scope. I still think that notations are better for some things (like very common mathematical symbols) while modules are good for some things (organizing mathematical theories).

DanGrayson:
Okay, thanks, it seems we've explored this issue thoroughly!

JasonGross:
Abbreviations are like lightweight definitions.  They can be accessed by qualified names, and you "bring them into scope" by importing a module.  As your comment might be suggesting, you could do something like:
```coq
Module Import ShortCategoryNames.
  Notation im := ...
  ...
End.
```
Then you must import the module if you want access to the short names in other files.  Alas, Coq has no equivalent of OCaml's `let open M in ...`

DanGrayson:
Yes, but rather than everything having a long name and a short name, how about everything having name, and it's short?

PR #447, opened 2016-09-06
improve TAGS file

DanGrayson:
so that it refers also to the Tactic Notations "use" and "assert" defined in Preamble.v

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,9 +56,33 @@ ENHANCEDDOCTARGET = enhanced-html
 ENHANCEDDOCSOURCE = util/enhanced-doc
 LATEXDIR = latex
 COQDOCLATEXOPTIONS := -latex -utf8 --body-only
-COQDEFS := --language=none																																					   \
-	-r '/^[[:space:]]*\(\(Local\|Global\)[[:space:]]+\)?\(Axiom\|Theorem\|Class\|Instance\|Let\|Ltac\|Definition\|Identity Coercion\|Lemma\|Record\|Remark\|Structure\|Fixpoint\|Fact\|Corollary\|Let\|Inductive\|CoInductive\|Notation\|Proposition\|Module[[:space:]]+Import\|Module\)[[:space:]]+\([[:alnum:]'\''_]+\)/\4/' \
-	-r "/^[[:space:]]*Notation.* \"'\([[:alnum:]]+\)'/\1/"																																	   \
+
+DEFINERS := 
+DEFINERS := $(DEFINERS)Axiom\|
+DEFINERS := $(DEFINERS)Class\|
+DEFINERS := $(DEFINERS)CoInductive\|
+DEFINERS := $(DEFINERS)Corollary\|
+DEFINERS := $(DEFINERS)Definition\|
+DEFINERS := $(DEFINERS)Fact\|
+DEFINERS := $(DEFINERS)Fixpoint\|
+DEFINERS := $(DEFINERS)Identity[[:space:]]+Coercion\|
+DEFINERS := $(DEFINERS)Inductive\|
+DEFINERS := $(DEFINERS)Instance\|
+DEFINERS := $(DEFINERS)Lemma\|
+DEFINERS := $(DEFINERS)Let\|
+DEFINERS := $(DEFINERS)Let\|
+DEFINERS := $(DEFINERS)Ltac\|
+DEFINERS := $(DEFINERS)Module[[:space:]]+Import\|
+DEFINERS := $(DEFINERS)Module\|
+DEFINERS := $(DEFINERS)Notation\|
+DEFINERS := $(DEFINERS)Proposition\|
+DEFINERS := $(DEFINERS)Record\|
+DEFINERS := $(DEFINERS)Remark\|
+DEFINERS := $(DEFINERS)Structure\|
+DEFINERS := $(DEFINERS)Theorem
+COQDEFS := --language=none												\
+	-r '/^[[:space:]]*\(\(Local\|Global\)[[:space:]]+\)?\($(DEFINERS)\)[[:space:]]+\([[:alnum:]'\''_]+\)/\4/'	\
+	-r "/^[[:space:]]*Notation.* \"'\([[:alnum:]]+\)'/\1/"								\
 	-r '/^[[:space:]]*Tactic Notation.* "\([[:alnum:]]+\)" /\1/'
 $(foreach P,$(PACKAGES),$(eval TAGS-$P: $(filter UniMath/$P/%,$(VFILES)); etags -o $$@ $$^))
 $(VFILES:.v=.vo) : $(COQBIN)coqc
--- a/Makefile
+++ b/Makefile
@@ -60,16 +60,18 @@ COQDOCLATEXOPTIONS := -latex -utf8 --body-only
 DEFINERS := 
 DEFINERS := $(DEFINERS)Axiom\|
 DEFINERS := $(DEFINERS)Class\|
+DEFINERS := $(DEFINERS)CoFixpoint\|
 DEFINERS := $(DEFINERS)CoInductive\|
 DEFINERS := $(DEFINERS)Corollary\|
 DEFINERS := $(DEFINERS)Definition\|
+DEFINERS := $(DEFINERS)Example\|
 DEFINERS := $(DEFINERS)Fact\|
 DEFINERS := $(DEFINERS)Fixpoint\|
+DEFINERS := $(DEFINERS)Function\|
 DEFINERS := $(DEFINERS)Identity[[:space:]]+Coercion\|
 DEFINERS := $(DEFINERS)Inductive\|
 DEFINERS := $(DEFINERS)Instance\|
 DEFINERS := $(DEFINERS)Lemma\|
-DEFINERS := $(DEFINERS)Let\|
 DEFINERS := $(DEFINERS)Ltac\|
 DEFINERS := $(DEFINERS)Module[[:space:]]+Import\|
 DEFINERS := $(DEFINERS)Module\|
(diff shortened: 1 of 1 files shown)


benediktahrens:
I don't use the TAGS mechanism, so I don't know how to evaluate/test this PR.

DanGrayson:
Do you want to test it?  (I could explain how.)  Or just rely on me to fix it for others if anyone complains?

I use the feature all the time -- it's essential.

benediktahrens:
On 09/08/2016 02:34 PM, Daniel R. Grayson wrote:
> Do you want to test it?  (I could explain how.)  Or just rely on me
> to fix it for others if anyone complains?
> 
> I use the feature all the time -- it's essential.

Actually we have a wiki page describing emacs tags - thanks, @mortberg.
I'll try it next time I feel the need to rename something.

In the meantime, we can pull this PR.

[1] https://github.com/UniMath/UniMath/wiki/Emacs-tags

DanGrayson:
Re: "I'll try it next time I feel the need to rename something."

Perhaps you think the main use is for global search and replace.  I think the main
use is for jumping quickly to the definition of an identifier.  For example, if I want to know
what "chain" does, a keystroke takes me to the point in the source code where it's defined.

Makefile
```diff
@@ -56,7 +56,10 @@ ENHANCEDDOCTARGET = enhanced-html
 ENHANCEDDOCSOURCE = util/enhanced-doc
 LATEXDIR = latex
 COQDOCLATEXOPTIONS := -latex -utf8 --body-only
-COQDEFS := --language=none -r '/^[[:space:]]*\(Local[[:space:]]+\)?\(Axiom\|Theorem\|Class\|Instance\|Let\|Ltac\|Definition\|Identity Coercion\|Lemma\|Record\|Remark\|Structure\|Fixpoint\|Fact\|Corollary\|Let\|Inductive\|Coinductive\|Notation\|Proposition\|Module[[:space:]]+Import\|Module\)[[:space:]]+\([[:alnum:]'\''_]+\)/\3/'
+COQDEFS := --language=none																																			       \
+	-r '/^[[:space:]]*\(Local[[:space:]]+\)?\(Axiom\|Theorem\|Class\|Instance\|Let\|Ltac\|Definition\|Identity Coercion\|Lemma\|Record\|Remark\|Structure\|Fixpoint\|Fact\|Corollary\|Let\|Inductive\|Coinductive\|Notation\|Proposition\|Module[[:space:]]+Import\|Module\)[[:space:]]+\([[:alnum:]'\''_]+\)/\3/' \
```
mortberg: Speaking of long lines...

benediktahrens:
On 09/08/2016 03:14 PM, Daniel R. Grayson wrote:
> Re: "I'll try it next time I feel the need to rename something."
> 
> Perhaps you think the main use is for global search and replace.

It is the only use documented in the wiki.

> I think the main use is for jumping quickly to the definition of an 
> identifier.  For example, if I want to know what "chain" does, a 
> keystroke takes me to the point in the source code where it's 
> defined.

That sounds awesome.
Could you explain more? I'll document the stuff in the wiki.

mortberg:
I think `Coinductive` should be `CoInductive` not that it should match anything in UniMath, but still.

mortberg:
Also maybe it should include `Global` as well as `Local`? I'm not sure if anyone is using that either though

DanGrayson:
Okay.  (By the way, global search is important, too.)

Suppose you are looking at 
```
Definition counit_pointwise_iso_from_adj_equivalence {A B : precategory}
  {F : functor A B} (HF : adj_equivalence_of_precats F) :
    Π b, iso (F (HF^^-1 b)) b.
  intro b.
  exists (counit_from_left_adjoint (pr1 HF) b).
  exact (pr2 (pr2 HF) b).
Defined.
```
and you wonder about `counit_from_left_adjoint`.  Put the cursor on it and press `M-.`.
In the most recent version of emacs you'll be taken instantly to here:
```
Definition counit_from_left_adjoint {A B : precategory}
  {F : functor A B}   (H : is_left_adjoint F)  :
 nat_trans (functor_composite (right_adjoint H) F) (functor_identity B)
   := pr2 (pr1 (pr2 H)).
```
Now suppose you wonder about `nat_trans`.  Do the same and it will take you here:
```
Definition nat_trans {C C' : precategory_data}
  (F F' : functor_data C C') := total2 (
   fun t : Π x : ob C, F x -->  F' x => is_nat_trans F F' t).
```
To get back one step, use `M-,`.

There are variations where you get to enter the name of the identifier yourself, with completion.

There is also a command for going to the next instance of a definition of that identifier, and one for specifying a class of identifiers as a regular expression.

DanGrayson:
Here's a use for global tags search:

I want to see all the tactics we've defined, so I search for "Ltac".

Or I want to see whether I can delete a lemma, so I search for all uses of its name.

DanGrayson:
... Or I want to move a lemma to a later file, so I search for its name.  The TAGS file
is ordered appropriately, so it finds uses in dependency sequence.  If you want to move something from PartA to PartD, this will help you.

Makefile
DanGrayson: Yes, that's long, but it's not covered by our guidelines, which apply only to proof files.

DanGrayson: But at least I didn't make the line longer in this PR!

mortberg: Would it be possible to split it up like you do for the two new rules? That way one could easier spot typos, like the one in CoInductive...

DanGrayson: One could split it up, but that would make it harder to detect other typos, those where one of the many
lines differs from another.  I recommend not doing it.
```
	-r '/^[[:space:]]*\(\(Local\|Global\)[[:space:]]+\)?Theorem[[:space:]]+\([[:alnum:]'\''_]+\)/\4/' \
	-r '/^[[:space:]]*\(\(Local\|Global\)[[:space:]]+\)?Lemma[[:space:]]+\([[:alnum:]'\''_]+\)/\4/' \
	...
```

mortberg: Ok, seems reasonable. Is it possible to create a constant for the list of things like Axiom\|Theorem\|... and then write something like:

STUFF := Axiom\|Theorem\|Class\|Instance\|Let\|Ltac\|Definition\|Identity Coercion\|Lemma\|Record\|Remark\|Structure\|Fixpoint\|Fact\|Corollary\|Let\|Inductive\|Coinductive\|Notation\|Proposition\|Module

-r '/^[[:space:]]*\(\(Local\|Global\)[[:space:]]+\)?$(STUFF)[[:space:]]+\([[:alnum:]'\''_]+\)/\4/' \

mortberg:
Maybe you should add `Arguments` as well?

mortberg:
I wonder if we can get a list of all reserved keywords and just put them all in there... Adding them one by one like this is a little dangerous.

Makefile
DanGrayson: I offer an improvement in this respect.

DanGrayson:
`Arguments` does not define anything, it just modifies a previous definition.

DanGrayson:
It's not all reserved keywords, it's all things that make definitions in the same way that "Lemma" does, with Local or Global optionally before it and the identifier after it.  But, take a look in
`sub/coq` at the file `parsing/g_vernac.ml4` and see if you can figure it out.

Makefile
```diff
@@ -56,7 +56,34 @@ ENHANCEDDOCTARGET = enhanced-html
...
+DEFINERS := $(DEFINERS)Corollary\|
+DEFINERS := $(DEFINERS)Definition\|
+DEFINERS := $(DEFINERS)Fact\|
+DEFINERS := $(DEFINERS)Fixpoint\|
+DEFINERS := $(DEFINERS)Identity[[:space:]]+Coercion\|
+DEFINERS := $(DEFINERS)Inductive\|
+DEFINERS := $(DEFINERS)Instance\|
+DEFINERS := $(DEFINERS)Lemma\|
+DEFINERS := $(DEFINERS)Let\|
+DEFINERS := $(DEFINERS)Let\|
```
mortberg: Let occurs twice

mortberg:
Ah yes of course, my bad. I'll take a look at the file

mortberg:
Ok, here comes a bunch of things that maybe should go into the TAGS file. Most of these are definitely forbidden from UniMath, but it might still be nice to have a complete TAGS file. 

There are some more ways to define things: `CoFixpoint`, `Example`, `Function`, `Infix`

There is also `Axioms` that maybe should be included? Also what about `Section` (we do have `Module` in there)?

Finally we have the following modifiers that can be put in front of things:
`Private` (as in `Private Inductive`)
`Program` (as in `Program Definition`)
`Canonical` (as in `Canonical Structure`)

I have probably missed a few (for example `Scheme` could be an example, but I've never used it so I don't know if it should be in there or not)...

DanGrayson:
Okay, I added the ones that are sensible.  (Infix is for symbols, not identifiers, so it won't work with tags.)

mortberg:
Ok, great! LGTM

benediktahrens:
On 09/08/2016 08:09 PM, Anders Mörtberg wrote:
> Ok, here comes a bunch of things that maybe should go into the TAGS
> file. Most of these are definitely forbidden from UniMath, but it
> might still be nice to have a complete TAGS file.
> 
> There are some more ways to define things: `CoFixpoint`, `Example`,
> `Function`, `Infix`
> 
> There is also `Axioms` that maybe should be included? Also what about
> `Section` (we do have `Module` in there)?
> 
> Finally we have the following modifiers that can be put in front of
> things: `Private` (as in `Private Inductive`) `Program` (as in
> `Program Definition`) `Canonical` (as in `Canonical Structure`)
> 
> I have probably missed a few (for example `Scheme` could be an
> example, but I've never used it so I don't know if it should be in
> there or not)...

Slightly off-topic: why is the tags stuff not part of PG? Isn't that
where it naturally belongs?

DanGrayson:
It has nothing to do with PG -- it's useful for anyone using emacs to edit Coq source files.

rmatthes:
A colleague of mine in Toulouse suggested to me to use company-coq as an extension of Proof General that also lets the user jump to the source code defining an identifier (also with M-.). There are many other features to ease the life of a user of huge libraries. E.g., Ctrl + left mouse shows the print information of an identifier right below the cursor in the buffer. The bulleted items are clickable to hide them. Symbols are clickable to show how they are typed in. Etc. Not to speak of completion features.

[Available at Github](https://github.com/cpitclaudel/company-coq) and installed in minutes, comes with an online tutorial
Author: Clément Pit-Claudel

DanGrayson:
Company-coq is pretty cool, thanks for the suggestion.  We should mention it in the README.

PR #449, opened 2016-09-07
decidable proofs

DanGrayson:
Here is one of the ideas of #257, implemented in a way closer to spirit of
Vladimir's original code about going back and forth between decidable relations
and relations with boolean values.  The way I implemented it in #257 was using
decidable relations whose values are always one of two propositions:
`hTrue` or `hFalse`.  The two ways are equivalent.

This should simplify #257.

When trying to prove a statement that is decidable, it is not necessary
to think about the proof, because the proof of decidability implements
an algorithm for deciding.  All we need to do is to figure out how to
"run" it.  The same applies to proving the negation of such a statement.

Here we offer some tactics that make the notation for doing so easy.

The idea is similar to the "canonical term" notation "ct", and we put it
in the same place in the code.

We illustrate the method by cleaning up 5 proofs about rational numbers
in RealNumbers/Prelim.v, each of which seemed ad hoc.  The new proofs
seem more uniform and involve no thought when constructing them.

### DIFF
--- a/UniMath/Foundations/Basics/Sets.v
+++ b/UniMath/Foundations/Basics/Sets.v
@@ -1141,6 +1141,25 @@ Defined.
 Notation " 'ct' ( R , is , x , y ) " := (ctlong R is x y (idpath true))
                                           (at level 70).
 
+(* An alternative to [ct], with tactics and negations *)
+
+Definition deceq_to_decrel {X:UU} : isdeceq X -> decrel X.
+Proof. intros ? i. use decrelpair.
+       - intros x y. exists (x=y). now apply isasetifdeceq.
+       - exact i.
+Defined.
+
+Ltac exact_op x := (* from Jason Gross: same as "exact", but with unification the opposite way *)
+  let T := type of x in
+  let G := match goal with |- ?G => constr:(G) end in
+  exact ((@id G : T -> G) x).
+
+(* I don't know why exact_op works better here, but with "exact", the code in RealNumbers/Prelim.v breaks *)
+Ltac confirm_yes d x y := exact_op (pathstor d x y (idpath true)).
+Ltac confirm_no  d x y := exact_op (pathstonegr d x y (idpath false)).
+Ltac confirm_equal     i := match goal with |- ?x = ?y => confirm_yes (deceq_to_decrel i) x y end.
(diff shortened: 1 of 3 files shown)


benediktahrens:
lgtm

DanGrayson:
I wonder if this is acceptable now, Vladimir.

vladimirias:
This looks very nice. Just one request: could you pease explain to me exact_op especially line 1155?

DanGrayson:
Here's Jason's tactic
```
Ltac exact_op x := (* from Jason Gross: same as "exact", but with unification the opposite way *)
   let T := type of x in
   let G := match goal with |- ?G => constr:(G) end in
   exact ((@id G : T -> G) x).
```
Here T is the type of x and G is the goal type, so `exact x` succeeds if G and T can be unified,
i.e., if the unification algorithm `unify T G` returns true, let's say.  Here "unify T G" is just
temporary notation: perhaps internally to Coq the name is something else and the parameters
are in the other order.  In any case the identity function`@id G : T -> G` is of type `T -> G`
and can be applied to`x` to yield something of the type needed by the goal, so all we need
is for the identity function to typecheck correctly.  The type of `@id G` is `G -> G`, so
the problem is to unify `T -> G` with `G -> G`, to see whether the type cast works.
Apparently this is done by evaluating `unify (G -> G) (T -> G)` -- it goes by structural
induction and succeeds only if `unify G T` succeeds -- the useful point is that the parameters
are in the opposite order.

Now this can be regarded as a useful accident about the difference between the way the
"exact" tactic works and the way the type cast ":" works.  Suppose it had not worked out:
i.e., that the type cast were checked by evaluating `unify (T -> G) (G -> G)` instead.  Then
we could have used `exact ((@id T : T -> G) x)`, which would have led to evaluating
`unify (T -> G) (T -> T)`, which would succeed only if `unify G T` succeeds.  So one of
the two was bound to work.

@JasonGross can correct me, if needed.

vladimirias:
Thanks! I have one request - please use the Foundations name for the identity function, which is idfun. It is chosen to be more informative than “id” which also reminds of the identity type.

DanGrayson:
Vladimir, what about a mini-project where we replace the core of Coq by something minimal,
containing just what we use, something without "id"?

vladimirias:
You mean changing Coq.init library? Let's open it as an issue.

Issue #316, opened 2016-03-08
Bibliography for the UniMath


vladimirias:
A lot of what we do when we write code for the UniMath is original research (e.g. all of the Basics is) but there are important cases when we formalize definitions formulated by others or use some source books to help us structure the material.

In such cases it is important that we start including literature references. Therefore, we need a bibliography file. Let us discuss how to make a workable version of it soon so that we can all start using it. It should be such that if we want to make some changes later we won't have to go through all the files and find references by hand.

DanGrayson:
Currently I'm just making citations in comments, like this: https://github.com/UniMath/UniMath/blob/master/UniMath/Foundations/Algebra/IteratedBinaryOperations.v#L57

benediktahrens:
On 03/09/2016 12:45 PM, Daniel R. Grayson wrote:
> Currently I'm just making citations in comments, like this:
> https://github.com/UniMath/UniMath/blob/master/UniMath/Foundations/Algebra/IteratedBinaryOperations.v#L57

If I understand Vladimir's message correctly, his goal is exactly to 
avoid this.

DanGrayson:
This was addressed in commits a090d5de90843f97369970ffbbd8a6e829a47b50, d165c72d18622aace5a5b367bc57414218ab1f29, and 878dc3d444b71d2b59605eefd6b7164504a292df.

What else remains to be done?

vladimirias:
There is a lot of stuff in these commits. I have no idea what it does. Where can I see written how to use the references in my comments?

DanGrayson:
There is no documentation, and there is just one example, in IteratedBinaryOperations:
```
  (** This proof comes from the Associativity theorem, % \cite[section 1.3, Theorem 1, page 4]{BourbakiAlgebraI}. \par % *)
```
It looks good in the pdf file but vanishes in the html version, so there is some work to do.  The bibtex items are to be inserted in the file `latex/references.bib`.

PR #451, opened 2016-09-07
Clean cocontinuous functors

mortberg:
This PR cleans the formalization of cocontinuous functors by:

* Merge all the theory about cocontinuous functors into CocontFunctors.v
* Add more documentation and make HTML output nicer
* Weaken a hypothesis of is_omega_cocont_binproduct_functor as suggested by Benedikt (this will be improved further in a future PR)
* Lots of cleaning, get rid of some `match`'s and add `Local` to some Definitions


UniMath/CategoryTheory/CocontFunctors.v
```diff
@@ -78,75 +82,514 @@ Require Import UniMath.CategoryTheory.CommaCategories.
...
+(** * Adámek's theorem for constructing initial algebras of omega-cocontinuous functors *)
+(* This section proves that (L,α : F L -> L) is the initial algebra
+    where L is the colimit of the inital chain:
+
+         !          F !           F^2 !
+     0 -----> F 0 ------> F^2 0 --------> F^3 0 ---> ...
+
+This result is also known as Adámek's theorem:
+
+  https://ncatlab.org/nlab/show/initial+algebra+of+an+endofunctor#AdameksTheorem
```
benediktahrens: Could you insert a reference to Adamek's article here, as discussed in #316 ?

UniMath/CategoryTheory/CocontFunctors.v
```diff
@@ -78,75 +82,514 @@ Require Import UniMath.CategoryTheory.CommaCategories.
...
+  destruct (natlehchoice4 _ _ Hij).
+  + rewrite <- (IHj h), <- assoc.
+    apply maponpaths, coconeInCommutes.
+  + destruct p.
+    apply coconeInCommutes.
+Qed.
+
+Lemma chain_mor_right {C : precategory} {c : chain C} {i j} (Hij : i < j) (HSij : S i < j) :
+  dmor c (idpath (S i)) ;; chain_mor c HSij = chain_mor c Hij.
+Proof.
```
benediktahrens: There seems to be a redundant hypothesis in this lemma.

mortberg: Yes, but I think it might be more convenient to use the lemma if it looks like this. Stated this way the lemma can be used for any proofs of these inequalities, however if I remove one hypothesis one of the `chain_mor`'s is going be applied to a specific proof created from that hypothesis. Now all of these proofs will be equal, but it might be painful using the lemma if I always have to massage things to look correctly. Or what do you think?

benediktahrens: On 09/07/2016 11:05 PM, Anders Mörtberg wrote:
>> +  chain_mor c Hij ;; coconeIn cc j = coconeIn cc i. +Proof. 
>> +induction j. +- intros Hi0. +  destruct (negnatlthn0 _ Hi0). +-
>> intros Hij; simpl. +  destruct (natlehchoice4 _ _ Hij). +  +
>> rewrite <- (IHj h), <- assoc. +    apply maponpaths,
>> coconeInCommutes. +  + destruct p. +    apply coconeInCommutes. 
>> +Qed. + +Lemma chain_mor_right {C : precategory} {c : chain C} {i
>> j} (Hij : i < j) (HSij : S i < j) : +  dmor c (idpath (S i)) ;;
>> chain_mor c HSij = chain_mor c Hij. +Proof.
> 
> Yes, but I think it might be more convenient to use the lemma if it
> looks like this. Stated this way the lemma can be used for any proofs
> of these inequalities, however if I remove one hypothesis one of the
> `chain_mor`'s is going be applied to a specific proof created from
> that hypothesis. Now all of these proofs will be equal, but it might
> be painful using the lemma if I always have to massage things to look
> correctly. Or what do you think?

Agreed. Then it would be good to explain this in a comment in the code.

mortberg: Yes, will do!

UniMath/CategoryTheory/CocontFunctors.v
```diff
@@ -78,75 +82,514 @@ Require Import UniMath.CategoryTheory.CommaCategories.
...
+Lemma mapcocone_chain_coconeIn {g : graph} {c : diagram g C} {x : C}
+  (cx : cocone c x) (n : vertex g) :
+  coconeIn (mapcocone c cx) n = # F (coconeIn cx n).
+Proof.
+apply idpath.
+Qed.
+
+Definition preserves_colimits {g : graph} (d : diagram g C) (L : C)
+  (cc : cocone d L) : UU :=
+  isColimCocone d L cc -> isColimCocone (mapdiagram d) (F L) (mapcocone d cc).
```
benediktahrens: Actually, this should indeed stay `preserves_colimit` (singular). In this definition, you define what it means for `F` to preserve the given colimiting cocone `cc`.

benediktahrens: What you are thinking of is the definition of `is_cocont`, which could equally well be called `preserves_colimits` (plural), as in "preserves any colimit" (also: of any shape).

mortberg: Yup, that's what made me confused today. I'll change it.

UniMath/CategoryTheory/CocontFunctors.v
mortberg: I think this is the proper reference:

https://eudml.org/doc/16649

I have trouble understanding it though as it is a little old... I found a reference to that paper from theorem 2.1.9 in:

https://www.tu-braunschweig.de/Medien-DB/iti/survey_full.pdf

Also I remember that the actual proof we formalized was written up in some slides by someone...

mortberg: It's not so clear to me what to do. Should I just put a reference like this in the text:

Adámek, Jiří. "Free algebras and automata realizations in the language of categories." Commentationes Mathematicae Universitatis Carolinae 015.4 (1974): 589-602.

Or should I add an item to the bibtex file and make a LaTeX style reference using \cite?

UniMath/CategoryTheory/CocontFunctors.v
mortberg: Done!

UniMath/CategoryTheory/CocontFunctors.v
mortberg: Done!

UniMath/CategoryTheory/CocontFunctors.v
benediktahrens: On 09/08/2016 01:54 AM, Anders Mörtberg wrote:
>> +
>> +End omega_cocont.
>> +
>> +Local Notation "'chain'" := (diagram nat_graph).
>> +
>> +
>> +(** * Adámek's theorem for constructing initial algebras of omega-cocontinuous functors *)
>> +(* This section proves that (L,α : F L -> L) is the initial algebra
>> +    where L is the colimit of the inital chain:
>> +
>> +         !          F !           F^2 !
>> +     0 -----> F 0 ------> F^2 0 --------> F^3 0 ---> ...
>> +
>> +This result is also known as Adámek's theorem:
>> +
>> +  https://ncatlab.org/nlab/show/initial+algebra+of+an+endofunctor#AdameksTheorem
> 
> It's not so clear to me what to do. Should I just put a reference like this in the text:
> 
> Adámek, Jiří. "Free algebras and automata realizations in the language of categories." Commentationes Mathematicae Universitatis Carolinae 015.4 (1974): 589-602.
> 
> Or should I add an item to the bibtex file and make a LaTeX style reference using \cite?

Both would be good: an HTML comment for the HTML documentation and a
\cite thing for the PDF.

benediktahrens:
Build failure due to network error, it seems. Restarting build service.

UniMath/CategoryTheory/CocontFunctors.v
```diff
@@ -78,75 +82,520 @@ Require Import UniMath.CategoryTheory.CommaCategories.
...
+  + generalize Hij; rewrite p; intros H.
+    destruct (isirreflnatlth _ H).
+Qed.
+
+(** Construct the initial chain:
+
+         !          F!            F^2 !
+     0 -----> F 0 ------> F^2 0 --------> F^3 0 ---> ...
+
+*)
```
benediktahrens: That comment is misleading since you don't show that that chain is initial in some category.
Please remove the word "initial".

mortberg: Good point. These ascii "art" comments look very bad when converted to html as well. Maybe I should remove make them not appear in the html documentation?

benediktahrens: On 09/08/2016 05:42 PM, Anders Mörtberg wrote:
>> +  + destruct (natlehchoice4 _ _ h). +    * destruct (natlehchoice4
>> _ _ Hij); [|destruct p, (isirreflnatlth _ h0)]. +      apply
>> cancel_postcomposition, cancel_postcomposition, maponpaths,
>> isasetbool. +    * destruct p; simpl. +      destruct
>> (natlehchoice4 _ _ Hij); [destruct (isirreflnatlth _ h0)|]. +
>> apply cancel_postcomposition, maponpaths, isasetnat. +  +
>> generalize Hij; rewrite p; intros H. +    destruct (isirreflnatlth
>> _ H). +Qed. + +(** Construct the initial chain: + +         !
>> F!            F^2 ! +     0 -----> F 0 ------> F^2 0 --------> F^3
>> 0 ---> ... + +*)
> 
> Good point. These ascii "art" comments look very bad when converted
> to html as well. Maybe I should remove make them not appear in the
> html documentation?

Is there a "verbatim" command?

mortberg: Looks like it, search for verbatim in: 

https://coq.inria.fr/refman/Reference-Manual017.html

I'll try it.

mortberg: This worked beautifully.

UniMath/CategoryTheory/CocontFunctors.v
mortberg: Fixed

mortberg:
I've fixed all of your comments now, so unless there are any more I think it can be merged once travis is done?

benediktahrens:
lgtm

UniMath/CategoryTheory/CocontFunctors.v
vladimirias: The reference should be like the usual biblio reference with the name of the paper, journal (if any) and the year. A link may be also given but in general one can find a link from the name but not the other way around if the link stopped working.

vladimirias: Could you please do both?

mortberg: I've done both now.

benediktahrens:
Can this PR be merged?

PR #452, opened 2016-09-07
resolve #450

### LINKED ISSUE #450: short "using" identifiers
Change one letter "using" identifiers in Ktheory, such as "L", to something longer, such as "_L_".

DanGrayson:
avoid some one letter identifiers


benediktahrens:
lvgtm

vladimirias:
In file Representation.v, line 243 says: [ abstract (apply weqproperty) using _T_. ] . I am curious why would you want to use [ abstract ] here. This _T_ is never referred to below and in any case referring to these things is, as we have postulated, prohibited.

DanGrayson:
I like short names so when I print out the proof term, it looks smaller.

PR #456, opened 2016-09-09
More on cocontinuous functors and (co)limits

mortberg:
This PR contains some new theory about (co)limits and more cleaning:

* Given two isomorphic functor F and G, if F preserves colimits then G also does. This is then used to drop one of the hypotheses for `is_omega_cocont_binproduct_functor`.
* Generalize assumption for the proofs about `binproduct_pair_functor` and `pair_functor` so that they hold for arbitrary diagrams and not just for chains (this makes it possible to prove that these functors are cocontinuous and not just only omega-cocontinuous).  
* A proof that `isaprop Colims` (and the same for both versions of limits) for univalent/saturated categories. 
* More cleaning (more implicit arguments, use tactics like `use`, `mkpair`...)

### DIFF
--- a/UniMath/CategoryTheory/CocontFunctors.v
+++ b/UniMath/CategoryTheory/CocontFunctors.v
@@ -747,50 +747,26 @@ Proof.
 now intros c L ccL M H; apply isColimCocone_pr2_functor.
 Defined.
 
-Local Definition cFAX {gr : graph} (cAB : diagram gr (binproduct_precategory A B))
-  (xy : C × D) (ccxy : cocone (mapdiagram (binproduct_pair_functor F G) cAB) xy) :
-  cocone (mapdiagram F (mapdiagram (pr1_functor A B) cAB)) (pr1 xy).
-Proof.
-use mk_cocone.
-- intro n; apply (pr1 (pr1 ccxy n)).
-- abstract (intros m n e; apply (maponpaths dirprod_pr1 (pr2 ccxy m n e))).
-Defined.
-
-Local Definition cGBY {gr : graph} (cAB : diagram gr (binproduct_precategory A B))
-  (xy : C × D) (ccxy : cocone (mapdiagram (binproduct_pair_functor F G) cAB) xy) :
-  (cocone (mapdiagram G (mapdiagram (pr2_functor A B) cAB)) (pr2 xy)).
-Proof.
-use mk_cocone.
-- intro n; apply (pr2 (pr1 ccxy n)).
-- abstract (intros m n e; apply (maponpaths dirprod_pr2 (pr2 ccxy m n e))).
-Defined.
-
-Lemma is_cocont_binproduct_pair_functor (HF : is_cocont F) (HG : is_cocont G) :
-  is_cocont (binproduct_pair_functor F G).
-Proof.
-intros gr cAB ml ccml Hccml xy ccxy.
-set (cFAX := cFAX cAB xy ccxy); set (cGBY := cGBY cAB xy ccxy).
-destruct (HF _ _ _ _ (isColimCocone_pr1_functor cAB ml ccml Hccml) _ cFAX) as [[f hf1] hf2].
-destruct (HG _ _ _ _ (isColimCocone_pr2_functor cAB ml ccml Hccml) _ cGBY) as [[g hg1] hg2].
-simpl in *.
-mkpair.
-- apply (tpair _ (f,,g)).
-  abstract (intro n; unfold binprodcatmor, compose; simpl;
-            now rewrite hf1, hg1, (paireta (coconeIn ccxy n))).
-- abstract (intro t; apply subtypeEquality; simpl;
-             [ intro x; apply impred; intro; apply isaset_dirprod; [ apply hsC | apply hsD ]
-             | induction t as [[f1 f2] p]; simpl in *; apply pathsdirprod;
-               [ apply (maponpaths pr1 (hf2 (f1,, (λ n, maponpaths pr1 (p n)))))
-               | apply (maponpaths pr1 (hg2 (f2,, (λ n, maponpaths dirprod_pr2 (p n)))))]]).
-Defined.
-
-(** Note that this proof is more less the same as the above one. However we cannot use it as the
-    assumptions are too strong *)
-Lemma is_omega_cocont_binproduct_pair_functor (HF : is_omega_cocont F) (HG : is_omega_cocont G) :
-  is_omega_cocont (binproduct_pair_functor F G).
+Lemma isColimCocone_binproduct_pair_functor {gr : graph}
+  (HF : Π (d : diagram gr A) (c : A) (cc : cocone d c) (h : isColimCocone d c cc),
+        isColimCocone _ _ (mapcocone F d cc))
+  (HG : Π (d : diagram gr B) (c : B) (cc : cocone d c) (h : isColimCocone d c cc),
+        isColimCocone _ _ (mapcocone G d cc)) :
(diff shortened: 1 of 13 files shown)


UniMath/CategoryTheory/CocontFunctors.v
```diff
@@ -463,6 +461,65 @@ Defined.
...
+    [ apply (α L ;; f')
+    | abstract (intro v; rewrite <- Hf', !assoc; apply cancel_postcomposition, nat_trans_ax)]
+  |];
+  apply pathsinv0; generalize (maponpaths pr1 (HHf HH)); simpl; intro Htemp;
+  rewrite <- Htemp, assoc;
+  eapply pathscomp0;
+    [apply cancel_postcomposition,
+           (nat_trans_eq_pointwise (@iso_after_iso_inv [C,D,hsD] _ _ (isopair _ Hα)))|];
+  now apply id_left).
+Defined.
```
benediktahrens: In this proof, would it be possible to break (some of) the large abstract blocks into lemmas of their own?
I'd hope that that would improve maintainability.

mortberg: Yup, that made it a little nicer, see:

https://github.com/UniMath/UniMath/pull/456/files#diff-b70ad25f7a9c285b676b8b89159c63beR468

I also got rid of the `destruct`'s.

mortberg:
I don't understand why the travis build failed?

DanGrayson:
The website for downloading the submodule lablgtk was down.  There should be a way to restart the build.  @benediktahrens ?

benediktahrens:
In order to understand why Travis fails, click on "Details". This will bring you to a website which lists the packages and their build status. Clicking on the failing package gives you a log of the failed build.
Travis stops when there are 10 minutes without output to stdout. In this case, the cloning of some repo failed (was hanging), and no output was produced during 10 min.

In order to restart the build, click on "Details". This will bring you to a website which lists the packages and their build status. Clicking on "Log in with Github" will allow you to log into Travis, provided that you are logged into Github. 
Once you are logged in, you see a button with a rounded arrow at the end of the line with the number of the PR. Put differently, the button is in the up-right corner of the large rectangle of the "Details" page. Hovering over that button will show "Restart Build". Click that button to restart the build.
Note that you need write access to the UniMath repo in order to restart a build.

benediktahrens:
I did not check the logs of the failures. Is it the case that we are building CoqIDE for every package? Otherwise I wouldn't understand why we are compiling lablgtk for every package, which is apparently what was happening (and failing) for a few of the packages.

DanGrayson:
The details on the previous failures are now gone, but there were at least 4 failures due to lablgtk not getting downloaded.  I don't know the syntax of the file `.travis.yml` file.  Here it is:
```
language:       ocaml
sudo: 		required
dist: 		precise
group: 		legacy
before_install: sudo apt-get update -qq
install:        sudo apt-get install -y ocaml ocaml-nox ocaml-native-compilers camlp5 camlp4-extra time libgtk2.0 libgtksourceview2.0
env:
  - PACKAGE=Foundations
  - PACKAGE=CategoryTheory
  - PACKAGE=Ktheory
  - PACKAGE=Topology
  - PACKAGE=RealNumbers
  - PACKAGE=SubstitutionSystems
  - PACKAGE=Tactics
  - PACKAGE=Folds
script:
  - time make BUILD_COQIDE=yes build-coq
  - time make TIMECMD=time $PACKAGE
```

DanGrayson:
If I guess that there is a single script and it gets run once for each setting of the environment, then yes, Coqide is getting built all the time.  Maybe this would fix it:
```
env:
  - PACKAGE=Foundations BUILD_COQIDE=yes
  - PACKAGE=CategoryTheory
  - PACKAGE=Ktheory
  - PACKAGE=Topology
  - PACKAGE=RealNumbers
  - PACKAGE=SubstitutionSystems
  - PACKAGE=Tactics
  - PACKAGE=Folds
script:
  - time make build-coq
  - time make TIMECMD=time $PACKAGE
```

DanGrayson:
Once travis' machine provides Coq 8.5, we can stop building it.  Ubuntu 16.04, which it might be, is at 8.4pl4 for Coq currently.

benediktahrens:
On 09/09/2016 10:00 PM, Daniel R. Grayson wrote:
> If I guess that there is a single script and it gets run once for each setting of the environment, then yes, Coqide is getting built all the time.  Maybe this would fix it:
> ```
> env:
>   - PACKAGE=Foundations BUILD_COQIDE=yes
>   - PACKAGE=CategoryTheory
>   - PACKAGE=Ktheory
>   - PACKAGE=Topology
>   - PACKAGE=RealNumbers
>   - PACKAGE=SubstitutionSystems
>   - PACKAGE=Tactics
>   - PACKAGE=Folds
> script:
>   - time make build-coq
>   - time make TIMECMD=time $PACKAGE
> ```

Yes, this looks very good!

DanGrayson:
I'll submit a pull request to test it.

benediktahrens:
On 09/09/2016 10:05 PM, Daniel R. Grayson wrote:
> Once travis' machine provides Coq 8.5, we can stop building it.
> Ubuntu 16.04, which it might be, is at 8.4pl4 for Coq currently.

The Coq version Ubuntu ships won't be updated, I think.
The only new software that enters an existing Ubuntu version is a new
kernel and xserver stack (they call it HWE, HardWare Enablement).

UniMath/CategoryTheory/CocontFunctors.v
```diff
@@ -463,6 +465,88 @@ Defined.
...
+  apply (nat_trans_eq_pointwise (@iso_after_iso_inv [C,D,hsD] _ _ (isopair _ Hα))).
+now apply id_left.
+Qed.
+
+Lemma preserves_colimit_iso  : preserves_colimit G d L cc.
+Proof.
+intros HccL y ccGy.
+set (H := HF HccL y (ccFy y ccGy)).
+set (f := pr1 (pr1 H)); set (Hf := pr2 (pr1 H)); set (HHf := pr2 H).
+mkpair.
```
benediktahrens: I feel that your proof could simplified even further by using `use unique_exists.` here instead of `mkpair.`.
This will change the subgoals, so your lemmas from before won't apply anymore. But I still suggest you try it, the subgoals should still be solved by some similar proofs as the ones you have above.

benediktahrens: Also, since what you are proving is a proposition, you can weaken the hypothesis from a specified iso to F and G being isomorphic.
We don't need this stronger lemma for now, though.

JasonGross:
I have a ppa with Coq 8.5; see https://github.com/mit-plv/fiat-crypto/blob/master/.travis.yml

DanGrayson:
Thanks, Jason, that will speed things up.

UniMath/CategoryTheory/CocontFunctors.v
mortberg: I used `use unique_exists` which made things a little nicer. I also added a comment about maybe not using a specified isomorphism.

benediktahrens:
```
time "sub/coq/bin/coqc"  -q  -Q "UniMath" UniMath  -indices-matter -type-in-type -w none  UniMath/CategoryTheory/CocontFunctors.v

File "./UniMath/CategoryTheory/CocontFunctors.v", line 474, characters 17-38:

Error:

In environment

C : precategory

D : precategory

hsD : has_homsets D

F : functor C D

G : functor C D

The term "is_iso" has type "[C, D, hsD] ⟦ F, G ⟧ → Type"

which should be Set, Prop or Type.

```

mortberg:
Oops, fixed this now.

benediktahrens:
lgtm

PR #457, opened 2016-09-09
add chimney symbol to latex preamble, requires inclusion of some late…

benediktahrens:
…x package

fixes the `make world` target


mortberg:
LGTM

PR #458, opened 2016-09-09
improve the travis script

DanGrayson:
- update to the latest ubuntu release
- compile coqide only once, with one package
- show the OS version
- add some comments

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,11 @@
-language:       ocaml
+# validate this file at http://lint.travis-ci.org/
+# ocaml and coq are not on the list of languages, and ruby is the default, so:
+language:       ruby
 sudo: 		required
-dist: 		precise
-group: 		legacy
+dist: 		trusty
 before_install: sudo apt-get update -qq
 install:        sudo apt-get install -y ocaml ocaml-nox ocaml-native-compilers camlp5 camlp4-extra time libgtk2.0 libgtksourceview2.0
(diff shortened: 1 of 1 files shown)


DanGrayson:
It seems to have worked.

DanGrayson:
I can improve it further with Jason Gross' coq package, so we don't have to build it, see https://github.com/UniMath/UniMath/pull/456#issuecomment-246141830

benediktahrens:
On 09/11/2016 04:36 AM, Daniel R. Grayson wrote:
> I can improve it further with Jason Gross' coq package, so we don't
> have to build it, see
> https://github.com/UniMath/UniMath/pull/456#issuecomment-246141830
> 

It seems to me that in, e.g.,
https://travis-ci.org/UniMath/UniMath/jobs/158773362 , the time to build
Coq (including CoqIDE) was only 149 - 111 = 38 seconds.
I don't think we can save much time by downloading binary packages.

Also, we should check as much as possible of the functionality of our
Makefile in Travis.


One tiny improvement we could do is compile Coq in the `before_script`
part of the travis build cycle (see [1]), which, I hope, would fold the
huge output of the Coq build and only leave the interesting output of
the Unimath library build unfolded.
I have not tested this, I am just assuming that this would happen.


[1] https://docs.travis-ci.com/user/customizing-the-build

DanGrayson:
Okay, I've updated the script some more...

PR #459, opened 2016-09-11
improve travis

DanGrayson:
building Coq in a separate stage might fold up some of the output in the build log, so let's see


benediktahrens:
Making `TAGS` fails with
```
/bin/sh: 1: etags: not found

make: *** [TAGS] Error 127
```

PR #466, opened 2016-09-21
improve travis test script

### LINKED ISSUE #465: Cloning lablgtk frequently fails on Travis
We clone lablgtk every time, even when we don't compile CoqIDE.
Is there a way to not do that?

DanGrayson:
... so it doesn't check out the submodules in every build

This will resolve #465

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ language:       generic
 sudo: 		required
 # "trusty" is the most modern release of Ubuntu supported
 dist: 		trusty
+git:
+  submodules:   false
 before_install:
   - sudo apt-get update -qq
 install:


benediktahrens:
Perfect, thanks!

DanGrayson:
Step 1 completed and lablgtk was not downloaded.  So it works.

benediktahrens:
On 09/21/2016 04:05 PM, Daniel R. Grayson wrote:
> Step 1 completed and lablgtk was not downloaded.  So it works.

Right. And lablgtk is cloned "at the right moment" here:
https://travis-ci.org/UniMath/UniMath/jobs/161632665#L830

benediktahrens:
Note that the recursive cloning takes a significant amount of time, see, e.g., here:
https://travis-ci.org/UniMath/UniMath/jobs/161527555#L105  - 134seconds. 
Good thing we now only fetch what we need.

PR #414, opened 2016-07-30
Added short exact sequences

tpannila:
This pull request contains formalization of short exact sequences. There are
4 definitions:
- ShortShortExact for  A -> B -> C.
- LeftShortExact and RightShortExact for 0 -> A -> B -> C and A -> B -> C -> 0,
  respectively.
- ShortExact for 0 -> A -> B -> C -> 0

Using factorization property for morphisms of Abelian_precategory, the
following are formalized:
- Construct the canonical ShortExact from ShortShortExact by using image and
  coimage.
- A criteria for constructing ShortShortExact from data similar to the
  canonical ShortExact obtained from ShortShortExact.

All comments are welcome.

Issue #395, opened 2016-07-01
Obligatory use of the existing definitions


vladimirias:
I suggest the following addition to our rules:

1. Formalizing a result or a group of results one must use the existing definitions wherever they are available. For example, if the ring of integers appears in the formulation of a theorem or in a definition of a more complex object the existing the definition of [hz] must be used.

If a different model of [hz] is more convenient for the proof, including intermediate lemmas, such a definition may be introduced and used as long as a proof of the theorem formulated using the existing definition is also produced.

DanGrayson:
In general, this sounds like a good principle, but I have some code for limits in categories based on representable functors in `Ktheory` that duplicates some code in `CategoryTheory` (with a different approach), and I'd like to keep it.

vladimirias:
Which particular definitions are duplicated?

DanGrayson:
at least initial object, terminal object, zero object, coproduct, product, direct sum, kernel, cokernel, binary products, binary coproducts, limits, colimits, cone, pushout, pullback.

DanGrayson:
My main motivation was to reduce duplication of proofs.  In this case, uniqueness of universal constructions up to isomorphism is proved just once, as uniqueness of the object representing a functor.

DanGrayson:
I'd be happy to label my development as experimental somehow, so others don't use it.  Meanwhile, if @tpannila has the energy to formalize exact categories (in the sense of Quillen) after he does abelian categories, that would save me a bunch of trouble, and I could base my K-theory on that.

tpannila:
@DanGrayson , I have some problems in formalizing abelian categories. If
these problems are difficult to solve, as I expect, then it takes a lot of time to
formalize abelian categories. I don't know anything about exact categories,
but I can take a look at them after we have formalized abelian categories.

benediktahrens:
On 07/03/2016 02:07 PM, Daniel R. Grayson wrote:
> My main motivation was to reduce duplication of proofs.  In this
> case, uniqueness of universal constructions up to isomorphism is
> proved just once, as uniqueness of the object representing a
> functor.

It is one of the fundamental principles of category theory that every
concept can be expressed in multiple different ways.
The different concepts might be useful in different situations.

Also, it might turn out that some concepts work under fewer assumptions
than others. For instance, the homs-are-sets assumption is not necessary
for direct formalizations of specific limits, but necessary for working
with (co)cones.
Hence, things that are equivalent classically can turn out to be not
[...]

Issue #362, opened 2016-06-12
Re-integrate the homs-are-sets assumption in the definition of precategory; make inverses compute better


benediktahrens:
I suggest the following changes to the CategoryTheory library:
1. The homs-are-sets assumption should be re-integrated into the definition of precategories.
2. The definition of isomorphism should be the "normal" one.

The first point should be rather trivial to solve. The second one requires more coding, but not too much either, since in practice one often resorts to working with the normal definition anyway.

The current definitions have several inconveniences:
1. Extra arguments of type `has_homsets ...` have to be given to functions, making the code more difficult to read
2. We lose convertibility when computing the inverse of an isomorphism, see, e.g., #279 . Another example is the fact that 'taking inverse' and 'taking pointwise iso' of a natural isomorphism do not commute up to convertibility.

The current definitions have few advantages:
1. We do not, to my knowledge, have an example of a precategory that does not have hom-sets.
2. We do not have a lot of interesting constructions that do not require the homs-are-sets condition.

DanGrayson:
I'd be in favor of this.

Here's an example of a precategory without hom-sets: the category of small types, where the arrows
are the functions.

JasonGross:
I'm curious why 2 is related to the separation of the homset argument.  What can you do with it that you can't do without?

benediktahrens:
On 06/12/2016 05:46 PM, Jason Gross wrote:
> I'm curious why 2 is related to the separation of the homset
> argument.  What can you do with it that you can't do without?

The usual definition of isomorphism, called "is_z_iso" in UniMath, is
the right one only for Precategories (i.e., for precategories whose
hom-types are sets):
only for Precategories is "is_z_iso(f)" a proposition.

The current definition of "is_iso(f)" is a proposition also for
precategories, but it lacks some definitional equalities that one has
for z_isos.

vladimirias:
Maybe this will work:

1. precategory - no assumptions on the h-levels of objets or morphisms. Examples: types and functions, a type and paths in it, precategory of precategories, categories etc.

2. category - a precategory with morphism types being sets.

3. setcategory - a category with the object type being a set.

4. univ_precategory - a precategory such that the functions from paths to isomorphisms (as currently defined) are equivalences.

5. univ_category - same with morphism types being sets.

To this we add coercions 5 -> 4 -> 1, 5 -> 2 , 3 -> 2 -> 1 and in theorems use concepts with minimal requirements.

Vladimir.

[...]

Issue #176, opened 2015-09-30
identifier names, a proposal going forward


DanGrayson:
It would be nice to have a consistent style in future for the naming of exported identifiers.  Here's one, which I like:
- identifiers should be obtained by concatenating English words, capitalizing the word to 
  the right of each border between two words, without underscores;
- one may also use previously defined identifiers as components of a name, with an underscore
  as the border to the left of an identifier is not capitalized;
- the first letter must be capitalized if the identifier refers to a type;
- the first letter may be capitalized if the identifier refers to a function returning a type;
- otherwise, elements should start with lower case;
- a function that returns a proposition may begin with "has" or "is";
- If "Foo" is the name of a type, then "hasFoo" is its propositional truncation.

cathlelay:
If I understand,

   - hnnr must be renamed as NonNegativeReals.
   - hnnq become NonNegativeRationals.
   - Dcuts (the future type for Dedekind cuts) become DedekindCuts.

Is it correct ?

2015-09-30 13:03 GMT-04:00 Daniel R. Grayson <notifications@github.com>:

DanGrayson:
Yes.  Or hnnr could be PositiveReals, using the French convention for "positive".  That would be
much prettier.

Long names might be a bit of a hassle for users, but then notations can be introduced, and word completion in emacs with M-/ works well, too.

vladimirias:
No, positive means >0 and non-negative means >=0 whether it is pretty or not.

DanGrayson:
Point taken.

vladimirias:
I am not sure what Dan calls “exported” identifiers but renaming hz into Integers and hnat in NaturalNumbers and using such long names all over the place does not seem to be a good idea.

DanGrayson:
This proposal is for the future names.

DanGrayson:
It can't be NonNegativeReals, because Non is not a word.  So it must be NonnegativeReals.

vladimirias:
I guess one can use these long names and then immediately introduce a short notation for it and then use this short notation.

cathlelay:
And for functions ?

   - leNonnegativeRationals to replace hnnq_le
   - plusNonnegativeRationals to replace hnnq_plus

Is it correct ?

2015-09-30 14:35 GMT-04:00 Vladimir Voevodsky <notifications@github.com>:

DanGrayson:
A nonexported symbol is one not visible to those who import your file.  The way to make one
[...]

PR #467, opened 2016-09-23
upstream Precategory

benediktahrens:
- upstream Precategory (precategory whose homs are sets) to CT/precategories
- set implicit arguments for cat axioms in CT/precategories

### DIFF
--- a/UniMath/CategoryTheory/functor_categories.v
+++ b/UniMath/CategoryTheory/functor_categories.v
@@ -1193,11 +1193,20 @@ Proof.
   apply hs.
 Qed.
 
+
+Definition functorPrecategory (C : precategory) (D : Precategory)
+  : Precategory.
+Proof.
+  exists (functor_precategory C D (homset_property D)).
+  apply functor_category_has_homsets.
+Defined.
+
 Lemma functor_identity_left (C D : precategory) (F : functor C D) :
   functor_composite (functor_identity C) F = F.
 Proof.
-  destruct F as [ [ Fob Fmor ] is ] . destruct is as [ idax compax ] . apply idpath .
-
+  destruct F as [ [ Fob Fmor ] is ].
+  destruct is as [ idax compax ] .
+  apply idpath .
 Defined.
 
 
@@ -1278,7 +1287,7 @@ Defined.
 
 Definition iter_functor {C : precategory} (F : functor C C) (n : nat) : functor C C.
 Proof.
-induction n.
-apply functor_identity.
-apply (functor_composite IHn F).
+  induction n.
+  apply functor_identity.
+  apply (functor_composite IHn F).
 Defined.
--- a/UniMath/CategoryTheory/functor_categories.v
+++ b/UniMath/CategoryTheory/functor_categories.v
@@ -1287,7 +1287,7 @@ Defined.
 
 Definition iter_functor {C : precategory} (F : functor C C) (n : nat) : functor C C.
 Proof.
-  induction n.
-  apply functor_identity.
-  apply (functor_composite IHn F).
+  induction n as [ | n IHn].
+  - apply functor_identity.
+  - apply (functor_composite IHn F).
 Defined.
--- a/UniMath/CategoryTheory/limits/graphs/limits.v
+++ b/UniMath/CategoryTheory/limits/graphs/limits.v
@@ -243,7 +243,7 @@ simple refine (tpair _ _ _).
   + simpl; intro u.
     assert (XR:=limArrowCommutes CC x cx u).
     eapply pathscomp0; [| apply XR].
-    eapply pathscomp0; [ apply (!assoc _ _ _ _ _ _ _ _ ) |].
+    eapply pathscomp0; [ apply (!assoc _ _ _ ) |].
     apply maponpaths.
     apply z_iso_inv_on_right.
     apply pathsinv0, limArrowCommutes.
@@ -254,7 +254,7 @@ simple refine (tpair _ _ _).
     apply pathsinv0, limArrowUnique; intro u.
     cbn in *.
     eapply pathscomp0; [| apply Hf].
-    eapply pathscomp0. apply (!assoc _ _ _ _ _ _ _ _ ).
+    eapply pathscomp0. apply (!assoc _ _ _ ).
     apply maponpaths.
     apply limArrowCommutes.
 Defined.
@@ -787,7 +787,7 @@ simple refine (tpair _ _ _).
   + simpl; intro u.
     assert (XR:=limArrowCommutes CC x cx u).
     eapply pathscomp0; [| apply XR].
-    eapply pathscomp0; [ apply (!assoc _ _ _ _ _ _ _ _ ) |].
+    eapply pathscomp0; [ apply (!assoc _ _ _ ) |].
     apply maponpaths.
     apply z_iso_inv_on_right.
     apply pathsinv0, limArrowCommutes.
@@ -798,7 +798,7 @@ simple refine (tpair _ _ _).
     apply pathsinv0, limArrowUnique; intro u.
     cbn in *.
     eapply pathscomp0; [| apply Hf].
-    eapply pathscomp0. apply (!assoc _ _ _ _ _ _ _ _ ).
+    eapply pathscomp0. apply (!assoc _ _ _ ).
     apply maponpaths.
     apply limArrowCommutes.
 Defined.
--- a/UniMath/CategoryTheory/Abelian.v
+++ b/UniMath/CategoryTheory/Abelian.v
@@ -562,7 +562,7 @@ Section abelian_equalizers.
     rewrite <- assoc in H. rewrite <- assoc in H.
     set (com1 := BinProductPr1Commutes A _ _ BinProd x (identity x) f).
     rewrite com1 in H.
-    rewrite (id_right A _ _ h1) in H. rewrite (id_right A _ _ h2) in H.
+    rewrite (id_right h1) in H. rewrite (id_right h2) in H.
     exact H.
   Qed.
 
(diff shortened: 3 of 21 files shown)


mortberg:
This looks very nice to me and I suggest merging it as soon as possible to avoid conflicts with other PRs.

There are many places where we have a precategory and the assumption that it has homsets, how about changing all of these to Precategory instead? A tags search for `has_homsets` should find these places. 

There are some more functions that probably could have more implicit arguments, try `grep "_ _ _" *v` in the CategoryTheory subfolder. This also finds some proofs that would look nicer with `mkpair` and `use`.

UniMath/CategoryTheory/limits/graphs/limits.v
```diff
@@ -243,7 +243,7 @@ simple refine (tpair _ _ _).
   + simpl; intro u.
     assert (XR:=limArrowCommutes CC x cx u).
     eapply pathscomp0; [| apply XR].
-    eapply pathscomp0; [ apply (!assoc _ _ _ _ _ _ _ _ ) |].
+    eapply pathscomp0; [ apply (!assoc _ _ _ ) |].
```
mortberg: What does `apply` with a ! do?

benediktahrens:
On 09/23/2016 05:02 PM, Anders Mörtberg wrote:
> This looks very nice to me and I suggest merging it as soon as
> possible to avoid conflicts with other PRs.
> 
> There are many places where we have a precategory and the assumption
> that it has homsets, how about changing all of these to Precategory
> instead? A tags search for `has_homsets` should find these places.

I don't think we need to change code that is working.
It is sufficient to write better code from now on. The old code will be
improved eventually, gradually.

> There are some more functions that probably could have more implicit
> arguments, try `grep "_ _ _" *v` in the CategoryTheory subfolder.
> This also finds some proofs that would look nicer with `mkpair` and
> `use`.

See above.

benediktahrens:
On 09/23/2016 05:04 PM, Anders Mörtberg wrote:
> 
> 
> 
>> @@ -243,7 +243,7 @@ simple refine (tpair _ _ _).
>    + simpl; intro u.
>      assert (XR:=limArrowCommutes CC x cx u).
>      eapply pathscomp0; [| apply XR].
> -    eapply pathscomp0; [ apply (!assoc _ _ _ _ _ _ _ _ ) |].
> +    eapply pathscomp0; [ apply (!assoc _ _ _ ) |].
> 
> What does `apply` with a ! do? 

! is path inversion.

It would be great if the underscores were not needed, but that would
require a tactic similar to `use`, which tries consecutively
`apply (! H)`, `apply (! H _ )` and so on.

benediktahrens:
On 09/23/2016 05:02 PM, Anders Mörtberg wrote:
> This looks very nice to me and I suggest merging it as soon as
> possible to avoid conflicts with other PRs.

Note that this PR might be in conflict with older PRs as well.

UniMath/CategoryTheory/functor_categories.v
```diff
@@ -1278,7 +1287,7 @@ Defined.
 
 Definition iter_functor {C : precategory} (F : functor C C) (n : nat) : functor C C.
 Proof.
-induction n.
-apply functor_identity.
-apply (functor_composite IHn F).
+  induction n.
+  apply functor_identity.
+  apply (functor_composite IHn F).
```
mortberg: I wrote the original code and it is pretty bad, sorry about that. The definition relies on automatic naming for `IHn` and it doesn't use any bullets...

DanGrayson:
Re: `apply (!assoc _ _ _ _ _ _ _ _ )`

Wouldn't something like `apply pathsinv0, assoc` work?

benediktahrens:
On 09/23/2016 05:13 PM, Daniel R. Grayson wrote:
> Re: `apply (!assoc _ _ _ _ _ _ _ _ )`
> 
> Wouldn't something like `apply pathsinv0, assoc` work?

That seems likely, yes.

mortberg:
Using ! for path inversion is maybe not so good as ! is used in rewrite to mean rewrite as many times as possible. I thought that was what was going on here...

benediktahrens:
On 09/23/2016 05:12 PM, Anders Mörtberg wrote:
> 
>> @@ -1278,7 +1287,7 @@ Defined.
>  
>  Definition iter_functor {C : precategory} (F : functor C C) (n : nat) : functor C C.
>  Proof.
> -induction n.
> -apply functor_identity.
> -apply (functor_composite IHn F).
> +  induction n.
> +  apply functor_identity.
> +  apply (functor_composite IHn F).
> 
> I wrote the original code and it is pretty bad, sorry about that. The definition relies on automatic naming for `IHn` and it doesn't use any bullets...

Fixed in latest commit to this PR.

DanGrayson:
Re: "Using ! for path inversion is maybe not so good as ! is used in rewrite to mean rewrite as many times as possible. I thought that was what was going on here..."

One is inside the parens and one is outside.

benediktahrens:
On 09/23/2016 05:16 PM, Anders Mörtberg wrote:
> Using ! for path inversion is maybe not so good as ! is used in
> rewrite to mean rewrite as many times as possible. I thought that was
> what was going on here...

I did not know about the use of ! in combination with `rewrite`.

Anyway, we have not had any problems with that notation so far, probably
because they occur in different places.

I suggest we don't change it unless we experience a problem.

mortberg:
Ok, it's just a little confusing to have the same symbol meaning different things. I wonder if one can do `rewrite !(!assoc _ _ _)` which might be equivalent to `rewrite <- !assoc`...

Anyway now that the confusion has been cleared up the PR looks good to me. I agree that we can do gradual improvement of the code instead of changing everything in one go.

benediktahrens:
On 09/23/2016 05:51 PM, Anders Mörtberg wrote:
> Ok, it's just a little confusing to have the same symbol meaning
> different things. I wonder if one can do `rewrite !(!assoc _ _ _)`
> which might be equivalent to `rewrite <- !assoc`...

The point, I think, is that one should not use stuff like `repeat
rewrite foo` and `rewrite ! foo`.

Usage of such tacticals makes proof code fragile, since when the number
of successful rewrites changes, it does so invisibly, and may produce
errors that occur only later.

Now, I am not innocent concerning those tacticals: incidentially, that
came up in this PR, where I had to change a `repeat rewrite foo` into
`repeat rewrite @foo`. The proof only failed about 10 lines later.

mortberg:
Interesting. It's really a pity that Coq's rewrite fails so often... I never had any problem with this when I was using SSReflect's rewrite, hopefully one day they will change the Coq matching algorithm and we don't have to bother about this kind of things any more.

benediktahrens:
Turns out this PR will indeed conflict with #414 . I suggest we first merge that one.

mortberg:
LGTM if travis is happy

DanGrayson:
LGTM.  I like these changes, but we should address @vladimirias's comment: https://github.com/UniMath/UniMath/issues/395#issuecomment-230326541 , saying that the difference caused by capitalization ( `Precategory` vs. `precategory` ) should be consistent.

I also am still fond of my proposal #176, under which `Precategory` would be a type (or a function returning a type) and `precategory` would be a function returning an element of a type.  To convince yourself that capitalization is a good way to pass from elements to types, consider such common usages as `c:C` for an object of a category, `n:ℕ` for a natural number `n`, or `x:X` for an element `x` of a set `X`.

What about renaming `precategory` to something else that is capitalized?  It could be long, because we haven't yet figured out whether the notion is useful for anything.  It could be `Prehomcategory` or `Nonsetcategory` or `Homtypecategory` or something similar.  If the notion becomes useful in the future, a better name will occur to us from the application, and the name can be changed trivially.

mortberg:
You're right, the names are not very good now. One possibility would be to do something along the lines of what Vladimir suggested in:

https://github.com/UniMath/UniMath/issues/362#issuecomment-225556634

I would suggest the following names: Precategory, Category, SetCategory, UnivPrecategory (or UnivalentPrecategory) and UnivCategory (or UnivalentCategory).

DanGrayson:
What would all those things mean?

People learned with "scheme" and "prescheme" that it's easier to add properties than to subtract them, so we should consider just eliminating "precategory" from the language.

I like taking `Category` to be the type of things with homsets but without univalence, and `UnivalentCategory` to be the things with homsets and univalence, because these seem to be the important notions.  I would have no use for `SetCategory`.

Does anyone dislike `HomtypeCategory` for the things without homsets?  We won't use it much and it can always be changed.  It broadcasts its meaning clearly.

mortberg:
I thought all of those notions were already formalized somewhere, but I guess I was wrong and I don't think we should add some new notions that we never use. Anyway I agree that it might be a good idea to rename `precategory` if we are going to have `Category` and `UnivalentCategory`. I think `HomtypeCategory` is a good name and like you say it won't appear in many places (given that we do the change that I suggested in my first message about changing every place where we now have a `precategory` and an assumption that it `has_homsets`).

benediktahrens:
On 09/23/2016 07:36 PM, Daniel R. Grayson wrote:
> LGTM.  I like these changes, but we should address @vladimirias's
> comment

Sounds good with me, but I suggest we do that in a new issue/PR.

Also, if you guys are planning to do a mass renaming, I ask you to first
merge the outstanding PRs.

vladimirias:
I agree with last comment from Benedikt. "Precategory" and "precategory" are Ok for now. They serve the purpose of introducing a name for precategories that have hom_sets. The rest we should open as an issue and discuss at leisure. 

BTW Do we have a pregroupoid such that "pregroupoid_data" = "setoid" ?

mortberg:
I agree with Benedikt as well. It's nice to keep PRs small and have massive renamings in separate PRs.

So let's merge this PR and make an issue about the names and change them once the outstanding PRs have been merged?

benediktahrens:
On 09/23/2016 09:40 PM, Vladimir Voevodsky wrote:
> 
> BTW Do we have a pregroupoid such that "pregroupoid_data" = "setoid"
> ?

Dan has a notion of groupoid that is the saturated version of what you
are seeking, I suppose:
```
Definition is_groupoid (C : Precategory) :=
  Π a b : ob C, isweq (fun p : a = b => idtomor a b p).
```
(see Ktheory/StandardCategories)

I don't think we have any more general notion of "pre"groupoid.

vladimirias:
In terms of pronunciation UniCategory is better than UnivCategory and same with other such cases.

The main reason for having precategories is to define a (2,0)-precategory as a precategory with a structure, namely the associativity pentagon. This is all  the structure one needs.

For the (2,1)-precategory (aka bicategory) one needs a (2,0)-precategory together with precategory structures on the hom-types (btw, probably it is better to use “mor” not “hom” in the context of general category theory). This is a little less straightforward and here the approach by @mvr has its merit. However, I am still uncomfortable with the fact that in his approach a bicategory is not a precategory.

vladimirias:
Dan’s definition has extra assumptions. It is sufficient to assume C to be precategory_data (in fact only identities are needed, not compositions but I think we don’t have a name for such a structure).

Can someone update the definition?

PR #470, opened 2016-09-25
Explicit induction variables

DanGrayson:
I located all the uses of "induction" without explicit names that
involve names starting with "IH" and made the names explicit.


benediktahrens:
lgtm

vladimirias:
LGTM and merging.

PR #473, opened 2016-09-26
Added localizing classes

tpannila:
Hi all,

this pull request is not ready to be merged. I'm confused and I need to ask
some questions. This pull request contains
- Formalization of localizing class
- Construction of the localization of a category using roofs when the
  category of morphisms one wants to invert is a localizing class. (Note : I
  have not formalized the universal property, yet. )

The thing why I'm confused is that I have not assumed the collection of objects
in the category to be a set, like Carlos Simpson did in "Explaining 
Gabriel-Zisman localization to the computer",
https://arxiv.org/abs/math/0506471 , see 3.1 page 7. In ZFC one
needs to use Grothendieck universes to show that the localization exists, in
general? See for example Krause's paper https://arxiv.org/abs/0806.1324 ,
Example 4.15.

Have I abused universes? Can someone derive a contradiction from the
construction? Or is this a feature of Univalent foundations where being a set
(or any other h-level) is a local property compared to ZFC where being a set
is something else (an element of a universe of sets) ? I don't know anything
about universes in type theory.

All comments are welcome.

### DIFF
--- a/UniMath/CategoryTheory/.package/files
+++ b/UniMath/CategoryTheory/.package/files
@@ -76,3 +76,4 @@ AbelianToAdditive.v
 ShortExactSequences.v
 Morphisms.v
 AdditiveFunctors.v
+LocalizingClass.v
--- a/UniMath/CategoryTheory/LocalizingClass.v
+++ b/UniMath/CategoryTheory/LocalizingClass.v
@@ -1698,4 +1698,100 @@ Section def_roofs.
   Defined.
   Opaque LocalizationUniversalProperty.
 
+  (** The functor FunctorToLocalization maps morphisms in SOM to isomorphisms *)
+  Definition FunctorToLocalization_is_iso :
+    Π (x y : C) (f : x --> y) (s : SOM x y f), is_iso (# FunctorToLocalization f).
+  Proof.
+    intros x y f s. cbn.
+    use (@is_iso_qinv loc_precategory).
+    - exact (RoofEqclassFromRoof (InvRoofFromMorInSom f s)).
+    - split.
+      + set (sqr := isLocClassSqr2 iLC (InvRoofFromMorInSom f s) (MorToRoof f) y
+                                   (RoofMor1 (InvRoofFromMorInSom f s))
+                                   (RoofMor1Is (InvRoofFromMorInSom f s))
+                                   (RoofMor2 (MorToRoof f))).
+        set (PostS := isLocClassPost iLC sqr x y (LocSqr2Mor2 sqr) (LocSqr2Mor1 sqr) f s
+                                     (! (LocSqr2Comm sqr))).
+        rewrite RoofEqclassCompToRoofComp.
+        use RoofEqClassEq1.
+        intros P X. apply X. clear X P.
+        unfold roof_comp. fold sqr.
+        use mk_RoofTop.
+        * exact PostS.
+        * exact (PostSwitchMor PostS).
+        * exact (PostSwitchMor PostS ;; LocSqr2Mor2 sqr).
+        * use (isLocClassComp iLC).
+          -- exact (PostSwitchMorIs PostS).
+          -- use (isLocClassComp iLC).
+             ++ exact (LocSqr2Mor2Is sqr).
+             ++ exact (isLocClassIs iLC x).
+        * cbn. rewrite assoc. apply idpath.
+        * cbn. rewrite assoc. rewrite id_right. rewrite id_right.
+          exact (! (PostSwitchEq PostS)).
+      + set (sqr := isLocClassSqr2 iLC (MorToRoof f) (InvRoofFromMorInSom f s) x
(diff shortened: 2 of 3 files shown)


benediktahrens:
On 09/26/2016 06:48 PM, Tomi Pannila wrote:
> Hi all,
[...]
> Can someone derive a contradiction from the
> construction? 

It should be quite easy to derive a contradiction in UniMath, with or
without using your construction.

In univalent foundations, being a set means something different,
orthogonal, to what it means in set theory.
The question of something is a set or a proper class, that is asked in
set theory, corresponds, in our setting, to the question of whether a
construction (of thing B from thing A) raises the universe level.
But since in UniMath universes are effectively disabled, UniMath is not
a good tool to investigate this kind of question - we gloss over "size
issues".

vladimirias:
Do not worry about the problem that you mention. As Benedikt explained, the meaning of “a set” in the univalent foundations is about the h-level, not about the size. We have a hierarchy of universes and there are sets and not sets in all of these universes.

However, I am very interested in whether the roof construction actually gives you localization, that is, in whether the universal property holds. Please try to prove it!

vladimirias:
> 
> It should be quite easy to derive a contradiction in UniMath, with or
> without using your construction.

What Benedikt means, I guess, is that it is easy to derive contradiction because we switched off the universe consistency checking and must do the universe discipline “manually”.

In your case all should be Ok with the universes. If C is a (pre)category such that Ob C is in the given universe UU and for all X,Y, Mor X Y is also in UU then, when you do the roofs localization, the morphisms in the localized category are defined as the quotient of something that is in UU. Assuming the resizing rules that one needs for the quotients to remain in the same universe morphisms in the localization will again be in the UU.

This shows that roofs construction should not increase the universe level.

It is consistent with Carlos Simpson’s result since assuming that Ob C : UU is equivalent, in the UniMath, to assuming in the ZFC that Ob C is a set as opposed to a proper class.

BTW For the category of SET or the category of abelian groups the judgement Ob C : UU does not hold, instead Ob C : UU+1.  This would be a problem if we had a special type of “large categories” where Ob C : UU+1 but Mor X Y : UU. For such an object its localization will not be an object of the same type, instead Mor X Y in the localization will be in UU+1.

tpannila:
Thanks for the explanation Benedikt and Vladimir.

I managed to prove the universal property in case where the objects in 
the target category D form a set. See line 1597 of LocalizingClass.v 
(I had no time to clean the file, sorry). More precisely, suppose C is a 
precategory, `loc_precategory` the roof stuff, D a precategory with objects
isaset, and F : C -> D a functor which sends every morphism in S to an 
isomorphism in D. Then there exists a unique functor 
H : loc_precategory -> D such that functor_comp i H = F, where
i : C -> loc_precategory is the natural inclusion functor.

The only place where I need the fact that objects of D form a set is where I 
need to show `isaprop(F1 = F2)`, with F1 and F2 functors. Thus, uniqueness 
and commutativity of the functor H might hold in case where the objects of D 
do not form a set, but uniqueness of commutativity not(?). I try to formalize this 
version tomorrow.

tpannila:
I think I formalized the universal property.

Results LocalizationUniversalFunctor, line 1605, LocalizationUniversalFunctorComm,
line 1618, and LocalizationUniversalFunctorUnique, line 1643, construct the functor
which makes the universal property diagram to commute and show that it is unique. 
Result LocalizationUniversalCategory, line 1756, shows that any category with the 
previous results is isomorphic to the precategory loc_precategory (the roof stuff). 
Thus, loc_precategory is localization of the category C with respect to the collection 
of subsets of morphisms SOM.

Does these claims match the code?

tpannila:
What is the status of this result/pull request? Is there something I can do to help to review it?

vladimirias:
Sorry for the delay. I want to understand how your “localizing class” is related to the “left localizing” and “right localizing” (I think it is the names) considered by Gabriel and Zisman in their book. I did not have time to do it yet. Can you check yourself? I do not remember this stuff very well it was very long ago that I read it.

tpannila:
This pull request follows Gelfand and Manin Methods of Homological Algebra,
second edition, pages 148 (7. Remarks) to 151 (End of the proof), and adds
the missing details. The definition of localizing class in this pull request should 
be equal to Definition 6, page 147.

The universal property proved in this pull request is the universal property of
localization stated in page 6, 1.1, [1]. Left and right fractions are defined
in page 12, 2.3, [1]. The localizing class in this pull request is both left and
right fractions at the same time. In particular, quasi-isomorphisms in K(A), 
the naive homotopy category over an abelian category A, is both left and 
right fractions. Thus, the results in this pull request should be enough to 
define derived categories (this is what I want to do).

[1] Gabriel and Zisman, Calculus of Fractions and Homotopy Theory

vladimirias:
LGTM I suggest we merge as many PR's as we can in the next few days. It will be easier to check things after the PRs are accepted.

PR #474, opened 2016-09-26
Cleaning of translation from binding signatures to monads

mortberg:
This PR contains a lot of cleaning and documentation for the code that goes into the translation from binding signatures to monads. The main changes apart from cleaning/renaming and documentation are in BindingSigToMonad.v where:

- The translation has been decomposed into a few simpler functions. 
- The translation has been generalized from HSET to a category with suitable structure. At the moment some of the assumptions on this category are redundant (for instance the category is required to have both (binary) (co)products and all (co)limits). A future PR will probably reduce the number of assumptions, but for now I want to keep it like it is as it allows for some flexibility when instantiating the translation (one can use direct implementation of (co)products if one wants).


mortberg:
What do you think about moving the definition of lists over a type A as iterated products of elements in A from CategoryTheory to Foundations? This definition isn't depending on anything in CategoryTheory and there are probably going to be other non-CT developments that might need it (e.g. polynomials, matrices...). If you think this is a good idea, where should I put it?

UniMath/SubstitutionSystems/BindingSigToMonad.v
```diff
@@ -0,0 +1,311 @@
...
+use (Monad_from_hss _ hsC BCC).
+- apply (BindingSigToSignature sig CC).
+- apply SignatureToHSS.
+  apply SignatureInitialAlgebra.
+  apply (is_omega_cocont_BindingSigToSignature _ _ PC H).
+Defined.
+
+End BindingSigToMonad.
+
+(** Specialized versions of some of the above functions for HSET *)
```
benediktahrens: This file is very important. Could you change some of the HTML comments so that they give section titles?

UniMath/SubstitutionSystems/LamFromBindingSig.v
```diff
@@ -234,4 +218,4 @@ eapply pathscomp0; eapply maponpaths.
 apply idpath.
 Defined.
```
benediktahrens: We will need to expand on this example: 
- write down the computation rules for `foldr_map` and prove them (pointwise)
- define an operation of type `LC(X+1) * LC(X) -> LC(X)` that is a special case of the monadic substitution. It replaces the distinguished variable in the first term by the second term. This operation can be generically defined for any monad on a category with binary coproduct and terminal object, and I'll write the specification down for you. The code to write herefor is hence very generic. It also involves one way of passing between the two different, but equivalent, definitions of monad.

vladimirias:
Looks like a good idea to me. It should go to Combinatorics, probably in a separate file called Lists.

UniMath/SubstitutionSystems/LamFromBindingSig.v
mortberg: Ok, I'll do this in a future PR.

UniMath/SubstitutionSystems/BindingSigToMonad.v
mortberg: done

mortberg:
I've moved to code on lists to a separate file now.

benediktahrens:
Merging to not stall Anders - comments can still be addressed in later PR

PR #475, opened 2016-09-27
Substitution operation for the lambda calculus

mortberg:
This PR adds:

- A "bind" operation for monads (as in Haskell)
- A substitution operation for monads and from this a substitution operation for the lambda calculus

### DIFF
--- a/UniMath/CategoryTheory/Monads.v
+++ b/UniMath/CategoryTheory/Monads.v
@@ -193,10 +193,10 @@ Definition precategory_Monad (C : precategory) (hs : has_homsets C) : precategor
 
 End Monad_precategory.
 
-(** * Theory about monads *)
-Section monad_theory.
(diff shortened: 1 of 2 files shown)


UniMath/CategoryTheory/Monads.v
```diff
@@ -73,6 +72,25 @@ Definition Monad (C : precategory) : UU := Σ T : Monad_data C, Monad_laws T.
 
 Coercion Monad_data_from_Monad (C : precategory) (T : Monad C) : Monad_data C := pr1 T.
 
+Lemma Monad_law1 {C : precategory} {T : Monad C} : Π c : C, η T (T c) ;; μ T c = identity (T c).
```
mortberg: These names are not the best. Any suggestions for better ones?

benediktahrens:
lgtm - we'll define point-full versions of the computation rules later.

PR #476, opened 2016-10-02
Maximum, minimum, and addition.. and Lattice

cathlelay:
New theorems about
- minimum in non-negative rationals
- minimum and maximum in non-negative reals

Better proofs for addition of non-negative reals using Dcuts_max_val and its properties


UniMath/RealNumbers/NonnegativeReals.v
```diff
@@ -722,233 +1162,135 @@ Section Dcuts_plus.
...
+  generalize (X_open _ (pr1 (pr2 (pr2 xy)))) (Y_open _ (pr2 (pr2 (pr2 xy)))).
+  apply hinhfun2.
+  intros nx ny.
+  exists (pr1 nx + pr1 ny).
+  split.
+  - apply hinhpr ; exists (pr1 nx ,, pr1 ny).
+    repeat split.
+    * exact (pr1 (pr2 nx)).
+    * exact (pr1 (pr2 ny)).
+  - pattern r at 1 ;
```
mortberg: Would it be possible to remove the uses of `pattern` with an index? They make the proof scripts fragile as they might break if Coq changes, as shown by this: https://github.com/UniMath/UniMath/pull/445

UniMath/RealNumbers/NonnegativeReals.v
```diff
@@ -1814,91 +2158,80 @@ Qed.
...
+   = Dcuts_plus_val (pr1 x) (Dcuts_plus_val (pr1 y) (pr1 z))).
+  rewrite !Dcuts_plus_max_val, !isassoc_Dcuts_max_val.
+  apply maponpaths, maponpaths.
+  rewrite iscomm_Dcuts_max_val, !isassoc_Dcuts_max_val.
+  apply maponpaths.
+  apply funextfun ; intros r.
+  apply hPropUnivalence ; apply hinhuniv, sumofmaps.
+  - apply hinhuniv.
+    intros xy.
+    generalize (pr1 (pr1 xy)) (pr2 (pr1 xy)) (pr1 (pr2 xy)) (pr1 (pr2 (pr2 xy))) (pr2 (pr2 (pr2 xy))) ;
```
mortberg: I don't know what `xy` is, but maybe having some access functions instead of just `pr1` and `pr2` would make this more readable?

mortberg: Another point of having access functions is that it makes code easier to maintain, if one reorganizes some definition only the access functions have to be changed. If there are no access functions one instead has to change every place where one uses combinations of `pr1` and `pr2`.

UniMath/RealNumbers/NonnegativeReals.v
```diff
@@ -1814,91 +2158,80 @@ Qed.
...
+        apply hinhpr, ii2, hinhpr.
+        exists (rx ,, ry + rz).
+        repeat split.
+        apply isassoc_plusNonnegativeRationals.
+        exact Xr.
+        apply hinhpr, ii2, hinhpr, ii2, hinhpr.
+        now exists (ry,,rz).
+  - apply hinhuniv ; intros xy.
+    generalize (pr1 (pr1 xy)) (pr2 (pr1 xy)) (pr1 (pr2 xy)) (pr1 (pr2 (pr2 xy))) (pr2 (pr2 (pr2 xy))) ;
+      clear xy ; intros rx ry -> Xr Yr.
```
mortberg: Maybe this would be easier to read with `set` instead of `generalize` and `intros`? One difference is that this will create a let bindings instead of a beta redexes.

UniMath/RealNumbers/NonnegativeReals.v
```diff
@@ -3210,70 +3369,125 @@ Lemma isldistr_Dcuts_max_plus :
...
+Proof.
+  intros x y z.
+  apply Dcuts_eq_is_eq.
+  intros r ; split.
+  - apply hinhuniv, sumofmaps.
+    apply sumofmaps ; intros Hr ; split.
+    now apply hinhpr, ii1, ii1.
+    now apply hinhpr, ii1, ii1.
+    now apply hinhpr, ii1, ii2, (pr1 Hr).
+    now apply hinhpr, ii1, ii2, (pr2 Hr).
```
mortberg: This would be easier to read (and maintain) with bullets. It's very hard to see which of the subgoals are being closed by the `now`'s and if some tactic/lemma changes it will be hard to figure out how to reorganize them. It is hard to tell by just looking at the code, but I think there are many other places where bullets could be used as well.

mortberg:
I can't really judge the mathematical content, but I put some comments here and there in the code and once these have been addressed this LGTM.

vladimirias:
I agree, it looks like here access functions can be very useful.

cathlelay:
I will try to make more access functions.
But, in NonnegativeReals.v them seem to me too specific, then I did not try to do it.
Most of time, when I do not define access function, it was because I was working under a troncation.
I began to prove intermediate lemmas for the addition and I will need to prove others for the multiplication and for the multiplicative inverse.

UniMath/RealNumbers/NonnegativeReals.v
cathlelay: I already had problems with "set" when I needed to use the tactic simpl.
It is an equivalent in coq of the ssreflect code for : "move:... =>{...} ..."

mortberg: Ok, interesting. Let's take a look at it together? Sometimes `cbn` works much better than `simpl` for me.

UniMath/RealNumbers/NonnegativeReals.v
cathlelay: I use `pattern` when I need to rewrite only some occurrences of an element. Unfortunately, I do not have other easy solution.

UniMath/RealNumbers/NonnegativeReals.v
cathlelay: `xy` is under a `hinhuniv`, I do not have any ideas to define useful access functions in this case.

vladimirias:
Define access functions from the expression under ishinh to its parts. Then use then them after applying hinhuniv.

vladimirias:
One way is, if you want to rewrite along e:a=b the first a in (a+a) to do the following:

set c:=a.
change (c=b) in e.
change (c+a).
rewrite e.

Maybe we can make a tactic for it?

PR #477, opened 2016-10-04
Multisorted signatures to "signature functors"

mortberg:
This PR contains:

- Definition of discrete precategories.
- Definition of multisorted signatures.
- Construction of a "signature functor" from a multisorted signature.

For now everything is in one file, I plan to put things in appropriate files soon. Also for now I've assumed that the sort has decidable equality, I'll try to get rid of this assumption later. I also had to assume that the indices form a set, I will probably have to assume that they even have decidable equality for proving that the indexed coproduct in the last definition is omega cocontinuous.

### DIFF
--- a/UniMath/SubstitutionSystems/MultiSorted.v
+++ b/UniMath/SubstitutionSystems/MultiSorted.v
@@ -47,27 +47,11 @@ Require Import UniMath.SubstitutionSystems.MonadsFromSubstitutionSystems.
 Local Notation "[ C , D ]" := (functor_Precategory C D).
 Local Notation "# F" := (functor_on_morphisms F)(at level 3).
 
-(* Lemma test4 (C1 C2 : precategory) (D1 D2 : Precategory) (F : functor [C1,D1] [C2,D2]) (G : [C1,D1]) : *)
-(*   adj_equivalence_of_precats F -> *)
-(*   is_omega_cocont G -> *)
-(*   is_omega_cocont (F G). *)
-(* Proof. *)
-(* intros α HG d L ccL HccL x ccx. *)
-(* unfold is_omega_cocont in HG. *)
-(* simpl in *. *)
-(* set (αinv := adj_equivalence_inv α). *)
-(* simpl in *. *)
-(* apply (test2 hsC hsD _ H _ _ _ _ HccL). *)
-(* Defined. *)
-
-
 (** Swapping of functor arguments *)
 (* TODO: Move upstream? *)
 Section functor_swap.
 
-Context {C D : precategory} {E : Precategory}.
-
-Lemma functor_swap : functor C [D,E] → functor D [C,E].
+Lemma functor_swap {C D : precategory} {E : Precategory} : functor C [D,E] → functor D [C,E].
 Proof.
 intros F.
 mkpair.
@@ -92,11 +76,8 @@ mkpair.
   | intros a b c f g; apply (nat_trans_eq (homset_property E)); intro x; simpl; apply functor_comp]).
 Defined.
 
-(* Lemma is_omega_cocont_functor_swap *)
-(*   (F : functor C [D,E]) (HF : forall c, is_omega_cocont (F c)) : *)
-(*    is_omega_cocont (functor_swap F). *)
-
-Lemma functor_cat_swap_nat_trans (F G : functor C [D, E]) (α : nat_trans F G) :
+Lemma functor_cat_swap_nat_trans {C D : precategory} {E : Precategory}
+  (F G : functor C [D, E]) (α : nat_trans F G) :
   nat_trans (functor_swap F) (functor_swap G).
 Proof.
 mkpair.
@@ -107,10 +88,12 @@ mkpair.
 + abstract (intros a b f; apply (nat_trans_eq (homset_property E)); intro c; simpl; apply nat_trans_ax).
 Defined.
 
-Lemma functor_cat_swap : functor [C, [D, E]] [D, [C, E]].
+Lemma functor_cat_swap (C D : precategory) (E : Precategory) : functor [C, [D, E]] [D, [C, E]].
 Proof.
 mkpair.
-- apply (tpair _ functor_swap functor_cat_swap_nat_trans).
+- mkpair.
+  + apply functor_swap.
+  + apply functor_cat_swap_nat_trans.
 - abstract (split;
   [ intro F; apply (nat_trans_eq (functor_category_has_homsets _ _ (homset_property E))); simpl; intro d;
     now apply (nat_trans_eq (homset_property E))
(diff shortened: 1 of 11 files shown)


UniMath/SubstitutionSystems/MultiSorted.v
```diff
@@ -0,0 +1,355 @@
...
+
+Require Import UniMath.SubstitutionSystems.Signatures.
+Require Import UniMath.SubstitutionSystems.SignatureExamples.
+Require Import UniMath.SubstitutionSystems.SumOfSignatures.
+Require Import UniMath.SubstitutionSystems.BinProductOfSignatures.
+Require Import UniMath.SubstitutionSystems.SubstitutionSystems.
+Require Import UniMath.SubstitutionSystems.LiftingInitial.
+Require Import UniMath.SubstitutionSystems.MonadsFromSubstitutionSystems.
+
+Local Notation "[ C , D , hs ]" := (functor_precategory C D hs).
```
benediktahrens: Note that we also have `Precategory` and `functor_Precategory` (?) for precats whose homs are sets.

UniMath/SubstitutionSystems/MultiSorted.v
```diff
@@ -0,0 +1,355 @@
...
+Defined.
+
+Definition is_precategory_DiscretePrecategory_data : is_precategory DiscretePrecategory_data.
+Proof.
+split; [split|]; trivial; intros.
++ apply pathscomp0rid.
++ apply path_assoc.
+Qed.
+
+Definition DiscretePrecategory : precategory :=
```
benediktahrens: Naming is bound to create confusion: `DiscretePrecategory` should be a `Precategory`, not a `precategory`. (I know, Dan suggested a naming convention that would avoid that confusion...)

UniMath/SubstitutionSystems/MultiSorted.v
```diff
@@ -0,0 +1,355 @@
...
+Section MBindingSig.
+
+Variable (sort : UU).
+Variable (eq : isdeceq sort). (* Can we eliminate this assumption? *)
+
+(** Define the discrete category of sorts *)
+Let sort_cat : precategory := DiscretePrecategory sort.
+
+(** This represents "sort → HSET" *)
+Let sortToHSET : precategory := [sort_cat,HSET,has_homsets_HSET].
```
benediktahrens: We will get cleaner code, I think, if from the beginning on we work with a category C instead of HSET. That is, the category of interest is the category `[sort_cat, C]`, with C having a terminal object, binary coproducts etc.

UniMath/SubstitutionSystems/MultiSorted.v
```diff
@@ -0,0 +1,355 @@
...
+          [sortToHSET,HSET,has_homsets_HSET].
+Proof.
+(* Apply the exp functor to every element of the list *)
+set (XS := map exp_functor xs).
+(* If the list is empty we output the constant functor *)
+set (T := constant_functor [sortToHSET, sortToHSET, has_homsets_sortToHSET]
+                           [sortToHSET, HSET, has_homsets_HSET]
+                           (constant_functor sortToHSET HSET emptyHSET)).
+(* TODO: Maybe use indexed finite products instead of a fold? *)
+(* TODO: Should we really use BinProduct_of_functors? Can we prove omega-cocont? *)
```
benediktahrens: We should use the same constructions we used for the untyped case. Hopefully we will only need very few extra lemmas, e.g., about cocontinuity of the projection from `[sort,C]` to `C` for a given `s : sort`.

UniMath/SubstitutionSystems/MultiSorted.v
benediktahrens: Also, Dan Grayson has a similar definition already, under the name of `path_pregroupoid`, in Ktheory/StandardCategories. Maybe this is too complicated to fix in the current PR, but the duplication should be addressed in a subsequent PR.

UniMath/SubstitutionSystems/MultiSorted.v
mortberg: It was called `functorPrecategory` for some reason. I'll rename it to `functor_Precategory` to match `functor_precategory` and use it.

UniMath/SubstitutionSystems/MultiSorted.v
mortberg: Yes, we use the this there so it should work. I'm not sure why I was confused about this... The comment has been removed.

UniMath/SubstitutionSystems/MultiSorted.v
mortberg: I renamed this to `discrete_precategory`. I agree that we should do some merging at some time, but this is probably not the right time to do it.

UniMath/SubstitutionSystems/MultiSorted.v
```diff
@@ -47,6 +47,20 @@ Require Import UniMath.SubstitutionSystems.MonadsFromSubstitutionSystems.
 Local Notation "[ C , D ]" := (functor_Precategory C D).
```
rmatthes: I do not understand why Coq tells me that the reference functor_Precategory was not found in the current
environment. I used the mechanism described in [https://github.com/UniMath/UniMath/wiki/Git-configuration-and-usage](https://github.com/UniMath/UniMath/wiki/Git-configuration-and-usage) to obtain the branch with the pull request.

benediktahrens:
On 10/07/2016 06:53 PM, Ralph Matthes wrote:
> rmatthes commented on this pull request.
> 
> 
> 
>> @@ -47,6 +47,20 @@ Require Import
>> UniMath.SubstitutionSystems.MonadsFromSubstitutionSystems.
> Local Notation "[ C , D ]" := (functor_Precategory C D).
> 
> I do not understand why Coq tells me that the reference
> functor_Precategory was not found in the current environment. I used
> the mechanism described in
> [https://github.com/UniMath/UniMath/wiki/Git-configuration-and-usage](https://github.com/UniMath/UniMath/wiki/Git-configuration-and-usage)
> to obtain the branch with the pull request.
> 

Did you recompile after pulling?

rmatthes:
I recompiled, but wrongly: I studied the new files but went back into master before compilation - and the system did not complain since master does not have the file I was looking at. Now okay. Sorry for that noise.

UniMath/SubstitutionSystems/MultiSorted.v
```diff
@@ -0,0 +1,792 @@
...
+    apply nat_trans_eq; [apply homset_property|]; intro G; simpl.
+    apply functor_id.
+  - intros F G H α β; simpl.
+    apply nat_trans_eq; [apply homset_property|]; intro γ; simpl.
+    apply functor_comp.
+Defined.
+
+(** Given a sort s this applies the sortToC to s and returns C *)
+Definition projSortToC (s : sort) : functor sortToC C.
+Proof.
```
rmatthes: Should this not just be an instance of `proj_gen_fun`?

mortberg:
I should warn you that this PR is in a very rough state right now and that I'm pushing patches to it regularly. So the best might be to wait to look at it until I've figured some things out.

UniMath/SubstitutionSystems/MultiSorted.v
mortberg: This has been fixed now.

mortberg:
I'll close this for now. I'll reopen it once I've figured out how to define the functor so that proving that it is omega cocontinuous is not too complicated.

PR #478, opened 2016-10-14
Lattice and truncated minus

cathlelay:
A new pull request with only the definitions of lattice and truncated minus.

### DIFF
--- a/UniMath/Foundations/Algebra/Lattice.v
+++ b/UniMath/Foundations/Algebra/Lattice.v
@@ -1,0 +1,417 @@
+(** * Lattice *)
+
+Require Export UniMath.Foundations.Algebra.BinaryOperations.
+Require Export UniMath.Foundations.Algebra.Monoids_and_Groups.
+
+(** ** Definition *)
+
+Definition islatticeop {X : hSet} (min max : binop X) :=
+  ((isassoc min) × (iscomm min))
+    × ((isassoc max) × (iscomm max))
+    × (Π x y : X, min x (max x y) = x)
+    × (Π x y : X, max x (min x y) = x).
+Definition islattice (X : hSet) := Σ min max : binop X, islatticeop min max.
+Definition lattice := Σ X : setwith2binop, islatticeop (X := X) op1 op2.
+
+Definition pr1lattice : lattice -> setwith2binop := pr1.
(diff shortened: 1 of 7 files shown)


UniMath/Foundations/Algebra/Lattice.v
```diff
@@ -0,0 +1,417 @@
...
+  λ X : lattice, op1,, op2,, pr2 X.
+Definition islattice2lattice : Π X : hSet, islattice X → lattice :=
+λ (X : hSet) (is : islattice X), mklattice (pr2 (pr2 is)).
+
+Section lattice_pty.
+
+Context {L : hSet}
+        (is : islattice L).
+
+Definition Lmin : binop L := pr1 is.
```
mortberg: Maybe a notation for min and max (/\ and \/) would look nice?

UniMath/Foundations/Algebra/Lattice.v
```diff
@@ -0,0 +1,417 @@
...
+  Π x y : L, Lmax x (Lmin x y) = x.
+Proof.
+  exact (pr2 (pr2 (pr2 (pr2 (pr2 is))))).
+Qed.
+
+Lemma Lmin_id :
+  Π x : L, Lmin x x = x.
+Proof.
+  intros x.
+  pattern x at 2 ; rewrite <- (Lmax_absorb x x).
```
mortberg: Like we discussed before pattern at a specific index is not the best as it produces fragile proof scripts. I investigated this one and found that what is done here can be achieved by:

`eapply pathscomp0; [apply maponpaths, (!Lmax_absorb x x)|].`

This is a little more verbose, but hopefully more stable. However I think the ideal solution would be to use a better `rewrite` tactic than the standard one of Coq (I'm of course thinking of the one of SSReflect/MathComp).

mortberg:
LGTM modulo my comments above.

UniMath/Foundations/Algebra/Lattice.v
cathlelay: `lattice` is not easy to use, then I have to use `is : islattice L` as a parameter of `Lmin`, `Lmax` and `Lle`.
These notations look nicer but are not useful when we want to work with algebraic structures as monoids.

PR #479, opened 2016-10-14
Category of signatures with strength

mortberg:
This PR defines the category of signatures with strength and proves that it has binary products and (indexed) coproducts.

### DIFF
--- a/UniMath/SubstitutionSystems/SignatureCategory.v
+++ b/UniMath/SubstitutionSystems/SignatureCategory.v
@@ -16,13 +16,8 @@ Section SignatureCategory.
 
 Variables (C : precategory) (hsC : has_homsets C).
 
-(** The forgetful functor from pointed endofunctors to endofunctors *)
 Local Notation "'U'" := (functor_ptd_forget C hsC).
-(** The precategory of pointed endofunctors on [C] *)
 Local Notation "'Ptd'" := (precategory_Ptd C hsC).
-(** The category of endofunctors on [C] *)
-Local Notation "'EndC'":= ([C, C, hsC]) .
-
 
 (* Define the commutative diagram used in the morphisms *)
 Section Signature_category_mor.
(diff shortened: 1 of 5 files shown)


UniMath/SubstitutionSystems/SignatureCategory.v
```diff
@@ -0,0 +1,305 @@
...
+  apply (nat_trans_eq (functor_category_has_homsets _ _ hsC)); intros X; apply assoc.
+Qed.
+
+Definition Signature_precategory : precategory :=
+ (Signature_precategory_data,,is_precategory_Signature_precategory_data).
+
+Lemma has_homsets_Signature_precategory : has_homsets Signature_precategory.
+Proof.
+intros Ht1 Ht2.
+change isaset with (isofhlevel 2); apply isofhleveltotal2.
```
benediktahrens: you can replace this by `apply (isofhleveltotal2 2).`

UniMath/SubstitutionSystems/SignatureCategory.v
```diff
@@ -0,0 +1,305 @@
...
+
+Require Import UniMath.SubstitutionSystems.Notation.
+Require Import UniMath.SubstitutionSystems.Signatures.
+Require Import UniMath.SubstitutionSystems.BinProductOfSignatures.
+Require Import UniMath.SubstitutionSystems.SumOfSignatures.
+
+(** * The category of signatures with strength *)
+Section SignatureCategory.
+
+Variables (C : precategory) (hsC : has_homsets C).
```
benediktahrens: Maybe using `Precategory` would be helpful?

UniMath/SubstitutionSystems/SignatureCategory.v
mortberg: Ok, I actually figured out how to do what I did by looking at the proof of `isaset_nat_trans`. I changed this to what you suggest both in my file and functor_categories.v, while I was doing this I also shortened and introduced bullets in a few proofs in functor_categories.

UniMath/SubstitutionSystems/SignatureCategory.v
mortberg: I tried this and noticed something very interesting: the time to check the Qed of `BinProductArrow_diagram` went up from 62s to 87s. Also applying the tactics took much longer.

mortberg: However, I just figured out how to prove that lemma without any `simpl` and `rewrite` and the Qed is now instant... Why is this?

I wonder if we can do the same to the other slow proofs in SubstitutionSystems?

mortberg: This humble patch

https://github.com/UniMath/UniMath/pull/479/commits/a59ec6fa52504aee1a7333e27ddf7e8ac6d37251

speeds up the slow proofs so that the total compilation time of the file goes from  1m33s to 0m16s.

benediktahrens:
On 10/17/2016 05:30 PM, Anders Mörtberg wrote:
> 
> However, I just figured out how to prove that lemma without any
> `simpl` and `rewrite` and the Qed is now instant... Why is this?
> 
> I wonder if we can do the same to the other slow proofs in
> SubstitutionSystems?

That would be great!
I recommend introducing a tactic
```
etrans := eapply pathscomp0
```
before systematically using this way.

benediktahrens:
lgtm (the forgetful functor isn't there yet, but that is not urgent)

mortberg:
I'm going to add the forgetful functor now. 

I'll also check if we get a slowdown if I use Precategory with the faster proofs as well.

mortberg:
If I add such a tactic, where should I put it? I can put it directly after `pathscomp0` together with the `intermediate_path` tactic.

mortberg:
Using Precategory makes the file compile in 15s instead of 12s. However it makes it nicer to use functor categories, so I'll keep it. 

I guess this experiment is an indication that our way of packaging structures is quite inefficient (I just add one more layer to our nested sigma type and the compilation time goes up 25%). So maybe we should consider a more sophisticated approach for packaging our structures if things one day get too slow (one could try reimplementing CT using mixins á la Mathematical Components, I'm not sure if we can do that without Records though...).

mortberg:
I added the etrans tactic and the forgetful functor now.

PR #488, opened 2016-10-29
Concatenate

tpannila:
This PR contains proofs of `concatenate_0_alt` and `concatenateStep_alt` which use alternative
definition `concatenate_alt` of `concatenate`. Other than that, the statements of these coincide
with statements of `concatenate_0` and `concatenateStep`. I have not verified that the function
`weqfromcoprodofstn_inv_map`, used in the definition `concatenate_alt`, is a weak equivalence, but
I expect it to be.

### DIFF
--- a/UniMath/Foundations/Combinatorics/FiniteSequences.v
+++ b/UniMath/Foundations/Combinatorics/FiniteSequences.v
@@ -363,23 +363,12 @@ Lemma append_length {X} (x:Sequence X) (y:X) :
   length (append x y) = S (length x).
 Proof. intros. reflexivity. Defined.
 
-Definition concatenate {X} : binop (Sequence X).
-Proof.
-  intros ? x y.
-  exists (length x + length y).
-  intros i.
-  induction (invweq (weqfromcoprodofstn (length x) (length y)) i) as [j|k].
-  - exact (x j).
-  - exact (y k).
-Defined.
-
-(** Alternative definition of concatenate. *)
-Definition concatenate_alt {X : UU} : binop (Sequence X).
+Definition concatenate {X : UU} : binop (Sequence X).
 Proof.
   intros X x y.
   use tpair.
   - exact (length x + length y).
-  - intros i. induction (weqfromcoprodofstn_inv_map _ _ i) as [j | k].
+  - intros i. induction (weqfromcoprodofstn_invmap _ _ i) as [j | k].
     + exact (x j).
     + exact (y k).
 Defined.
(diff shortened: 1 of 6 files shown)


vladimirias:
natplusnsms is essentially equivalent to natplusnsm (also in NaturalNumbers).

vladimirias:
minusminusplus is essentially equivalent to natsubsub (in the same file). The proof of natsubsub looks better, but it would be good if you improve it further by removing the two step inductions ( [ | n N ] and [ | i _ ]) . Also the name natminusminus is a good name for natsubsub and natminusplus for your minusminusplus.

vladimirias:
This: 

(*
Local Variables:
compile-command: "make -C ../.. UniMath/Foundations/Sequences.vo"
End:
*)

probably should not be at the end of finite sequences. Instead (* End of the file *) would be good.

vladimirias:
Why do you use induction in the proof of weqfromcoprodofstn_inv_map_lt when you have a direct proof using natlthandminusl, natlthandplusm and plusminusnmm?

vladimirias:
The goal in weqfromcoprodofstn_inv_map_empty should be (the) empty. In the proof first rewrite H in the other direction, then use stnlt i. You will get i<n and i>=n. Then apply natlthtonegnatgeh.

vladimirias:
I think the two cases in the proof of weqfromcoprodofstn_inv_map after (natchoice0 m) should be done entirely in nat and without external lemmas.

tpannila:
I made the following changes

- Cleaned a few long lines
- Cleaned some proofs as you suggested. Rewrite did not work in all cases, so I
  used induction
- Renamed things as you suggested
- Some test cases in Combinatorics/Tests.v do not work (see the changes in
  Tests.v)
- The construction in weqcoprodfromstn_invmap cannot be moved to nat after
  natchoice0 because of the ii2 part. The idea to use external lemmas is to
  separate construction part from proposition part. If we don't use external
  lemmas, then unfolding weqcoprodfromstn_invmap shows the proofs of the
  propositions, which slows down the proof assistant, and makes sometimes the
  output of cbn and simpl harder to read.
- Changed a few lines in Topology/Filters.v and Topology/Topology.v to make
  them to work with the new definition of concatenate.
- Added reverse for finite sequences and dualelement for i : stn n, which is
  given by n - 1 - i.

vladimirias:
This looks much better. More work may be good in the future to clean it up further. Now I suggest to accept.

PR #489, opened 2016-11-04
Added isaset_hfiber corollary

mweav:
Vladimir recommended I submit this PR which is just adding a corollary `isaset_hfiber` as I am pretty sure there is no equivalent existing lemma.


vladimirias:
LGTM

Issue #444, opened 2016-09-05
Illegal use of `Inductive` in NaturalNumbers


benediktahrens:
When reviewing #427 , I stumbled across what looks like an illegal use of `Inductive` in the file Basics/NumberSystems/NaturalNumbers:

```
Inductive leF { T : UU } ( F : T -> T ) ( t : T ) : T -> UU := leF_O : leF F t t | leF_S : Π t' : T , leF F t t' -> leF F t ( F t' ) .
```

I suggest this be removed once #427 is merged.

vladimirias:
This is intentional. This Inductive defines an object that is used in the proof that the inductive definition of <= on natural numbers given in coq.init defines an object in hProp. Then it is easy to show that it is equivalent to natleh.

In fact, this use of “Inductive” is even mentioned in the description of Foundations. I don’t know if it is present in our README.

It can not be removed without removing the comparison result, which anyway refers to the inductive definition of <= in coq.init.

We can isolate it in a separate file and create a special package for results that prove comparisons of the UniMath constructions with the existing inductive constructions created by earlier Coq developers. No results from this package can be used anywhere else in the UniMath.

Vladimir.

benediktahrens:
fixed by moving the inductive type to a separate file, closing.

PR #490, opened 2016-11-06
moved inductive predicate le into separate file

benediktahrens:
Work on issue #444 (Illegal use of Inductive in NaturalNumbers):

- remove the inductive definition and following stuff from NaturalNumbers
- put the removed content in file NaturalNumbers_le_Inductive so that by typing the filename, one gets suspicious.


vladimirias:
LGTM

PR #492, opened 2016-11-08
Added modification of category of elements specialized for presheaves

mweav:
I added a copy of `Elements.v` that modifies the definition of the category of elements to specifically be over a presheaf instead of a covariant functor. In particular, when defined over a presheaf `P : [C^op, SET]`, the category of elements as defined in `ElementsOp.v` is a subcategory (modulo a forgetful functor) of `C` instead of `C^op` (as would be the case using the definition in `Elements.v`).


vladimirias:
LGTM Are there any other comments? @DanGrayson?

PR #493, opened 2016-11-10
lazy evaluation works on maps from lists to base types

benediktahrens:
- maps defined by initiality on our "inductive" type of lists over A (initial algebra of X => 1 + A x X) compute when the target type is a "base type".
- such maps don't compute, of course, when the target type is an inductive type (and hence a quotient) itself.

### DIFF
--- a/UniMath/CategoryTheory/Inductives/Lists.v
+++ b/UniMath/CategoryTheory/Inductives/Lists.v
@@ -196,6 +196,9 @@ apply listIndProp.
   now rewrite !foldr_cons, <- Hl.
 Qed.
 
+Definition concatenate : pr1 List -> pr1 List -> pr1 List :=
+  fun l l' => foldr _ l cons l'.
+
 End lists.
 
 (** Some examples of computations with lists over nat *)
@@ -227,6 +230,8 @@ Eval lazy in length _ testlist.
 Eval lazy in length _ testlistS.
 Eval lazy in sum testlist.
 Eval lazy in sum testlistS.
+Eval lazy in length _ (concatenate _ testlist testlistS).
+Eval lazy in sum (concatenate _ testlist testlistS).
 
 Goal (Π l, length _ (2 :: l) = S (length _ l)).
 simpl.
--- a/UniMath/CategoryTheory/Inductives/Lists.v
+++ b/UniMath/CategoryTheory/Inductives/Lists.v
@@ -221,6 +221,13 @@ Definition sum : pr1 (List natHSET) -> nat :=
 (* Eval vm_compute in sum testlist. *)
 (* Eval vm_compute in sum testlistS. *)
 
+(* All of these compute *)
+Eval lazy in length _ (nil natHSET).
+Eval lazy in length _ testlist.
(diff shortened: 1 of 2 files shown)


vladimirias:
> On Nov 10, 2016, at 11:51 AM, Benedikt Ahrens <notifications@github.com> wrote:
> 
> when the target type is an inductive type (and hence a quotient) itself.

What does this mean? Why an inductive type is a quotient and what is a base type if not an inductive type?

benediktahrens:
On 11/10/2016 07:00 PM, Vladimir Voevodsky wrote:
> 
>> On Nov 10, 2016, at 11:51 AM, Benedikt Ahrens
>> <notifications@github.com> wrote:
>> 
>> when the target type is an inductive type (and hence a quotient)
>> itself.
> 
> What does this mean? Why an inductive type is a quotient and what is
> a base type if not an inductive type?

Here, I call "base type" a type that is implemented as a Coq Inductive datatype (Nat, Bool, empty, unit) and an "inductive type" an initial algebra constructed as the colimit (hence quotient) of a chain. 
As examples of the latter we have lists over a fixed type, the lambda calculus etc.

Maps from lists to, say, Nat, such as length, do compute to a value.
This is shown by the examples this PR gives.

Maps from lists to lists, such as appending, do not compute to anything in reasonable time, since the target type is a quotient. So we don't have any examples of computation for those maps.

vladimirias:
I see. It would be good to call it something a little different than just inductive type. This way there is too much confusion with the Inductive Datatype.

Speaking of computation, what if we consider lists to lists to nat?

benediktahrens:
On 11/10/2016 07:49 PM, Vladimir Voevodsky wrote:
> I see. It would be good to call it something a little different than
> just inductive type. This way there is too much confusion with the
> Inductive Datatype.
> 
> Speaking of computation, what if we consider lists to lists to nat?

I tested a few more examples:

- a map of the type you suggest: take two lists, concatenate them, then take the length resp. the sum (when the entries are Nats). Does compute when evaluating lazily.

- The Trees of Inductives/Trees.v are similar to lists, but each node has two subtrees instead of just one. That is, the signature functor for trees over base type A is
F(X) := 1 + A x X x X
I have implemented a map from Trees over A to Lists over A, which flattens a Tree.
We can then compose flattening with taking the sum to obtain a map from Trees over Nat to Nat. Does compute when evaluating lazily.

vladimirias:
Cool!

mortberg:
This is nice and I think it should be merged. But maybe we should comment the Eval lines to reduce noise in the output when compiling?

I found it very strange that `lazy` works but not `cbn` (as lazy computation usually denote call-by-name with memoization/sharing) so I asked some of the Coq experts here. Apparently `lazy` is part of the kernel (more specifically to test conversion) and computes WHNF using a lazy strategy. `cbn` on the other hand is a tactic which is like a call-by-need variant of `simpl` (so it might do some crazy things, while `lazy` is part of the trusted core so it should be correct).

vladimirias:
LGTM and merging.

Issue #495, opened 2016-11-11
direction of = in slicecategory


vladimirias:
How difficult would it be to change the direction of equality in the definition of a morphism h from f to g in slicecategory from the current h ;; g = f to f = h ;; g?

There are some reasons to consider the latter to be more natural.

benediktahrens:
Slice categories are not used in any other file, neither in UniMath/UniMath nor in UniMath/TypeTheory. 
So it should be a minor change, and not difficult at all.

benediktahrens:
resolved in #497

PR #497, opened 2016-11-12
Switched the direction of the equality in slicecat morphisms

mweav:
I changed the direction of the equality in the definition of slice category morphisms so the composition is on the right, as requested in #495.


benediktahrens:
LGTM

PR #498, opened 2016-11-14
Direct proof that the precomposition functor is cocontinuous

mortberg:
This PR implements:

- A direct proof that the precomposition functor is cocontinuous (not relying on construction of right Kan extensions from limits)
- Move the construction of right Kan extensions from limits to the file on Kan extensions
- Make some arguments implicit in the construction of (co)limits in functor categories


benediktahrens:
lgtm

benediktahrens:
This PR is a first step to fixing a problem. I am merging in order to speed up work on that fix.

PR #499, opened 2016-11-15
Archimedean property

cathlelay:
- move proofs that nat, hz, and hq are Archimedean in them respective files
- rename some theorems in Archimedean.v


vladimirias:
LGTM

PR #500, opened 2016-11-17
Lattice - definition

cathlelay:
A lattice is a set with two binary operators `min` and `max` such that : 
- `min` and `max` are associative and commutative
- `Π x y : X, min x (max x y) = x`
- `Π x y : X, max x (min x y) = x`
In a lattice, we can define a partial order : 
`Lle {X : hSet} (is : islattice X) := λ (x y : X), Lmin is x y = x`.

A lattice with a strong order `gt` is a lattice such that : 
- `Π x y : X, (¬ (gt x y)) <-> Lle is x y`
- `Π x y z : X, gt x z → gt y z → gt (Lmin is x y) z`
- `Π x y z : X, gt z x → gt z y → gt z (Lmax is x y)`

### DIFF
--- a/UniMath/Foundations/.package/files
+++ b/UniMath/Foundations/.package/files
@@ -16,6 +16,7 @@ Algebra/DivisionRig.v
 Algebra/Apartness.v
 Algebra/ConstructiveStructures.v
 Algebra/Archimedean.v
+Algebra/Lattice.v
 NumberSystems/NaturalNumbers.v
 NumberSystems/NaturalNumbers_le_Inductive.v
 Combinatorics/StandardFiniteSets.v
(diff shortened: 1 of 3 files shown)


vladimirias:
I suggest instead of Lemmas introduce objects such as isassoc_Lmin in definitions given in one line without using "Proof".

vladimirias:
Could you please find a reference for this definition of a lattice and include it into the file as a (** comment? There are many books about lattices and it would be good to have your results, where they are not new, connected by references to their source (even if you did not look into any books when proving them). Then new results will be made seen by the absence of references. Also, please make a header for the file with your name and the date.

vladimirias:
Also, please make a table of content as we have made in in the files in Basics, especially PartA so that by opening the file one can see what it contains.

vladimirias:
In the definition of a lattice with a strong order, do you really need both implications in (not(gt x y)) <-> Lle is x y?

vladimirias:
LGTM

PR #501, opened 2016-11-18
Colimits of a specific shape

mortberg:
This PR introduce `Colims_of_shape g C` where g is a graph in order to get better control over the shape of the diagrams in C that we can compute colimits of (and similarily for limits). We then use this everywhere possible in order to make sure that the graphs we quantify over are small.


benediktahrens:
lgtm

PR #502, opened 2016-11-18
Multisorted extras

mortberg:
This PR collects the results that I finished while working on the generalization to multisorted signatures which is not finished yet. In particular it contains:

- Definition of discrete precategories over a type

- Swapping of functor category arguments: [C,[D,E]] -> [D,[C,E]].   This functor is adjoint with itself with arguments swapped, so it's is cocontinuous.

- (bin)(co)products from (co)lims of suitable shape

- Colimits in slice categories

- Various rewriting lemmas for lists (defined as iterated products)

- Renaming and reorganization of definitions.


UniMath/CategoryTheory/slicecat.v
```diff
@@ -326,8 +324,69 @@ assert (H2 : Π h', h' = h ->
...
++ mkpair.
+  - apply pr1.
+  - intros a b; apply pr1.
++ abstract (now split).
+Defined.
+
+Lemma slice_precat_ColimCocone (d : diagram g (C / x))
+  (H : ColimCocone (mapdiagram slicecat_to_cat d)) :
+  ColimCocone d.
+Proof.
```
benediktahrens: I think that this lemma would be better stated by saying that the forgetful functor `slicecat_to_cat' creates limits, that is, by saying that, given a diagram in C/X and a cocone CC under that diagram, then CC is colimiting if the image of CC under the forgetful functor is.

UniMath/CategoryTheory/DiscretePrecategory.v
```diff
@@ -0,0 +1,57 @@
...
+  functor discrete_precategory D.
+Proof.
+mkpair.
++ mkpair.
+  - apply f.
+  - intros s t []; apply identity.
++ abstract (now split; [intro|intros a b c [] []; simpl; rewrite id_left]).
+Defined.
+
+End DiscretePrecategory.
```
benediktahrens: There might be some code on discrete precategories in Ktheory. If this is the case, then it would be good to move it to CategoryTheory, if that is easily possible.

UniMath/CategoryTheory/equivalences.v
```diff
@@ -30,6 +31,7 @@ Require Import UniMath.Foundations.Basics.Sets.
 Require Import UniMath.CategoryTheory.precategories.
 Require Import UniMath.CategoryTheory.functor_categories.
 Require Import UniMath.CategoryTheory.UnicodeNotations.
+Require Import UniMath.CategoryTheory.whiskering.
```
benediktahrens: It seems that this import was added without reason, since no code was added to the file.

UniMath/CategoryTheory/functor_categories.v
```diff
@@ -63,7 +63,7 @@ Ltac pathvia b := (apply (@pathscomp0 _ _ b _ )).
 
 
 (** * Functors : Morphisms of precategories *)
-
+Section functors.
```
benediktahrens: Why is a section being opened here?

UniMath/CategoryTheory/slicecat.v
```diff
@@ -304,9 +312,9 @@ apply (functor_eq _ _ (has_homsets_slice_precat _ _ _)); simpl.
 unfold slicecat_functor_data; simpl.
 unfold functor_composite_data; simpl.
 apply (total2_paths2 (slicecat_functor_comp_ob _ _ _ _ _)).
-apply funextsec; intro a; case a; clear a; intros a fax.
-apply funextsec; intro b; case b; clear b; intros b fbx.
-apply funextsec; intro h; case h; clear h; intros h hh.
+apply funextsec; intros [a fax].
+apply funextsec; intros [b fbx].
+apply funextsec; intros [h hh].
```
benediktahrens: I think that we have a rule that forbids such intro patterns.

UniMath/CategoryTheory/slicecat.v
```diff
@@ -326,8 +334,106 @@ assert (H2 : Π h', h' = h ->
...
+  isColimCocone d a cc.
+Proof.
+set (CC := mk_ColimCocone _ _ _ H).
+intros y ccy.
+  use unique_exists.
+  + mkpair; simpl.
+    * apply (colimArrow CC), (mapcocone U _ ccy).
+    *
+apply pathsinv0.
+eapply pathscomp0.
```
benediktahrens: Here is weird formatting: no indentation.

UniMath/CategoryTheory/slicecat.v
```diff
@@ -326,8 +334,106 @@ assert (H2 : Π h', h' = h ->
...
+
+Lemma slice_precat_colims_of_shape (CC : Colims_of_shape g C) : Colims_of_shape g (C / x).
+Proof.
+now intros d; apply slice_precat_ColimCocone, CC.
+Defined.
+
+End slicecat_colimits.
+
+Lemma slice_precat_colims {C : precategory} (hsC : has_homsets C) (x : C) (CC : Colims C) :
+  Colims (slice_precat C x hsC).
```
benediktahrens: This would be better phrased in terms of `colims_of_shape'.

vladimirias:
Matthew Weaver, I think, also has a lemma that any family of morphisms defines a morphism of functors.

vladimirias:
Some of us (Cath, Tommi) use quite sophisticated patters a lot. I let it go for now since it does make code look good and it is my understanding that translating such patterns into “inductions” is straightforward. So I guess it depends on how strict is the language in a given package.

UniMath/CategoryTheory/equivalences.v
mortberg: I had some code there that depended on this, but then I removed the code. I'll remove the import for now.

UniMath/CategoryTheory/functor_categories.v
mortberg: I thought it looked nicer to encapsulate the definitions about functors in a section. When that section closes the next that opens is the one about natural transformations. I guess we could also make C and C' section variables (and maybe implicit) if we want to, but I didn't bother. This change was more for documentation/organization purposes (another nice thing with encapsulating similar definitions in a section is that you can easily jump to the end by grepping for `End foo` so navigating the files in emacs gets a little easier).

mortberg:
I think the sophisticated intro patterns are quite nice (and I recently learned that they will get even better soon), and as long as they appear in a lemma that is ended with Qed they shouldn't be problematic. The reason I like them is that this kind of bookkeeping (destructing, naming, removing hypotheses) is usually not really interesting, so it shouldn't take up much room in the formal proof.

UniMath/CategoryTheory/slicecat.v
mortberg: That lemma was stated just above, I moved it down to have them together.

UniMath/CategoryTheory/slicecat.v
mortberg: Sorry, I finished the lemma last night and didn't have the energy to clean it. It has been cleaned now.

UniMath/CategoryTheory/DiscretePrecategory.v
benediktahrens: I think that there should be a coordinated effort to move stuff from Ktheory to more basic packages. Not in this PR, but in one or more separate ones.

benediktahrens:
ok, lgtm

vladimirias:
The problem with them has been and remains that they are unfolded into the match expressions of the form that has not been really verified to admit a univalent model. It should be possible to unfold them into expressions based on the eliminators, but there remains the question of whether all the patterns that we use can be unfolded in such a way.

DanGrayson:
Maybe we should modify Coq so the intro patterns are implemented using eliminators instead of "match".  If we make it a command-line option that turns that on or off, then the Coq team would accept the patch.

vladimirias:
Do you think you can make such patch?

DanGrayson:
I haven't looked at the code, so I don't know for sure, but it might be
possible.

benediktahrens:
@mortberg There is now a merge conflict in the files list. Could you fix that, plz?

vladimirias:
Yes, please fix the conflicts and merge.

mortberg:
I fixed the conflicts now. I don't have rights to merge, but please do it once travis is happy.

PR #503, opened 2016-11-22
Alternative definition of GenMendlerIteration and LiftingInitial

mortberg:
Implement alternative versions of GenMendlerIteration and LiftingInitial. These new files produce the same results, but have a different set of hypotheses. In particular there is no need for the construction of Kan extension but we can rely on omega cocontinuity instead. 

This is used in BindingSigToMonad in order to eliminate the hypotheses of general limits which was needed for the construction of Kan extensions. The code there now only needs the existence of colimits for omega chains.

### DIFF
--- a/UniMath/SubstitutionSystems/GenMendlerIteration_alt.v
+++ b/UniMath/SubstitutionSystems/GenMendlerIteration_alt.v
@@ -168,9 +168,79 @@ Proof.
 apply (colimArrowCommutes temp).
 Qed.
 
+Lemma is_iso_inF : is_iso inF.
+Proof.
+(* Use Lambek's lemma, this could be extracted from the concrete proof as well *)
+apply (initialAlg_is_iso _ hsC), pr2.
+Defined.
+
+Let inF_iso : iso (F μF) μF := isopair _ is_iso_inF.
+Let inF_inv : C⟦μF,F μF⟧ := inv_from_iso inF_iso.
+
 (* The direction ** -> * *)
-Lemma SS_imp_S h (H : Π n, # L (e n) ;; h = Pow n IC z) : # L inF ;; h = ψ μF h.
-Admitted.
+Lemma SS_imp_S (H : Π n, # L (e n) ;; preIt = Pow n IC z) : # L inF ;; preIt = ψ μF preIt.
+Proof.
+assert (H' : Π n, # L (e (S n)) ;; # L inF_inv ;; ψ μF preIt = pr1 (Pow (S n)) _ z).
+{ intro n.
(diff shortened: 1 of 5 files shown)


UniMath/SubstitutionSystems/GenMendlerIteration_alt.v
```diff
@@ -0,0 +1,285 @@
...
+Context {C : precategory} (hsC : has_homsets C) (IC : Initial C)
+        (CC : Colims_of_shape nat_graph C) (F : functor C C)
+        (HF : is_omega_cocont F).
+
+Local Notation "0" := (InitialObject IC).
+
+Let AF := FunctorAlg F hsC.
+Let chnF := initChain IC F.
+
+Definition μF_Initial : Initial AF := colimAlgInitial hsC IC HF (CC chnF).
```
benediktahrens: I'd prefer this to be a notation instead of a definition, because as a definition, it is just a worse version of `colimAlgInitial'.

UniMath/SubstitutionSystems/GenMendlerIteration_alt.v
```diff
@@ -0,0 +1,285 @@
+(** **********************************************************
+
+Contents:
+
+- Derivation of Generalized Iteration in Mendler-style Instantiation to a special case, Specialized
+- Mendler Iteration Proof of a fusion law à la Bird-Paterson (Generalised folds for nested
+- datatypes) for Generalized Iteration in Mendler-style
```
benediktahrens: More specifically, this is Theorem 1 of that paper.

mortberg: Ok, I copied this from the GenMendlerIteration file

UniMath/SubstitutionSystems/GenMendlerIteration_alt.v
mortberg: I used a Let for this. Everything around there are Let's so it looked nicer that way.

mortberg:
The FromBindingSigToMonad_summary file doesn't rely on anything in these files yet. I plan to update it soon based on what is in the paper now. That will probably go in another PR.

benediktahrens:
lgtm

benediktahrens:
Actually, could you please add `InitialHSS' to the summary file? Be careful to get the right one...

mortberg:
I added InitialHSS from LiftingInitial_alt now

benediktahrens:
I am merging this PR, but hope that @rmatthes is going to take a look at it anyway.

PR #504, opened 2016-11-26
Minor changes to code for "from binding signatures to monads" paper

mortberg:
Mostly cosmetic changes to make the code for the paper look better. The only interesting  change is to now use the new version of LiftingInitial in LamFromBindingSig, LamHSET and MLTT79 so that they don't depend on the existence of Kan extensions anymore.

### DIFF
--- a/UniMath/CategoryTheory/CocontFunctors.v
+++ b/UniMath/CategoryTheory/CocontFunctors.v
@@ -130,8 +130,9 @@ Section omega_cocont.
 >>
    with exactly one arrow from n to S n.
 *)
+
 Definition nat_graph : graph :=
-  tpair (λ D : UU, D → D → UU) nat (λ m n, S m = n).
+  mk_graph nat (λ m n, 1 + m = n).
 
 Local Notation "'chain'" := (diagram nat_graph).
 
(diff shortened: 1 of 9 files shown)


rmatthes:
LGTM
It would be good to state in the header of LamFromBindingSig.v that this is only the instance for the category of sets although it has a name without HSET. Same with MLTT79.v.

benediktahrens:
Could you remove `RightKanExtension_from_limits' from the file FromBindingSigsToMonads_Summary.v ?

mortberg:
I did what you suggested now.

benediktahrens:
@mortberg  Compilation error in the file CT/Inductives/List.v .

vladimirias:
Could you clarify? What do you mean by it being an instance?

rmatthes:
What I meant by instance is that FromBindingSigsToMonads_Summary.v lists the requirements for a base category C for which the whole chain of constructions works. HSET fulfills all of them, but one might think of another category as basis for the semantics of MLTT79.

mortberg:
I fixed the problems now, I hope travis will be happy.

vladimirias:
I see.

PS I would not call it semantics of MLTT79. It is about raw syntax of MLTT79.

PR #505, opened 2016-11-27
remove broken import and thus make UniMath compile again

benediktahrens:
Merging #473 has broken UniMath due to an obsolete Import.
This PR removes that obsolete import.


vladimirias:
LGTM and merging.

PR #507, opened 2016-11-29
Right Modules over a monad

amblafont:


benediktahrens:
Two comments:

- The formatting could be improved a bit.
- Would it make sense to express the fact that the modules you are implementing are left or right, or whatever? It might also be the opportunity to think about better vocabulary than 'left' and 'right'...

benediktahrens:
lgtm

PR #512, opened 2016-12-06
Update the summary file for the binding signatures to monads paper

mortberg:
This PR updates the summary file for our recent paper on the translation from binding signatures to monads. It now lists all the major results and definitions of the paper with links to where they are in the development. While doing this I noticed that we didn't specify which universe `is_cocont` was supposed to be in so that it ended up in `Type` instead of `UU`, this has now been fixed.

### DIFF
--- a/UniMath/CategoryTheory/CocontFunctors.v
+++ b/UniMath/CategoryTheory/CocontFunctors.v
@@ -43,7 +43,7 @@ This file also contains proofs that the following functors are (omega-)cocontinu
 - Product of functors: F * G : C -> D, x |-> (x,x) |-> (F x,G x) |-> F x * G x
   [is_omega_cocont_BinProduct_of_functors_alt] [is_omega_cocont_BinProduct_of_functors]
 - Precomposition functor: _ o K : ⟦C,A⟧ -> ⟦M,A⟧ for K : M -> C
-  [is_cocont_pre_composition_functor] [is_omega_cocont_pre_composition_functor]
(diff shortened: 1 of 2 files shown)


benediktahrens:
lgtm, but I'd be happy if @rmatthes could also take a look

rmatthes:
I wanted to take a closer look - from the perspective of a reader of the paper. Please be patient until tomorrow.

rmatthes:
a stupid comment: FromBindingSigsToMonads_Summary.v is missing the final newline, making it impossible to use ediff in Emacs

rmatthes:
the file is now a very good help for the reader of the paper

some small comments for improvement:
- Def. 10 and Lemmas 11 and 12 are not reflected in the summary, how about just referring to the file where they sit (Signature_Examples.v which is maybe not obvious), in the style of mentioning Example 39?
- Theorem 42 in the paper is even stronger than what is proved in the file that is referred to; would there have been any problem in establishing the stronger form in UniMath?
- CocontFunctors.v still has lemmas with right Kan extensions (not pertaining to this PR, though)
- Theorem 48 only says that we obtain an HSS, not that it is initial in that category (although this was important in the HSS paper for TYPES'15). Maybe it would be better to refer also to InitHSS where the HSS itself is constructed.

mortberg:
I updated the summary file now to take Ralph's comments into account. 

I didn't do anything about the Kan extensions in the CocontFunctors file. I think it would be good to do what Benedikt suggested at some point and have a dummy argument `small C` just to make sure that we don't use the dangerous lemmas incorrectly (until we have proper universe handling).

PR #513, opened 2016-12-07
Various results on adjunctions

mortberg:
This PR modify the definition of adjunctions and add some new results:

- Add `is_right_adjoint` and make some results depend on `are_adjoints` instead of `is_left_adjoint`.
- Add `preserves_limit` and prove that right adjoints preserve limits. 
- Prove that postcomposition with a left adjoint is a left adjoint (and hence (omega)-cocontinuous)).
- Generalize the monoidal structure on endofunctors to arbitrary functors. The file EndofunctorsMonoidal.v now only contains instances of the more general definitions.

### DIFF
--- a/UniMath/CategoryTheory/equivalences.v
+++ b/UniMath/CategoryTheory/equivalences.v
@@ -32,86 +32,102 @@ Require Import UniMath.CategoryTheory.precategories.
 Require Import UniMath.CategoryTheory.functor_categories.
 Require Import UniMath.CategoryTheory.UnicodeNotations.
 
-Ltac pathvia b := (apply (@pathscomp0 _ _ b _ )).
-
-Local Notation "a --> b" := (precategory_morphisms a b)(at level 50).
-(*Local Notation "'hom' C" := (precategory_morphisms (C := C)) (at level 2).*)
-(* Local Notation "f ;; g" := (compose f g) (at level 50, format "f  ;;  g"). *)
 Local Notation "[ C , D , hs ]" := (functor_precategory C D hs).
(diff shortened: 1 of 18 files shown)


UniMath/CategoryTheory/equivalences.v
```diff
@@ -573,6 +584,51 @@ mkpair; simpl.
...
+              (nat_trans_functor_assoc _ _ _)
+              (nat_trans_comp _ _ _ (pre_whisker F' ε) (nat_trans_functor_id_left _))).
+  * abstract (intros F1 F2 α; apply (nat_trans_eq hsE); intro c; simpl in *;
+              now rewrite !id_right, !id_left; apply (nat_trans_ax ε _ _ (α c))).
+- abstract (split; simpl; intro F';
+    [ apply (nat_trans_eq hsE); simpl; intro c;
+      now rewrite !id_left, !id_right; apply H1
+    | apply (nat_trans_eq hsD); simpl; intro c;
+      now rewrite !id_left, !id_right; apply H2]).
+Defined.
```
benediktahrens: Would it make sense to break this construction up into several pieces?

mortberg: Not really... I think the way to fix it would be to make the constructors handle the opacification as I suggested the other day. We could then remove these horrible abstract blocks. 

Should I do this in this PR or in a separate on?

benediktahrens:
now a merge conflict in CocontFunctors.v

mortberg:
Fixed the conflict

benediktahrens:
lgtm

PR #515, opened 2016-12-09
Update README.md after the paper by Ahrens, Matthes & Mörtberg

rmatthes:
the README.md did not keep up with the development for the mentioned paper (file names changed, disappeared, emerged, ...)
now every item revisited and fully updated


rmatthes:
my PR does not concern any UniMath code, so you can safely ignore the travis timeout problem for this PR

PR #517, opened 2016-12-11
Opposite categories, cone, cylinder, and translation functors

tpannila:
This pull request contains the following changes
- Copied the changes in opp_precat.v from the Ktheory folder
- The new file limits/Opp.v contains translations of some special (co)limits in C to special
  (co)limits in C^op and vice versa.
- Opposite category of an abelian category is an abelian category, Abelian_opp, line 1559 in
  Abelian.v 
- Redefined kernels and cokernels so that they are independent of equalizers and coequalizers.
  In particular, one now uses is_Kernel and is_Cokernel instead of is_Equalizer and is_Coequalizer.
- Constructed translation functors in C(A) and K(A), see section translation_functor, line 2476 in
  Complexes.v.
- Constructed mapping cone and mapping cylinder in C(A), see sections mapping_cone, line 3573 in
  Complexes.v and mapping_cylinder, line 3829 in Complexes.v.
- Proved that for a morphism of complexes f : X -> Y, Y and Cyl(f) are isomorphic in K(A)
  (These are not isomorphic in general in C(A)!). See section mapping_cylinder_KA_iso, line 4058 in
  Complexes.v   
- Renamed PrecategoriesWithBinops to precategoriesWithBinops, made MMor a coercion in  Complexes.v,
  renamed CEq to DSq, and other fixes following suggestions of Vladimir.
- Proved that the translation functors C(A) -> C(A) and K(A) -> K(A) are isomorphisms. See
  `TranslationInvTranslation`, line 2957, `InvTranslationTranslation`, line 3008,
  `TranslationHInvTranslationH`, line 3477, `InvTranslationHTranslation`, line 3516, Complexes.v

To prove that the translation functors C(A) -> C(A), `TranslationFunctor` and
`InvTranslationFunctor`, are isomorphisms, I had to assume a new axiom. This axiom is
`transportf_funextfun`, line 90, functor_categories.v. The underlying reason why I need this axiom
is that I need to get the original pointwise equalities back from funextfun. I believe this
new axiom is not provable in UniMath because univalence, and hence funextfun, is not provable.
I think it should be provable in cubicaltt. (In experiments/uafunext1.ctt can one recover the
argument pe from funext? )

I don't know where to put the axiom `transportf_funextfun`. I think it should not be in
functor_categories.v and suggestions on where it should be, if it is accepted, are welcome.

### DIFF
--- a/UniMath/CategoryTheory/PrecategoriesWithAbgrops.v
+++ b/UniMath/CategoryTheory/PrecategoriesWithAbgrops.v
@@ -184,21 +184,21 @@ Section transport_morphisms.
 
   Variable PA : PrecategoryWithAbgrops.
 
-  Lemma transportf_to_inv {x y z : ob PA} (f : x --> y) (e : y = z) :
+  Lemma transport_target_to_inv {x y z : ob PA} (f : x --> y) (e : y = z) :
     to_inv (transportf (precategory_morphisms x) e f) =
     transportf (precategory_morphisms x) e (to_inv f).
   Proof.
     induction e. apply idpath.
   Qed.
 
-  Lemma transportb_to_inv {x y z : ob PA} (f : y --> z) (e : x = y) :
-    to_inv (transportb (fun x' : ob PA => precategory_morphisms x' z) e f) =
-    transportb (fun x' : ob PA => precategory_morphisms x' z) e (to_inv f).
+  Lemma transport_source_to_inv {x y z : ob PA} (f : y --> z) (e : y = x) :
+    to_inv (transportf (fun x' : ob PA => precategory_morphisms x' z) e f) =
+    transportf (fun x' : ob PA => precategory_morphisms x' z) e (to_inv f).
   Proof.
     induction e. apply idpath.
   Qed.
 
-  Lemma transportf_to_binop {x y z : ob PA} (f g : x --> y) (e : y = z) :
+  Lemma transport_target_to_binop {x y z : ob PA} (f g : x --> y) (e : y = z) :
     to_binop _ _ (transportf (precategory_morphisms x) e f)
              (transportf (precategory_morphisms x) e g) =
     transportf (precategory_morphisms x) e (to_binop _ _ f g).
@@ -206,10 +206,10 @@ Section transport_morphisms.
     induction e. apply idpath.
   Qed.
 
-  Lemma transportb_to_binop {x y z : ob PA} (f g : y --> z) (e : x = y) :
-    to_binop _ _ (transportb (fun x' : ob PA => precategory_morphisms x' z) e f)
-             (transportb (fun x' : ob PA => precategory_morphisms x' z) e g) =
-    transportb (fun x' : ob PA => precategory_morphisms x' z) e (to_binop _ _ f g).
+  Lemma transport_source_to_binop {x y z : ob PA} (f g : y --> z) (e : y = x) :
+    to_binop _ _ (transportf (fun x' : ob PA => precategory_morphisms x' z) e f)
+             (transportf (fun x' : ob PA => precategory_morphisms x' z) e g) =
+    transportf (fun x' : ob PA => precategory_morphisms x' z) e (to_binop _ _ f g).
   Proof.
     induction e. apply idpath.
   Qed.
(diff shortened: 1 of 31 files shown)


benediktahrens:
Here is a proof of your axiom:
```
Lemma toforallpaths_induction (X Y: UU) (f g : X -> Y) 
  (P : (forall x, f x = g x) -> UU)
  (H : forall e : f = g, P (toforallpaths _ _ _ e)) 
  : forall i, P i.
Proof.
  intros.
  set (XR := homotweqinvweq (weqtoforallpaths _ f g)). cbn in XR. simpl in XR.
  rewrite <- XR.
  apply H.
Defined.

Definition transportf_funextfun : forall {X Y : UU} (P : Y -> UU) (F F' : X -> Y)
           (H : Π (x : X), F x = F' x) (x : X) (f : P (F x)),
  transportf (λ x0 : X → Y, P (x0 x)) (funextsec _ F F' (λ x0 : X, H x0)) f =
  transportf (λ x0 : Y, P x0) (H x) f.
Proof.
  intros X Y P F F' H x f.
  apply (toforallpaths_induction _ _ F F' (fun H' =>  transportf (λ x0 : X → Y, P (x0 x))
    (funextsec (λ _ : X, Y) F F' (λ x0 : X, H' x0)) f =
                                    transportf (λ x0 : Y, P x0) (H' x) f)).
  intro e. clear H.
  set (XR := homotinvweqweq (weqtoforallpaths _ F F') e).
  match goal with [|- transportf ?PP ?HH _ = _ ] =>
        set (H := HH); set (P' := PP) end.
  etrans. 
    eapply (maponpaths (fun X => transportf P' X f)).
    apply XR.  
  destruct e. apply idpath.
Defined.
```

tpannila:
Wow!!! It works! I added your code to the pull request.

UniMath/Foundations/Basics/PartA.v
```diff
@@ -628,6 +628,12 @@ Proof.
   reflexivity.
 Defined.
 
+Definition idpath_transportb {X : UU} (P : X -> UU) {x : X} (p : P x) :
```
benediktahrens: Is this lemma really useful?

mortberg:
It's not optimal to just copy things from Ktheory to CategoryTheory. Would it be possible to only use the version of opposite category from CT in Ktheory and remove the now duplicate definitions from Ktheory?

mortberg:
Apart from the issue with code duplication I think this PR looks very good

tpannila:
- Removed idpath_transportb from PartA.v and replaced all transportb in precategories, Complexes,
  and CohomologyComplex by transportf. Renamed transportb_* to transport_source_* , and
  transportf_* to transport_target_*.
- Moved `toforallpaths_induction` and `transportf_funextfun` to Basics/UnivalenceAxiom.v
- Moved the other transport stuff to precategories
- Can't remove the opposite category stuff in KTheory/Precategories.v that I copied (and changed 
  a little), because if I do, then there is an error in KTheory/Representation.v.

mortberg:
Ok, I experienced similar issues when trying to use the category of elements in Ktheory with what we have in CT for another project. I ended up implementing my own version of it which worked better with the CT package. I also noticed that there is a version of it in the TypeTheory project of Benedikt, Peter and Vladimir. I'm not sure what the best way to proceed with the overlap between CT and Ktheory is, but this PR is not the place to solve it, nor is this the place to discuss it.

UniMath/CategoryTheory/PrecategoriesWithAbgrops.v
```diff
@@ -114,6 +140,30 @@ Section def_precategory_with_abgrops.
     apply (grrcan (to_abgrop x y) h H).
   Qed.
 
+  Lemma to_lrw {x y : PA} (f g h : PA⟦x, y⟧) (e : f = g) : to_binop x y f h = to_binop x y g h.
+  Proof.
+    induction e. apply idpath.
+  Qed.
+
+  Lemma to_rrw {x y : PA} (f g h : PA⟦x, y⟧) (e : g = h) : to_binop x y f g = to_binop x y f h.
```
benediktahrens: In cases like this, it might be more convenient to work with `maponpaths f` directly instead of having a specialized lemma for the given function `f`. In this way, you can avoid a proliferation of trivial lemmas.

tpannila:
- Replaced the paths by `=` in Integers.v
- I would like to keep `transport_target_funextfun` and `transport_source_funextfun` because 
  these follow a naming convention used for transports of morphisms in precategory.v, and are 
  thus easier to remember/find.
- `to_rrw` gives symmetry to `to_lrw`. Thus, I would like to keep it for now.

PR #518, opened 2016-12-14
Equivalence of precategories Set / X and Set ^ X

mweav:
This pull request adds a proof that the precategories Set / X and Set ^ X are equivalent.


UniMath/CategoryTheory/set_slice_fam_equiv.v
```diff
@@ -0,0 +1,225 @@
...
+    is_nat_trans (pr1 ((functor_identity_data _) f))
+                 (pr1 ((functor_composite_data fam_to_slice_data slice_to_fam_data) f))
+                 (fam_unit_fun_fun f) :=
+    is_nat_trans_discrete_precategory X has_homsets_HSET (fam_unit_fun_fun f).
+
+  Definition fam_unit_fun (f : fam X) :
+    (functor_identity_data _) f --> (functor_composite_data fam_to_slice_data slice_to_fam_data) f :=
+    (fam_unit_fun_fun f) ,, (fam_unit_fun_is_nat_trans f).
+
+  Definition fam_unit_is_nat_trans : is_nat_trans _ _ fam_unit_fun.
```
benediktahrens: Your naming convention is a bit confusing: why do you turn the order of the expressions around?
That is, wouldn't `is_nat_trans_fam_unit_fun` be more consistent with the naming in UniMath so far?

UniMath/CategoryTheory/set_slice_fam_equiv.v
```diff
@@ -0,0 +1,225 @@
...
+  Proof.
+    intro x.
+    apply iso_to_slice_precat_iso.
+    simpl.
+    assert (H : fromcoconusf (pr2 x) =
+                (fun h : total2 (fun x' : X =>
+                                   hfiber (@pr2 _ (fun a : hSet => forall _ : a, X) x) x')
+                 => pr1 (pr2 h))).
+    + reflexivity.
+    + rewrite <- H.
```
mortberg: This can be done nicer with the `change ... with ...` tactic

mortberg:
I see that you're using `Defined` everywhere, it's often a good idea to use `Qed` to make things opaque for efficiency reasons. Often when you're proving something that starts with a `is_` (e.g. `is_functor`) it's ok to use Qed as the term is not needed and can be hidden in order to speed up comparisons later on (this way Coq won't unfold these terms when testing conversion). If you have a short subgoal that should be opaque you can use the `abstract` tactic which is equivalent to making a lemma ended with `Qed` for that subgoal.

UniMath/CategoryTheory/set_slice_fam_equiv.v
```diff
@@ -0,0 +1,225 @@
...
+        UniMath.CategoryTheory.category_hset
+        UniMath.CategoryTheory.slicecat
+        UniMath.CategoryTheory.equivalences
+        UniMath.CategoryTheory.DiscretePrecategory
+        UniMath.CategoryTheory.UnicodeNotations.
+
+Section set_slice_fam_equiv.
+
+  Definition hfiber_hSet {X : hSet} {Y : hSet} (f : X → Y) (y : Y) : hSet :=
+    hSetpair (hfiber f y) (isaset_hfiber f y (pr2 X) (pr2 Y)).
```
mortberg: I think this should be moved to Foundations. I have wanted this many times myself and I guess it has many more uses than in CT (for example in algebra).

vladimirias:
I may add that using Defined. Opaque … is sometimes better than the Qed. since then one can open the proof term up with “Transparent.” without re-compiling the whole thing.

benediktahrens:
lgtm

vladimirias:
lgtm and merging.

UniMath/CategoryTheory/slicecat.v
```diff
@@ -431,3 +431,16 @@ Lemma slice_precat_colims {C : precategory} (hsC : has_homsets C) (x : C) (CC :
 Proof.
 now intros g d; apply slice_precat_ColimCocone, CC.
 Defined.
+
+(** morphisms of slice precategories with homsests equal if first projection equal *)
+Lemma slice_precat_morphisms_pr1_eq {C : precategory} (hsC : has_homsets C) (x : ob C)
```
mortberg: This lemma is already formalized and is called `eq_mor_slicecat`. I'll remove this one in an upcoming PR of mine.

PR #521, opened 2016-12-16
update of the README.md

rmatthes:
taking into account that many more vernacular files have been added to this package (before creation of new package on homological algebra by Tomi Pannila)


benediktahrens:
lgtm

Issue #520, opened 2016-12-16
Own package for complexes?


benediktahrens:
I would like to discuss two suggestions concerning @tpannila 's work:

1. Would it make sense to create an own package for that material? The CT package is getting very big and long to compile.
2. Could the file CohomologyComplex.v be split into several files? We sometimes get Travis timeouts due to 10 minutes without output during the compilation of that file. A hacky alternative to that would be to insert a few commands in that file that produce output, such as a `Check` command.

tpannila:
If I add `Unset Kernel Term Sharing.` to CohomologyComplex.v,  before the imports, line 14,
compilation time on my system goes from 4m24.307s to 2m54.998s. 
I can create a new package, HomologicalAlgebra, in my next PR.

benediktahrens:
Both your suggestions sound excellent to me!

benediktahrens:
done

PR #522, opened 2016-12-16
Five lemma and HomologicalAlgebra package

tpannila:
This pull request contains formalization of the 5-lemma and a new package, HomologicalAlgebra, see
#520 . Here is a summary of the changes
- Some new stuff on short exact sequences. Most of these are used to prove the 5-lemma. Some 
  of these will be used later to develop derived functors.
- AbelianPushoutPullback.v contains results on pushouts of monics and pullbacks of epis. These are
  used to prove the 5-lemma.
- PseudoElements.v contains formalization of pseudo elements. These are used to prove the  
  5-lemma. 
- FiveLemma.v contains formalization of the 5-lemma. As a corollary of 5-lemma, there is a version
  of it for short exact sequences.

I moved the following files from CategoryTheory to HomologicalAlgebra
- Complexes.v
- CohomologyComplex.v


benediktahrens:
Tomi, could you please also add your new package to the file .travis.yml?

DanGrayson:
It's good that this will no longer be part of CategoryTheory, since it isn't category theory.

DanGrayson:
`UniMath/README.md` should document that the name of each new package should also be added to `.travis.yml`.

benediktahrens:
@tpannila could you please merge master into your branch, so that the Travis configuration file will be updated in this branch?

PR #526, opened 2016-12-21
Various results on the category of sets and slice categories

mortberg:
This PR includes:

- Direct definition of pullbacks in HSET
- Definition of binary products in C/c when C has pullbacks
- Terminal object in C/c
- Direct proof that HSET is locally cartesian closed
- Definition of base change functor (given c -> c' it gives a functor C/c' -> C/c)
- Definition of the left adjoint to the base change functor (dependent sum functor)
- Proof that if the base change functor has a right adjoint (dependent product functor) then C/c has exponentials
- Make the second argument opaque in the constructors `mk_functor`, `mk_nat_trans` and `mk_are_adjoints`. This has the benefit that we can avoid make separate Qed'd lemmas for subproofs or using large `abstract` blocks when we want the second component opaque when constructing these things.

### DIFF
--- a/UniMath/CategoryTheory/CocontFunctors.v
+++ b/UniMath/CategoryTheory/CocontFunctors.v
@@ -83,6 +83,7 @@ Require Import UniMath.CategoryTheory.AdjunctionHomTypesWeq.
 Require Import UniMath.CategoryTheory.exponentials.
 Require Import UniMath.CategoryTheory.whiskering.
 Require Import UniMath.CategoryTheory.RightKanExtension.
+Require Import UniMath.CategoryTheory.slicecat.
 
 Local Notation "# F" := (functor_on_morphisms F) (at level 3).
 Local Notation "[ C , D , hs ]" := (functor_precategory C D hs).
--- a/UniMath/CategoryTheory/CocontFunctors.v
+++ b/UniMath/CategoryTheory/CocontFunctors.v
@@ -699,7 +699,7 @@ transparent assert (HHH : (cocone cAB (pr1 ab,, x))).
 { use mk_cocone.
   - simpl; intro n; split;
       [ apply (# (pr1_functor A B) (pr1 ccab n)) | apply (pr1 ccx n) ].
-  - abstract (simpl; intros m n e; rewrite (paireta (dmor cAB e)); apply pathsdirprod;
+  - abstract (simpl; intros m n e; rewrite (tppr (dmor cAB e)); apply pathsdirprod;
                 [ apply (maponpaths pr1 (pr2 ccab m n e)) | apply (pr2 ccx m n e) ]).
  }
 destruct (Hccab _ HHH) as [[[x1 x2] p1] p2].
@@ -750,7 +750,7 @@ simpl in *.
 mkpair.
 - apply (tpair _ (f,,g)).
   abstract (intro n; unfold precatbinprodmor, compose; simpl;
-            now rewrite hf1, hg1, (paireta (coconeIn ccxy n))).
+            now rewrite hf1, hg1, (tppr (coconeIn ccxy n))).
 - abstract (intro t; apply subtypeEquality; simpl;
              [ intro x; apply impred; intro; apply isaset_dirprod; [ apply hsC | apply hsD ]
              | induction t as [[f1 f2] p]; simpl in *; apply pathsdirprod;
--- a/UniMath/CategoryTheory/CocontFunctors.v
+++ b/UniMath/CategoryTheory/CocontFunctors.v
@@ -48,6 +48,8 @@ This file also contains proofs that the following functors are (omega-)cocontinu
   [is_cocont_post_composition_functor] [is_omega_cocont_post_composition_functor]
 - Swapping of functor category arguments:
   [is_cocont_functor_cat_swap] [is_omega_cocont_functor_cat_swap]
+- The forgetful functor from Set/X to Set preserves colimits
+  ([preserves_colimit_slicecat_to_cat_HSET])
 
 Written by: Anders Mörtberg and Benedikt Ahrens, 2015-2016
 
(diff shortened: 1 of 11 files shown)


benediktahrens:
Good stuff! lgtm

rmatthes:
Very nice. But it seems that work is duplicated: dependent_sum_functor seems to be the same thing as slicecat_functor.

mortberg:
@rmatthes : Good catch! They are indeed the same, I had totally forgotten about `slicecat_functor` but it looks like I proved some useful lemmas about it. 

I'll merge the two, but I'm not sure which name to use. Do you prefer `slicecat_functor` or `dependent_sum_functor`?

rmatthes:
I prefer the former name. The sums are rather an interpretation.

vladimirias:
I agree that slicecat_functor is better. Its connection with dependent sums is rather complicated.

mortberg:
I did the renaming as you suggested now

mortberg:
I now also added binary and general coproducts in slice categories to this PR

rmatthes:
All is looking fine now - you did a  good merge of the two copies. The coproducts are appreciated as well.

mortberg:
Thanks @rmatthes. 

I'd like to prove that Set/x has general indexed products as well. But I can only see how to do this if I have an explicit inhabitant of the indexing set. Suggestions?

mortberg:
I pushed my proof that Set/x has nonempty indexed products. The requirement of non-emptiness is annoying and it would be nice if it could be removed. 

This is the last result I'll need to finish the multisorted signatures and the consequence of the non-emptiness is that the set of operations is non-empty.

rmatthes:
No suggestion on getting rid of the annoying inhabitant of I for the products. But a question: is the slice category complete? This would follow according to Wikipedia if the constant functor was continuous. It is not co-continuous as you remarked during your formalization of co-continuous functors. How about continuity?

vladimirias:
The empty product is the final object. Set/x certainly has final objects e.g. (x,id_x).

vladimirias:
To get the product for a general family F:I -> (Set/X) what if try to start with X\times (prod_{Sets} F) ?

vladimirias:
> On Dec 22, 2016, at 8:55 AM, Ralph Matthes <notifications@github.com> wrote:
> 
> No suggestion on getting rid of the annoying inhabitant of I for the products. But a question: is the slice category complete? This would follow according to Wikipedia if the constant functor was continuous. It is not co-continuous as you remarked during your formalization of co-continuous functors. 

Why so? It seems to me that colimits exist and the forgetting functor to SET is co-continuous.

rmatthes:
I am sorry that I don't have the details: this is a remark on p.17 l.-8 in the recently submitted paper in the arXiv version at https://arxiv.org/abs/1612.00693
However, that statement is formulated for rather arbitrary categories, not HSET.

vladimirias:
Hi Ralph,

I am not sure what you have replied to.

Vladimir.

rmatthes:
Hi Vladimir, this was meant as a reply to your reply to my remark:
Anders noted in the paper at https://arxiv.org/abs/1612.00693 on p.17 l.-8 that the constant functor is not in general cocontinuous, and I wondered if it was at least continuous because then, the slice categories are complete according to a result mentioned on Wikipedia and maybe easy to transport to UniMath. The "not in general" did not refer to the specific situation of sets, though. Unfortunately, I had not made the effort to understand the problem, and omega-cocontinuity was good enough for the purposes of the paper.

mortberg:
@rmatthes: What's the wikipedia reference you mentioned?

Set/X is complete as it is equivalent to [X,Set] and (co)limits in functor categories are computed pointwise. So one way to get all of these (co)limits would be to use the equivalence formalized be @mweav recently, however it is often easier to use a simple direct definition in Coq which is why I'm trying to formalize them all directly.

@vladimirias : Thanks, I'll try your suggestion for the products. I'll also try to prove that the forgetful functor to Set is cocontinuous as it will be very useful for proving that the "proj" functor from the multisorted signatures is omega-cocontinuous.

rmatthes:
@mortberg: 
- second paragraph under the heading "Limits and universal morphisms" on page https://en.wikipedia.org/wiki/Comma_category
- maybe, for dealing with the projection in multisorted signatures, you could profit from the remark that I added yesterday to the paper draft

mortberg:
I'll look at your comment later, but I proved this morning that `proj` is omega cocont. However my proof is not so nice and I think I see how to do it nicer in case I manage to prove that the forgetful functor from Set/X to Set is omega-cocont.

vladimirias:
A colimit of an empty diagram is an initial object. If the constant functor takes everything to a non-initial object it does not preserve colimits of empty diagrams.

rmatthes:
@vladimiras: thank you, now I seem to recall that Anders said this was the reason

But: now in the case of limits (which suffers from a dual problem), is there a way to get around this corner case when trying to establish completeness of the slice category on basis of that general result mentioned on Wikipedia? The terminal object is the one already provided in UniMath and recalled by Vladimir 3 hours ago.

mortberg:
I didn't think about establishing completeness yet, but I managed to construct Products in Set/x without the annoying non-emptiness criteria. Given a family F : I -> C/X the key was to take:
```
Σ x : X, Π i : I, hfiber (pr2 (F i)) x
```
as the product. I don't think this is what Vladimir suggested, but I actually let Coq compute it for me by applying the equivalence of Matthew to the product in [X,C]. I'm quite impressed that it worked so nicely.

vladimirias:
No, it’s not what I suggested. Still, I am curious if what I suggested works.

UniMath/CategoryTheory/category_hset_structures.v
```diff
@@ -866,3 +864,53 @@ use mk_ProductCone.
...
 
 End products_set_slice.
+
+Lemma is_left_adjoint_slicecat_to_cat_HSET (X : HSET) :
+  is_left_adjoint (slicecat_to_cat has_homsets_HSET X).
+Proof.
+mkpair.
+- use mk_functor.
+  + mkpair.
+    * intro Y. {
```
rmatthes: is there a standard reference for this construction?

rmatthes:
looks still good to me

mortberg:
Thanks, I have one last small result that I want to push. Then I'll be done with this PR and I will have everything needed to finish the proof that the multisorted signature functor is omega-cocont.

UniMath/CategoryTheory/category_hset_structures.v
mortberg: Not sure, but I see now that the sigma type below is simply `X \times Y`

vladimirias:
I am not sure about a reference. The facts are as follows: the slice_cat_functor (this is the name for C/X -> C/Y for X -> Y, right?) is left adjoint to the pullback functor Z/Y -> Z\times_Y X/X if the latter are given, the forgetting functor from C/X to C is the left adjoint to the functor defined by the binary product with X if the latter are given. In each case it is sufficient to give the products on objects, the functor structure can then be constructed.

UniMath/CategoryTheory/CocontFunctors.v
```diff
@@ -1578,6 +1581,93 @@ Defined.
...
+}
+use unique_exists.
+- intros l; apply (pr2 (pr1 (colimArrow CC c cc) l)).
+- simpl; intro n.
+  apply funextsec; intro x; cbn.
+  now etrans; [apply maponpaths,
+                 (toforallpaths _ _ _ (maponpaths pr1 (colimArrowCommutes CC c cc n)) x)|].
+- intros z; apply impred_isaprop; intro n; apply setproperty.
+- simpl; intros f Hf.
+apply funextsec; intro l.
```
benediktahrens: Indentation can be improved starting from here.

UniMath/CategoryTheory/slicecat.v
```diff
@@ -432,15 +440,245 @@ Proof.
...
+Defined.
+
+End slicecat_bincoproducts.
+
+(** * Coproducts in slice categories of categories with coproducts *)
+Section slicecat_coproducts.
+
+Context {C : precategory} (hsC : has_homsets C) (I : UU) (BC : Coproducts I C).
+
+Local Notation "C / X" := (slice_precat C X hsC).
```
benediktahrens: Probably `slice_precat` should be defined as a map on `Precategory` from the beginning. But that is for later...

vladimirias:
Merging.

Issue #525, opened 2016-12-17
Elements and ElementsOp


vladimirias:
should be in CategoryTheory and free from using Ktheory.Utilities

mortberg:
I'll do this now

PR #527, opened 2016-12-21
a replacement for apevalat

vladimirias:
This PR is a step towards resolving #525. I have created eqtohomot function that replaces apevalat and placed it to an appropriate position in BasicsA.v 

I replaced all the uses of apevalat in Elements and ElementsOp with eqtohomot. This closed the dependency of these two files on Ktheory.Utilities. They still depend on Ktheory.Precategories which should be corrected before they can be moved to CategoryTheory.


vladimirias:
Note: the only relevant file changes are in BasicsA, Elements and ElementsOp. The rest is due to sub/coq issues that Dan can comment on.

benediktahrens:
This setting might be useful for the future:
https://www.ivankristianto.com/git-ignore-file-permission/

Concerning the current PR, it would be great to get rid of the change of permissions.
On https://www.builtinbruges.com/2013/03/git-tricks-permission-reset/ there is an explanation for how to get a list of permission changes between two commits, and to revert those permission changes.
Could you please try the following:
```
git  diff -p 1421286 8f63153 \
    | grep -E '^(diff|old mode|new mode)' \
    | sed -e 's/^old/NEW/;s/^new/old/;s/^NEW/new/' \
    | git apply
```
What it does is checking for difference in file permissions (option `-p`) and reverting them by exchanging "old" and "new".

benediktahrens:
There is a slightly shorter solution not requiring sed:
```
git diff -p -R --no-color 1421286 8f63153 \
    | grep -E "^(diff|(old|new) mode)" --color=never  \
    | git apply
```
found on https://stackoverflow.com/questions/2517339/how-to-recover-the-file-permissions-to-what-git-thinks-the-file-should-be

vladimirias:
Closing because of the Benedikt's modification.

PR #528, opened 2016-12-21
Copy of PR #527 without change of file permissions

benediktahrens:
This PR is essentially the same as #527 , but with a few extra commits to revert the change of file permissions.
I suggest merging the present one, and discarding #527.


mortberg:
LGTM

PR #529, opened 2016-12-23
Multisorted binding signatures

mortberg:
This PR contains:

- Definition of multisorted binding signatures
- Construction of an endofunctor on [Set/X,Set/X] from a multisorted binding signature
- Proof that this functor is omega cocontinuous

This PR relies on: https://github.com/UniMath/UniMath/pull/526

In the end of the file there is a failed attempt att constructing an endofunctor on [[X,Set],[X,Set]] instead. This code is currently in a comment as it contains some admits.


mortberg:
As this PR relies on another PR the diff is huge. The only new file is UniMath/SubstitutionSystems/MultiSorted.v

rmatthes:
I see that the project is advancing at a fast pace. I think that, currently, the precise definition of the option functor does not yet matter (nothing is proven about the obtained syntax). But are you sure about option_fun? It is not the definition from the paper draft in that it checks the f component of Xf for mapping to the given sort s. This is reminiscent of Vladimir's suggestion for solving the problem with option in the first approach in the draft. To my mind, one should just take X+1 as the new set and sumofmaps (pr2 Xf) (termfun s) as the function. To validate this ultimately, the example of the lambda calculus would have to be implemented.

vladimirias:
Option functor should not change the type of sorts. We have important examples of two sorted binding signatures. The only reasonable way that I see is that there should be many option functors, one for each T -> X (where X is the type of sorts).

rmatthes:
@vladimirias: The set sort is currently plainly a parameter (a variable at the beginning of the section that contains all the development in Multisorted.v), and my concern was how option(s) has to calculate the set component of the object in Set/sort from the given Xf which represents (X,f:X->sort). To my mind, it should just be 1+X and not even consider f. Of course, the function component of the object will then extend f with constantly s.

vladimirias:
I am not sure what you mean. The set component of the option functor associated with a given (X,f:X -> sort) should be  \lambda Y, Y+X, right? It makes sense that if X consists of more than one element then the associated option corresponds to the fixation (distinguishing) of several variables.

rmatthes:
In the current formalization, option_functor(s) is supposed to add one single variable of sort s. Then, its extension to lists of sorts, option_list(xs) adds one variable for each element of the list xs of sorts. We did not consider adding a given set of names Y of variables that all have a given sort s.

vladimirias:
I see. Now the first comment

> I see that the project is advancing at a fast pace. I think that, currently, the precise definition of the option functor does not yet matter (nothing is proven about the obtained syntax). But are you sure about option_fun? It is not the definition from the paper draft in that it checks the f component of Xf for mapping to the given sort s. This is reminiscent of Vladimir's suggestion for solving the problem with option in the first approach in the draft. To my mind, one should just take X+1 as the new set and sumofmaps (pr2 Xf) (termfun s) as the function. To validate this ultimately, the example of the lambda calculus would have to be implemented.

makes sense. 

The more general version seems to be a very natural generalization that does not add any complexity to the proofs. It can be called generalized_option_functor and the option_functor to be defined as generalized_option_functor (unit,, termfun s) .

mortberg:
Indeed, I spent this week exclusively on this project which made the formalization advance at a high pace. I really wanted to get the code I wrote in October finished and without admit's before going on Christmas vacation, and I finally managed to do this today. It might be that the definition of the functor has to be adapted a little when we develop an example, but I hope most of the necessary results should be in place for me to do this easily if necessary. 

As I wrote in the other comment I don't have time to program anymore until after the holidays, so feel free to merge this now or wait until I have formalized the generalized option functor and maybe adapted the definition a bit.

UniMath/SubstitutionSystems/MultiSorted.v
```diff
@@ -0,0 +1,682 @@
...
+
+Let SET : Precategory := (HSET,,has_homsets_HSET).
+
+Local Definition SET_over_sort : Precategory.
+Proof.
+exists (SET / sort).
+now apply has_homsets_slice_precat.
+Defined.
+
+Local Lemma has_homsets_Csort : has_homsets SET_over_sort.
```
benediktahrens: Is it ever useful to have this lemma? Why not simply use `homset_property`?

UniMath/SubstitutionSystems/MultiSorted.v
```diff
@@ -0,0 +1,682 @@
...
+- exists (proj_fun s).
+  intros X Y f p.
+  exists (pr1 f (pr1 p)).
+  abstract (now destruct f as [h hh]; destruct p as [x hx]; simpl in *; rewrite <- hx, hh).
+- abstract (split; [intros X|intros X Y Z f g];
+            apply funextsec; intro p; apply subtypeEquality; trivial;
+            intros x; apply setproperty).
+Defined.
+
+Definition hat_functor (t : sort) : functor SET (SET / sort).
```
benediktahrens: Since the name is so non-descriptive, a comment explaining what this functor does would be great.

UniMath/SubstitutionSystems/MultiSorted.v
```diff
@@ -0,0 +1,682 @@
...
+(** option_functor for lists (also called option in the note) *)
+Definition option_list (xs : list sort) : functor (SET / sort) (SET / sort).
+Proof.
+use (foldr _ _ xs).
++ intros s F.
+  apply (functor_composite (option_functor s) F).
++ apply functor_identity.
+Defined.
+
+(** This is X^a as a functor between functor categories *)
```
benediktahrens: In this comment would be useful to explain what `X` and `a` are.

UniMath/SubstitutionSystems/MultiSorted.v
```diff
@@ -0,0 +1,682 @@
...
+  (H : Colims_of_shape nat_graph SET_over_sort) :
+  is_omega_cocont (exp_functor a).
+Proof.
+apply is_omega_cocont_functor_composite.
++ apply functor_category_has_homsets.
++ apply is_omega_cocont_pre_composition_functor, H.
++ apply is_omega_cocont_post_comp_proj.
+Defined.
+
+Lemma is_omega_cocont_exp_functors (xs : list (list sort × sort))
```
benediktahrens: Can this lemma be split into several lemmas in a useful way, i.e. such that each lemma is interesting/useful by itself?

UniMath/SubstitutionSystems/MultiSorted.v
```diff
@@ -0,0 +1,682 @@
...
+Lemma exp_functor (a : list sort × sort) : functor [SET_over_sort,SET_over_sort] [SET_over_sort,SET].
+Proof.
+eapply functor_composite.
+- apply (pre_composition_functor _ _ _ has_homsets_Csort _ (option_list (pr1 a))).
+- apply post_comp, (proj_functor (pr2 a)).
+Defined.
+
+(* This defines X^as where as is a list. Outputs a product of
+   functors if the list is nonempty and otherwise the constant functor. *)
+Definition exp_functors (xs : list (list sort × sort)) :
```
benediktahrens: Why is the name of this definition in plural?

UniMath/SubstitutionSystems/MultiSorted.v
```diff
@@ -0,0 +1,682 @@
...
+(* If the list is empty we output the constant functor *)
+set (T := constant_functor [SET_over_sort,SET_over_sort] [SET_over_sort,SET]
+                           (constant_functor SET_over_sort HSET TerminalHSET)).
+(* TODO: Maybe use indexed finite products instead of a fold? *)
+use (foldr1 (fun F G => BinProduct_of_functors _ _ _ F G) T XS).
+apply BinProducts_functor_precat.
+apply BinProductsHSET.
+Defined.
+
+Definition hat_exp_functors (xst : list (list sort × sort) × sort) :
```
benediktahrens: Same question here: why plural?

UniMath/CategoryTheory/CocontFunctors.v
```diff
@@ -1578,6 +1581,93 @@ Defined.
...
+Defined.
+
+Lemma is_omega_cocont_slicecat_to_cat (X : HSET) :
+  is_omega_cocont (slicecat_to_cat has_homsets_HSET X).
+Proof.
+intros d L cc.
+now apply preserves_colimit_slicecat_to_cat_HSET.
+Defined.
+
+(** Direct proof that the forgetful functor Set/X to Set preserves colimits *)
```
benediktahrens: Is there a reason why you have both a direct proof of preservation of colimits as well as a proof via an adjunction?

mortberg: I first implemented the direct proof and then realized that I could even construct an adjunction. I figured that it would be worth keeping the direct proof if we want to experiment with computations. Even if it works to compute with `cbn` or `lazy` it could be fun to experiment with `cbv` and the various `compute` tactics which could be more efficient. As far as I remember we had more luck when computing with call by value with direct proofs.

UniMath/SubstitutionSystems/MultiSorted.v
mortberg: Nope, I'll remove it.

DanGrayson:
If the travis check fails again, you should try merging with the current master branch -- I think you have something strange going on with the `lablgtk` package, which is no longer included in UniMath as a submodule.  It's evident only when building coqide, which you could also try at home.

DanGrayson:
PS: you have to merge with master, anyway, because Benedikt and are trying out that as a general requirement for pull requests.

UniMath/SubstitutionSystems/MultiSorted.v
mortberg: I added a comment saying what this is now.

UniMath/SubstitutionSystems/MultiSorted.v
mortberg: This comment was from the other approach, I changed it to something about the definition.

mortberg:
That seems like a reasonble requirement, I'll do it tomorrow when I have addressed all of Benedikt's comments.

UniMath/SubstitutionSystems/MultiSorted.v
mortberg: I went for plural because it takes a list, but maybe `exp_functor_list` would be better...

UniMath/SubstitutionSystems/MultiSorted.v
mortberg: Same answer as above.

UniMath/SubstitutionSystems/MultiSorted.v
mortberg: I don't think so, but I can shorten it a little. I think many of the definitions and lemmas in this file should be made local. A potential user would probably only want to see the final map from binding sigs to a functor and the proof that this functor is omega-cocontinuous. However if I make them local I might not be able to refer to them in the summary file? I can experiment a little with this when I make the summary file once this PR as been merged.

DanGrayson: In the summary file, whatever that is, perhaps you can just qualify the local names, e.g., use `Sets.setquotuniv2_iscomprelfun` instead of `setquotuniv2_iscomprelfun`.

benediktahrens:
On 01/09/2017 09:49 AM, Anders Mörtberg wrote:
> 
> 
> 
>> +  - now intros X; apply (eq_mor_slicecat has_homsets_HSET).
> +  - intros X; apply funextsec; intro x. +    now apply
> subtypeEquality; trivial; intros x'; apply setproperty. +Defined. + 
> +Lemma is_omega_cocont_exp_functor (a : list sort × sort) +  (H :
> Colims_of_shape nat_graph SET_over_sort) : +  is_omega_cocont
> (exp_functor a). +Proof. +apply is_omega_cocont_functor_composite. ++
> apply functor_category_has_homsets. ++ apply
> is_omega_cocont_pre_composition_functor, H. ++ apply
> is_omega_cocont_post_comp_proj. +Defined. + +Lemma
> is_omega_cocont_exp_functors (xs : list (list sort × sort))
> 
> I don't think so, but I can shorten it a little. I think many of the
> definitions and lemmas in this file should be made local. A potential
> user would probably only want to see the final map from binding sigs
> to a functor and the proof that this functor is omega-cocontinuous.
> However if I make them local I might not be able to refer to them in
> the summary file? 

Sure you can, but you'll have to use the fully qualified name.

UniMath/SubstitutionSystems/MultiSorted.v
mortberg: Ok, thanks Dan and Benedikt for quick answers. I'll make them local.

mortberg:
I've now rebased on master and addressed Benedikt's comments, so feel free to merge in case you think it looks ok. 

I uncommented the partis of the earlier attempt to work with  [[X,Set],[X,Set]] which don't have any admits and put this in a separate Module in the end of the file to avoid name clashes. This way this code won't break when UniMath develops which could be good if we ever need to use it.

UniMath/SubstitutionSystems/MultiSorted.v
```diff
@@ -0,0 +1,640 @@
...
+  - apply (pr2 Xf t).
+  - apply s.
+Defined.
+
+Local Definition option_functor_data (s : sort) : functor_data (SET / sort) (SET / sort).
+Proof.
+exists (option_fun s).
+intros X Y f.
+mkpair.
+- intros F; induction F as [t h]; simpl.
```
benediktahrens: The (old) rule of not destructing pairs in the definition of structure should still be observed, until we get eta for Sigmas.

mortberg: Fixed

DanGrayson: Also after we get eta.

PR #530, opened 2016-12-31
simplify some proofs

DanGrayson:
I've simplified the proofs of these recent lemmas and made them not
depend any longer on funextemptyAxiom.

 Lemma hzbooleqisi (i : hz) : hzbooleq i (i + 1) = false.
 Lemma hzbooleqissi (i : hz) : hzbooleq i (i + 1 + 1) = false.
 Lemma hzeqpii {i : hz} : i - 1 != i.

### DIFF
--- a/UniMath/Foundations/Basics/Sets.v
+++ b/UniMath/Foundations/Basics/Sets.v
@@ -1144,12 +1144,50 @@ Defined.
 Notation " 'ct' ( R , is , x , y ) " := (ctlong R is x y (idpath true))
                                           (at level 70).
 
-(* An alternative to [ct], with tactics and negations *)
+(*
+
+  Tactics for computing with decidable relations
+
+  A tactic alternative to [ct], with negation
+
+  If [R x y] is a decidable relation on a type [X], then, by definition, we have a proof of [d : R x
+  y ⨿ ¬ R x y].  If [x] and [y] are constants (definable in an empty context), then simplification
+  of [d] will yield either a term of the form [inl p] or [inr q], depending on whether [R x y] is
+  true or false.  Let us suppose the answer turns out to be true.  Then if our current proof goal is
+  [R x y], then we can automate the proof by first converting [R] to a binary relation [S], so that
+  [S x y] simplifies to [true].  Then we know that a proof of [S x y = true] is [idpath true], and
+  we can convert that into a proof of [R x y] using a proof of the implication [S x y = true → R x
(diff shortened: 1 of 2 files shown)


mortberg:
LGTM

UniMath/Foundations/NumberSystems/Integers.v
```diff
@@ -957,22 +957,17 @@ Qed.
...
 Proof.
-  intros i.
-  unfold hzbooleq, decreltobrel. cbn.
-  induction (isdecrelhzeq i (i + 1)) as [e | n].
-  - apply fromempty. apply (ct (hzneq, isdecrelhzneq, 1, 0)).
-    apply (hzpluslcan 1 0 i (! (pathscomp0 (hzplusr0 i) e))).
-  - apply idpath.
+  intros i. apply negrtopaths.
+  apply (negf (λ e, hzpluslcan _ _ _ (! (hzplusr0 i @ e)))); clear i.
+  confirm_not_equal isdecrelhzeq.
```
mortberg: This seems like a useful tactic, but I can't understand what it is doing by just looking at its definition. Maybe a comment explain what the "confirm_" tactics in Sets.v do could be useful?

UniMath/Foundations/Basics/Sets.v
```diff
@@ -1144,24 +1144,63 @@ Defined.
...
+
+  The tactics [confirm_equal] and [confirm_not_equal] make this strategy concise when the goal is of
+  the form [x = y] or [x != y] and we have a proof [i] that equality is decidable.
+
+  (It would be redundant to also have a proof that inequality is decidable.)
+
+  To evaluate whether having a tactic simplifies things, compare the proofs of [hzbooleqisi] and
+  [hzbooleqisi'] in Integers.v.
+
+ *)
```
mortberg: Thanks, this is very cool and useful! I think some of my proofs can be simplified using this.

vladimirias: 1. Instead of "we have a proof of [d : R x +  y ⨿ ¬ R x y]. " it's better to say "we have a proof [d] of [R x +  y ⨿ ¬ R x y].

2. "(It would be redundant to also have a proof that inequality is decidable.)" - what does it mean?

DanGrayson: I mean that decidability of P is equivalent to decidability of ¬ P, and that only one of the two is ever needed, in practice.

PR #532, opened 2017-01-02
remove submodule lablgtk

### LINKED ISSUE #383: Submodule `lablgtk` still necessary?
As per #115 , the submodule was necessary to compile CoqIDE on MacOS because of some incompleteness of a homebrew recipe.
Is that still the case? If not, could we remove that stuff?

DanGrayson:
Resolves #383.

It turns out that there is a homebrew "tap" called "mht208/formal"
that contains a recipe for building "ocamlfind", which is now required
by the makefile that builds "coq".  Here we take advantage of that to
remove the submodule "lablgtk", saving some time.

Note that "lablgtk" is used only if we build coqide.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,10 +2,6 @@
 	path = sub/coq
 	url = https://github.com/coq/coq.git
 	branch = v8.5
-[submodule "sub/lablgtk"]
-	path = sub/lablgtk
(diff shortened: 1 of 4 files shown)


DanGrayson:
I see how to fix this for travis, which is running Ubuntu, and will do so.

PR #533, opened 2017-01-03
effective epis, epis as pushouts

amblafont:


UniMath/CategoryTheory/Epis.v
```diff
@@ -166,3 +172,66 @@ Section epis_functorcategories.
...
+
+(*
+Proof that f: A -> B is an epi is the same as saying that the diagram
+A ---> B
+|      |
+|      |  id         is a pushout
+‌v     ‌‌ v
+B----> B
+  id
+*)
```
mortberg: This comment won't show up in the documentation, to make it show open it with `(**` instead. If you do this the diagram won't render nicely, to make it do so put `<<` before and `>>` after. See [this](https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/CocontFunctors.v#L208) for an example.

UniMath/CategoryTheory/Epis.v
```diff
@@ -166,3 +172,66 @@ Section epis_functorcategories.
...
+  Lemma epi_to_pushout : isEpi f -> isPushout f f (identity _) (identity _) (idpath _).
+  Proof.
+    intro h.
+    red.
+    intros x p1 p2 eqx.
+    assert (hp : p1 = p2).
+    { now apply h. }
+    destruct hp.
+    apply (unique_exists p1).
+    rewrite id_left.
```
mortberg: This proof would look nicer with some bullets here.

UniMath/CategoryTheory/Epis.v
```diff
@@ -166,3 +172,66 @@ Section epis_functorcategories.
...
+  Context {C:Precategory} {A B:C} (f:C⟦A,B ⟧).
+
+  Lemma epi_to_pushout : isEpi f -> isPushout f f (identity _) (identity _) (idpath _).
+  Proof.
+    intro h.
+    red.
+    intros x p1 p2 eqx.
+    assert (hp : p1 = p2).
+    { now apply h. }
+    destruct hp.
```
mortberg: I think it would be better UniMath style to use `induction` rather than `destruct` here

UniMath/CategoryTheory/Epis.v
mortberg: Oops, I mean after applying `(unique_exists p1)`

UniMath/CategoryTheory/Epis.v
```diff
@@ -166,3 +172,66 @@ Section epis_functorcategories.
...
+    apply hf in hp.
+    destruct hp as [[p [hp1 hp2]] _].
+    now rewrite <- hp1,hp2.
+  Qed.
+
+End EpiPushoutId.
+
+
+
+(* Definition of an effective epimorphism.
```
mortberg: Please use `(**` here as well

mortberg:
Building this PR fails due to circular imports. The files coequalizers.v and pushouts.v already imports Epis.v. I'm not sure what the best way to solve this is.

amblafont:
Thanks for the feedback. I also just commited some changes in the file Basics/Sets.v. Should I put them in another pull request ?

amblafont:
I was also wondering whether I should move the definition of kernel pairs elsewhere

mortberg:
I think PRs ideally should contain results that are related to make them easier to review, but it looks like all of these results are related (they are all about epis and surjections) so I think it's fine. 

Maybe the definition of kernel pairs can go into pullbacks.v? @benediktahrens: what do you think?

amblafont:
All the new results in Epis.v are related to limits, so I could move them to a new file limits/epis.v and there would be no circularity. What do you think of that?

mortberg:
I'm not a huge fan of adding another file with the same name (even if it's in another directory), there are already a lot of files in the CT package...

Another alternative would be to move the code on epis from coequalizers.v and pushouts.v to Epis.v so that those files don't depend on Epis.v. As far as I can tell the results that would be moved are pushouts of epis are epis and that the coequalizer arrow is an epi. It seems fine to me to have these results in epis.v instead, but maybe @tpannila has some opinion on this as he wrote those proofs?

tpannila:
I suggest to move the section `EpiPushtoutId` to limits/pushouts.v and create a file
limits/EffectiveEpis.v which would contain the section `EffectiveEpi` and the definition
`HasEffectiveEpis`.

amblafont:
I have done as tpannila suggested.

UniMath/CategoryTheory/limits/EffectiveEpis.v
```diff
@@ -0,0 +1,31 @@
...
+  
+  Definition kernel_pair := Pullback  f f.
+
+  Definition isEffective :=
+    Σ  g:kernel_pair,
+         (isCoequalizer (PullbackPr1 g)
+                        (PullbackPr2 g) f (PullbackSqrCommutes g)).
+End EffectiveEpi.
+
+Definition HasEffectiveEpis (C:precategory) :=
```
benediktahrens: Wouldn't it be better to call this type `EpisAreEffective` instead of `HasEffectiveEpis`?

UniMath/Foundations/Basics/Sets.v
```diff
@@ -1346,7 +1348,143 @@ Proof.
...
+    apply weqtopathshProp.
+    apply logeqweq.
+    + apply hinhuniv.
+      intros [y eqx].
+      rewrite eqx.
+      apply (hfiberpr2 _ _ y).
+    + intro eqx.
+      apply hinhpr.
+      use tpair;[now exists b|exact eqx].
+Qed.
```
benediktahrens: It might be worth thinking about a proof that does not need univalence for propositions. I don't know if such a proof exists. Maybe you can add a comment (a TODO) to that end.

UniMath/Foundations/Basics/Sets.v
```diff
@@ -1346,7 +1348,143 @@ Proof.
...
+       destruct y2 as [ [z2 h2] h2' ].
+       rewrite <- h1' ,<-h2'.
+       apply comp_f_epi;simpl.
+       rewrite h1,h2.
+       apply idpath.    
+    - apply prtoimage. apply H.
+  Defined.
+
+  Definition univ_surj : B -> C :=
+    (fun b => (pr1 (pr1 (comp_f_epi_hprop b)))).
```
benediktahrens: More parentheses than necessary.

UniMath/CategoryTheory/category_hset.v
```diff
@@ -37,12 +37,14 @@ Local Notation "# F" := (functor_on_morphisms F) (at level 3).
 (** * Precategory of hSets *)
 Section HSET_precategory.
 
+(* moved to Basics/Sets.v
```
benediktahrens: This comment should be deleted.

mortberg:
LGTM

PR #535, opened 2017-01-06
specific notion of equalities between diagrams

amblafont:
This customized notion of equality between diagrams is useful to make the link between standard 
diagrams (pushouts, coequalizers, ...) in a functor category and the induced pointwise diagram given an object of the source category

### DIFF
--- a/UniMath/CategoryTheory/limits/graphs/eqdiag.v
+++ b/UniMath/CategoryTheory/limits/graphs/eqdiag.v
@@ -79,7 +79,7 @@ Lemma transport_swap: Π {X Y : UU} (P : X -> Y → UU) {x x':X} {y  y' : Y}
                         (e : x = x') (e' : y = y') (p : P x y),
                   transportf (fun a => P _ a) e' (transportf (fun a => P a _) e p) =
                   transportf (fun a => P a _) e (transportf (fun a => P _ a) e' p) .
-Proof
+Proof.
   intros.
   induction e.
   induction e'.
--- a/UniMath/CategoryTheory/limits/graphs/eqdiag.v
+++ b/UniMath/CategoryTheory/limits/graphs/eqdiag.v
@@ -81,8 +81,8 @@ Lemma transport_swap: Π {X Y : UU} (P : X -> Y → UU) {x x':X} {y  y' : Y}
                   transportf (fun a => P a _) e (transportf (fun a => P _ a) e' p) .
 Proof
   intros.
-  destruct e.
-  destruct e'.
+  induction e.
+  induction e'.
   apply idpath.
 Qed.
 
@@ -98,7 +98,7 @@ Lemma transportf_transpose {X : UU} {P : X → UU}
   {x x' : X} (e : x = x') (y : P x) (y' : P x')
 : transportb P e y' = y -> y' = transportf P e y.
 Proof.
-  intro H; destruct e; exact H.
+  intro H; induction e; exact H.
 Defined.
 
 
--- a/UniMath/CategoryTheory/limits/graphs/eqdiag.v
+++ b/UniMath/CategoryTheory/limits/graphs/eqdiag.v
@@ -1,0 +1,393 @@
+(**
+
+- Custom notion of equality between diagrams (eq_diag) over the same graph
+- Transports of cones and cocones between equal diagrams.
+- Limits/Colimits are the same for equal diagrams.
+
+This notion of equality is useful to make the link between standard 
+diagrams (pushouts, coequalizers, ...) in a functor category and
+the induced pointwise diagram given an object of the source category
+
+Example of the binary product :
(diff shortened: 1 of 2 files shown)


DanGrayson:
@amblafont -- I'd be happy to learn who you are.

vladimirias:
Hi! What is "symmetry" and what is "etrans"?

amblafont:
My name is Ambroise Lafont, I am a phD student in the CoqHott team. I am currently working on syntax and semantics with André Hirschowitz and Benedikt helps me formalize some proof.

etrans is a tactic that Benedikt told me about that uses the transitivity of equality to transform a goal a = b into a goal a= ?x and ?x = b where ?x is a fresh existential variable.

symmetry is a tactic that transform a goal a = b into a goal b = a. Maybe there is a more canonical way to do this in UniMath ?

mortberg:
Instead of `symmetry` I usually use `apply pathsinv0`

DanGrayson:
@amblafont - Dear Ambroise, how about adding your name and portrait to your github profile?  Then we can remember who you are.  I also have my web site URL and my email address on mine.

vladimirias:
Instead of etrans one can use “simple refine (pathscomp0 _ _)” (correct me if I am wrong in the number of the underscores). 

We can make our own hsymm and htrans tactics to type less.

I am worried about etrans symm because they don’t know about pathscomp0 and pathsinv0  so what do they use instead?

DanGrayson:
`etrans` does know about `pathscomp0`.  Its definition is `Ltac etrans := eapply pathscomp0.`.

Here's what `symmetry` does.  I tried using it in this trivial example:
```
Lemma A : Π X (x y:X), x=y → y=x.
  intros.
  symmetry.
  assumption.
Defined.
```
Then `Print A` yields
```
A = 
λ (X : Type) (x y : X) (X0 : x = y) (H:=X0),
(λ X1 : x = y, match X1 in (_ = y0) return (y0 = x) with
               | idpath _ => idpath x
               end) H
     : Π (X : Type) (x y : X), x = y → y = x
```
So it uses `match`, whereas `pathsinv0` uses `paths_rect`.

Our current UniMath coding style rules in the file UniMath/README.md includes advice designed to reduce the dependence on `match`.

DanGrayson:
I notice the new file has lines ending with spaces.  This might cause confusion if anyone uses emacs to touch your files, as our file `.dir-locals.el` makes those spaces go away automatically.  The confusion would come from seeing a new pull request where lots of differences are due only to white space changes.

amblafont:
Thanks for the feedback. Should I replace etrans with Vladimir's 'refine' then ?

mortberg:
No, `etrans` is a UniMath tactic defined by me (as suggested by Benedikt) in Basics/PartA.v and it's just `eapply pathscomp0` so it should be ok to use?

I like Vladimir's suggestion to define our own `symmetry` tactic. Would it be possible to override the Coq one by defining our own in Foundations so that someone who already knows Coq don't use the wrong one? I don't know Ltac very well, but maybe we can simply write `Ltac symmetry := apply pathsinv0.` which will then override the Coq one? How about doing the same for other tactics like `reflexivity` and `transitivity`?

DanGrayson:
`reflexivity` uses `idpath`, so there is no reason to replace it.  However, `transitivity` uses `match` -- we could replace it by renaming our `intermediate_path` to `transitivity`.  Here is a test to show that that overriding works:
```
Ltac transitivity x := apply (pathscomp0 (b := x)).

Lemma A : Π X (x y z:X), x=y -> y=z -> x=z.
  intros ? ? ? ? p q.
  transitivity y; assumption.
Defined.

Print A.
```
The output is
```
transitivity is defined
A is defined
A = 
λ (X : Type) (x y z : X) (p : x = y) (q : y = z), p @ q
     : Π (X : Type) (x y z : X), x = y → y = z → x = z
```

DanGrayson:
I suggest replacement of symmetry and transitivity be done in a future pull request, since many other files will get edited.

UniMath/CategoryTheory/limits/graphs/eqdiag.v
```diff
@@ -0,0 +1,390 @@
...
+Proof.
+  exact(pr2 (pr1 H)).
+Qed.
+
+
+Lemma transport_swap: Π {X Y : UU} (P : X -> Y → UU) {x x':X} {y  y' : Y}
+                        (e : x = x') (e' : y = y') (p : P x y),
+                  transportf (fun a => P _ a) e' (transportf (fun a => P a _) e p) =
+                  transportf (fun a => P a _) e (transportf (fun a => P _ a) e' p) .
+  intros.
```
benediktahrens: There is `Proof` missing here, which is used as a keyword to hide proofs in the enhanced documentation.

PR #536, opened 2017-01-09
Define new lattices using weq and abmonoidfrac

cathlelay:


DanGrayson:
Could you merge with the current master branch?

DanGrayson:
Most of the code in `Section so_pty` is repeated in Lattice and in Sets.  There should be just one copy.

DanGrayson:
The proof of this lemma
```
Lemma isStrongOrder_weq {X Y : UU} (H : weq Y X) (gt : hrel X) :
  isStrongOrder gt → isStrongOrder (λ x y : Y, gt (H x) (H y)).
```
does not use that H is an equivalence, so rephrase to
```
Lemma isStrongOrder_weq {X Y : UU} (H : Y -> X) (gt : hrel X) :
  isStrongOrder gt → isStrongOrder (λ x y : Y, gt (H x) (H y)).
```

DanGrayson:
To go along with existing convention, you may wish to rename `latticeop` to `islatticeop`.

DanGrayson:
Similarly, we might rename `latticewithgtrel` to `islatticewithgtrel`.

DanGrayson:
Except for my minor remarks above, it looks good to me.

vladimirias:
I see no need for isStrongOrder_weq and StrongOrder_weq. The coercion weq >-> Funclass should apply automatically when you use isStrongOrder_fun and StrongOrder_fun with H in place of H. I would also use letters _bck instead of _fun as the constructions "pull" a strong order backwards along a function rather than define what a "strong order function"  is.

vladimirias:
ln 112 is too long. It does not fit to the screen in my GitHub preview (I don't see how many symbols it is).

vladimirias:
Is StrongOrder_abgrdiff really a particular case of StrongOrder_abgrdiff?

vladimirias:
Sorry, for a typo. Is StrongOrder_abgrdiff really a particular case of StrongOrder_abmonoidfrac?

vladimirias:
It seems to me that in StrongOrder_correct_commrngfrac the logical equivalence can be strengthened to an equality in hrel. Indeed, both sides are obtained using abmonoidfracrel with the same arguments except for the last one, Hgt in one case and something obtained using ispartbinopcommrngfracgt in the other. This argument belongs to is.... type that must be an hprop (it is either proved somewhere or easy to prove). Therefore any two elements of it are equal and so are the objects obtained by applying a function to these two elements.

vladimirias:
ln. 676. The type of is0 is equivalent to the type called isinvbinophrel. See BinaryOperations.v

vladimirias:
ln. 975-981 why not to use one line is : lattice X. Then your access functions will give you all of the parts named in 975-981 when you need them?

vladimirias:
On ln. 983-1022 you have some stuff that should be in BinaryOperations. On ln. 72 there introduce a small comment section ****  Transfer of binary operations relative to weak equivalences  **** and define two functions

binop_weq_fwd .. ( w : weq X Y ) : binop X -> binop Y.

binop_weq_bck .. ( w : weq X Y ) : binop Y -> bionop X

using the formula for weq_min, weq_max in the second case and an obvious analog in the first. Then use them in lattice.

vladimirias:
Next, there should be two lemmas in BinaryOperations corresponding to your isassoc_weq_min for *each* property of a binary operation introduced there and *each* property of two binary operations. For example for isassoc there should be isassoc_weq_frw and isassoc_weq_bck stating (isassoc opp -> isassoc (bionop_weq_fwd opp) etc.

This is a lot of typing, but a worthy thing to do.

vladimirias:
Next, I suggest to move the definition of isabsorb opp1 opp2 to BinaryOperations where pairs of operations are considered. It should be provided with all the usual constructions and properties that are given for pairs of operations in BinaryOperations, including the new ones isabsorb_weq_fwd and isabsorb_weq_bck.

vladimirias:
BTW I think, that binop_weq_fwd and bin_op_weq_bck are better defined with w : weq and invweq w exchanged places in the definition as opposed to one being defined as another applied to invweq.

vladimirias:
What I suggest will allow you to remove Section lattice_weq and prove islatticeop_weq and lattice_weq directly using the new constructions that you will add to BinaryOperations.

DanGrayson:
The operation `(λ x y : Y, gt (f x) (f y))` on an hrel gt could be given a name an located in Sets.v where hrel is defined.  The name could be something like `hrelcomp`.

vladimirias:
Maybe fun_hrel_comp f gt (with underscores or as one word). This will show that it is the composition of a function with an hrel.

cathlelay:
It is not exactly the case because `abgrdiff M` is logically
equivalent to `abmonoidfrac
M (totalsubmonoid M)` but not equal.
I could use `StrongOrder_abmonoidfrac` and `StrongOrder_weq`, as I will do
to define lattice on abgrdiff, but here the proof is really small.

2017-01-29 15:06 GMT-05:00 Vladimir Voevodsky <notifications@github.com>:

cathlelay:
Because all lemmas in this section only depend on the corresponding
property for min and max.
They, they can be reused on min_weq and max_weq even if we don't have a
lattice.

2017-01-29 15:34 GMT-05:00 Vladimir Voevodsky <notifications@github.com>:

vladimirias:
I see, that makes sense.

vladimirias:
Why are access functions used in the proof of isStrongOrder_setquot but not in the proof of isStrongOrder_bck ?

vladimirias:
They (access functions) are also not used in the other proofs. I understand that they are much longer than pr1/pr2 but the change can be made now throughout the file and for a reader it is better to see the access function than pr1/pr2 even though it is longer.

vladimirias:
The line-breaks are chaotic. Please make them after : or := whenever possible as determined by the rule that the lines must not be longer that 100 characters. It will make it easier to see at a glance what has been done.

vladimirias:
Lemmas about the strong order are above the definition of a lattice in the file lattice.v This can't be right. Could you please move it to another file?

vladimirias:
It might be convenient to introduce a notation for 

hProppair (x = y) (setproperty X _ _)))

It should probably be done in hSet.v and all in UniMath should be informed.

DanGrayson:
There is such a notation already:

```
Notation "a = b" := (eqset a b) (at level 70, no associativity) : set.
```

It's in scope `set`.

I was wondering recently whether it should be x==y instead of x=y, so as to conflict less.

DanGrayson:
I am also experimenting with notation for decidable equality.

vladimirias:
There is a general definition isrcancelable for what is equivalent to isinvbinophrel for the relation =. I would prefer this definition to be used with the isinvbinophrel, if needed, derived as a corollary (that may be placed in BinaryOperations).

vladimirias:
It is preferable, I think, to group the properties that only concern the abmonoid operation with this operation and to have the lattice structure and its properties grouped together as well. On the other hand I can understand the approach where the structures are grouped together and the properties together. What is the common opinion?

vladimirias:
Can you have the lattice structure named something other than "is" since "is" is usually used for elements of types that start with "is"? The lattice structure can be called "lat".

DanGrayson:
I recommend eliminating `mkStrongOrder`:
```
diff --git a/UniMath/Algebra/Lattice.v b/UniMath/Algebra/Lattice.v
index 33c5d0b2..4b812045 100644
--- a/UniMath/Algebra/Lattice.v
+++ b/UniMath/Algebra/Lattice.v
@@ -44,9 +44,6 @@ Unset Automatic Introduction.
 
 Definition isStrongOrder {X : UU} (R : hrel X) : UU :=
   istrans R × iscotrans R × isirrefl R.
-Definition mkStrongOrder {X : UU} (R : hrel X)
-           (Htrans : istrans R) (Hcotrans : iscotrans R) (Hirrefl : isirrefl R) :
-  isStrongOrder R := Htrans,,Hcotrans,,Hirrefl.
 Definition StrongOrder (X : UU) := ∑ R : hrel X, isStrongOrder R.
 Definition pairStrongOrder {X : UU} (R : hrel X) (is : isStrongOrder R) : StrongOrder X :=
   R,,is.
@@ -75,7 +72,7 @@ Lemma isStrongOrder_bck {X Y : UU} (f : Y → X) (gt : hrel X) :
   isStrongOrder gt → isStrongOrder (fun_hrel_comp f gt).
 Proof.
   intros X Y H gt is.
-  apply mkStrongOrder.
+  repeat split.
   - intros x y z.
     apply (istrans_StrongOrder (_,,is)).
   - intros x y z.
```

DanGrayson:
The frequency of constructions like `istrans_StrongOrder (_,,H)` suggests that `istrans_StrongOrder` should be redefined to accept `H` directly.  Similarly for the others.
```
--- a/UniMath/Algebra/Lattice.v
+++ b/UniMath/Algebra/Lattice.v
@@ -56,14 +56,15 @@ Coercion  pr1StrongOrder : StrongOrder >-> hrel.
 Section so_pty.
 
 Context {X : UU}.
-Context (R : StrongOrder X).
+Context {R : hrel X}.
+Context (is : isStrongOrder R).
 
 Definition istrans_StrongOrder : istrans R :=
-  pr1 (pr2 R).
+  pr1 is.
 Definition iscotrans_StrongOrder : iscotrans R :=
-  pr1 (pr2 (pr2 R)).
+  pr1 (pr2 is).
 Definition isirrefl_StrongOrder : isirrefl R :=
-  pr2 (pr2 (pr2 R)).
+  pr2 (pr2 is).
 Definition isasymm_StrongOrder : isasymm R :=
   istransandirrefltoasymm
     istrans_StrongOrder
@@ -77,11 +78,11 @@ Proof.
   intros X Y H gt is.
   apply mkStrongOrder.
   - intros x y z.
-    apply (istrans_StrongOrder (_,,is)).
+    apply (istrans_StrongOrder is).
   - intros x y z.
-    apply (iscotrans_StrongOrder (_,,is)).
+    apply (iscotrans_StrongOrder is).
   - intros x.
-    apply (isirrefl_StrongOrder (_,,is)).
+    apply (isirrefl_StrongOrder is).
 Qed.
 Definition StrongOrder_bck {X Y : UU} (f : Y → X) (gt : StrongOrder X) : StrongOrder Y :=
   (fun_hrel_comp f gt) ,, isStrongOrder_bck f _ (pr2 gt).
```

DanGrayson:
The purpose of
```
  { apply isapropdirprod ;
    apply impred_isaprop ; intros a ;
    apply impred_isaprop ; intros b ;
    apply impred_isaprop ; intros c ;
    apply isapropimpl, isapropimpl, (pr2 (gt _ _)). }
```
is to prove `isaprop (ispartbinophrel Y gt)`.  That should be abstract to a lemma and put just after the definition of `ispartbinophrel` so others can use it.

Is the general feeling that making such dinky proofs is preferable to packaging up the properties as hProps in the first place?  For example, using `hconj` instead of `dirprod`, and so on.

DanGrayson:
I second Vladimir's suggestion about parameters called `is`, as here:
```
Context {X : hSet}
        (is : lattice X).
```

DanGrayson:
Using this tactic can make things simpler:
```
--- a/UniMath/Algebra/Lattice.v
+++ b/UniMath/Algebra/Lattice.v
@@ -182,7 +182,7 @@ Lemma Lmin_id :
   ∏ x : X, Lmin is x x = x.
 Proof.
   intros x.
-  apply (pathscomp0 (b := Lmin is x (Lmax is x (Lmin is x x)))).
+  intermediate_path (Lmin is x (Lmax is x (Lmin is x x))).
   - apply maponpaths, pathsinv0, Lmax_absorb.
   - apply Lmin_absorb.
 Qed.
```

DanGrayson:
Here's a minor suggestion:
```
--- a/UniMath/Algebra/Lattice.v
+++ b/UniMath/Algebra/Lattice.v
@@ -230,7 +230,7 @@ Qed.
 Lemma isPartialOrder_Lle :
   isPartialOrder (Lle is).
 Proof.
-  split ; [ split | ].
+  repeat split.
   - exact istrans_Lle.
   - exact isrefl_Lle.
   - exact isantisymm_Lle.
```

DanGrayson:
Other than all that, it seems fine.

vladimirias:
Why? It looks like the usual thing that we do when we introduce an important type that is a total2 or an iterated total2.

DanGrayson:
Re: "Why? It looks like the usual thing that we do when we introduce an important type that is a total2 or an iterated total2."

Because it's trivial to do without it, rendering proofs easier to read.  (One less identifier to look up.)

DanGrayson:
By the way, my minor suggestion above about "repeat split" contradict my recently accepted proposal that should avoid "repeat".

DanGrayson:
@cathlelay , this PR of yours seems to be on hold waiting for a response from you.

DanGrayson:
Hi, Catherine.  Were there many conflicts to deal with?  I have five files in conflict on a branch of mine, too.

UniMath/Foundations/Sets.v
```diff
@@ -511,6 +514,9 @@ Definition iscotrans {X : UU} (R : hrel X) : UU
 Definition isdeccotrans {X : UU} (R : hrel X) : UU
   := ∏ x1 x2 x3, R x1 x3 -> R x1 x2 ⨿ R x2 x3.
 
+Definition isStrongOrder {X : UU} (R : hrel X) : UU :=
```
DanGrayson: Is this definition a standard one from the mathematical literature?

DanGrayson:
The only error from Travis is that there are changes to Foundations.  We may consider them, though.

DanGrayson:
Let us know when this pull request is in good shape and ready to be considered again.

cathlelay:
This pull request is ready to be considered again.
Changes in Foundations are the one you suggested.

DanGrayson:
I wonder about "strong order", and whether that is new terminology, or is to be found in the mathematical literature.

cathlelay:
When I formalized this order, we couldn't find a better name. StrictOrder is perhaps closer to the mathematical literature

cathlelay:
Can we close this PR? I can change the name of StrongOrder in a future PR

benediktahrens:
On 12/19/2017 12:40 PM, cathlelay wrote:
> Can we close this PR? I can change the name of StrongOrder in a future PR
> 

Do you mean "close" or "merge"?

DanGrayson:
I think Catherine meant "merge".

Since this definition doesn't seem to occur in the mathematical literature, there's no clear way to find a better name, so it may not make sense to assign ourselves the impossible task of changing it later.  (A strict order is something else.)

I'd be satisfied if a comment would be added near the definition of isStrongOrder, to the effect that "strong order" is a newly invented term, and saying why it's worthwhile to invent it and to place it in Foundations.  Provide also your primary example of a strong order.

peterlefanulumsdaine:
I’ve updated this old PR to build again, and addressed the main outstanding comments from reviews.  Its material looks to me still good and not obsoleted by anything more recent.

benediktahrens (approved):
lgtm

PR #538, opened 2017-01-11
remove lemma apevalat

DanGrayson:
In pull request #527 Vladimir introduced the function `eqtohomot` to replace
`apevalat` partially.  In this pull request I eliminate `apevalat` in favor of
`eqhomotop`, generalizing `eqtohomot` from functions to sections to make that
possible.  I also moved the definition of `toforallpaths` to PartA, to point
out that the only difference between it and `eqtohomot` is the number of
implicit arguments.

Probably we should eliminate `toforallpaths` in favor of `eqtohomot` later on.

### DIFF
--- a/UniMath/Foundations/Basics/PartA.v
+++ b/UniMath/Foundations/Basics/PartA.v
@@ -504,9 +504,14 @@ Definition homotfun {X Y Z : UU} {f f' : X -> Y} (h : f ~ f')
 
 (** *** Equality between functions defines a homotopy *)
 
-Definition eqtohomot {X Y : UU}{f f' : X -> Y}(e : f = f') : homot f f' :=
-  fun x => maponpaths ( fun f => f x ) e.
+Definition toforallpaths {T:UU} (P:T->UU) (f g:Π t:T, P t) : f = g -> f ~ g.
+Proof. intros ? ? ? ? h t. induction h.  apply (idpath _). Defined.
 
+Definition eqtohomot     {T:UU} {P:T->UU} {f g:Π t:T, P t} : f = g -> f ~ g.
+(* the same as toforallpaths, but with different implicit arguments *)
+Proof.
+  intros ? ? ? ? e t. induction e. reflexivity.
+Defined.
 
 (** *** [ maponpaths ] for a function homotopic to the identity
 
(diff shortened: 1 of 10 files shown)


UniMath/Foundations/Basics/PartA.v
```diff
@@ -504,9 +504,14 @@ Definition homotfun {X Y Z : UU} {f f' : X -> Y} (h : f ~ f')
 
 (** *** Equality between functions defines a homotopy *)
 
-Definition eqtohomot {X Y : UU}{f f' : X -> Y}(e : f = f') : homot f f' :=
-  fun x => maponpaths ( fun f => f x ) e.
+Definition toforallpaths {T:UU} (P:T->UU) (f g:Π t:T, P t) : f = g -> f ~ g.
+Proof. intros ? ? ? ? h t. induction h.  apply (idpath _). Defined.
 
+Definition eqtohomot     {T:UU} {P:T->UU} {f g:Π t:T, P t} : f = g -> f ~ g.
```
benediktahrens: Wouldn't it be better to use `toforallpaths` in the definition of `eqtohomot`?

DanGrayson: I suppose, but the next thing I'd like to do is to eliminate `toforallpaths`, so it's moot.

mortberg:
Does `eqhomotop` have the maximal number of arguments implicit?

I always get frustrated when using `toforallpaths` and have to remember how many underscores to write (I can of course ask Coq, but it's annoying to always have to do this every time I can't remember). Maybe we should discuss having some rules for implicit arguments in UniMath? I think it is nice to always be able to predict which arguments are implicit when using a function. I'm personally a fan of having the maximal number of arguments implicit (and there is a way to get Coq do this automatically for us). This way one has to write the minimal number of underscores and if some of the implicit arguments cannot be inferred by Coq one can always use `@` and insert it manually.

DanGrayson:
Yes, it has the maximum number of arguments implicit.

Another way to give an implicit argument a value, instead of using `@`, is to type something like `(x := a)`.

Turning on automatic maximal implicit arguments might be a good idea.  Same for turning on automatic introduction.

Issue #350, opened 2016-04-18
Foundations should contain foundations and nothing else


benediktahrens:
Following up on a conversation with Vladimir, I suggest to move out of the directory Foundations everything that is not in Basics. The contents of Basics can be moved one directory up, since Basics really means Foundations.
Besides the clarification this modification brings, it also brings more fine-grained dependencies for package-wise compilation.

DanGrayson:
Sounds good to me.

What about moving "Tactics" to "Padics/Tactics" or somewhere?  It's not a mathematical topic, but its siblings are.

vladimirias:
Did not see this message when I wrote mine :)

DanGrayson:
I'll prepare a pull request for this.  Lots of files will be renamed, so others may wish not
to rename files in preparation for their pull requests.

DanGrayson:
I'm working on separating Foundations into four top level packages, and it
seems to make sense for them to be linearly ordered as

   Basics < NumberSystems < Combinatorics < Algebra

and for none of the files within a package to require files from a later
package.

One obstruction to doing this currently is that NumberSystems/NaturalNumbers
requires Algebra/Domains_and_Fields and Algebra/Archimedean; it might be the
only obstruction.  That dependency involves results in
NumberSystems/NaturalNumbers about "nat" and its order operations both as a rig
and as an archimedean rig.  Probably it would be safe to postpone such high
level treatments of "nat" until later.

Let me know if anyone disagrees.

benediktahrens:
On 01/11/2017 02:14 PM, Daniel R. Grayson wrote:
> I'm working on separating Foundations into four top level packages, and it
> seems to make sense for them to be linearly ordered as
> 
>    Basics < NumberSystems < Combinatorics < Algebra
> 
> and for none of the files within a package to require files from a later
> package.
> 
> One obstruction to doing this currently is that NumberSystems/NaturalNumbers
> requires Algebra/Domains_and_Fields and Algebra/Archimedean; it might be the
> only obstruction.  That dependency involves results in
> NumberSystems/NaturalNumbers about "nat" and its order operations both as a rig
> and as an archimedean rig.  Probably it would be safe to postpone such high
> level treatments of "nat" until later.
> 
> Let me know if anyone disagrees.

I was annoyed recently by the dependency of NaturalNumbers on the files
you mention, because it means that a lot of files need to be compiled
before one can do anything with natural numbers.
In my case, I wanted to do something with signatures, and could not
[...]

PR #539, opened 2017-01-12
Reorganize foundations, part 1

DanGrayson:
Here we set the stage for reorganization of Foundations as envisioned in issue #350 

  Basics < Combinatorics < Algebra < NumberSystems

I.e., no file should require one from a subdirectory occurring later than its
own on this list.

To accomplish this we moved the basic material about natural numbers
not depending on Algebra into Basics and put the other material into
a new file called NumberSystems/NaturalNumbersAlgebra.v.


DanGrayson:
https://github.com/UniMath/UniMath/pull/539/commits/7aac82ff4377d029b2f106516fd768836d6f5849 enforces linear ordering of proof files.

benediktahrens:
@DanGrayson : Oh no, a conflict...

DanGrayson:
Okay.  Guess what: the new file eqdiag.v depends on something in Ktheory, so I'll move some lemmas to fix that.

benediktahrens:
How did you notice that? Has your new make target made you aware of that?

DanGrayson:
Yes, when it said that eqdiag.v should not depend on Ktheory/Utilities.

benediktahrens:
On 01/13/2017 05:47 PM, Daniel R. Grayson wrote:
> Yes, when it said that eqdiag.v should not depend on Ktheory/Utilities.

Can one see that somewhere?

DanGrayson:
It was on just my screen, but you can try it: put back the Require that used to be in eqdiag.v and type make on the current master branch.

DanGrayson:
I will have the next pull request soon.  Testing ...

PR #540, opened 2017-01-14
Reorganize foundations, part 2

### LINKED ISSUE #350: Foundations should contain foundations and nothing else

DanGrayson:
resolves #350 

Files are moved, components of qualified names are modified to reflect the new locations.

No proofs are changed.


benediktahrens:
It seems like the Travis file has not been updated?

PR #545, opened 2017-01-15
Construction of functions defined at a few points on types with decid…

DanGrayson:
…able equality

This seems to be useful notation, for example, in the section on homological algebra
when making chain complexes with specified objects in one or two locations.

Lemma decfun1 {X Y:UU} (i : isdeceq X) (x1:X) (y1 y':Y) : X → Y.
Lemma decfun2 {X Y:UU} (i : isdeceq X) (x1 x2:X) (y1 y2 y':Y) : X → Y.
Lemma decfun3 {X Y:UU} (i : isdeceq X) (x1 x2 x3:X) (y1 y2 y3 y':Y) : X → Y.


vladimirias:
I think you can use that x1, x2 etc. are isolated points instead. 

If I am not mistaken, one can do it for isolated points and then introduce a notation that will do the isdeceq case since if, for example, isolfun x1 x2 i1 i2 is the lemma where i1 i2 are proofs that x1 x2 are isolated points and the you have a proof i that the type has decidable equality, then 

decfun2 i x1 x2 =“isolfun x1 x2 (i x1) (i x2)”

will give decfun2 with exactly the same behavior as your decfun2.

There are many types with isolated points but without decidable equality such as the types coprod X unit that are used a lot in substitution systems.

DanGrayson:
Good idea.

PR #546, opened 2017-01-16
Pointwise epi

amblafont:
Proof that
- a natural transformation that is a pointwise effective epimorphism is an effective epimorphism
- if the target category has pushouts, then an epimorphic natural transformation is pointwise epimorphic

The last proof is not related to the notion of effective epimorphisms but I don't know in which file the proof should go. It cannot go in limits/pushouts.v because it uses limits/graphs/pushouts.v which depends on limits/pushouts.v


benediktahrens:
Can the last proof go into CategoryTheory.Epi ? It is (kind of) a converse to an existing result which says that a nat. trans that is pointwise epi is epi.

amblafont:
I don't think so. I would need to import limits/pushouts.v which depends on CategoryTheory.Epi. I had the same circularity in the dependencies in my previous pull request because I was trying to put the definition of epimorphism as a pushout diagram in CategoryTheory.Epi (https://github.com/UniMath/UniMath/pull/533)

benediktahrens:
How about renaming the file EffectiveEpis into something more general, that subsumes the second result?

amblafont:
I would rename into Epi.v as I suggested once, but having two files named Epi could be confusing ?

benediktahrens:
On 01/16/2017 07:45 PM, Ambroise wrote:
> I would rename into Epi.v as I suggested once, but having two files
> named Epi could be confusing ?

Yes, that is confusing.
You can call it EpiFacts or something similar.
Also:

- in file Epis, you can insert a comment saying that more stuff on epis
can be found in file EpiFacts (or whatever name you choose)
- there is no need for the file EpiFacts to be in the subdirectory
limits. Not all the files that import limits need to be in that subdir.

vladimirias:
A suggestion. Denote by ## F , for F a presheaf of sets, the following expression (F : _ -> hSet). Then one can write ## F X and a : ## F X without this very annoying pr1’s.

DanGrayson:
spelling: "wich as a kernel pair"

DanGrayson:
"transfo nat" could be more understandable as "natural transformation"

DanGrayson:
Modulo my two comments, it looks good to me.

DanGrayson:
Re: https://github.com/UniMath/UniMath/pull/546/commits/405857a3a7bbe56d10701e585712738a9e3a6479

Hmm?  I meant, "transfo nat" would be more understandable if it were replace by "natural transformation", because the latter expression uses the proper English words.

amblafont:
Oops! I understood the opposite although I was not really convinced :)

DanGrayson:
Well, I said it in a confusing way...

UniMath/CategoryTheory/EpiFacts.v
```diff
@@ -0,0 +1,188 @@
...
+      (diagram_pointwise (homset_property D)
+                         (pushout_diagram CD a b) x).
+  Proof.
+    use tpair.
+    use StandardFiniteSets.three_rec_dep;  apply idpath.
+    use StandardFiniteSets.three_rec_dep;  use StandardFiniteSets.three_rec_dep;
+      exact (Empty_set_rect _ )||exact (fun _ => idpath _).
+  Defined.
+
+  Import graphs.pushouts.
```
benediktahrens: I suggest you remove this import, and instead replace `Pushouts D` by `graphs.pushouts.Pushouts D` in the statement of the Lemma `Pushouts_pw_epi`.

PR #549, opened 2017-01-19
rename some identifiers that are needlessly plural

DanGrayson:
subabmonoids -> subabmonoid
    submonoids -> submonoid
    hsubtypes -> hsubtype
    subgrs -> subgr
    subabgrs -> subabgr
    subrigs -> subrig
    subrngs -> subrng

Notice that for types that are not implemented as subtypes, the identifiers are
not plural, so there is an inconsistency.  These types include gr, monoid, rng,
hSet, hProp, rig, setwithbinop, and subset.

### DIFF
--- a/UniMath/Algebra/Archimedean.v
+++ b/UniMath/Algebra/Archimedean.v
@@ -752,7 +752,7 @@ Proof.
     exact Hc2.*)
 Defined.
 
-Lemma natmult_commrngfrac {X : commrng} {S : subabmonoids} :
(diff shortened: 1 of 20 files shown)


mortberg:
Sounds good to me. Maybe it would be good to have a page on the wiki describing naming conventions? (However, last time we discussed this I don't think we reached any conclusions...)

DanGrayson:
Sounds good to me.  A good way to work toward an agreement would be to continue the discussion at https://github.com/UniMath/UniMath/issues/176 .

PR #550, opened 2017-01-21
Summary: improve INSTALL (bash under MAC OS)

DanGrayson:
Now that the Makefile uses a feature of recent versions of bash, Mac OS X users
may have to install a recent version of bash, so here we modify the install
instructions.  The feature is associative arrays, declared with `declare -A`.

Here is homebrew's bash:
```
$ /usr/local/bin/bash --version
GNU bash, version 4.4.5(1)-release (x86_64-apple-darwin16.3.0)
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
```
Here is the Mac OS X bash:
```
This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

$ /bin/bash --version
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin16)
Copyright (C) 2007 Free Software Foundation, Inc.
```

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,12 +1,13 @@
 Installation of UniMath
 =======================
 
-Prepare for installation by installing the OCAML compiler on your system.
-Under Mac OS X, the most convenient way to do that is with "Homebrew",
-available from http://brew.sh/, with the following command:
+Prepare for installation by installing the OCAML compiler and a more modern
+version of `bash` on your system.  Under Mac OS X, the most convenient way to
+do that is with "Homebrew", available from http://brew.sh/, with the following
+command:
 
 ```bash
-$ brew install objective-caml camlp5 camlp4 lablgtk
+$ brew install objective-caml camlp5 camlp4 lablgtk bash
 ```
 
 Also install "ocamlfind" using "homebrew" with the following commands.
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,7 @@ COQBIN ?=
 COQIDE_OPTION ?= no
 ifeq "$(BUILD_COQ)" "yes"
 COQBIN=sub/coq/bin/
+all: check-first
 all: build-coq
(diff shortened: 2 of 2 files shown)


mortberg:
I'm not a MAC user, but this looks good to me

peterlefanulumsdaine:
I am a Mac user, and can confirm that this is good.

In more detail: the “enforce-linear-order” check broke the build on my system, with the message
```
bash: line 0: declare: -A: invalid option
declare: usage: declare [-afFirtx] [-p] [name[=value] ...]
make: *** [enforce-linear-ordering] Error 2
```
and installing a newer bash with `brew install bash` fixed it.

(Including the gory details so that other users with the same problem can hopefully find this thread.)

benediktahrens:
How about the makefile skips the check if a too old bash is detected?
If I were a Mac user, I'd be pretty upset about having to get a new bash just to compile UniMath with 
```
$ make
```

peterlefanulumsdaine:
Actually, I second Benedikt’s suggestion as even better.  It was a bit frustrating having my build broken by this, given that it’s not actually needed for compiling the library.  :-)

DanGrayson:
Okay, will do.  Travis has a recent bash, so at least Travis will check this, and developers should be told to install a recent bash, so their pull requests will pass the test.

DanGrayson:
From Peter's link: "Apple will not update Bash, because the latest version is licensed under GPLv3, which Apple cannot use. "

So that explains it, sigh!  Maybe I'll go back to gnu-linux for my next laptop.

DanGrayson:
Okay, that should do it.  Good thing I looked at it -- I found some other errors.  It wasn't doing the enforcement early enough, and it could be waylaid by a too-old version of "sed".

PR #551, opened 2017-01-21
reformat two proofs for easier reading

DanGrayson:


UniMath/Combinatorics/FiniteSequences.v
```diff
@@ -42,67 +42,39 @@ Definition seq_key_eq_lemma {X :UU}( g g' : seq X)(e_len : seq_len g = seq_len g
...
-    rewrite <- e_int1 in e.
-    rewrite <- e_int2 in e.
-
-    apply e. }
-
-  apply (seq_key_eq_lemma _ _ e_len e_el).
+  intros ? ? ? k r.
+  apply seq_key_eq_lemma.
+  * assumption.
+  * intros.
```
mortberg: Are the variables introduced here used later in the proof?

DanGrayson: Yes, all 5.  For example, "assumption" refers to k.

DanGrayson: Yes, they are part of the statement of the goal.

mortberg: I was referring to the last `intros` where the variables are not named. There are some variables later in the script that I don't see introduced anywhere, so I was thinking that they might have been automatically named by this `intros` which would go against the style guide.

DanGrayson: That part of the style guide refers to names invented by Coq, whereas these names are explicit in the statement we're proving, so they're nearby.

mortberg: Ok, I still prefer always naming things explicitly. What if you would change the proof at some earlier point so that you have a hypothesis called `i` in the context (or add a section variable called `i`)? If you do this the `intros` will call the first thing `i0` and your proof might break later on if you use `i` for something. By always writing the names explicitly Coq will report an error when trying to do the `intros` in case there is something already called `i` in context. This way the proof will be more robust and easier to maintain.

DanGrayson: The whole proof is only 8 lines long, so you can immediately see what went wrong, because you changed it 4 seconds earlier.  It's a self contained, locally scoped proof, so there won't be a problem.  The real problem is when Coq has changed, and you have something like `induction p3`, and you have no idea which of the 10 things in the context it might have referred to before.

mortberg: Sure, this example is quite simple and the probability that it breaks is rather small. I just want to emphasize that it is good coding style to always name things you are going to use later explicitly when using `intros`. This way the proofs become more maintable (for the reason I mentioned above) and easier to read (you see directly how many things the `intros` introduce and what they are called).

DanGrayson: Oh, well, I uniformly adopt the opposite coding style, naming just the intro variables I have to name.  That way the names appear just once.

DanGrayson: But you might have a point.  I'll pay attention in the future and apply your advice if it makes sense.

mortberg: Ok, sounds reasonable to me.

mortberg:
LGTM

PR #553, opened 2017-01-23
Rmodules complements

amblafont:
- monad Precategory
- (faithful) forgetful functor from monads to endofunctors
- tautological right modules
- pullbacks of right modules along a monad morphism

### DIFF
--- a/UniMath/CategoryTheory/Monads.v
+++ b/UniMath/CategoryTheory/Monads.v
@@ -3,7 +3,9 @@
 Contents:
 
         - Definition of monads ([Monad])
-        - Precategory of monads [precategory_Monad C] on [C]
+        - Precategory of monads [Precategory_Monad C] on [C]
+        - Forgetful functor [forgetfunctor_Monad] from monads
+             to endofunctors on [C]
         - Haskell style bind operation ([bind])
         - A substitution operator for monads ([monadSubst])
 
@@ -210,6 +212,23 @@ Definition Precategory_Monad (C:Precategory) : Precategory :=
   (precategory_Monad C (homset_property C) ,, has_homsets_Monad C ).
 
 
+Definition forgetfunctor_Monad (C:Precategory) :
+  functor (Precategory_Monad C) (functor_Precategory C C).
+Proof.
+  use mk_functor.
+  use mk_functor_data.
+  - exact (fun M => pr1 M:functor C C).
+  - exact (fun M N f => pr1 f).
+  - abstract (split; red; intros;  reflexivity).
+Defined.
+
+Lemma forgetMonad_faithful C : faithful (forgetfunctor_Monad C).
+Proof.
+  intros M N.
+  apply isinclpr1.
+  apply isaprop_Monad_Mor_laws.
+  apply homset_property.
+Qed.
 
 End Monad_precategory.
 
--- a/UniMath/CategoryTheory/Monads.v
+++ b/UniMath/CategoryTheory/Monads.v
@@ -216,9 +216,9 @@ Definition forgetfunctor_Monad (C:Precategory) :
   functor (Precategory_Monad C) (functor_Precategory C C).
 Proof.
   use mk_functor.
-  use mk_functor_data.
-  - exact (fun M => pr1 M:functor C C).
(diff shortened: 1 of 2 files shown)


UniMath/CategoryTheory/RModules.v
```diff
@@ -166,9 +172,138 @@ Proof.
...
-  := tpair _ _ (precategory_RModule_axioms C hs).
+Definition precategory_RModule (C : Precategory) : precategory
+  := tpair _ _ (precategory_RModule_axioms C (homset_property C)).
+
+Lemma has_homsets_RModule (C:Precategory) :
+  has_homsets (precategory_RModule C).
+Proof.
+  intros F G.
+  simpl.
+  apply isaset_total2 .
```
mortberg: Please use bullets after this

UniMath/CategoryTheory/RModules.v
```diff
@@ -166,9 +172,138 @@ Proof.
...
+right modules over M
+
+If T is a module over M', we call m* T the pullback module of T along m
+*)
+Section Pullback_module.
+
+  Lemma cancel_functor_on_morph (C C' : precategory_ob_mor)
+        (F : functor_data C C') (a b : C) (m m': C ⟦ a, b ⟧) : m = m' -> #F m = #F m'.
+  Proof.
+    now induction 1.
```
mortberg: I think it is better practice to name the hypothesis and then do induction on the name instead. In this case I would do:

`intros e; induction e.` 

Or directly put e among the hypothesis of the lemma in order to avoid having to introduce it.

UniMath/CategoryTheory/RModules.v
```diff
@@ -166,9 +172,138 @@ Proof.
...
+    tpair _ (T:functor B C) pb_RModule_σ.
+
+  Lemma pb_RModule_laws : RModule_laws M pb_RModule_data.
+  Proof.
+    split.
+    - simpl.
+      assert (hT1:= RModule_law2 _ (T:=T)).
+      assert (hT2:= RModule_law1 _ (T:=T)).
+      intro c.
+      rewrite <- hT2; clear hT2.
```
mortberg: Can't you just do `rewrite <- (RModule_law1 _ (T:=T))` instead of asserting it and then clearing it?

UniMath/CategoryTheory/Monads.v
```diff
@@ -191,6 +193,43 @@ Qed.
...
+
+Definition forgetfunctor_Monad (C:Precategory) :
+  functor (Precategory_Monad C) (functor_Precategory C C).
+Proof.
+  use mk_functor.
+  use mk_functor_data.
+  - exact (fun M => pr1 M:functor C C).
+  - exact (fun M N f => pr1 f).
+  - abstract (split; red; intros;  reflexivity).
+Defined.
```
mortberg: Isn't the structure of the bullets a bit weird in this proof? I would imagine something like this instead:

```
use mk_functor.
- use mk_functor_data.
  + exact (fun M => pr1 M:functor C C).
  + exact (fun M N f => pr1 f).
- abstract (split; red; intros; reflexivity).
```

mortberg:
LGTM (modulo minor comments)

amblafont:
Thanks for the feedback

UniMath/CategoryTheory/RModules.v
```diff
@@ -166,9 +172,135 @@ Proof.
...
+(** Let m : M -> M' a monad morphism.
+
+m induces a functor m* between the category of right modules over M' and the category of
+right modules over M
+
+If T is a module over M', we call m* T the pullback module of T along m
+*)
+Section Pullback_module.
+
+  Lemma cancel_functor_on_morph (C C' : precategory_ob_mor)
```
benediktahrens: This is just `maponpaths`. What do you need this lemma for?

DanGrayson:
lgtm

UniMath/CategoryTheory/RModules.v
amblafont: this explains why there is no such lemma :)

PR #557, opened 2017-01-24
In this commit I propose adding advice to the style guide that names …

DanGrayson:
…of assert

variables that are referred to should be named explicitly, instead of depending
on Coq to invent a name.

I searched for such assertions and fixed them.

The idea to do this was prompted by the discussion in #551.


benediktahrens:
Nice!

PR #559, opened 2017-01-25
simplify a few proofs

DanGrayson:


vladimirias:
LGTM (What is "now apply?")

DanGrayson:
"now X" means apply tactic "X" and then try "trivial" to finish the current goal.

PR #561, opened 2017-01-27
replace Π with ∏ and Σ with ∑

### LINKED ISSUE #560: Π versus ∏ and ∑ versus Σ
I've just noticed that the Greek letter Π (\Pi) and the symbol for product ∏
(\prod) in Unicode are different.  We should switch, since we're using the
wrong one.  Similarly for sum.

DanGrayson:
because we should use the unicode math symbols, not the unicode Greek letters.

This resolves #560.

### DIFF
--- a/UniMath/Algebra/Apartness.v
+++ b/UniMath/Algebra/Apartness.v
@@ -53,11 +53,11 @@ Proof.
   apply isapropiscotrans.
 Qed.
 
-Definition aprel (X : UU) := Σ ap : hrel X, isaprel ap.
+Definition aprel (X : UU) := ∑ ap : hrel X, isaprel ap.
 Definition aprel_pr1 {X : UU} (ap : aprel X) : hrel X := pr1 ap.
 Coercion aprel_pr1 : aprel >-> hrel.
 
-Definition apSet := Σ X : hSet, aprel X.
+Definition apSet := ∑ X : hSet, aprel X.
 Definition apSet_pr1 (X : apSet) : hSet := pr1 X.
 Coercion apSet_pr1 : apSet >-> hSet.
 Arguments apSet_pr1 X: simpl never.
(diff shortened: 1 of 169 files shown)


DanGrayson:
This should be the next one to be pulled, since it will conflict with almost any new code that anyone writes in the meantime.  Too bad I didn't notice these symbols at the time of our original introduction of unicode.  

Comments?

benediktahrens:
Will it make other PRs conflict?

Am 29. Januar 2017 17:50:06 MEZ schrieb "Daniel R. Grayson" <notifications@github.com>:
>This should be the next one to be pulled, since it will conflict with
>almost any new code that anyone writes in the meantime.  Too bad I
>didn't notice these symbols at the time of our original introduction of
>unicode.  
>
>Comments?

DanGrayson:
Actually, it will be the test that fails: any PR that introduces one of the Greek letters will have to have it changed to the symbols.

But there will always be other PRs, no matter when we do this.

DanGrayson:
Fixing a conflict is harder than fixing a failing test.  If another PR fails the test because it introduced the Greek letters, a single emacs command fixes it.  If this PR gets a conflict with the master branch, then I have to go through and remove all the conflict markers and repair them manually, and there might be dozens of them.

Of course, another PR could also get a conflict by modifying a line near a line with one of the Greek letters.

DanGrayson:
On the other hand, I could always prepare a fresh version of the pull request by starting from the master branch, without having to resolve conflicts.  But when would be a good time to do that?

DanGrayson:
Comments, anyone?  This one seems completely straightforward.

vladimirias:
No special comments, but I thought that it would be good to wait with this one until as many as possible of the other PRs are closed.  I guess it is now such a time, since the remaining PRs are all fairly complicated and can not be closed right away, so LGTM.

PR #564, opened 2017-02-01
Transfer of binary operations relative to weak equivalences

cathlelay:
The PR #536 depend on this one.


vladimirias:
LGTM

DanGrayson:
LGTM, except `opp` is not a good shortening of "operator", because it has too many p's.

vladimirias:
This is the abbreviation that I have introduced when I started “Algebra”. I agree that it is a strange one, but fixing it is better done separately from this PR. I am merging, so that the next one can be discussed.

PR #565, opened 2017-02-02
generalize total2_paths2

DanGrayson:
```
    Lemma total2_paths2 {A : UU} {B : A -> UU} {a1 : A} {b1 : B a1}
	  {a2 : A} {b2 : B a2} (p : a1 = a2)
	  (q : transportf B p b1 = b2) : a1,,b1 = a2,,b2.
```
turns out to be a special case of this:
```
    Lemma two_arg_paths_f {A : UU} {B : A -> UU} {C:UU} {f : ∏ a, B a -> C} {a1 : A} {b1 : B a1}
	  {a2 : A} {b2 : B a2} (p : a1 = a2)
	  (q : transportf B p b1 = b2) : f a1 b1 = f a2 b2.
    Proof.
      intros. induction p. induction q. reflexivity.
    Defined.
```
, and similarly for the backwards version.  We introduce the
generalizations and use it, where possible.

### DIFF
--- a/UniMath/Foundations/PartA.v
+++ b/UniMath/Foundations/PartA.v
@@ -766,31 +766,15 @@ Proof.
   induction s as [a b].
   induction s' as [a' b']; simpl in *.
   induction p.
-  induction (!q).
+  change _ with (b=b') in q.
+  induction q.
   reflexivity.
-Defined.
-
-Lemma total2_paths2 {A : UU} {B : A -> UU} {a1 : A} {b1 : B a1}
-      {a2 : A} {b2 : B a2} (p : a1 = a2)
-      (q : transportf B p b1 = b2) : a1,,b1 = a2,,b2.
-Proof.
-  intros.
-  apply (@total2_paths _ _ (tpair (fun x => B x) a1 b1)
-                       (tpair (fun x => B x) a2 b2) p q).
-Defined.
-
-Lemma total2_paths2_b {A : UU} {B : A -> UU} {a1 : A} {b1 : B a1}
-  {a2 : A} {b2 : B a2} (p : a1 = a2)
-  (q : b1 = transportb B p b2) : a1,,b1 = a2,,b2.
-Proof.
-  intros.
-  apply (@total2_paths_b _ _
-    (tpair (fun x => B x) a1 b1) (tpair (fun x => B x) a2 b2) p q).
 Defined.
 
 Lemma two_arg_paths_f {A : UU} {B : A -> UU} {C:UU} {f : ∏ a, B a -> C} {a1 : A} {b1 : B a1}
       {a2 : A} {b2 : B a2} (p : a1 = a2)
       (q : transportf B p b1 = b2) : f a1 b1 = f a2 b2.
+(* This lemma replaced [total2_paths2], which is the special case f := tpair _ *)
 Proof.
   intros. induction p. induction q. reflexivity.
 Defined.
@@ -798,6 +782,7 @@ Defined.
 Lemma two_arg_paths_b {A : UU} {B : A -> UU} {C:UU} {f : ∏ a, B a -> C} {a1 : A} {b1 : B a1}
   {a2 : A} {b2 : B a2} (p : a1 = a2)
   (q : b1 = transportb B p b2) : f a1 b1 = f a2 b2.
+(* This lemma replaced [total2_paths2_b], which is the special case f := tpair _ *)
 Proof.
   intros. induction p. change _ with (b1 = b2) in q. induction q. reflexivity.
 Defined.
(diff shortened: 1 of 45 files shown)


vladimirias:
I don't understand what is the advantage of this. The "generalization" can be easily obtained by applying to the original version the function maponpaths relative to the function [ fun ab : total2 B => C (pr1 ab) (pr2 ab) ]. Clearly, the original version is a fundamental fact while the generalization is a derived corollary.

DanGrayson:
I think it makes more sense to think of the generalization as going in the opposite direction, because the original statement is the special case where `f` is `tpair _`.  The point is that Coq can usually infer that f should be `tpair _`, except in a few cases.  Probably Coq can not make the analogous inference in the direction you propose.

In any case, the statement as it stands maybe be useful.

vladimirias:
I see. I guess the way to argue for it is to say that in your form it avoids the use of total2 and so it is more general since it can be used in a type theory without the total2. Maybe you could mention it in a comment in the .v file?

DanGrayson:
Okay, I've added such comments, and, by improving a proof, was able to remove the
less general old lemmas.

vladimirias:
1. Shouldn't it be two_args_ instead of two_arg_ ?
2. It seems that two_arg_paths_f ( f := tpair _ ) occurs in many cases. Could we have the old identifier total2_paths2 as a notation for two_arg_paths_f ( f := tpair _ ) ? This will make the code nicer in many places and will give those who are used to total2_paths2 a way to use a familiar identifier.

DanGrayson:
1. In English, you can say "two argument paths", but not "two arguments paths".

2. ok

DanGrayson:
Okay, that should do it.

vladimirias:
Dan, I could not find the comment that you said you have made about the generality of two_arg_paths being related to the fact that it does not use total2 and therefore makes sense in type theories without dependent sums. Could you point me out where it is?

DanGrayson:
In PartA, line 757, I say this:
```
(* This lemma generalizes [total2_paths2_f], which is the special case f := tpair _ *)
```
Perhaps you want more than that?  We have so many thousands of statements that don't refer to total2, I don't see why this one is especially important compared the others.

vladimirias:
Could you please put in my comment that attracts attention to the fact that this generalization removes the use of total2? 

It is not important that this statement does not refer to total2, but that it removes the reference to the total2 from the previous version of the statement.

DanGrayson:
Like this?:
```
(* This lemma is a replacement for and a generalization of [total2_paths2_f] that
    does not refer to [total2].   The lemma  [total2_paths2_f]  can be obtained as
    the special case [f := tpair _], and Coq can often infer the value for [f], which is
    declared as an implicit argument. *)
```
Or is it also worthwhile saying that a function of type [total2 P -> Z] is essentially 
the same thing as the sort of function of two arguments dealt with here?
(Our functions [curry] and [uncurry] implement going back and forth.)

vladimirias:
Like this is good. Thanks you!

DanGrayson:
Okay, you're welcome!

UniMath/Foundations/PartA.v
```diff
@@ -746,50 +746,83 @@ Proof.
...
+Lemma two_arg_paths {A B C:UU} {f : A -> B -> C} {a1 b1 a2 b2} (p : a1 = a2)
+      (q : b1 = b2) : f a1 b1 = f a2 b2.
+(* This lemma is an analogue of [maponpaths] for functions of two arguments. *)
+Proof.
+  intros. induction p. induction q. reflexivity.
+Defined.
+
+Lemma two_arg_paths_f {A : UU} {B : A -> UU} {C:UU} {f : ∏ a, B a -> C} {a1 b1 a2  b2}
+      (p : a1 = a2) (q : transportf B p b1 = b2) : f a1 b1 = f a2 b2.
+(* This lemma is a replacement for and a generalization of [total2_paths2_f], formerly called
```
DanGrayson: Here is the comment requested.

vladimirias:
LGTM

benediktahrens:
Turns out that with this more general version `two_arg_paths_f` the inference of implicit arguments fails sometimes... :-(

benediktahrens:
It would be great if people would not remove working stuff. If you want to add a generalization, please make sure that the less general case stays in place.

DanGrayson:
The comment there explains that the old version is still available under the new name total2_paths2_f

mortberg:
What's the rationale behind this naming convention? I don't understand why the name of the lemmas are appended with `_f`?

DanGrayson:
They come in pairs corresponding to transportf and transportb.

PR #567, opened 2017-02-07
Associativity pull request

DanGrayson:
In this pull request we prove associativity for sequences and show
that lists are equivalent to sequences.


benediktahrens:
Merge conflicts...

UniMath/Algebra/IteratedBinaryOperations.v
```diff
@@ -4,158 +4,318 @@ Require Export UniMath.Algebra.Monoids_and_Groups.
...
+(* end of move upstream *)
+
+Local Notation "[]" := Lists.nil (at level 0, format "[]").
 Local Infix "::" := cons.
 
+(** general associativity for binary operations on types *)
+
+Section BinaryOperations.
+
+  Context {X:UU} (unit:X) (op:binop X).
```
benediktahrens: We have a type called `unit`, it is a bit confusing to have a section variable of the same name.

UniMath/Algebra/IteratedBinaryOperations.v
```diff
@@ -4,158 +4,318 @@ Require Export UniMath.Algebra.Monoids_and_Groups.
...
+    apply iterop_fun_nat.
+  - simple refine (nat_plus_associativity' _ _ _ @ _).
+    apply maponpaths.
+    unfold flatten'.
+    unfold lexicalEnumeration.
+    apply (maponpaths ((λ f, uncurry x ∘ f)
+                       : (stn (stnsum m) → ∑ i : stn n, stn (m i)) → stn (stnsum m) → nat )).
+    change (invmap (weqstnsum1 m) = weqstnsum_invmap m).
+    (* we need to construct weqstnsum1 with weqstnsum_invmap *)
+Abort.
```
benediktahrens: I forgot what the conclusion of the discussion about use of `Abort` was. Was there a conclusion?

UniMath/Combinatorics/FiniteSequences.v
```diff
@@ -1,26 +1,97 @@
 Require Export UniMath.Combinatorics.StandardFiniteSets.
+Require Export UniMath.Combinatorics.Lists.
 Unset Automatic Introduction.
 
 (* move upstream *)
-(* end of move upstream *)
 
+Local Open Scope transport.
+
+Lemma two_arg_paths_f {X:UU} {Y:X->UU} {Z:UU} (f : ∏ x, Y x -> Z)
```
benediktahrens: This sequence of lemmas might fit better in Foundations.

UniMath/Combinatorics/StandardFiniteSets.v
```diff
@@ -686,27 +734,55 @@ Proof.
   apply natplusr0.
 Defined.
 
-Definition weqstnsum_map { n : nat } (f : stn n -> nat) : (∑ i, stn (f i)) -> stn (stnsum f).
+Lemma two_arg_paths_f {X:UU} {Y:X->UU} {Z:UU} (f : ∏ x, Y x -> Z)
```
benediktahrens: This lemma is also in FiniteSequences. Same for the one below, I think.

UniMath/Algebra/IteratedBinaryOperations.v
DanGrayson: No consensus.  I think Abort can be useful.  In any case, my next pull request will eliminate it.

UniMath/Combinatorics/FiniteSequences.v
DanGrayson: In fact, I already did submit it and it got included.  Removing...

UniMath/Algebra/IteratedBinaryOperations.v
DanGrayson: fixing ...

vladimirias:
> On Feb 15, 2017, at 10:42 AM, Benedikt Ahrens <notifications@github.com> wrote:
> 
> I forgot what the conclusion of the discussion about use of Abort was. Was there a conclusion?

I am not sure, but I suggest that Aborted code should be at least commented out.

DanGrayson:
I suggest that Aborted code not be commented out, so the partial proof that's there will continue to work.  Letting it break would be a great loss, in many cases, if the partial proof was hard to obtain.

vladimirias:
Makes sense. I agree.

DanGrayson:
I've just added the following theorem to this pull request:

```
Definition UnorderedSequence (X:UU) := ∑ I:FiniteSet, I -> X.

Definition iterop_finseq_mon {M:abmonoid} : UnorderedSequence M -> M.
(* iterate the monoid operation over a finite sequence of elements of M *)
```

The proof uses the generalized commutativity theorem (a finite product is
independent of the ordering).  The proof of commutativity uses the generalized
associativity theorem (a product of products is a product).  I almost had this
proof more than a year ago.

vladimirias:
Impressive! I wouldn’t have guessed it.  The second construction should, I think, use, substantially, that M is a set. Or not? It is an open question whether a binary product structure on a category defines finite product structure. More generally, I guess, it may be stated as whether a binary, associative, commutative, monoidal structure on a category defines monoidal products of unordered finite sequences.

DanGrayson:
Yes, it depends heavily on M being a set:  It uses 
```
Definition squash_to_set {X Y : UU} (is : isaset Y) (f : X -> Y) : (∏ x x', f x = f x') -> ∥ X ∥ -> Y.
```
, taking X to be `finstruct I`.  But the theorem itself is part of standard math, so one needn't guess it.

Getting the proof was like pulling teeth, because the maps in the (old) proof of this theorem were
hard to compute with:
```
Definition weqstnsum1 {n} (f : stn n -> nat) : (∑ i, stn (f i)) ≃ stn (stnsum f).
```
The new proof I give of it takes the old proof and replaces both forward and backwards maps by ones I define "by hand", to be easier to work with and more closely connected the the process of 
converting a sequence of sequences to a sequence, for using in associativity:
```
Proof.
  intros. simple refine (remakeweqboth (weqstnsum1_prelim_eq _) (weqstnsum1_prelim_eq' _)).
Defined.
```
A year ago I was stuck with the proof of this:

```
Definition flattenStep' {X n}
           (m : stn (S n) → nat)
           (x : ∏ i : stn (S n), stn (m i) → X)
           (m' := m ∘ dni lastelement)
           (x' := x ∘ dni lastelement) :
  flatten' x = concatenate' (flatten' x') (x lastelement).
```
, but the new proof of `weqstnsum1` makes `flatten'` easier to deal with:
```
Definition flatten' {X n} {m:stn n->nat} : (∏ (i:stn n), stn (m i) -> X) -> (stn (stnsum m) -> X).
Proof.
  intros ? ? ? g. exact (uncurry g ∘ invmap (weqstnsum1 m)).
Defined.
```
The old definition of the weq provided `weqstnsum1` gets it as a composite of three others, one of the inverted.  The new definition of the forward map is quite simple:
```
Local Definition weqstnsum_map { n : nat } (m : stn n -> nat) : (∑ i, stn (m i)) -> stn (stnsum m).
Proof.
  intros ? ? ij.
  exact (stnpair _ (stnsum (m ∘ stn_left'' (stnlt (pr1 ij))) + pr2 ij) (_c_ ij)).
Defined.
```
The challenge was to prove the new forward and backwards maps are homotopic to the old ones, which took 115 lines and blocked me a year ago.  The main new idea there was to use these new little lemmas to move a weak equivalence to the other side of the equation:
```
homotweqinv:
  ∏ (X : Type) (Y : UU) (Z : Type) (f : X → Z) (w : X ≃ Y) (g : Y → Z),
  f ~ g ∘ w → f ∘ invmap w ~ g
homotweqinv':
  ∏ (X : Type) (Y : UU) (Z : Type) (f : X → Z) (w : X ≃ Y) (g : Y → Z),
  f ∘ invmap w ~ g → f ~ g ∘ w
```
Finally, it was this new useful lemma (the proof does not use `gradth`!) that fixes up the two maps in the weq:
```
remakeweqboth: ∏ (X Y : UU) (f : X ≃ Y) (g : X → Y) (h : Y → X), f ~ g → invmap f ~ h → X ≃ Y
```
These lemmas check that it worked:
```
Lemma remakeweqboth_eq {X Y : UU} (f:X≃Y) (g:X->Y) (h:Y->X) (r:f~g) (s:invmap f ~ h) :
  pr1weq (remakeweqboth r s) = g.
Proof.
  reflexivity.
Defined.

Lemma remakeweqboth_eq' {X Y : UU} (f:X≃Y) (g:X->Y) (h:Y->X) (r:f~g) (s:invmap f ~ h) :
  invmap (remakeweqboth r s) = h.
Proof.
  reflexivity.
Defined.
```

DanGrayson:
Re: "It is an open question whether a binary product structure on a category defines finite product structure."

I suppose that's not true.  It seems that all the standard examples of additive categories come with finite product structure, and it's useful, so probably the right definition of additive category would start with that instead of with binary product structure.  @tpannila may have a comment about that.

DanGrayson:
PS: One of the reasons I want this theorem
```
Definition iterop_finseq_mon {M:abmonoid} : UnorderedSequence M -> M.
```
is so I can give the definition of a direct sum diagram of a finite unordered sequence of objects in a linear category via equations.  The monoids will be the additive monoids associated to the Hom sets in the category.

DanGrayson:
PS: I was careful in the definition of repeated iteration of an binary operation with unit element so the unit element would be used only in the case `n=0`, and not appear at all when `n>0`.   That's why there's induction twice here:
```
  Definition iterop_fun {n} (x:stn n->X) : X.
  Proof.
    intros.
    induction n as [|n _].
    { exact unel. }
    { induction n as [|n I].
      { exact (x lastelement). }
      { exact (op (I (x ∘ dni lastelement)) (x lastelement)). }}
  Defined.
```
It might make some proofs simpler if I did the corresponding thing with `stnsum`, which
uses only a single induction:
```
Definition stnsum {n : nat} (f : stn n -> nat) : nat.
Proof.
  intro n. induction n as [ | n IHn].
  - intro. exact 0.
  - intro f. exact (IHn (λ i, f (dni lastelement i)) + f lastelement).
Defined.
```
, so I might try that.

DanGrayson:
PS: Being careful, as above, with the definition of `iterop_fun {n} (x:stn n->X) : X` means that an operation is a monoid operation if and only if it's associative in the sense that a product of products is a product, for then `x * 1 = (x) * () = (x) = x`.  Here I use () to denote the product of the empty sequence.

Here are the definitions of associativity and commutativity:
```
  Definition isAssociative_fun :=
    ∏ n (m:stn n -> nat) (x : ∏ i (j:stn (m i)), X), iterop_fun (StandardFiniteSets.flatten' x) = iterop_fun_fun x.

  Definition isAssociative_seq :=
    ∏ (x : Sequence (Sequence X)), iterop_seq (FiniteSequences.flatten x) = iterop_seq_seq x.

  Definition isCommutative_fun :=
    ∏ n (x:⟦n⟧ -> X) (f:⟦n⟧≃⟦n⟧), iterop_fun (x ∘ f) = iterop_fun x.
```

DanGrayson:
Re: "It is an open question whether a binary product structure on a category defines finite product structure."

Before I said : "I suppose that's not true."  But products satisfy a universal property so are unique up to a unique isomorphism, rendering all coherency conditions true.  So it should be true, after all.

@tpannila

tpannila:
Dan, every additive category has finite ordered products and finite ordered coproducts.
These follow from `FinOrdProducts_from_Terminal_and_BinProducts` in
limits/FinOrdProducts.v and `FinOrdCoproducts_from_Initial_and_BinCoproducts` in
limits/FinOrdCoproducts.v, because Zero is both initial and terminal, and BinDirectSum 
is both BinProduct and BinCoproduct.

vladimirias:
Yes, this is clear. We are discussing unordered ones.

DanGrayson:
Re: "It is an open question whether a binary product structure on a category defines finite product structure."

Okay, here's the proof.  Consider a univalent category C and a functor F : C --> Set.  Define the type Rep F whose elements are representations of F, i.e., an object c : C with an isomorphism h^c = F, where h^c ( d ) = Hom(c,d).  Any two representations are actually (as opposed to merely) isomorphic, hence actually equal.  Thus Rep F is a proposition.  Now suppose that F is designed so that a representation amounts to a product of an unordered sequence of elements.  Use `squash_to_prop` to reduce the proof to the ordered case.

benediktahrens:
On 02/16/2017 06:00 PM, Daniel R. Grayson wrote:
> PS: See https://github.com/UniMath/TypeTheory/commit/6f240a7 @benediktahrens 


??

DanGrayson:
It makes sense in the context of the discussion on my pull request on github, so to see it, look there.

benediktahrens:
On 02/16/2017 07:24 PM, Daniel R. Grayson wrote:
> It makes sense in the context of the discussion on my pull request on
> github, so to see it, look there.


You are discussing a lot of stuff in that PR. What exactly are you
referring to?

DanGrayson:
The message immediately before: https://github.com/UniMath/UniMath/pull/567#issuecomment-280368656

vladimirias:
I agree. I was thinking about a general, not necessarily a univalent, category.

DanGrayson:
Re: "I agree. I was thinking about a general, not necessarily a univalent, category."

Oh, okay, here's a counterexample, perhaps.  Let C be a skeletal presubcategory of the univalent category F of finite (unordered) sets and suppose C has finite unordered sums.  Let * denote the one element set in C.  If J is a finite set, we may consider the unordered sum `G J := ∑ j:J, *`, formed in C.  It gives a functor G : F --> C which is a retraction for the inclusion C --> J, and is an equivalence of precategories.  But F is a univalent category, so any nontrivial permutation f of J corresponds to a nontrivial loop m.  Since C has only trivial loops, G sends m to a trivial loop and G sends f to a nontrivial automorphism.  But G must preserves the relationship between loops and their induced automorphisms, because it is a functor, contradiction.

DanGrayson:
Here's another thought about associativity and commutativity.  It seems the traditional definition of symmetric monoidal category is bulky and inconvenient.  Wouldn't a better definition be that it is a category C equipped with functors (∏ i:I, C) --> C for each finite (unordered) set satisfying the evident generalized associativity rule arising from considering a finite family of finite sets?

vladimirias:
I am not sure how to make this idea precise. The associativity requires the pentagon diagram. What will it follow from?

DanGrayson:
Re: "The associativity requires the pentagon diagram. What will it follow from?"

You're right, coherence is missing.

benediktahrens:
There is a file called `foo.v` as part of your PR. Maybe rename that file?

benediktahrens:
Your PR seems to change sub/coq.

DanGrayson:
Oops, fixed, thanks.

UniMath/Folds/folds_isomorphism.v
```diff
@@ -230,7 +230,8 @@ Proof.
       * apply funextfun; intro f.
         eapply pathscomp0.
         { apply ϕ₁_is_comp. }
-        symmetry.
+        symmetry
+        .
```
benediktahrens: Here's a weird newline.

UniMath/Foundations/UnivalenceAxiom.v
```diff
@@ -435,3 +435,16 @@ Proof.
   - use transportf_paths. exact XR.
   - induction e. apply idpath.
 Defined.
+
+(** induction tactic for the universe *)
+
+Theorem UU_rect (X Y : UU) (P : X ≃ Y -> UU) :
```
benediktahrens: This kind of induction is often very useful, for instance for univalent categories.
Should we agree on calling those lemmas `*_rect`, as you do here?

benediktahrens: See here for the analog for univalent categories:
https://github.com/UniMath/TypeTheory/blob/master/TypeTheory/Auxiliary/Auxiliary.v#L828

vladimirias:
UU_rec is better. Hopefully they will eventually get rid of _rect in Coq and have _rec for recursion in type and _recs for recursion in Set, if they still want set.

We should not follow what was a mistake in Coq naming.

UniMath/Folds/folds_isomorphism.v
DanGrayson: oops, fixed

DanGrayson:
Using "_rec" instead of "_rect" is fine with me.  I could also add notations to rename all of Coq's *_rect identifiers to *_rec.  It would also be nice to have a uniform naming convention for the map that takes paths to isomorphisms and for its inverse.

But I think we also need something more, to give all these principles an identical interface, not just identical names.  Hmm...

By the way, the tactics I make for using these are useful.  Here's my favorite instance:
```
(* now we try to mimic this construction:

    Inductive PosetEquivalence (X Y:Poset) : Type :=
                  pathToEq : (X=Y) -> PosetEquivalence X Y.

    PosetEquivalence_rect
         : ∏ (X Y : Poset) (P : PosetEquivalence X Y -> Type),
           (∏ e : X = Y, P (pathToEq X Y e)) ->
           ∏ p : PosetEquivalence X Y, P p

*)

Theorem PosetEquivalence_rect (X Y : Poset) (P : X ≅ Y -> UU) :
  (∏ e : X = Y, P (Poset_univalence_map e)) -> ∏ f, P f.
Proof.
  intros ? ? ? ih ?.
  set (p := ih (invmap (Poset_univalence _ _) f)).
  set (h := homotweqinvweq (Poset_univalence _ _) f).
  exact (transportf P h p).
Defined.

Ltac poset_induction f e :=
  generalize f; apply PosetEquivalence_rect; intro e; clear f.
```

DanGrayson:
Let's not change _rect to _rec in this pull request.  That identifier UU_rect appeared in the diff just because I was moving needed things upstream to where they belong and that got dragged along.

benediktahrens:
There are now merge conflicts in this PR, I am afraid...

DanGrayson:
I'd be glad to fix it if it's ready to be pulled.

benediktahrens:
On 02/19/2017 01:41 PM, Daniel R. Grayson wrote:
> I'd be glad to fix it if it's ready to be pulled.

In Combinatorics/Lists.v there is a lemma marked for moving upstream

```
Lemma dirprodEquality {X Y} {w w':X × Y} : pr1 w = pr1 w' -> pr2 w = pr2
w' -> w = w'.
(* move upstream *)
```
but otherwise it looks fine to me.

DanGrayson:
Okay, I fixed the conflicts and moved that upstream.

vladimirias:
Such a lemma exists in BasicsA. It is called dirprodeq, l.427.

DanGrayson:
Thanks, see #582

PR #573, opened 2017-02-13
a *different* definition of the option functor which is simpler

rmatthes:
the previous definition uses hfiber and thus even forces that the created new environment yields s everywhere


DanGrayson:
I'll leave this to others to evaluate.

benediktahrens:
lgtm

PR #576, opened 2017-02-17
Strength for multisorted binding signatures

rmatthes:
For the function MultiSortedSigToFunctor, there is now also a corresponding function MultiSortedSigToSignature constructing the signature with strength.
Key to doing this is a heterogeneous notion of signature with strength: the final target category can be different (maybe it should be called pre-signature, and only when both categories are equal, we call it signature). For us, the difference is that between the slice category and its base category.
The heterogeneous notion needed more general strength laws on basis of monoidal operations on functors that are not endofunctors (the respective file is now misnomed).
The addition of a variable with a sort is reasonably described by a generalized option functor, this time not with a terminal object added but an arbitrary one.
Also more structure was needed for the distributive laws - in the end, it seemed better to have the functor as argument and not as part of the structure.
For simplicity, I replaced the iterated product based on foldr1 by one on foldr. Is anything lost by this?

### DIFF
--- a/UniMath/CategoryTheory/EndofunctorsMonoidal.v
+++ b/UniMath/CategoryTheory/EndofunctorsMonoidal.v
@@ -28,14 +28,26 @@ Require Import UniMath.CategoryTheory.precategories.
 Require Import UniMath.CategoryTheory.functor_categories.
 Require Import UniMath.CategoryTheory.UnicodeNotations.
 
-(** There is a monoidal structure on endofunctors, given by composition.
-    While this is considered to be strict in set-theoretic category theory,
-    it ain't strict in type theory with respect to convertibility.
-    So we consider it to be a weak monoidal structure instead.
+(** There is a monoidal structure on endofunctors, given by composition.  While
+    this is considered to be strict in set-theoretic category theory, it ain't
+    strict in type theory with respect to convertibility.  So we consider it to
+    be a weak monoidal structure instead. However, pointwise, it suffices to
+    take the identity for all those natural transformations (the identity is
+    also behind the definition of nat_trans_functor_assoc).
+
+    To understand the need for this structure even better, notice that the
+    proofs of functor axioms for one composition in the unitality and
+    associativity properties are slightly different from the proofs for another
+    and because of it the composition of functors is not strictly unital or
+    associative. However, these proofs are not used in the definition of natural
+    transformations, to be precise only functor_data is used, and the
+    composition of functor_data is strictly unital and associative.
+
 *)
 Section Monoidal_Structure_on_Endofunctors.
 
-(** while this is normally used for endofunctors, it can be done more generally *)
+(** while this is normally used for endofunctors, it can be done more generally,
+    but already for endofunctors, this is crucial for the development of substitution systems *)
 
 Context {C D : precategory}.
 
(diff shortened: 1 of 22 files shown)


vladimirias:
Hi Ralph,

simple question: 

in EndofunctorsMonoidal.v <https://github.com/UniMath/UniMath/pull/576#diff-dbfb7862f9a09cd80bd2ae9ec4b04c41> ll. 40-52. 

We define objects of different types but define them all as \rho_functor X. This must mean that the types are actually the same.

Id functor composition strictly unital on both sides? That is, is \rho_functor actually the identity natural transformation?

The same question about \lambda_functor. Is it the identity as well, which is why you can use it as an element of what appears to be different types?

benediktahrens:
On 02/17/2017 06:48 PM, Vladimir Voevodsky wrote:
> Hi Ralph,
> 
> simple question:
> 
> in EndofunctorsMonoidal.v
> <https://github.com/UniMath/UniMath/pull/576#diff-dbfb7862f9a09cd80bd2ae9ec4b04c41>
> ll. 40-52.
> 
> We define objects of different types but define them all as
> \rho_functor X. This must mean that the types are actually the same.
> 
> Id functor composition strictly unital on both sides? That is, is
> \rho_functor actually the identity natural transformation?
> 
> The same question about \lambda_functor. Is it the identity as well,
> which is why you can use it as an element of what appears to be
> different types?

Yes.
We do *not* have that (F o G) o H is convertible to F o (G o H), but we
do have that, for any functor X,
nat_trans ((F o G) o H) X
is convertible to
nat_trans (F o (G o H)) X
since the functors F o (G o H) and (F o G) o H are *pointwise*
convertible, and they only occur pointwise in the definition of the type
of natural transformations.

Here, all those coherence natural isos are pointwise the identity morphism.

rmatthes:
Benedikt was too quick for me. I can add that, for our previous work on unsorted systems, these definitions were already needed for endofunctors. Without them, it looked impossible to get proper definitions of the strength laws (in https://github.com/UniMath/UniMath/blob/master/UniMath/SubstitutionSystems/Signatures.v#L195 and https://github.com/UniMath/UniMath/blob/master/UniMath/SubstitutionSystems/Signatures.v#L245), and we baked our own versions, in particular https://github.com/UniMath/UniMath/blob/master/UniMath/SubstitutionSystems/Signatures.v#L237 which is not motivated mathematically. However, both versions could be shown to be equivalent. EndofunctorsMonoidal.v is thus only about getting statements type-check in a natural form.

DanGrayson:
About using foldr or foldr1 for iterated products, I wonder whether the iterated product for a list of length 1 turns out to a product with the terminal object.  It might be nice to get rid of it, if so.

DanGrayson:
lgtm

rmatthes:
@DanGrayson: thanks for your observation; so, one could gain in going back to foldr1, but there is not yet a dedicated induction principle for reasoning about it nor a dependently-typed version for definitions (the latter should both be the same principle)

rmatthes:
@vladimirias: thanks for your proposal of a comment to EndofunctorsMonoidal.v, I have used this for a more elaborate comment

@benediktahrens: no need to wait for Travis, I only changed contents of comments

PR #578, opened 2017-02-17
show that the monoidal axioms do hold for propositional equality

rmatthes:
these lemmas are just there as extra motivation and are not meant to be used

### DIFF
--- a/UniMath/CategoryTheory/EndofunctorsMonoidal.v
+++ b/UniMath/CategoryTheory/EndofunctorsMonoidal.v
@@ -80,36 +80,21 @@ Definition α_functor_inv (X : functor C D)(Y : functor D E)(Z : functor E F) :
     of the three pairs of functors; the extra assumption on having homsets is only used in order
     to have simple proofs, it is not necessary, as shown in Section "functor_equalities" in
     functor_categories.v: Lemmas functor_identity_left, functor_identity_right and functor_assoc *)
-Lemma motivation_ρ_functor (hsD : has_homsets D)(X : functor C D) : functor_composite X (functor_identity D) = X.
+Local Lemma motivation_ρ_functor (hsD : has_homsets D)(X : functor C D) : functor_composite X (functor_identity D) = X.
 Proof.
-  apply functor_eq.
-  exact hsD.
-  simpl.
-  destruct X as [data laws].
-  destruct data as [onobs onmorphs].
-  apply idpath.
-Qed.
+  now apply (functor_eq _ _ hsD); induction X as [data laws]; induction data as [onobs onmorphs].
+Defined.
 
-Lemma motivation_λ_functor (hsD : has_homsets D)(X : functor C D) : functor_composite (functor_identity C) X = X.
+Local Lemma motivation_λ_functor (hsD : has_homsets D)(X : functor C D) : functor_composite (functor_identity C) X = X.
 Proof.
-  apply functor_eq.
-  exact hsD.
-  simpl.
-  destruct X as [data laws].
-  destruct data as [onobs onmorphs].
-  apply idpath.
-Qed.
+  now apply (functor_eq _ _ hsD); induction X as [data laws]; induction data as [onobs onmorphs].
+Defined.
 
-Lemma motivation_α_functor (hsF : has_homsets F)(X : functor C D)(Y : functor D E)(Z : functor E F) :
+Local Lemma motivation_α_functor (hsF : has_homsets F)(X : functor C D)(Y : functor D E)(Z : functor E F) :
   functor_composite (functor_composite X Y) Z = functor_composite X (functor_composite Y Z).
 Proof.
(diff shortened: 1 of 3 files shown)


benediktahrens:
We do actually have those results already, without requiring the homs-are-sets assumption. See the section "functor_equalities" in functor_categories.v.

rmatthes:
Thanks, I did not even have a look since propositional equality between functors seemed to be something one should rather not look at. I made a clear comment pointing to those lemmas with
rather difficult proofs. As the restricted lemmas here were only given for motivational purposes, maybe this PR still makes sense.

benediktahrens:
lgtm

UniMath/CategoryTheory/EndofunctorsMonoidal.v
```diff
@@ -74,4 +74,43 @@ Definition α_functor_inv (X : functor C D)(Y : functor D E)(Z : functor E F) :
...
+    functor_categories.v: Lemmas functor_identity_left, functor_identity_right and functor_assoc *)
+Lemma motivation_ρ_functor (hsD : has_homsets D)(X : functor C D) : functor_composite X (functor_identity D) = X.
+Proof.
+  apply functor_eq.
+  exact hsD.
+  simpl.
+  destruct X as [data laws].
+  destruct data as [onobs onmorphs].
+  apply idpath.
+Qed.
```
mortberg: This is not properly using bullets. However for such a simple result I would do everything on one line, for example by:

`now apply (functor_eq _ _ hsD); induction X as [[]].`

As a side remark I think `functor_eq` needs some more implicit arguments.

mortberg:
Maybe put `Local` in front of each of these so that they are not exported? This way someone won't incidentally use them.

vladimirias:
What does Coq do when one writes "induction X as [[]].”? Is the proof term expressed in terms of the induction principle?

DanGrayson:
Re: What does Coq do when one writes "induction X as [[]].”? Is the proof term expressed in terms of the induction principle?

No, the outer one is by induction, using the induction principle, and the inner ones are by destruction, using match.  So that formulation should not be used.

rmatthes:
I took your advice seriously and also changed two files of PR #576 that used such induction constructs. I did not want to create a new PR just for that.

benediktahrens:
From https://travis-ci.org/UniMath/UniMath/jobs/202947924#L5196 : 
```
COQC UniMath/SubstitutionSystems/MultiSorted.v

File "./UniMath/SubstitutionSystems/MultiSorted.v", line 514, characters 4-6:

Syntax error: '.' expected after [vernac:command] (in [vernac_aux]).

real	0m6.148s

user	0m6.040s

sys	0m0.108s

make: *** [UniMath/SubstitutionSystems/MultiSorted.vo] Error 1
```

rmatthes:
I am sorry, but I did not wait until compilation ended at home (trusting in Travis).

DanGrayson:
I don't know why you're apologizing:  pushing another commit to your branch just causes Travis to start over again -- your code will still be checked, so it's not an issue of trusting Travis.

rmatthes:
I wanted to say that I apologize for not having checked by myself (the stupid typo came from editing a comment) before pushing commits, and so I attracted your attention to a triviality. And I implied that I would not have done so had I not known that Travis will anyway check. Since we are on this: is it considered acceptable practice to push before doing a full compilation in case one is in good faith about that? (My computer at home is much slower than the one in my office where I can more easily wait until make -j7 has completed - which I would not even run with -j7 on my computer at home.)

vladimirias:
I would say “no”. 

There is usually no hurry to push, why not to wait?

DanGrayson:
Oh, yes, I misunderstood your apology.  I always compile everything at home before creating or changing a pull request.

benediktahrens:
On 02/18/2017 09:01 PM, Ralph Matthes wrote:
> I wanted to say that I apologize for not having checked by myself
> (the stupid typo came from editing a comment) before pushing commits,
> and so I attracted your attention to a triviality. And I implied that
> I would not have done so had I not known that Travis will anyway
> check. Since we are on this: is it considered acceptable practice to
> push before doing a full compilation in case one is in good faith
> about that? (My computer at home is much slower than the one in my
> office where I can more easily wait until make -j7 has completed -
> which I would not even run with -j7 on my computer at home.)

Two tipps:

a) You can set up Travis for your own fork. This is very simple, since
the difficult part, writing the .travis.yml, is already done. I can help
you do that, if you want.

b) If you need to push urgently (because you change location, or
whatever), you can push to a different branch on your fork on Github, by
doing

$ git push origin devel:new-devel

This creates a new branch "new-devel" in the origin repository. Your PR,
created from your branch devel on origin, will not receive the new
commits until you also push them to devel on origin.

Combining a) and b) lets you check your changes by Travis before you
push those changes to an existing PR.

PR #579, opened 2017-02-18
add a little lemma, potentially useful

DanGrayson:
He explains the reason for the name:

" ... and “Voldemort’s theorem” stating that certain objects exist but cannot be named. For example, there is no natural point on a geometric circle — no point on a geometric circle can be named by a well-typed expression.  Similarly it is not possible to name any particular basis for a vector space or any particular isomorphism of a finite dimensional vector space with its dual. "

Well, it doesn't state that, but it does imply it.

### DIFF
--- a/UniMath/Foundations/PartA.v
+++ b/UniMath/Foundations/PartA.v
@@ -869,6 +869,12 @@ Proof.
   apply idpath.
 Defined.
 
+Lemma total2_section_path {X:UU} {Y:X->UU} (a:X) (b:Y a) (e:∏ x, Y x) : (a,,e a) = (a,,b) -> e a = b.
+(* this is called "Voldemort's theorem" by David McAllester, https://arxiv.org/pdf/1407.7274.pdf *)
+Proof.
+  intros ? ? ? ? ? p. simple refine (_ @ fiber_paths p). unfold base_paths. simpl.
+  apply pathsinv0, transport_section.
+Defined.
 
 (** *** Lemmas about transport adapted from the HoTT library and the HoTT book *)
 


vladimirias:
By a well-typed *closed* expression. This also applies to a point of an abstract set with two points. It can, however, be named easily as a variable of the corresponding type. 

The theorem is nice.

DanGrayson:
Yes.  I prefer to think of it as a mathematical theorem, rather than a metamathematical one about well-typed expressions, namely this:
```
Corollary noname : ¬ (∏ F, cardinalityFiniteSet F = 2 -> F).
```
I've got the proof half formalized, if anyone is interested in it, encourage me to finish it.

Somehow he makes do without univalence.  In our context, to apply it to a type like the type X of 2-element sets, we would use univalence and apply it as in the structure identity principle to get a path in X rather than in the universe.  It works for any type that parametrizes objects that have enough automorphisms so any point is moved by one of them.

In that paper he defines a type theory with two notions of equality, one strict and one propositional, and reflection to the strict one for sets.  Maryanthe Malliaris mentioned it to me.

vladimirias:
The lemma says that if a point of the total space is homotopic to the point of a section lying in the same fiber, then the fiber component of this point is homotopic to the fiber component of the section point inside the fiber. 

In homotopy theory, two points in the same fiber of a fibration are homotopic in the total space if and only if their classes in the pi_0 of the fiber are in the same orbit of the action of the pi_1 of the base on the pi_0 of the fiber. A point lying on a section is in the element of the pi_0 of the fiber that is invariant under the action of the pi_1 of the base. So if another point is in the same orbit of the action it is in the same element of the pi_0 of the fiber.

It can be applied to the universal fibration over a universe and then one needs the univalence axiom to connect the action of the pi_1 for this fibration with the action of automorphisms of the fiber.  The lemma itself does not refer to the universe and so does not need the UA.

DanGrayson:
Right, but the corollary needs UA.

DanGrayson:
It strikes me that it will be necessary to explain the statement of the corollary to a mathematician:
```
Corollary noname : ¬ (∏ F, cardinalityFiniteSet F = 2 -> F).
```
for it will be confused with
```
Corollary noname' : ¬ (∀ F, cardinalityFiniteSet F = 2 -> ∥ F ∥).
```
That suggests we follow the lead of McAllester and say that the meaning of `¬ (∏ F, cardinalityFiniteSet F = 2 -> F)` is that there is no *natural* or *canonical* way to select an element from each set of cardinality 2.

Has it been pointed out before that we finally have a definition of *natural* that doesn't involve any category theory, that it's just a matter of logic?

PR #581, opened 2017-02-19
clean up, add Section and isProofIrrelevant

DanGrayson:
introduce `Definition Section {X:UU} (P:X->UU) := ∏ x:X, P x.` and use
it in a few places

`homot` was generalized previously to handle sections, now generalize
`homotcomp`, and a few others

introduce `Definition isProofIrrelevant (X:UU) := ∏ x y:X, x = y.`

simplify proofs of `iscontraprop1inv` and `invproofirrelevance`

prove `Lemma isaprop_isProofIrrelevant X : isaprop (isProofIrrelevant X).`

remove `Coercion isapropifcontr : iscontr >-> isaprop.`.  It was used
only twice and it's an untraditional use for coercion.


benediktahrens:
`Section` is a Coq keyword. Syntax highlighting will be confusing, at least.

benediktahrens:
I don't think that abbreviating such small things like `∏ x:X, P x` is useful - actually I think that it makes things more difficult. Imagine you are searching for a proof that a pi-type is a proposition. It is very natural to type
```
Search (isaprop (∏ _ , _ )).
```
Now if your lemma is stated as `isaprop (Section P)`, then the above search will not succeed, because the `Search` does not unfold definitions. So you always will have to search for two different patterns instead of one.

DanGrayson:
Yes, you're right, it does get screwed up by emacs' Coq mode.  The difference in color is not great.  I suggest the name is more important than the precise color.

DanGrayson:
Yes, `Search` may not find statements if there are multiple ways to phrase statements, but that's a general remark about user interface tools, which are always going to have difficulties, but might be improved shortly, whereas "section" is an important mathematical concept which will always be important.

vladimirias:
I agree with Benedikt. 

1. The name Section should definitely not be used for an identifier.  

2. I suggest not to introduce a new name for “forall”. One may argue that it is a little better since it does not require the base of the family to be specified (an implicit argument) and that it makes it more similar to the “toral2”. On the other hand it will make it necessary to write e.g. “section (fun x:A => B)” where now we write “forall x:A, B”. It will also not have all the convenient fancy features such as being able to write “forall(x:A)(y:B),C”.

DanGrayson:
Okay, I'm convinced, will fix.

benediktahrens:
lgtm

PR #589, opened 2017-02-21
objects are made implicit arguments for functor_comp

### LINKED ISSUE #584: functor_comp
The first three arguments of functor_comp should be implicit.

benediktahrens:
resolves #584


DanGrayson:
lgtm

Issue #587, opened 2017-02-21
ElementsOp


vladimirias:
@mweav In el_data there is this "!ap (#X (pr1 f)) (!(pr2 g)))". Why? Shouldn't it be "ap(#X(pr1 f))(pr2 g)?

DanGrayson:
resolved by #590, closing

Issue #586, opened 2017-02-21
ElementsOp (and probably Elements?)


vladimirias:
@mweav: In the definition of the identity morphism, the equality from the axioms of presheaves is first inversed and then eqtohomot applied to it.
In the definition of the composition of morphisms the eqtohomot is first applied to the equality from the axioms of presheaves and then the resulting equality is inversed. 

It should be the same in both cases.

DanGrayson:
resolved by #590, closing

PR #590, opened 2017-02-21
Small changes to ElementsOp.v

mweav:
Addressing issues #586 and #587.


benediktahrens:
There is now a merge conflict in this PR. Also, some arguments of `functor_comp` have been made implicit, so you should remove the underscores. Also, a few parentheses can be removed, I think.

mortberg:
Maybe you can resolve this issue as well in this PR: https://github.com/UniMath/UniMath/issues/525

Also, do we really need both Elements.v and ElementsOp.v?

DanGrayson:
ElementsOp.v seems redundant.  Did I create it?

I think I made Elements.v as a way of defining representable functors, but it wasn't efficient, so now I don't use it at all.

mortberg:
I think @mweav added ElementsOp as he's working with the category of elements of presheaves? I know from working on limits and colimits that the op often got in the way when we tried to do things just by dualizing (i.e. defining limits using colimits with a lot of op's), so in the end it was easier just have two unrelated files and not fiddle with any op's. This led to some duplication, but using the library was a lot easier in the end. 

However for the category of elements having the op's around when working with presheaves is maybe not that much of an issue? I think a good test would be to check if Matthew's proofs go through with Elements instead of ElementsOp.

DanGrayson:
We currently have no files that use `ElementsOp`.  According to the log it was created by Vladimir here:
```
commit baab969d6eca599e6747522e4a6f03ca52de9433
Author: Vladimir Voevodsky <vladimir @ ias . edu>
Date:   Sat Nov 19 08:31:06 2016 -0500

    Replaced my copy of UniMath/UniMath with the copy from the UniMath repo
```
, but that makes no sense.

@mweav -- do you need that file?  If not, we can just delete it.

DanGrayson:
Let's not use the word "presheaf" for a contravariant functor unless there's a topology around and sheafification is a prospect.

benediktahrens:
On 02/22/2017 02:23 PM, Daniel R. Grayson wrote:
> We currently have no files that use `ElementsOp`.  According to the log it was created by Vladimir here:
> ```
> commit baab969d6eca599e6747522e4a6f03ca52de9433
> Author: Vladimir Voevodsky <vladimir @ ias . edu>
> Date:   Sat Nov 19 08:31:06 2016 -0500
> 
>     Replaced my copy of UniMath/UniMath with the copy from the UniMath repo
> ```
> , but that makes no sense.
> 
> @mweav -- do you need that file?  If not, we can just delete it.
> 

A general request: before you decide to remove any files from
UniMath/UniMath, please check that it is not used in UniMath/TypeTheory.

mortberg:
I think he does, this file on his fork is using it:

https://github.com/mweav/UniMath/blob/elems-slice-equiv/UniMath/CategoryTheory/elems_slice_equiv.v


About "presheaf": isn't it standard to use it for functors from C^op into Set (see https://en.wikipedia.org/wiki/Presheaf_%28category_theory%29)? If you don't like it what do you think we should call them? I want to add a file with basic theory about these (it will be useful for simplicial and cubical sets to have basic results proved somewhere).

DanGrayson:
The connotation of the word "presheaf" is that it is about to become a sheaf, so we should reserve it for that situation.  Is there something wrong with the word "functor", which includes both contravariant and covariant ones?

mortberg:
The point is that it's a functor into Set, so the functor category of these inherit lots of good properties from Set. This discussion is diverging from this PR, so let's not continue it here and instead continue it when I have prepared a PR with my file on [C^op,Set].

DanGrayson:
Well, no, the word "presheaf" doesn't imply that the target category is Set.  For that you have to say "presheaf of sets".

vladimirias:
However, it is usually assumed that a “a presheaf” is a presheaf of sets unless some other target category was specified. For example, PreShv(C) is the notation for the category of presheaves of sets on C that I use in my papers.

vladimirias:
The word presheaf has long overgrown its original meaning as “not quite a sheaf” and became a separate concept. It is a good word for a contravariant functor to sets.

vladimirias:
>  I want to add a file with basic theory about these (it will be useful for simplicial and cubical sets to have basic results proved somewhere).

This is a good idea.

vladimirias:
This is not a good idea. The word presheaf, as I have already remarked, is now a standard word in many contexts where no sheafification is intended. It applies especially to categorical logic.

vladimirias:
The file ElementsOp was created on purpose, precisely because working with Elements for presheaves was extremely inconvenient. Indeed, observe that a morphism in El(F) where F is covariant is (X,a) -> (X’,a’). If P is a presheaf then F is the corresponding covariant functor C^{op} -> SET. Then a morphism in El(F) is (X,a) -> (X’,a’) where X -> X’ is in C^{op}. The category El(P) is then the category (El(F))^{op} not El(F).

This led to a proliferation of op’s, that, as we know, do not satisfy (C^{op})^{op}=C definitionally.

DanGrayson:
Re: "The file ElementsOp was created on purpose, precisely because working with Elements for presheaves was extremely inconvenient. "

If so, then at least an attempt should be made to derive the results of ElementsOp as corollaries from the results of Elements, so we don't have two copies of the code evolving independently.

mortberg:
Ok, thanks for this explanation. It was what I suspected, so I think it
makes sense to keep both.

--
Anders

mortberg:
It would be great if this worked out, but from my experience working on
limits and colimits I'm not too optimistic. We tried to derive results
about limits from results on colimits in order to only develop the theory
once, but it turned out to be extremely inconvenient because of op's
getting in the way. This was mainly because things were not convertible,
Coq unfolded things too much and some notation sometimes hid which category
the morphisms were in which made the limits extremely confusing to work
with.  In the end we gave up and just had two independent files with more
or less the same theory. I get annoyed every time I need a result for
limits which has only been proved for colimits (and vice versa), but it's
not too hard to just copy the result I need and dualize manually. So it
turned out to not be too bad to have two files in the end, but it is also
not very satisfactory...


On Thu, Feb 23, 2017 at 3:19 AM, Daniel R. Grayson <notifications@github.com

DanGrayson:
Anders, it sounds like an interesting project to figure out how to avoid all that code
duplication, just for having every lemma stated also for the dual category.  Somehow
what little I did with my approach through representations seemed to work.  Here are
my definitions of product and coproduct:
```
Definition Product {C:Precategory} {I} (c:I -> ob C) := Representation (HomFamily C c).
Definition Sum {C:Precategory} {I} (c:I -> ob C) := Representation (HomFamily C^op c).
```
I'll make a new issue so someone can think about the narrow issue of what to do about 
Elements and ElementsOp as well as the broader issue about how to have category 
theory code not doubled in size any longer.

DanGrayson:
This is a 2 line pull request.  I have only one remark for Matthew: it's a change to ElementsOp.v, so why not make the corresponding change also to Elements.v?   Other than that, it looks good to me.

vladimirias:
One more comment on this. Presheaves on the category of elements of a presheaf (ElementsOp) are becoming a key tool in type theory. For example all of the cubical set model is based on these categories.

vladimirias:
I have the same question.

mweav:
The only thing I changed in ElementsOp was where I inverted the paths in cat_data, but as the functors in Elements are covariant the paths never need to be inverted in the first place, so there was no corresponding change to make.

DanGrayson:
lgtm.

Please bring it up to date with the master branch.  There is a conflict now, so I can't do it for you.

vladimirias:
lgtm

mortberg:
@mweav It would be great if you rebased on UniMath master so we can merge this. I think your change is so small that it won't lead to any conflicts, but if you have some I can help you.

Btw, I see that you are using your own master for this PR, I think that is a little dangerous and bad practice. So in the future it is better if you make a new branch for each PR and just try to keep your own master up to speed with the UniMath master.

PR #592, opened 2017-02-21
move HLevel_n_is_of_hlevel_Sn to Foundations

### LINKED ISSUE #523: Hlevel_n_is_of_hlevel_Sn
This file should be in Foundations.

benediktahrens:
resolves #523 

Also, put a few bullets in the proofs of that file.


UniMath/Foundations/HLevel_n_is_of_hlevel_Sn.v
```diff
@@ -159,10 +159,10 @@ Lemma isofhlevelpathspace : ∏ n : nat, ∏ X Y : UU,
 Proof.
   intros n.
   case n.
```
mortberg: What is the `case` tactic generating? Maybe use `induction` instead?

benediktahrens:
On 02/22/2017 02:17 PM, Anders Mörtberg wrote:
> 
> 
> 
>> @@ -159,10 +159,10 @@ Lemma isofhlevelpathspace : ∏ n : nat, ∏ X Y : UU,
>  Proof.
>    intros n.
>    case n.
> 
> What is the `case` tactic generating? Maybe use `induction` instead?

Case is producing a match, but is making clear that the proof is NOT by
induction.
`induction` would work, of course, but obscures the fact that the
inductive hypothesis is not used.

DanGrayson:
So we have to add "case" to the list of tactics not to use, because it generates a "match"...

mortberg:
Maybe it would make sense to have `nat_case` as in the standard library? This would be proved by `induction` but we can write `apply nat_case` instead to make clear that the induction hypothesis is not needed.

mortberg:
LGTM

DanGrayson:
Here's a better way to make it visible that the induction hypothesis is not needed:
```
   induction n as [|n _].
```

DanGrayson:
I'd like to see a more appropriate name for this file, such as `Hlevels.v`.  The current name is laden with notation.

DanGrayson:
Unicode would help readability:
```
diff --git a/UniMath/Foundations/HLevel_n_is_of_hlevel_Sn.v b/UniMath/Foundations/HLevel_n_is_of_hlevel_Sn.v
index c184e977..bb63835c 100644
--- a/UniMath/Foundations/HLevel_n_is_of_hlevel_Sn.v
+++ b/UniMath/Foundations/HLevel_n_is_of_hlevel_Sn.v
@@ -35,8 +35,8 @@ Require Import UniMath.Foundations.UnivalenceAxiom.
 *)
 
 Lemma weq1  (P : UU -> hProp) (X X' : UU) (pX : P X) (pX' : P X') :
-   weq (tpair _ X pX = tpair (fun x => P x) X' pX')
-       (total2 (fun w : X = X' => transportf (fun x => P x) w pX = pX')).
+  tpair P X pX = tpair P X' pX'
+     ≃ ∑ w : X = X', transportf P w pX = pX'.
 Proof.
   apply total2_paths_equiv.
 Defined.
```

mortberg:
Actually, why is this there at all? Why not just use total2_paths_equiv
directly? It's not too good that the two lemmas weq1 and weq2 get
exported...

On Wed, Feb 22, 2017 at 2:52 PM, Daniel R. Grayson <notifications@github.com

DanGrayson:
Replace `(fun x => P x)` by `P`, and similarly with `(fun X : UU => isofhlevel n X)` and `(fun X => isaprop X)`.

DanGrayson:
With \chimney it looks even more beautiful:
```
Lemma weq1  (P : UU -> hProp) (X X' : UU) (pX : P X) (pX' : P X') :
  tpair P X pX = tpair P X' pX'
     ≃
  tpair P X pX ╝ tpair P X' pX'.
Proof.
  apply total2_paths_equiv.
Defined.
```
Making this short names local would be good, as Anders says.

benediktahrens:
The file was written in autumn of 2012, I think. Please don't ask me why
things are the way they are.

DanGrayson:
lgtm

mortberg:
LGTM

vladimirias:
This is not a good name because it may appear that this is the file where h-levels are introduced. Maybe we will make more descriptive names for BasicsA-D eventually and then we will need the name hlevels.

benediktahrens:
On 02/23/2017 12:03 AM, Vladimir Voevodsky wrote:
> This is not a good name because it may appear that this is the file
> where h-levels are introduced. Maybe we will make more descriptive
> names for BasicsA-D eventually and then we will need the name
> hlevels.

This PR has been merged, but we can rename that file in a future PR.

vladimirias:
hlevel_of_HLevel ?

DanGrayson:
Re: "hlevel_of_HLevel"

That's still a name with identifiers in it, rather than English.  Probably a future reorganization
will assemble the results into meaningful categories, renaming and remixing PartA, PartB, PartC, PartD, and this one.

vladimirias:
> On Feb 22, 2017, at 8:53 PM, Daniel R. Grayson <notifications@github.com> wrote:
> 
> PartA, PartB, PartC, PartD,


Should not be re-mixed. The order of things there is exactly or nearly exactly what it is intended to be. It is not at all random even if might not always be obvious why this or that choice has been made. 

In fact I think that we should lock the  Basics and possibly some other parts of the original Foundations against any further changes except for the ones that I am going to introduce while adding comments and cleaning up. The results going “upstream” will go instead to a new layer around this Foundations_1. This way we will be able to develop the Foundations without destroying the order that exists in the original construction. 

Vladimir.

DanGrayson:
Well, yes, the sequence can't be disturbed.  But the dividing lines might be moved, and the
file currently under discussion might be incorporated into one of those.

In light of your remark, perhaps you have an opinion about #593 .

vladimirias:
Yes, I do. I did not have time today to look at it.

I think that my suggestion about locking Basics was quite timely.

Please do not introduce any changes to Basics. I can not explain for every line why it was chosen to be in this form and not in that one.

Let us think about how to design a new layer outside of the original Foundations that will be used to accommodate all of the upstream flow.

Vladimir.

DanGrayson:
It's not called Basics, it's called Foundations now.

Do you want to reject #593 ?

vladimirias:
For now yes. 

Sorry Dan,  but I can not spend my time on re-explaining each time a change is suggested for something in Foundations what were my reasons, not expressed explicitly in the code or comments, to make things the way they were made and in what way the proposed change will disturb it.

It is a hard problem in general, but in our case we do have what seems to me to be a really good solution through locking the original construction and building a layer outside with additions and, if need be, changes since Coq allows to shadow old identifiers.

Please tell me if I am missing something and there are some drawbacks to this suggestions that I do not see.

Vladimir.

PS After all a paper copy publication remains, unless burned, in the form in which it was intended by the author.

DanGrayson:
Okay, so perhaps we need a package inserted between Foundations and Combinatorics
in the list of packages:
```
    Foundations
    Combinatorics
    Algebra
    NumberSystems
    CategoryTheory
    Ktheory
    Topology
    RealNumbers
    Tactics
    SubstitutionSystems
    Folds
    HomologicalAlgebra
```
It would be a basin for useful results that are too general to fit into any later category.  At the moment I can't think of a good word.  Utilities?  Auxiliary?  Extra?  Miscellaneous?

DanGrayson:
Will locking Foundations imply that we never switch to Primitive Projections?

DanGrayson:
Re: "After all a paper copy publication remains, unless burned, in the form in which it was intended by the author."

The git repository retains a complete history, too, in every state, after every commit.  Here, for example:

https://github.com/UniMath/UniMath/tree/vv-master-for-coq8.4pl3patched

So don't worry about losing anything.

DanGrayson:
Re: "Please tell me if I am missing something and there are some drawbacks to this suggestions that I do not see."

No, your proposal is sensible.  Your original Foundations code was a great accomplishment, and sprinkling additional lemmas throughout it might distract the future reader who wants to step through the proofs for inspirational motivation and for an introduction to the new foundational scheme for encoding modern math.

vladimirias:
I would like to also move the portions of the files that we took out of Foundations that were originally there back to Foundations under the lock. There is a lot of structure there as well that I want to have preserved.

DanGrayson:
Re: "I would like to also move the portions of the files that we took out of Foundations that were originally there back to Foundations under the lock."

Are there any such portions?  Perhaps you could point them out.

vladimirias:
Not necessarily. I don’t see how switching to primitive projections would destroy any structure in Foundations.

I will still have access to Foundations and we will keep updating what is needed to stay up to date with Coq. 

Who knows, maybe the day will come when we will even have means in Coq to add resizing rules and fix the current lack in universe management.

DanGrayson:
Re: "After all a paper copy publication remains, unless burned, in the form in which it was intended by the author."

Perhaps it would be worthwhile to have a separate branch where we maintain the old Foundations code in exactly the form it was published, except for the minimal changes required to get it to run with the current public release of Coq, and adding something to the Makefile to generate the TAGS file, if requested, and adding enough emacs stuff to make ProofGeneral run Coq with the right command line options.  Then readers of the paper would be able to step through it interactively in ProofGeneral.

vladimirias:
This would be nice. It does not, however, replace the idea of having mostly locked Foundations_0 at the core of the UniMath with the adjustments made in the next layer e.g. Foundations_1.

DanGrayson:
Okay.  At some point it would be good if you would formulate your locking idea precisely, so it could be incorporated into the style guide.

Issue #365, opened 2016-06-18
Moving contents from Ktheory.Utilities to Foundations


benediktahrens:
The module Ktheory.Utilities has many useful lemmas, in particular about transport. They would be even more useful if they were in a more canonical place, such as Basics.PartX.
Any volunteers for moving them there?

DanGrayson:
Sure, I can do it.  The summer is about to finally "begin" for me tomorrow.

benediktahrens:
At the moment, there is also this in Utilities:
```
Axiom PostponedProof : empty.
Ltac postponeProof := apply fromempty, PostponedProof.
```
Is there an advantage of this custom tactic over the `admit.` tactic?
The `admit.`tactic has the advantage that it only allows `Admitted.` and `Abort.` at the end of a proof, thus making the status of the proof clear.

DanGrayson:
I guess not, since I didn't use it even once.

benediktahrens:
Also, Cathy uses stuff from 

- UniMath/Ktheory/Tactics
- UniMath/Ktheory/Utilities
- UniMath/Ktheory/QuotientSet

in her Topology, in particular she uses `Sequence`.

Could that go somewhere in Basics, too?

DanGrayson:
Okay.

benediktahrens:
There is some stuff about categories that would be good to have in CT instead of Ktheory, see, e.g., discussion in #502 .
To keep #502 tractable, let's defer this to later, but not too much later.

DanGrayson:
Closing, see https://github.com/UniMath/UniMath/pull/592#issuecomment-281873412

benediktahrens:
On 02/23/2017 04:10 AM, Daniel R. Grayson wrote:
> Closing, see https://github.com/UniMath/UniMath/pull/592#issuecomment-281873412

Instead of closing this issue, I vote for renaming it into "... to
Auxiliary", for a package "Auxiliary" (or whatever) still to be created.

DanGrayson:
Okay, I'll reopen it again when it's ready to be reconsidered.

"Auxiliary" is perhaps not so good, since it conveys the impression that the items within are not needed.  "Basics" is free, but that might confuse the history of the project.  I suggest "Miscellany" tentatively.  That conveys the idea that the items within don't fit into one of the other categories, and its position as the second item in the list of packages conveys the idea that the items within are supposed be useful to subsequent packages, which we can also say in the README.

[...]

PR #593, opened 2017-02-21
Move things upstream

DanGrayson:
Resolves https://github.com/UniMath/UniMath/issues/365

Also removed a file.


benediktahrens:
Comes with a change in sub/coq.

DanGrayson:
Thanks, fixed.

Makefile
```diff
@@ -103,7 +103,8 @@ COQDEFS := --language=none																\
 	-r '/^[[:space:]]*Tactic[[:space:]]+Notation.*[[:space:]]"\([[:alnum:]'\''_]+\)"[[:space:]]/\1/'										\
 	-r '/^[[:space:]]*Delimit[[:space:]]+Scope[[:space:]]+[[:alnum:]'\''_]+[[:space:]]+with[[:space:]]+\([[:alnum:]'\''_]+\)[[:space:]]*\./\1/'
 
-$(foreach P,$(PACKAGES),$(eval TAGS-$P: $(filter UniMath/$P/%,$(VFILES)); etags -o $$@ $$^))
+$(foreach P,$(PACKAGES),$(eval VFILES_$P := $(filter UniMath/$P/%,$(VFILES))))
```
benediktahrens: Could you explain this change to the Makefile, please?

UniMath/Foundations/PartA.v
```diff
@@ -746,7 +802,59 @@ Proof. intros. destruct e. reflexivity. Defined.
 Definition transportfbinv {T} (P:T->Type) {t u:T} (e:t = u) (p:P u) : e#e#'p = p.
 Proof. intros. destruct e. reflexivity. Defined.
 
-Close Scope transport.
+Definition transport_type_path {X Y:Type} (p:X = Y) (x:X) :
```
benediktahrens: Would it be good to make this the definition of `cast`, instead of a consequence?

UniMath/Foundations/PartA.v
```diff
@@ -746,7 +802,59 @@ Proof. intros. destruct e. reflexivity. Defined.
...
+Definition transportf_pathsinv0' {X} (P:X->UU) {x y:X} (p:x = y) (u:P x) (v:P y) :
+  p # u = v -> !p # v = u.
+Proof. intros ? ? ? ? ? ? ? e. induction p, e. reflexivity. Defined.
+
+Lemma transport_idfun {X} (P:X->UU) {x y:X} (p:x = y) (u:P x) :
+  transportf P p u = transportf (idfun _) (maponpaths P p) u.
+(* same as HoTT.PathGroupoids.transport_idmap_ap *)
+Proof. intros. induction p. reflexivity. Defined.
+
+Lemma transport_functions {X} {Y:X->Type} {Z:∏ x (y:Y x), Type}
```
benediktahrens: The use of `Type` is discouraged, I think.

UniMath/Foundations/PartA.v
```diff
@@ -746,7 +802,59 @@ Proof. intros. destruct e. reflexivity. Defined.
...
+      {f f':∏ x : X, Y x} (p:f = f') (z:∏ x, Z x (f x)) x :
+    transportf (fun f => ∏ x, Z x (f x)) p z x =
+    transportf (Z x) (toforallpaths _ _ _ p x) (z x).
+Proof. intros. induction p. reflexivity. Defined.
+
+Definition transport_funapp {T} {X Y:T->Type}
```
benediktahrens: Again, use of `Type`.
Also, what happens if you use `{T}` instead of `{T : UU}`? What type is assigned to `T`?

UniMath/Foundations/PartA.v
```diff
@@ -746,7 +802,59 @@ Proof. intros. destruct e. reflexivity. Defined.
...
+  (transportf (fun y => ∏ t, P t y) e k) t.
+Proof. intros. induction e. reflexivity. Defined.
+
+Definition helper_B {T} {Y} (f:T->Y) {y y':Y} (k:∏ t, y = f t) (e:y = y') t :
+  transportf (fun y => y = f t) e (k t)
+  =
+  (transportf (fun y => ∏ t, y = f t) e k) t.
+Proof. intros. exact (helper_A _ k e t). Defined.
+
+Definition maponpaths_pr1_pr2 {X} {P:X->UU} {Q:∏ x, P x->Type}
```
benediktahrens: `Type`

UniMath/Foundations/PartA.v
```diff
@@ -2388,6 +2544,15 @@ Proof.
     reflexivity.
 Defined.
 
+(* associativity of ∑ *)
+Definition totalAssociativity {X:UU} {Y: ∏ (x:X), UU} (Z: ∏ (x:X) (y:Y x), UU) : (∑ (x:X) (y:Y x), Z x y) ≃ (∑ (p:∑ (x:X), Y x), Z (pr1 p) (pr2 p)).
```
benediktahrens: We should stick to one of `total2` and `total`, not mix them.

UniMath/Foundations/PartD.v
```diff
@@ -419,6 +418,35 @@ Definition weqonsecbase {X Y : UU} (P : Y -> UU) (f : weq X Y)
   : weq (∏ y : Y, P y) (∏ x : X, P (f x))
   := weqpair _ (isweqmaponsec1 P f).
 
+Definition weqonsec {X Y} (P:X->Type) (Q:Y->Type)
```
benediktahrens: This one looks like a duplicate, but I am not sure. Also, `Type`...

UniMath/Foundations/Sets.v
```diff
@@ -437,6 +440,14 @@ Proof.
   exact (pr1 p).
 Defined.
 
+Lemma squash_pairs_to_set_compute {Y : UU} {F : Y -> UU}
+      (i:isaset Y) (e:∏ y y', F y -> F y' -> y = y') {y} (f : F y) :
+  squash_pairs_to_set _ i e (hinhpr (y,,f)) = y.
+(* upstream *)
```
benediktahrens: Comment `upstream` can be removed?

UniMath/Foundations/Sets.v
```diff
@@ -458,6 +469,14 @@ Proof.
   exact (pr1 p).
 Defined.
 
+Lemma squash_to_set_compute {X Y : UU} (is : isaset Y) (f : X -> Y)
+      (e:∏ x x', f x = f x') x :
+  squash_to_set is f e (hinhpr x) = f x.
+(* upstream *)
```
benediktahrens: Again, comment `upstream`.

UniMath/Foundations/UnivalenceAxiom.v
```diff
@@ -32,11 +32,24 @@ consequences of univalence, rather than the theorems giving the implications.
 
 Require Export UniMath.Foundations.PartB.
 
-(* everything related to eta correction is obsolete *)
-
 Definition eqweqmap { T1 T2 : UU } : T1 = T2 -> T1 ≃ T2.
 Proof. intro e. induction e. apply idweq. Defined.
 
+Definition eqweqmap_ap {T} (P:T->Type) {t t':T} (e:t = t') (f:∏ t:T, P t) :
+  eqweqmap (maponpaths P e) (f t) = f t'. (* move near eqweqmap *)
```
benediktahrens: remove comment

UniMath/Foundations/UnivalenceAxiom.v
```diff
@@ -32,11 +32,24 @@ consequences of univalence, rather than the theorems giving the implications.
...
+Proof. intros. induction e. reflexivity. Defined.
+
+Definition eqweqmap_ap' {T} (P:T->Type) {t t':T} (e:t = t') (f:∏ t:T, P t) :
+  invmap (eqweqmap (maponpaths P e)) (f t') = f t. (* move near eqweqmap *)
```
benediktahrens: remove comment

Makefile
DanGrayson: The Makefile already allowed for making not only the file TAGS, but also a tags file for each package, e.g., TAGS-Ktheory.  But the etags command for them was lacking the command line options that allowed it to locate the Coq definitions.  I also made those other TAGS file depend on the Makefile, so they would get remade if the Makefile changed, for example, if the command line options are changed.

UniMath/Combinatorics/FiniteSets.v
```diff
@@ -206,6 +206,12 @@ Defined.
 
 Definition cardinalityFiniteSet (X:FiniteSet) : nat := fincard (pr2 X).
 
+Lemma fincard_compute {n} {F:FiniteSet} : nelstruct n F -> cardinalityFiniteSet F = n.
+(* upstream *)
```
DanGrayson: That's an obsolete comment, will remove.

UniMath/Foundations/PartA.v
DanGrayson: That's a good idea to keep in mind for later.

UniMath/Foundations/PartA.v
DanGrayson: will fix

UniMath/Foundations/PartA.v
DanGrayson: Sometimes Type, sometimes UU.  It depends on whether T gets used as a parameter in something that accepts only UU.  Will fix.

If we eventually get universe management working again, Coq will tell us where we have a problem, or take care of it for us, so we don't have to worry about this sort of thing, since it won't, at the moment, affect whether a proof is acceptable.

UniMath/Foundations/PartA.v
DanGrayson: fixing...

UniMath/Foundations/PartA.v
DanGrayson: Well, but here I was using the English word "total", as in "total space", not the identifier "total2", in which case I would have put an underscore after it.

benediktahrens:
On 02/22/2017 01:28 PM, Daniel R. Grayson wrote:
> DanGrayson commented on this pull request.
> 
> 
> 
>> @@ -2388,6 +2544,15 @@ Proof.
> reflexivity. Defined.
> 
> +(* associativity of ∑ *) +Definition totalAssociativity {X:UU} {Y: ∏
> (x:X), UU} (Z: ∏ (x:X) (y:Y x), UU) : (∑ (x:X) (y:Y x), Z x y) ≃ (∑
> (p:∑ (x:X), Y x), Z (pr1 p) (pr2 p)).
> 
> Well, but here I was using the English word "total", as in "total
> space", not the identifier "total2", in which case I would have put
> an underscore after it.

Well, that explanation makes it much easier to remember...

Anyway, is this not the same as weqtotal2asstor or something like that?

UniMath/Foundations/PartD.v
DanGrayson: I think it's not a duplicate.

UniMath/Foundations/Sets.v
DanGrayson: ok

UniMath/Foundations/Sets.v
DanGrayson: ok

DanGrayson:
Yes, there are a zillion of those things.  I thought this one was different at the time...

DanGrayson:
Re: "In general, I'd prefer if changes to the Makefile were in separate PRs, for easier rollback in case of problems."

Okay, but rollback would still be almost trivial.  And I'll be around.

DanGrayson:
Closing, see https://github.com/UniMath/UniMath/pull/592#issuecomment-281873412

benediktahrens:
On 02/23/2017 03:54 AM, Daniel R. Grayson wrote:
> Closing, see https://github.com/UniMath/UniMath/pull/592#issuecomment-281873412

I would like us to pursue the idea with an additional package "between"
Foundations and the other packages.

Would it make sense to pursue that project in this PR?

In TypeTheory, we have a package "Auxiliary" where we have a few helper
files:

- a file listing "Require Export"s that we frequently need
- a file that serves as basin for basic results

Something similar should happen here.

vladimirias:
I still have not submitted the grant application and over the head in other stuff as well. I want to participate :)

Maybe something temporary just to get all of this “move upstream” work merged.

Issue #472, opened 2016-09-26
need notation


vladimirias:
We need to decide on a good infix notation for the composition written in the diagrammatic order.

benediktahrens:
On 09/26/2016 05:45 PM, Vladimir Voevodsky wrote:
> We need to decide on a good infix notation for the composition
> written in the diagrammatic order.

Are you talking about composition in a (pre)category, or about
composition of functions?

mortberg:
For morphisms in (pre)categories we usually write `f ;; g` which I like. The code to introduce this notation is:

https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/precategories.v#L98

Note the two spaces around `;;` in the format, without this I think Coq will print the notation without any spaces.

DanGrayson:
There are two diagrammatic orders, one derived from the diagram `X->Y->Z` and one derived from the diagram `Z<-Y<-X'.

DanGrayson:
Since there two diagrammatic orders, I suggest introducing both ` g ⇴ f ` and `
f ⬰ g `. The latter one would be the same as ` f ∘ g `.  The arrow suggests the diagram aspect and the circle suggests the composition aspect.

vladimirias:
There is only one diagrammatic order - the order that you obtain by following the arrows in the digram in the direction of the arrow. In your second diagram it is exactly the same as in the first.

How is  g ⇴ f called in the Agda mode?

DanGrayson:
Re: "There is only one diagrammatic order - the order that you obtain by following the arrows in the digram in the direction of the arrow."

What do you mean, exactly?  I'll spell out what I meant.  It depends on whether the direction of the arrows in the diagram is leftward or rightward.  For example, in the diagram `Z <--g-- Y <--f-- X` the diagrammatic composition order would be `g o f`, but in the diagram `X --f--> Y --g--> Z` the diagrammatic composition order would be `f o g`.

DanGrayson:
Re: "How is  g ⇴ f called in the Agda mode?"

It's accessible only from the right arrow menu with `\r`, row 4, entry 5, and there are 70 arrows in that menu.  But we can add a name as we did for `\chimney`.

vladimirias:
No, not at all. There is only one diagrammatic order. In both diagrams that you drew it is first f and then g.

[...]

PR #595, opened 2017-02-22
add notation for funcomp the other way

DanGrayson:
resolves issue #472

### DIFF
--- a/UniMath/.dir-locals.el
+++ b/UniMath/.dir-locals.el
@@ -14,8 +14,8 @@
 	     (modify-syntax-entry ?_ "w")
 	     (if (not (memq 'agda-input features))
 		 (load (concat unimath-topdir "emacs/agda/agda-input")))
-	     (if (not (member '("chimney" "╝") agda-input-user-translations))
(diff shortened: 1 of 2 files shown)


benediktahrens:
lgtm

benediktahrens:
Actually, I would like to retract my "lgtm". This symbol is barely readable. Now, some people will say that that is due to a bad font, not due to the symbol itself, but that is misleading: I cannot (easily) change font on Github, for instance.

DanGrayson:
We have a suggestion from Anders on another thread to use ";;" instead, which seems fine to me.  Vladmir?

I want to add the \cdot stuff to this pull request, anyway.  If we adopt ";;" we may have to stop using it completely for "compose", but that won't be a problem, except for the need to update TypeTheory.

vladimirias:
Let’s drop it for now. In Basics I may want to use another notation. For now funcomp is fine.

vladimirias:
\cdot stuff is good to make a separate PR since we all agree on it.

PR #596, opened 2017-02-22
Simplicial sets

### LINKED ISSUE #373: Remove `Admitted.``` in Foundations/Combinatorics/SimplicialSets.v

DanGrayson:
resolve #373


benediktahrens:
lgtm

vladimirias:
The set monfunstn m n is actually the set of monotone functions (monot_fun_stn might be better) from stn (S m) to stn (S n). It either should be called monot_fun_stn_sn or (S _) replaced by _ in the definition 

One can also have a notation Delta_mor m n for monote_fun_stn (S m) (S n).

DanGrayson:
What if I just remove those lemmas and incorporate the definitions into `precatDelta` instead?  From the category the Hom sets can be obtained, after all.

DanGrayson:
Re: "What if I just remove those lemmas and incorporate the definitions into precatDelta instead?"

Never mind, I figured out your comment and have made the change.

mortberg:
LGTM (modulo the fact that it's out-of-date with master). 

Once my PR about presheaves is merged sSet could be defined as `PreShv precatDelta`.

PR #598, opened 2017-02-22
proscribe tactic "case"

DanGrayson:


DanGrayson:
I cancelled the travis build, since this changes just `UniMath/INSTALL.md`, and there's no reason to waste machine time.

PR #599, opened 2017-02-22
From multisorted signatures to monads

mortberg:
Instantiate the framework for constructing a monad from a signature with strength whose underlying functor is omega cocontinuous with the signature obtained from a multisorted signature. 

To do this I also had to construct initial objects in slice categories.


DanGrayson:
lgtm

benediktahrens:
lgtm

rmatthes:
lgtm

PR #601, opened 2017-02-23
fix the Makefile

### LINKED ISSUE #555: problems with running "make" in parallel
These lines
```
sub/coq/configure sub/coq/configure.ml:
	git submodule update --init sub/coq
```
in our makefile cause a problem when running make in parallel, because it thinks it can run
two copies of git at the same time to make those two files -- one of the git processes
will fail, locked out by the other.  It also involves these lines:
```
sub/coq/config/coq_config.ml: sub/coq/configure sub/coq/configure.ml
	: making $@ because of $?
	cd sub/coq && ./configure -coqide "$(COQIDE_OPTION)" -with-doc no -annotate -debug -local
```
whose purpose is to run the Coq configure script again when necessary.

Work arounds:
   -  when using "git clone" initially, add the "--recurse-submodules" option
   -  when running "make" the first time, don't run it in parallel (no "-j" option)

I don't know the best way to deal with it.

DanGrayson:
resolves #555

These lines
```
sub/coq/configure sub/coq/configure.ml:
	git submodule update --init sub/coq
```
in our makefile cause a problem when running make in parallel, because it thinks it can run
two copies of git at the same time to make those two files -- one of the git processes
will fail, locked out by the other.  It also involves these lines:
```
sub/coq/config/coq_config.ml: sub/coq/configure sub/coq/configure.ml
	: making $@ because of $?
	cd sub/coq && ./configure -coqide "$(COQIDE_OPTION)" -with-doc no -annotate -debug -local
```
whose purpose is to run the Coq configure script again when necessary.

The solution is to have just one prerequisite for sub/coq/config/coq_config.ml.  We
take out sub/coq/configure, because it's more stable.


benediktahrens:
lgtm

PR #603, opened 2017-02-23
fix an abort

AnthonyBordg:
Remove an abort and give a complete proof.


DanGrayson:
The comment above it indicates that the purpose of that Goal is to demonstrate whether eta expansion for pairs works in your current proof of Coq.  If you "fix" the proof, it won't demonstrate that.  Eventually, when our code is compatible with it, we'll turn on the option "Primitive Projections", and that proof will start working.  So it should stay the way it is.

AnthonyBordg:
Hi Dan,

does it mean that currently eta expansion for pairs is not available in UniMath ?

DanGrayson:
Yes, it breaks some of our proofs.  I'm working on it today.

PR #607, opened 2017-02-23
Some theory about presheaf categories

mortberg:
Define the (set-valued) presheaf category Psh C as [C^op,HSET] and prove various things:

- Construction of some (co)limits (all of these are lifted from HSET to the functor category)
- Definition of the subobject classifier in Psh C as the set of sieves (there is no proof yet that this actually is the subobject classifier as this notion hasn't been defined in UniMath yet)
- Various minor results (morphisms from the terminal object are monic, morphisms to the initial object are epi and construction of the terminal object in functor categories from the target category)

### DIFF
--- a/UniMath/CategoryTheory/.package/files
+++ b/UniMath/CategoryTheory/.package/files
@@ -99,3 +99,4 @@ FiveLemma.v
 set_slice_fam_equiv.v
 Elements.v
 ElementsOp.v
+Presheaf.v
--- a/UniMath/CategoryTheory/Epis.v
+++ b/UniMath/CategoryTheory/Epis.v
@@ -17,8 +17,6 @@ Local Open Scope cat.
 Require Import UniMath.CategoryTheory.sub_precategories.
 Require Import UniMath.CategoryTheory.functor_categories.
 
-
-
 (** * Definition of Epis *)
 Section def_epi.
 
--- a/UniMath/CategoryTheory/Monics.v
+++ b/UniMath/CategoryTheory/Monics.v
@@ -14,7 +14,6 @@ Local Open Scope cat.
 Require Import UniMath.CategoryTheory.sub_precategories.
 Require Import UniMath.CategoryTheory.functor_categories.
 
-
 (** * Definition of Monics *)
 Section def_monic.
 
--- a/UniMath/CategoryTheory/Presheaf.v
+++ b/UniMath/CategoryTheory/Presheaf.v
@@ -14,7 +14,7 @@ Contents:
 - Terminal object ([Terminal_Psh])
 - Pullbacks ([Pullbacks_Psh])
 - Exponentials ([has_exponentials_Psh])
-- Definition of the subobject classifier (without proof) ([Ω_Psh])
+- Definition of the subobject classifier (without proof) ([Ω_Psh], [Ω_mor])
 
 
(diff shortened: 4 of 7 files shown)


DanGrayson:
Looks good to me.

But why not rename Psh to Presheaf?  It would be less cryptic.

vladimirias:
Or PreShv, this is what I use in my papers.

vladimirias:
I want to have a look. I had some ideas on how to make b : F Y  and # F f b for f:X-->Y well typed.

benediktahrens:
To add another variant, in UniMath/TypeTheory, we use `preShv`.

mortberg:
I got the abbreviation from: https://ncatlab.org/nlab/show/category+of+presheaves (But as you see I changed the capitalization...)

Many books also use \hat{C} (according to MacLane/Moerdijk this is the notation of the French school, which I guess refers to the Grothendieck school). This isn't easy to type in emacs though and the best approximation I can think of is C^ which is quite ugly. 

I like Vladimir's suggestion, it's just a little longer and a lot more informative. I'm not a fan of having a lowercase p as first letter as I like having lowercase first letter for elements of a category and uppercase for categories.

mortberg:
I renamed Psh to PreShv now and rebased on master. I think this should be ready to be merged as soon as travis is done

DanGrayson:
Vladimir said he wants to have a look first, so let's wait for him to say something before merging.

mortberg:
I've put this up to speed with the master branch now

mortberg:
I just added the discrete presheaf defined by a set A. This felt too simple to put in another PR so I added it to this one.

DanGrayson:
Isn't it called the constant presheaf?

mortberg:
I was thinking about discrete simplicial/cubical sets. There are constant functors to Set right? So I wasn't sure if I should call it a constant or discrete presheaf?

DanGrayson:
It's definitely called a constant presheaf, not a discrete presheaf.  Only when the category is something like Delta do we think of those things as spaces, so that a word like "discrete" would be applicable.

mortberg:
Ok, thanks. I updated the name now.

DanGrayson:
lgtm

mortberg:
Maybe we could merge this PR now? It has been open for 4 weeks and it would be great to get the code into UniMath. I have some lemmas that I think could be useful for other people (e.g. Matthew Weaver), but these should go into a new PR as I don't want to make this one any bigger. I would also like to experiment with making PreShv into a definition instead of a notation so that I can add a coercion to functor and maybe get rid of all the explicit pr1's that are needed now. 

Maybe Vladimir could try out his ideas on the merged code when he has time or make an issue about it so that we don't forget and can experiment later.

DanGrayson:
Not too long ago Vladimir emphasized to me that he still wants to take a look at this, which perhaps he will have time to do after March 25.  On the other hand, as you say, he could look at it even after the merge.  @vladimirias ?

benediktahrens:
This PR has been around for far too long.
I suggest merging this PR and making an issue about the problem given by the need for `pr1`s.

mortberg:
I merged master and fixed the conflicts now. I hope travis will be happy.

benediktahrens:
lgtm

PR #611, opened 2017-02-24
Adjointification of an equivalence of categories

benediktahrens:
This PR prepares for the adjointification. The purpose of this PR is to introduce a type `adjunction_data` (two functors, two nat trans) and to define two properties on that type: to form an adjunction, and to form an equivalence.

Sometimes one would like to have the left or the right adjoint as a parameter instead of as a component, or vice versa. The somewhat non-standard coercion introduced in this PR does this for one such case.

EDIT: I have finished the adjointification. There is room for improvement, organization-wise and speed-wise. In another PR...


DanGrayson:
lgtm

UniMath/CategoryTheory/equivalences.v
```diff
@@ -85,6 +98,167 @@ Proof.
...
+  - apply forms_equivalence_adjointification.
+  - apply adjointification_triangle_1.
+Qed.
+
+Definition adjointificiation : adj_equivalence_of_precats F.
+Proof.
+  mkpair.
+  - mkpair.
+    + apply G.
+    + mkpair. mkpair. apply ηntiso. apply ε'isont.
```
mortberg: Bullets missing after `mkpair`. A trick I use to not get too deep nested bullets is to use `exists foo` when I can give the first component directly.

UniMath/CategoryTheory/equivalences.v
```diff
@@ -85,6 +98,167 @@ Proof.
...
+      exists adjointification_triangle_1.
+      apply adjointification_triangle_2.
+  - apply forms_equivalence_adjointification.
+Defined.
```
mortberg: I have a feeling that this proof could look nicer if there was a `mk_adj_equivalence_of_precats` where you give the arguments "at the same level" instead of going "deeper and deeper" with nested mkpair.

mortberg:
There are a lot of `Defined` and I think some of them could be `Qed`, or at least parts could be `abstract`. Apart from this it LGTM

UniMath/CategoryTheory/equivalences.v
DanGrayson: Another trick to avoid nested bullets is to switch to braces.  Then the last new goal needn't be enclosed.

benediktahrens:
Thanks for the comments.
The `Defined` are important: being an iso is like being an equivalence, you want to extract the inverse.

benediktahrens:
Closing this in favor of #619.

PR #612, opened 2017-02-24
improve the Makefile

DanGrayson:
... by logging the success of checking whether travis.yml checks all the packages
and whether the *.v files are encountered in the correct linear order, so the
checks are not redone unless they need to be.


benediktahrens:
lgtm

PR #613, opened 2017-02-25
\cdot compose notation

DanGrayson:
... and put all the category notations into a scope called "cat"
... and remove the file UnicodeNotations.v
... and change all uses of "f;;g" to "f·g"

Users who formerly would just load UnicodeNotations.v should now load
the needed files from the ones formerly loaded by UnicodeNotations,
which were these:

    Require Export UniMath.Foundations.Propositions.
    Require Export UniMath.CategoryTheory.precategories.
    Require Export UniMath.CategoryTheory.functor_categories.
    Require Export UniMath.CategoryTheory.opp_precat.

, and also add

    Local Open Scope cat.

to make the notations within the scope active.

To use the scope without opening it, write a term of the form '(.....)%cat'.

The notation "f;;g" is still present, but is deprecated.

Other category theory notations were also added to the scope "cat",
namely these:

    Notation "f ;; g" := (compose f g) (at level 50, left associativity, only parsing) : cat. (* deprecated *)
    Notation "f · g" := (compose f g) (at level 50, format "f  ·  g", left associativity) : cat.
    Notation "# F" := (functor_on_morphisms F)(at level 3) : cat.
    Notation "G ∙ F" := (functor_composite _ _ _ F G) (at level 35) : cat.
    Notation "F ⟶ G" := (nat_trans F G) (at level 39) : cat.
    Notation "[ C , D , hs ]" := (functor_precategory C D hs) : cat.
    Notation "C '^op'" := (opp_precat C) (at level 3, format "C ^op") : cat.
    Notation "a --> b" := (precategory_morphisms a b)(at level 50) : cat.
    Notation "C ⟦ a , b ⟧" := (precategory_morphisms (C:=C) a b) (at level 50) : cat.


benediktahrens:
It is weird to write composition in categories in the good order, but to write composition of functors in the bad order.
I have used the notation `F \. G` for composition of functors in the files Adjunctions.v and equivalences.v, see #611. I suggest adopting this notation more globally.

benediktahrens:
This might also be the right moment to reconsider functor categories.
Should it rather be
```
Notation "[ C , D ]" :=  (functor_precategory C D) : cat.
```
with D being a `Precategory`?

benediktahrens:
This PR is going to produce quite some conflicts with other current PRs, and making new PRs does not make any sense while this one hasn't been merged. Could we move swiftly on this PR and the other outstanding ones that are concerned, as to not stall work on UniMath for longer than necessary?

UniMath/CategoryTheory/AbelianPushoutPullback.v
```diff
@@ -19,7 +19,7 @@ Require Import UniMath.CategoryTheory.limits.Opp.
 
 Require Import UniMath.CategoryTheory.precategories.
 Require Import UniMath.CategoryTheory.opp_precat.
-Require Import UniMath.CategoryTheory.UnicodeNotations.
+Local Open Scope cat.
```
mortberg: Could you please put the `Local Open Scope cat` after the imports? I think it looks better if one first lists imports and then which scopes are opened at the top of the file.

UniMath/CategoryTheory/CommaCategories.v
```diff
@@ -26,10 +26,8 @@ Require Import UniMath.Foundations.Sets.
 
 Require Import UniMath.CategoryTheory.precategories.
 Require Import UniMath.CategoryTheory.functor_categories.
-Require Import UniMath.CategoryTheory.UnicodeNotations.
+Local Open Scope cat.
 
-Local Notation "# F" := (functor_on_morphisms F)(at level 3).
-Local Notation "F ⟶ G" := (nat_trans F G) (at level 39).
 Local Notation "G □ F" := (functor_composite _ _ _ F G) (at level 35).
```
mortberg: Is this notation ever used?

mortberg:
Could you please add some comments explaining how to type the various new notations in Agda mode? Or if there is a standard way to figure out how to do it we should write it somewhere. I always get puzzled when there is some new symbol for something and I can't suddenly write what I used to (last night it was sum and prod..).

mortberg:
This PR is very very long, so I couldn't look at all of it. I agree with Benedikt that we should merge this ASAP, I also think that we shouldn't make any more big scale changes in this PR because it is already almost impossible to look at all the code that is changed. 

Regarding notation for functor categories, why not add the one Benedikt suggests as well? We can then phase out the old notation.

It would be nice to have some notations for slice categories (introduced in slicecat.v) in the scope.

mortberg:
I think it should be (note the uppercase P): 
```
Notation "[ C , D ]" :=  (functor_Precategory C D) : cat.
```

I would also add some formatting to it so that it gets displayed nicely (right now I think there are too many spaces around), but this can be done later.

vladimirias:
Can we, where a notation is introduced, also to say how to write it in he Agda mode?

vladimirias:
Yes, I have not noticed. The functor composition should to be in the direct order. Otherwise reading equations off the diagrams of functors is really difficultm, especially taking into account non-strict associativity!

vladimirias:
Can we have both?

DanGrayson:
I will add those comments when fixing the conflicts.

There is a way to use emacs to ask how to type a unicode character, if you have an example of one in front of you: make sure agda input mode is enabled (so you see ∏ in the mode line), position the cursor on the character and use C-u C-x =

In the response is an item labeled "to input", which shows how to type it in agda.

DanGrayson:
By the way, I found a way to alleviate some confusion when trying to make arrows in C^op, with this notation:
```
Notation "a <-- b" := (@precategory_morphisms (opp_precat _) a b) (at level 50) : cat.
```

DanGrayson:
Re: "The functor composition should to be in the direct order. Otherwise reading equations off the diagrams of functors is really difficult..."

One can also just draw the diagram right to left and bottom to top, like this "A <---- B <---- C <---- D", to avoid the difficulty.

vladimirias:
What if I see one in a PR?

DanGrayson:
Copy and paste it into emacs.

vladimirias:
I think this will confuse me even more. The direct order of composition is a convenient thing. It only creates a little extra writing when a function (or a functor) is applied to an element (an object) so that F(G(X))=(G F)(X), but I have not noticed any real problems with it and I have used the direct notation a lot now in many many formulas in my papers.

DanGrayson:
Closing temporarily to fix it up...

DanGrayson:
Okay, maybe the notation is good to go.  Minor changes would be easy to make, but it might be best to postpone them.  They won't stop being easy to make, but it might be good to encounter the conflicts as soon as possible.

vladimirias:
Dan, could you summarize for us what are the changes introduced by this PR? I can see the change of ;; into \cdot. What else?

DanGrayson:
See the initial pull request message at https://github.com/UniMath/UniMath/pull/613#issue-210204904 as well as the second commit message at https://github.com/UniMath/UniMath/pull/613/commits/0c888a1fd2b4f7866cc6e8f6f1b03486885ff936

vladimirias:
LGTM

benediktahrens:
Re "allowed both "f · g" and "g ∘ f" for "compose f g"": which one is used for display?

benediktahrens:
Re Re "allowed both "f · g" and "g ∘ f" for "compose f g"": which one is used for display?
I see that the latter is parse-only. That answers my question.

benediktahrens:
In hindsight, it would have been good to swap the notations for functors and natural transformations. After all, those are 1- and 2-cells of a bicategory, and it is customary to denote those with 1- and 2-arrows, respectively. At the moment, we are denoting them with 2- and 1-arrows, respectively.

PR #622, opened 2017-02-26
proscribe "repeat" in our style guide

DanGrayson:
Use `do` with a specific numerical count, rather than `repeat`, to make proofs
easier to repair.


DanGrayson:
I canceled the build because this changes no code.

PR #623, opened 2017-02-26
hush eval tests in CategoryTheory/Inductives/Lists.v

DanGrayson:


benediktahrens:
I would like to keep a note saying that `Eval lazy` works. Your PR eliminates that knowledge.

benediktahrens:
What exactly is the status of the `reflexvity` tactic in UniMath?

DanGrayson:
Why wouldn't Eval lazy work?  When would we ever want to use it for anything?

I like `reflexivity`.  We should convince coqdoc to color it.

DanGrayson:
Rather than using a vernacular Eval command, what about inserting the "lazy" tactic into the proofs of the goals that I've just created?  Would that test the same thing?

mortberg:
Yes, please use the lazy tactic in the Goal's. The reason we want to record this information is that other evaluation strategies failed (cbn, cbv, compute, vm_compute, native_compute) but interestingly lazy worked. 

About `reflexivity` I really think we should redefine it as I suggested in the other branch so that we are sure that it only does UniMath things. It's a useful tactic which is used in many places (and by other tactics), and as Dan keep saying it is displayed in nice red color in PG.

DanGrayson:
But "reflexive" already does just do UniMath things.

mortberg:
At the moment yes, but maybe it will change in the future (like `apply`...).

DanGrayson:
No.

In any case, I think it's not apply that is changing, it's the terms we are passing to it, revealing inadequacies in an old unmaintained algorithm.

mortberg:
Ok ok, let's keep `reflexivity` as it is for now.  

How about writing all of these examples as:
```
Goal length _ (nil natHSET) = 0.
lazy.
Abort.
```

There is no reason to Qed them as we only want to see what lazy does.

While you're at it could you also please update the comment in the header to something like:

"This file contains a formalization of lists as the initial
algebra of the list functor ([List]). We then prove that
this definition is equivalent to the lists defined as
iterated products ([weq_list])."

DanGrayson:
Okay.

But I'll Qed them, as people might search for "Abort" to find projects to work on.

DanGrayson:
Closing temporarily to fix it ...

Issue #616, opened 2017-02-25
initialize package EssentialFoundations


DanGrayson:
See discussion at https://github.com/UniMath/UniMath/issues/365#issuecomment-282501336 and above.

PR #624, opened 2017-02-26
Essential foundations

DanGrayson:
Resolves issue #616.

The interesting stuff is in the files FirstTypes.v and README.md.  There is also a file "All.v" for getting the entire contents of the package, which is called EssentialFoundations.

One feature is that the definition of "isaset" is the more natural expression "isofhlevel 2", which is judgmentally equal to the old one.  A side effect of that change is that a gazillion uses of "apply" broke when trying to show that something is a set or that some equality type is a proposition, so I fixed those, both under Coq 8.5 and 8.6, with many of the fixes using a new tactic I tentatively call "use_exact", which just applies "exact" with more and more underscores.

I give new definitions of "unit" and "coprod".  A side effect is that the alternate names "inl" and "inr" for "ii1" and "ii2" are gone.  I think that's an improvement.

If I were to continue, the next thing I might add would be the statement of the univalence axiom, not the assertion, just the definition of the type of the proposed axiom, since we don't want the axiom asserted until after PartB.

I also took the opportunity to use p and q for variables of type "paths _ _" instead of e1 and e2.  After all, "paths" begins with "p".


DanGrayson:
PS: I'd like to give a new definition of "nat", but I don't know how to get the natural number syntax plug-in to parse numerals and make elements of the new "nat".

DanGrayson:
PS: I also gave a new definition of "empty".

DanGrayson:
PS: Here's the coqdoc output for this : http://www.math.illinois.edu/~dan/essential-foundations/UniMath.EssentialFoundations.FirstSteps.html

DanGrayson:
PS: I also changed the definition of total2 to use a record instead of an inductively defined type.  The rationale is that soon we'll turn on primitive projections, and from then on it will be a record.

DanGrayson:
PS: The changes I made increased the speed by 13%.

benediktahrens:
What about all the places where you changed `apply foo` to `exact (foo _ _ _ )`? Is that because `use_exact foo` does not work there, or because you introduced `use_exact` only afterwards?

DanGrayson:
There actually were a few places where use_exact didn't work, and I have no idea why, but in many cases I just thought it would be more beautiful and more efficient to put the actual term there rather than to appeal to a tactic.

PR #626, opened 2017-02-27
Theory about subobjects

mortberg:
I define the set of subobjects of an element c as equivalence classes of monos into c. I define this by quotienting the subcategory of monos over c by a relation that relates (merely) isomorphic elements.  The main result is that this set is a poset.

### DIFF
--- a/UniMath/CategoryTheory/Subobjects.v
+++ b/UniMath/CategoryTheory/Subobjects.v
@@ -10,24 +10,23 @@ Require Import UniMath.CategoryTheory.functor_categories.
 Require Import UniMath.CategoryTheory.slicecat.
 Require Import UniMath.CategoryTheory.Monics.
 Require Import UniMath.CategoryTheory.sub_precategories.
-
+Require Import UniMath.CategoryTheory.category_hset.
 Require Import UniMath.CategoryTheory.limits.pullbacks.
 
 
 (** * Definition of subobjects *)
 Section def_subobjects.
 
-  Variable C : precategory.
-  Hypothesis hs : has_homsets C.
+  Context {C : precategory} (hsC : has_homsets C).
 
-  Definition SubobjectsPrecategory (c : C) : UU :=
-    slice_precat (subprecategory_of_monics C hs)
-                 (subprecategory_of_monics_ob C hs c)
-                 (has_homsets_subprecategory_of_monics C hs).
+  Definition SubobjectsPrecategory (c : C) : precategory :=
+    slice_precat (subprecategory_of_monics C hsC)
+                 (subprecategory_of_monics_ob C hsC c)
+                 (has_homsets_subprecategory_of_monics C hsC).
 
   (** Construction of a subobject from a monic *)
   Definition SubobjectsPrecategory_ob {c c' : C} (h : C⟦c', c⟧) (isM : isMonic h) :
-    SubobjectsPrecategory c := tpair _ (subprecategory_of_monics_ob C hs c') (tpair _ h isM).
+    SubobjectsPrecategory c := (subprecategory_of_monics_ob C hsC c',,(h,,isM)).
 
   Hypothesis hpb : @Pullbacks C.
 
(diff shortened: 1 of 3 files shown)


UniMath/CategoryTheory/Subobjects.v
```diff
@@ -1,46 +1,205 @@
...
     use SubobjectsPrecategory_ob.
     - exact pb.
-    - exact (limits.pullbacks.PullbackPr1 pb).
+    - exact (PullbackPr1 pb).
     - use MonicPullbackisMonic'.
   Defined.
 
 End def_subobjects.
+
+(** * Equivalence relation identifying isomorphic objects *)
```
mortberg: Maybe it would be useful to upstream this section?

UniMath/CategoryTheory/Subobjects.v
```diff
@@ -1,46 +1,205 @@
...
+Section iso_rel.
+
+Context (C : precategory).
+
+(** a and b are related if there merely exists an iso between them *)
+Definition iso_rel : hrel C := λ a b, ∥iso a b∥.
```
benediktahrens: A good name for this would be `are_isomorphic`.

UniMath/CategoryTheory/Subobjects.v
```diff
@@ -1,46 +1,205 @@
...
+set (Htemp := maponpaths pr1 (pr2 (pr1 h2))).
+apply pathsinv0; simpl in *.
+rewrite <-!assoc, <- Htemp, <- (pr2 f).
+etrans; [ apply maponpaths, (maponpaths pr1 (pr2 (pr1 h1))) |]; simpl.
+rewrite assoc.
+etrans; [ eapply cancel_postcomposition, (maponpaths pr1 (maponpaths pr1 Hh1)) |].
+apply id_left.
+Qed.
+
+(** Lift the relation on monos to subobjects *)
```
DanGrayson: "Lift" is the wrong word for getting a relation on a quotient set.

mortberg: Indeed. Maybe "Restrict"?

UniMath/CategoryTheory/Subobjects.v
DanGrayson: This is the kind of thing that would be much clearer and more beautiful in a univalent category, for then you could just define a subobject of b to be a mono from some other object to b -- isomorphic ones would be equal, and they would form a set.

UniMath/CategoryTheory/Subobjects.v
DanGrayson: VV calls it a quotient relation, i.e., a relation on a quotient, as in this lemma:
```
Definition quotrel {X : UU} {R L : hrel X} (is : iscomprelrel R L) :
  hrel (setquot R) := setquotuniv2 R hPropset L is.
```
, which perhaps you could simply use.

UniMath/CategoryTheory/Subobjects.v
mortberg: I renamed `iso_rel` to `are_isomorphic`.

Dan: Yes, I guess this remark is related to what you wrote in https://github.com/UniMath/UniMath/issues/630#issuecomment-283328612

mortberg:
I've addressed your comments now. Should I upstream the section on are_isomorphic? It's a quite basic thing which could go into precategories somewhere after isomoprhisms are introduced.

benediktahrens:
On 03/01/2017 02:02 PM, Anders Mörtberg wrote:
> I've addressed your comments now. Should I upstream the section on
> are_isomorphic? It's a quite basic thing which could go into
> precategories somewhere after isomoprhisms are introduced.

Sure, upstreaming would be nice. Please add a (sub)section for it, too,
so that it shows up nicely in the HTML doc.

mortberg:
I upstreamed the code now. I noticed the precategories.v don't contain any `Section`'s, one of these days we should maybe take a pass over it and add some together with more consequent use of implicit arguments (using `Context` variables) and use some more unicode notations. Maybe we should have an issue for this so that we don't forget.

mortberg:
I've addressed all of your comments now, so feel free to merge if travis is happy.

DanGrayson:
lgtm

PR #628, opened 2017-02-28
A formalization of vector spaces

AnthonyBordg:
In this file we define the set of endomorphisms of an abelian group and we prove it is endowed with the structure of a ring. 
We define a module over a commutative ring R as an abelian group G together with a ring morphism from R to the ring of endomorphisms of G. 
We conclude with the definition of a vector space over a field. 
Any feedback is welcome.


DanGrayson:
There's no reason to assume K is commutative, is there?  R would be better than K, by the way.

Defining an R-module M directly in terms of a multiplication operation R x M --> M seems more intuitive and might even run faster.

AnthonyBordg:
Hi Dan,

2017-02-28 19:06 GMT+01:00 Daniel R. Grayson <notifications@github.com>:

> There's no reason to assume K is commutative, is there? R would be better
> than K, by the way.
>
yes you are right, but I develop some notions in "call by need", namely as
I need them for my developments, and I'm mainly interested in the case
where R is commutative, even a field. Of course my file should be completed
and updated as I go along.

Yes, R would be better, sorry, it's just a typo, I used other letters in my
file anyways, namely X and Y, it might be better to use R and  G actually ☺
.

> Defining an R-module M directly in terms of a multiplication operation R x
> M --> M seems more intuitive and might even run faster.
>
Again, I agree on this, but here none of these criteria, intuition and fast
code, were my goals. There is nothing on these aspects in the readme
defining the coding style in UniMath. If those criteria are important, then
please let me know.
I take the opportunity to explain my choice. I'm interested in seeing if
UniMath allows "high-level" formalization of mathematics, and I think
UniMath allows it and I consider this as an important improvement or even a
breakthrough in the formalization of mathematics, with this respect I want
to experiment  and write mathematical notions and theorems "from The Book".
Of course I'm open to discussion and I'm eager to hear any advice.





-- 
PhD. Anthony Bordg
postdoctoral fellow
Mathematical Institute Charles University
https://sites.google.com/site/anthonybordg/

DanGrayson:
Dear Anthony,

It seems your goal is to pursue a personal formalization experiment in your own unique style, whereas ours is to build a broad base of tools usable by others, written in clean and efficient form, informed by collaboration with others.  It seems best for you to pursue your experiment as an application of UniMath rather than as a part of UniMath.  You could do that easily in a separate repository under your account, whose first file would be the one you have offered here.  Publicity for your repository can easily be obtained by advertising it on the Univalent Math google group.

Closing...

PR #631, opened 2017-02-28
change notation for functors and natural transformations

### LINKED ISSUE #620: Change notation for functors and natural transformations to make them ressemble 2-categorical notation
as per https://github.com/UniMath/UniMath/pull/613#issuecomment-282545442

One solution would simply be to swap the notations, but we can also think about a unicode variant of `==>`.

DanGrayson:
resolves #620

### DIFF
--- a/UniMath/CategoryTheory/functor_categories.v
+++ b/UniMath/CategoryTheory/functor_categories.v
@@ -143,7 +143,8 @@ Qed.
 Definition functor (C C' : precategory_data) : UU :=
   total2 ( fun F : functor_data C C' => is_functor F ).
 
-Notation "a ==> b" := (functor a b) (at level 50) : cat.
+Notation "a ⟶ b" := (functor a b) (at level 39) : cat.
+(* to input: type "\r-->" with Agda input method *)
 
 (** Note that this makes the second component opaque for efficiency reasons *)
 Definition mk_functor {C C' : precategory_data} (F : functor_data C C') (H : is_functor F) :
@@ -764,8 +765,7 @@ Qed.
 Definition nat_trans {C C' : precategory_data} (F F' : functor_data C C') : UU :=
   total2 (fun t : ∏ x : ob C, F x -->  F' x => is_nat_trans F F' t).
 
-Notation "F ⟶ G" := (nat_trans F G) (at level 39) : cat.
-(* to input: type "\r--" or "\r" or "\-->" with Agda input method *)
+Notation "F ⟹ G" := (nat_trans F G) (at level 39) : cat.
 
 (** Note that this makes the second component opaque for efficiency reasons *)
 Definition mk_nat_trans {C C' : precategory_data} (F F' : functor_data C C')
@@ -1458,8 +1458,7 @@ Notation "[ C , D , hs ]" := (functor_precategory C D hs) : cat.
 
 Notation "[ C , D ]" := (functor_Precategory C D) : cat.
 
-Notation "F ⟶ G" := (nat_trans F G) (at level 39) : cat.
-(* to input: type "\r--" or "\r" or "\-->" with Agda input method *)
+Notation "F ⟹ G" := (nat_trans F G) (at level 39) : cat.
 
 Notation "F ∙ G" := (functor_composite F G) (at level 35) : cat.
 (* to input: type "\." with Agda input method *)
@@ -1471,4 +1470,5 @@ Notation "G □ F" := (functor_composite F G) (at level 35, only parsing) : cat.
 Notation "G □ F" := (functor_composite (F:[_,_]) (G:[_,_]) : [_,_]) (at level 35) : Cat.
 (* to input: type "\Box" or "\square" or "\sqw" or "\sq" with Agda input method *)
 
-Notation "a ==> b" := (functor a b) (at level 50) : cat.
+Notation "a ⟶ b" := (functor a b) (at level 39) : cat.
+(* to input: type "\r-->" with Agda input method *)
--- a/UniMath/CategoryTheory/functor_categories.v
+++ b/UniMath/CategoryTheory/functor_categories.v
@@ -144,7 +144,7 @@ Definition functor (C C' : precategory_data) : UU :=
   total2 ( fun F : functor_data C C' => is_functor F ).
 
 Notation "a ⟶ b" := (functor a b) (at level 39) : cat.
-(* to input: type "\r-->" with Agda input method *)
+(* to input: type "\-->" with Agda input method *)
 
 (** Note that this makes the second component opaque for efficiency reasons *)
 Definition mk_functor {C C' : precategory_data} (F : functor_data C C') (H : is_functor F) :
@@ -766,6 +766,7 @@ Definition nat_trans {C C' : precategory_data} (F F' : functor_data C C') : UU :
   total2 (fun t : ∏ x : ob C, F x -->  F' x => is_nat_trans F F' t).
 
 Notation "F ⟹ G" := (nat_trans F G) (at level 39) : cat.
+(* to input: type "\==>" with Agda input method *)
 
 (** Note that this makes the second component opaque for efficiency reasons *)
 Definition mk_nat_trans {C C' : precategory_data} (F F' : functor_data C C')
@@ -1459,6 +1460,7 @@ Notation "[ C , D , hs ]" := (functor_precategory C D hs) : cat.
 Notation "[ C , D ]" := (functor_Precategory C D) : cat.
 
 Notation "F ⟹ G" := (nat_trans F G) (at level 39) : cat.
+(* to input: type "\==>" with Agda input method *)
 
 Notation "F ∙ G" := (functor_composite F G) (at level 35) : cat.
 (* to input: type "\." with Agda input method *)
@@ -1471,4 +1473,4 @@ Notation "G □ F" := (functor_composite (F:[_,_]) (G:[_,_]) : [_,_]) (at level 
 (* to input: type "\Box" or "\square" or "\sqw" or "\sq" with Agda input method *)
 
 Notation "a ⟶ b" := (functor a b) (at level 39) : cat.
-(* to input: type "\r-->" with Agda input method *)
+(* to input: type "\-->" with Agda input method *)
--- a/UniMath/Foundations/Preamble.v
+++ b/UniMath/Foundations/Preamble.v
@@ -297,8 +297,6 @@ Reserved Notation "g ∘ f"  (at level 50, left associativity).
 
 Reserved Notation "p #' x" (right associativity, at level 65, only parsing).
 
-Reserved Notation "a ==> b"  (at level 50).
-
 Reserved Notation "C '^op'" (at level 3, format "C ^op").
 
 Reserved Notation "a <-- b" (at level 50).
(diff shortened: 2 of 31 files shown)


UniMath/CategoryTheory/functor_categories.v
```diff
@@ -143,7 +143,8 @@ Qed.
 Definition functor (C C' : precategory_data) : UU :=
   total2 ( fun F : functor_data C C' => is_functor F ).
 
-Notation "a ==> b" := (functor a b) (at level 50) : cat.
+Notation "a ⟶ b" := (functor a b) (at level 39) : cat.
+(* to input: type "\r-->" with Agda input method *)
```
mortberg: Why is an r needed to type this?

UniMath/CategoryTheory/functor_categories.v
```diff
@@ -764,8 +765,7 @@ Qed.
 Definition nat_trans {C C' : precategory_data} (F F' : functor_data C C') : UU :=
   total2 (fun t : ∏ x : ob C, F x -->  F' x => is_nat_trans F F' t).
 
-Notation "F ⟶ G" := (nat_trans F G) (at level 39) : cat.
-(* to input: type "\r--" or "\r" or "\-->" with Agda input method *)
+Notation "F ⟹ G" := (nat_trans F G) (at level 39) : cat.
```
mortberg: Please add a comment how to type this

UniMath/CategoryTheory/functor_categories.v
```diff
@@ -1458,8 +1458,7 @@ Notation "[ C , D , hs ]" := (functor_precategory C D hs) : cat.
 
 Notation "[ C , D ]" := (functor_Precategory C D) : cat.
 
-Notation "F ⟶ G" := (nat_trans F G) (at level 39) : cat.
-(* to input: type "\r--" or "\r" or "\-->" with Agda input method *)
+Notation "F ⟹ G" := (nat_trans F G) (at level 39) : cat.
```
mortberg: Weren't this notation already to the cat scope above?

UniMath/CategoryTheory/functor_categories.v
```diff
@@ -1471,4 +1470,5 @@ Notation "G □ F" := (functor_composite F G) (at level 35, only parsing) : cat.
 Notation "G □ F" := (functor_composite (F:[_,_]) (G:[_,_]) : [_,_]) (at level 35) : Cat.
 (* to input: type "\Box" or "\square" or "\sqw" or "\sq" with Agda input method *)
 
-Notation "a ==> b" := (functor a b) (at level 50) : cat.
+Notation "a ⟶ b" := (functor a b) (at level 39) : cat.
```
mortberg: Wasn't this notation already added to the cat scope above?

UniMath/Foundations/Preamble.v
```diff
@@ -297,8 +297,6 @@ Reserved Notation "g ∘ f"  (at level 50, left associativity).
 
 Reserved Notation "p #' x" (right associativity, at level 65, only parsing).
```
mortberg: When you reserve a notation like this I don't think you have to add the associativity and level information later when you introduce the notation?

mortberg:
Can't we please call the `cat` scope `cat_scope` instead? This way it will be more informative. 

We can then add:
```
Delimit Scope cat_scope with cat.
```
which allows us to write `(...)%cat`

DanGrayson:
Re: "This way it will be more informative."

It's not more informative, just confusing.  It's better for our users if we just pretend that the name of the scope and its delimiter are the same thing.

DanGrayson:
... and we can already write `(...)%cat`.

UniMath/CategoryTheory/functor_categories.v
DanGrayson: The strange thing about notation is that it gets erased at the end of a section, so after the end of a section I have to define it again.

UniMath/CategoryTheory/functor_categories.v
DanGrayson: Will do

UniMath/CategoryTheory/functor_categories.v
DanGrayson: Will fix.

UniMath/CategoryTheory/functor_categories.v
mortberg: Ah yes, of course. Maybe you can add Local to the ones above to make this clearer (it will not do anything though)? 

Alternatively we can change the structure a bit in the file so that we first have the definitions of functors and nat_trans in one section, then close the section and add the notations, then the other sections can follow as they are. This way we only add the notations once.

mortberg:
To me it is confusing because I've been using a lot of other Coq libraries before UniMath which had the convention I suggested, but maybe I'm the only one who has this problem. 

Note that we might want to use `cat` or `Cat` for the category of small categories at some point, so there might be a clash of names. So I really think it is better to end all scopes by `_scope` so that they never clash with anything (I don't think any normal mathematical objects end with `scope`...).

mortberg:
Note that the Coq standard library are using the naming convention ending scopes with `_scope`:

https://coq.inria.fr/library/Coq.Init.Notations.html

UniMath/Foundations/Preamble.v
mortberg: It seems like I remembered this correctly:

https://coq.inria.fr/refman/Reference-Manual014.html#hevea_command218

So I suggest you drop all of this information from where the notations are declared in order to not have redundant information which could cause trouble if someone changes it at the wrong place.

DanGrayson:
Re: "So I really think it is better to end all scopes by _scope so that they never clash with anything (I don't think any normal mathematical objects end with scope...)."

I don't understand your point about clashing, which is impossible, since scope names are in their own scope.

And this :
```
Delimit Scope cat_scope with cat.
Open Scope cat. 
Notation "x ** y " := (Nat.mul x y) (at level 50) : cat_scope.
Definition cat := (3 ** 2)%cat.
```
looks more confusing than this:
```
Delimit Scope cat with cat.
Open Scope cat.
Notation "x ** y " := (Nat.mul x y) (at level 50) : cat.
Definition cat := (3 ** 2)%cat.
```

The idea of Coq to have scope names and scope delimiters be different seems useless and confusing.

DanGrayson:
Re: "Note that the Coq standard library are using the naming convention ending scopes with _scope"

Yes, but it's a bad idea, anyway.

mortberg:
Fair enough, maybe my clash argument wasn't too strong. But I think there is some value to the argument that UniMath shouldn't look too weird for normal Coq users. 

I wouldn't write what you are writing, I would write:
```
Delimit Scope cat_scope with cat.
Open Scope cat_scope. 
Notation "x ** y " := (Nat.mul x y) (at level 50) : cat_scope.
Definition cat := (3 ** 2)%cat.
```
I always use the `_scope` name for the scope and the delimited name only after `%`.

Anyway I don't think we're going to get any further in this discussion. If someone else has an opinion on whether we should stick to the Coq standard of ending scopes with `_scope` please weigh in, otherwise I guess we will go with Dan's suggestion (and I'll cringe every time I open a scope in UniMath... :) )

benediktahrens:
On 03/01/2017 03:00 PM, Anders Mörtberg wrote:
> Fair enough, maybe my clash argument wasn't too strong. But I think there is some value to the argument that UniMath shouldn't look too weird for normal Coq users. 
> 
> I wouldn't write what you are writing, I would write:
> ```
> Delimit Scope cat_scope with cat.
> Open Scope cat_scope. 
> Notation "x ** y " := (Nat.mul x y) (at level 50) : cat_scope.
> Definition cat := (3 ** 2)%cat.
> ```
> I always use the `_scope` name for the scope and the delimited name only after `%`.
> 
> Anyway I don't think we're going to get any further in this discussion. If someone else has an opinion on whether we should stick to the Coq standard of ending scopes with `_scope` please weigh in, otherwise I guess we will go with Dan's suggestion (and I'll cringe every time I open a scope in UniMath... :) )
> 

I agree with Dan that it is confusing to distinguish scopes from
delimiters for scopes.

I suppose the technical reason for that is that one can use the same
delimiter for different scopes, hence opening and closing several scopes
at the same time. I am happy to not make use of that feature.

DanGrayson:
Oh, yes, that was a misprint of mine.  Isn't that proof that even a sophisticated user can get confused if those names differ?

I hear you saying that you prefer to have those two names be different, but I don't hear any reason for your preference.  You mentioned clashing as a reason, but I pointed out clashing is impossible and gave an example.  The reason for my preference is that I think the "Delimit" command and the idea the delimiter and scope name are different was a bad design decision by the Coq team, since there is no need for the additional concept, and it adds to the burden of learning for the new user.

mortberg:
Ok, I rest my case. But if we ever run into some unforeseen trouble because we don't distinguish the name of the scopes from the delimiters remember this discussion. :)

DanGrayson:
Well, I think I can depend on you to remind me when the time comes, and my brain is almost 65 years old, so the memory of this discussion will certainly have disappeared within a month.  :-)

DanGrayson:
I think I've taken all comments so far into account in the last pull request, by the way.

mortberg:
How about my comments about `Reserved`? I think it is nice to only have information about levels and associativity when Reserving the notation in order to not duplicate this information (and hence introducing the risk that things break because one of them get changed but not the other).

DanGrayson:
Oh, right, that's a great idea, I'll implement it.

DanGrayson:
Okay, how's that?

mortberg:
That's great! 

While you're at it, can you also add:
```
format "C ⟦ a , b ⟧"
```
to that notation so that it's printed properly (without too many spaces). I had to do something similar for `f ;; g` to make sure that there are spaces around `;;`, I'm not sure if it is needed for some of the other binary operation.

DanGrayson:
It seems to do that already:
```
precategory_id_comp = λ C : precategory_ob_mor, (∏ c : C, C ⟦ c, c ⟧) × (∏ a b c : C, C ⟦ a, b ⟧ → C ⟦ b, c ⟧ → C ⟦ a, c ⟧)
     : precategory_ob_mor → Type
```

mortberg:
That's too much spaces in my opinion, I think it should look like:
```
precategory_id_comp = 
λ C : precategory_ob_mor, (∏ c : C, C⟦c,c⟧) × (∏ a b c : C, C⟦a,b⟧ → C⟦b,c⟧ → C⟦a,c⟧)
     : precategory_ob_mor → Type
```

By adding the format information I wrote above this will be the result. The reason I want this is that this notation is not for an operator so there shouldn't be too many spaces.

mortberg:
I don't understand when Coq decied to put spaces around special characters when printing notations. For example I had to add `format "f  ;;  g"` to get that notation printed correctly (note that I have 2 spaces on both sides of `;;`).

mortberg:
Oops, github erased the 2 spaces in my example. But trust me, there should be 2 of them on each side for it to be printed nicely.

DanGrayson:
I tried it with 0, 1, and 2 spaces, with no luck.  What if I let you futz with that later on?

mortberg:
Ok, I'll look at it later.

mortberg:
LGTM

UniMath/Ktheory/Elements.v
```diff
@@ -24,87 +24,87 @@ Definition cat_data {C} (X:C==>SET) : precategory_data.
...
-Definition cat_on_nat_trans {C:Precategory} {X Y:C==>SET} (p:X ⟶ Y) :
-  cat X ==> cat Y.
+Definition cat_on_nat_trans {C:Precategory} {X Y:C⟶SET} (p:X ⟹ Y) :
+  cat X ⟶ cat Y.
 Proof.
   intros.
   exact (cat_on_nat_trans_data p,, cat_on_nat_trans_is_nat_trans p). Defined.
 
-(* maybe make a functor [C,SET] ==> [category of Precategories] *)
+(* maybe make a functor [C,SET] ⟶ [category of Precategories] *)
```
benediktahrens: Off topic: that comment does not make sense, and should be corrected.

benediktahrens:
lgtm

PR #632, opened 2017-03-01
Proof that Omega in PreShv(C) is a bounded lattice

mortberg:
The main result of this PR is that Omega in PreShv(C) is a bounded lattice object (internally in PreShv(C)). To prove this I had to:

- Define bounded lattices
- Define internal lattice objects and bounded lattice objects in a category
- Prove that hProp is a bounded lattice (these results should be upstreamed, but I don't know where)

The lattice structure on hProp then gives a lattice structure on the set of sieves used in the definition of Omega which in turn gives that Omega is a lattice object in PreShv(C).

I also defined sublattice objects of a lattice object. I'm not completely sure that this definition is the right one so it would be great if someone looks at it. Given elements `M` and `L` with lattice operations and a mono `i : C[M,L]` which is a homomorphism for the lattice operations I prove that `M` is a lattice object if `L` is. So to give a sublattice object it suffices to give a mono `i : C[M,L]`, operations on `M`, a lattice object structure on `L` and a proof that `i` is a homomorphism for the lattice operations. I want to use this to construct the "face lattice" used in the cubical set model (in cSet the object |F is a sublattice of Omega with some extra structure).

Note that this PR relies on https://github.com/UniMath/UniMath/pull/607 which means that that PR should be merged first, this is also the reason why the diff is pretty big.

### DIFF
--- a/UniMath/CategoryTheory/LatticeObject.v
+++ b/UniMath/CategoryTheory/LatticeObject.v
@@ -20,6 +20,7 @@ Require Import UniMath.Algebra.Lattice.
 Require Import UniMath.CategoryTheory.precategories.
 Require Import UniMath.CategoryTheory.Monics.
 Require Import UniMath.CategoryTheory.limits.binproducts.
+Require Import UniMath.CategoryTheory.limits.terminal.
 
 Local Open Scope cat.
 
@@ -114,6 +115,46 @@ Definition mklatticeob {L : C} {meet_mor join_mor : C⟦L ⊗ L,L⟧} :
 
 Definition meet_mor {L : C} (isL : latticeob L) : C⟦L ⊗ L,L⟧ := pr1 isL.
 Definition join_mor {L : C} (isL : latticeob L) : C⟦L ⊗ L,L⟧ := pr1 (pr2 isL).
+
+(** Bounded lattice objects *)
+
+Context {TC : Terminal C}.
+
+Definition ι_1 {x : C} : C⟦x,TC ⊗ x⟧ :=
+  BinProductArrow _ _ (TerminalArrow _) (identity x).
+
+(** Given u : C⟦TC,L⟧ the equation witnessing the left unit law is given by the diagram:
+<<
+         ι_1
+     L ------> 1 ⊗ L
+     |           |
+   1 |           | u×1
+     V           V
+     L <------ L ⊗ L
+          f
+>>
+
+ *)
+
+Definition islunit_cat {L} (f : C⟦L ⊗ L,L⟧) (u : C⟦TC,L⟧) : UU := ι_1 · (u ×× 1) · f = 1.
+
+
+Definition bounded_latticeop {X : hSet} (l : lattice X) (bot top : X) :=
+  (islunit (Lmax l) bot) × (islunit (Lmin l) top).
+
+Definition bounded_lattice (X : hSet) :=
+  ∑ (l : lattice X) (bot top : X), bounded_latticeop l bot top.
+
+Definition mkbounded_lattice {X : hSet} {l : lattice X} {bot top : X} :
+  bounded_latticeop l bot top → bounded_lattice X := λ bl, l,, bot,, top,, bl.
+
+Definition bounded_lattice_to_lattice X : bounded_lattice X → lattice X := pr1.
+Coercion bounded_lattice_to_lattice : bounded_lattice >-> lattice.
+
+Definition Lbot {X : hSet} (is : bounded_lattice X) : X := pr1 (pr2 is).
+Definition Ltop {X : hSet} (is : bounded_lattice X) : X := pr1 (pr2 (pr2 is)).
+
+
 
 End LatticeObject_def.
 
--- a/UniMath/CategoryTheory/LatticeObject.v
+++ b/UniMath/CategoryTheory/LatticeObject.v
@@ -109,7 +109,7 @@ Definition latticeop_cat {L} (meet_mor join_mor : C⟦L ⊗ L,L⟧) :=
 Definition latticeob (L : C) : UU :=
   ∑ (meet_mor join_mor : C⟦L ⊗ L,L⟧), latticeop_cat meet_mor join_mor.
 
-Definition mklatticeob {L : C} {meet_mor join_mor : C⟦L ⊗ L,L⟧} :
+Definition mk_latticeob {L : C} {meet_mor join_mor : C⟦L ⊗ L,L⟧} :
   latticeop_cat meet_mor join_mor → latticeob L :=
     λ (isL : latticeop_cat meet_mor join_mor), meet_mor,, join_mor ,, isL.
 
@@ -145,7 +145,7 @@ Definition bounded_latticeop_cat {L} (l : latticeob L) (bot top : C⟦TC,L⟧) :
 Definition bounded_latticeob (L : C) : UU :=
   ∑ (l : latticeob L) (bot top : C⟦TC,L⟧), bounded_latticeop_cat l bot top.
 
-Definition mkbounded_latticeob {L} {l : latticeob L} {bot top : C⟦TC,L⟧} :
+Definition mk_bounded_latticeob {L} {l : latticeob L} {bot top : C⟦TC,L⟧} :
   bounded_latticeop_cat l bot top → bounded_latticeob L := λ bl, l,, bot,, top,, bl.
 
 Definition bounded_latticeob_to_latticeob X : bounded_latticeob X → latticeob X := pr1.
@@ -281,7 +281,7 @@ Qed.
 
 Definition sub_latticeob : latticeob BPC M.
 Proof.
-use mklatticeob.
+use mk_latticeob.
 - apply meet_mor_M.
 - apply join_mor_M.
 - repeat split.
@@ -327,7 +327,7 @@ Qed.
 
 Definition sub_bounded_latticeob : bounded_latticeob BPC TC M.
 Proof.
-use mkbounded_latticeob.
+use mk_bounded_latticeob.
 - exact (sub_latticeob BPC Hi l Hmeet Hjoin).
 - exact bot_mor_M.
 - exact top_mor_M.
--- a/UniMath/CategoryTheory/LatticeObject.v
+++ b/UniMath/CategoryTheory/LatticeObject.v
@@ -26,6 +26,7 @@ Require Import UniMath.CategoryTheory.limits.terminal.
 
 Local Open Scope cat.
 
+(** * Definition of lattice objects and bounded lattice objects *)
 Section LatticeObject_def.
 
 Context {C : precategory} {BPC : BinProducts C}.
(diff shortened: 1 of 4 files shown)


benediktahrens:
Could you give a reference for that stuff?

UniMath/CategoryTheory/LatticeObject.v
```diff
@@ -0,0 +1,339 @@
...
+Require Import UniMath.CategoryTheory.limits.binproducts.
+Require Import UniMath.CategoryTheory.limits.terminal.
+
+Local Open Scope cat.
+
+Section LatticeObject_def.
+
+Context {C : precategory} {BPC : BinProducts C}.
+
+Local Notation "c '⊗' d" := (BinProductObject C (BPC c d)) (at level 75) : cat.
```
benediktahrens: `C` should be made implicit argument of `BinProductObject`.

UniMath/CategoryTheory/LatticeObject.v
```diff
@@ -0,0 +1,339 @@
...
+- Lattice objects ([latticeob])
+- Bounded lattice objects ([bounded_latticeob])
+- Proof that a subobject of a (bounded) lattice object is a lattice object
+  ([sublatticeob], [sub_bounded_latticeob])
+
+Written by: Anders Mörtberg, 2017
+
+*********************************************************************************)
+
+Require Import UniMath.Foundations.PartD.
```
benediktahrens: The comments in this file are very good. Could you also add HTML sections?

UniMath/CategoryTheory/Presheaf.v
```diff
@@ -0,0 +1,439 @@
...
+Defined.
+
+End hProp_lattice.
+
+(** Various limits and colimits in PreShv C *)
+Section limits.
+
+Context {C : precategory}.
+
+Lemma Lims_PreShv : Lims (PreShv C).
```
benediktahrens: This lemma is dangerous. I'd feel more comfortable if it weren't there. What do you think?

UniMath/CategoryTheory/Presheaf.v
```diff
@@ -0,0 +1,439 @@
...
+  has_exponentials BinProducts_PreShv.
+Proof.
+now apply has_exponentials_functor_HSET, has_homsets_opp, hsC.
+Defined.
+
+End limits.
+
+(** Definition of the subobject classifier in a presheaf
+    TODO: Prove that Ω actually is the subobject classifier  *)
+Section Ω_PreShv.
```
benediktahrens: A reference would be good (Sheaves in geometry and logic?).

UniMath/CategoryTheory/Presheaf.v
```diff
@@ -0,0 +1,439 @@
...
+  + apply (pr2 S1 _ _ (pr1 S)).
+  + apply (pr2 S2 _ _ (pr2 S)).
+Defined.
+
+Definition union_sieve (c : C) : binop (sieve c).
+Proof.
+intros S1 S2.
+mkpair.
+- intros x f.
+  (* How can I get rid of the pr1's? *)
```
benediktahrens: Have you tried defining sieves differently, namely as a family of maps into `UU` together with a proof that any these things are propositions?
Obviously you would add a coercion from the whole to the family of maps into UU.

UniMath/CategoryTheory/Presheaf.v
DanGrayson: Do you think it's too close to taking advantage of the inconsistency of type-in-type?

benediktahrens:
Am 3. März 2017 01:09:06 MEZ schrieb "Daniel R. Grayson" <notifications@github.com>:
>DanGrayson commented on this pull request.
>
>
>
>> +- exact hProp_lattice.
>+- exact hfalse.
>+- exact htrue.
>+- split.
>+  + intros P; apply hfalse_hdisj.
>+  + intros P; apply htrue_hconj.
>+Defined.
>+
>+End hProp_lattice.
>+
>+(** Various limits and colimits in PreShv C *)
>+Section limits.
>+
>+Context {C : precategory}.
>+
>+Lemma Lims_PreShv : Lims (PreShv C).
>
>Do you think it's too close to taking advantage of the inconsistency of
>type-in-type?


Yes. A category that has ALL limits (not just all small limits) is in some sense degenerate - I don't know the details. But size matters here.

DanGrayson:
re: "Yes. A category that has ALL limits (not just all small limits) is in some sense degenerate - I don't know the details. But size matters here."

Maybe it's time to introduce a second universe and start at least stating theorems correctly, even if they won't be checked for universe consistency yet.

UniMath/CategoryTheory/LatticeObject.v
mortberg: Yes, another things I was thinking is whether we could introduce this kind of notation in binproducts.v? The problem is that for this to work Coq needs to find the BinProducts structure on C and I don't think it will without help?

UniMath/CategoryTheory/Presheaf.v
mortberg: Sure, I think they define Omega in PreShv(C) in the beginning. I'll double check and add a reference. I'll also add a reference to that book for internal lattice objects.

They also prove that Omega is a lattice, see Thm. 1 (Internal), page 201. But they prove it much more generally, a special case of their theorem is that the subobject classifier in any topos is an internal lattice. I first began proving this (which is why I did the stuff on subobjects as quotients), but then I realized that it would be far too much work and that I should just come up with the direct proof instead.

UniMath/CategoryTheory/Presheaf.v
mortberg: Good idea, I just copied these instances from category_hset_structures. Maybe I should comment it out there as well?

mortberg:
I think Benedikt suggested at some point that we could add a predicate:
```
Definition small (C : precategory) : UU := unit.
```
and add this as a parameter to theorems that requires that the input category is small (e.g. construction of Kan extensions). This way the user would be forced to think if the category is small before using the theorem. This seems like a good idea until we start having universe management (using type-in-type only on the files that use resizing could be a first step).

UniMath/CategoryTheory/LatticeObject.v
mortberg: Done

UniMath/CategoryTheory/Presheaf.v
mortberg: I commented this lemma and the one for Colims in this file now.

benediktahrens:
On 03/03/2017 09:39 AM, Anders Mörtberg wrote:
> 
> 
> 
>> +Require Import UniMath.Foundations.Sets.
> +
> +Require Import UniMath.Algebra.Lattice.
> +
> +Require Import UniMath.CategoryTheory.precategories.
> +Require Import UniMath.CategoryTheory.Monics.
> +Require Import UniMath.CategoryTheory.limits.binproducts.
> +Require Import UniMath.CategoryTheory.limits.terminal.
> +
> +Local Open Scope cat.
> +
> +Section LatticeObject_def.
> +
> +Context {C : precategory} {BPC : BinProducts C}.
> +
> +Local Notation "c '⊗' d" := (BinProductObject C (BPC c d)) (at level 75) : cat.
> 
> Yes, another things I was thinking is whether we could introduce this kind of notation in binproducts.v? The problem is that for this to work Coq needs to find the BinProducts structure on C and I don't think it will without help? 

Yes, this won't work. The only way I have found is the one you are
taking, where you make a notation relative to a section variable.
That is not too bad, in my opinion.

benediktahrens:
On 03/03/2017 09:48 AM, Anders Mörtberg wrote:
> 
> 
> 
>> +- exact hProp_lattice.
> +- exact hfalse. +- exact htrue. +- split. +  + intros P; apply
> hfalse_hdisj. +  + intros P; apply htrue_hconj. +Defined. + +End
> hProp_lattice. + +(** Various limits and colimits in PreShv C *) 
> +Section limits. + +Context {C : precategory}. + +Lemma Lims_PreShv :
> Lims (PreShv C).
> 
> Good idea, I just copied these instances from
> category_hset_structures. Maybe I should comment it out there as
> well?

Yes, I think that would be good.

DanGrayson:
Wait, why comment out the theorem?  Don't you want to use it to construct (small) limits?

mortberg:
We introduced the notion of having all limits of shape g:

https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/limits/graphs/limits.v#L331

The idea is that the user can then be careful to only instantiate this for small g. So for HSET the following result is the one that should be used until we have proper universe management:

https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/category_hset_structures.v#L361

UniMath/CategoryTheory/Presheaf.v
mortberg: I tried this, but it made some things more complicated. For instance to prove that the sieves form a set (at least I couldn't see how to do it as easily as before). So I don't think it's worth redefining it just to get rid of some projections.

UniMath/CategoryTheory/LatticeObject.v
mortberg: Let's make `C` implicit in another PR as it will affect so much code. I think there are implicit arguments missing in many of the files on limits actually, so maybe the best would be to make an issue and clean them all in one big PR.

benediktahrens:
On 03/07/2017 01:12 PM, Anders Mörtberg wrote:
> 
> 
> 
>> +mkpair.
> +- intros x f.
> +  (* How can I get rid of the pr1's? *)
> +  apply (pr1 S1 x f ∧ pr1 S2 x f).
> +- simpl; intros x f S y f'.
> +  split.
> +  + apply (pr2 S1 _ _ (pr1 S)).
> +  + apply (pr2 S2 _ _ (pr2 S)).
> +Defined.
> +
> +Definition union_sieve (c : C) : binop (sieve c).
> +Proof.
> +intros S1 S2.
> +mkpair.
> +- intros x f.
> +  (* How can I get rid of the pr1's? *)
> 
> I tried this, but it made some things more complicated. For instance to prove that the sieves form a set (at least I couldn't see how to do it as easily as before). So I don't think it's worth redefining it just to get rid of some projections. 
> 

I think it is worth separating data and properties in general.

mortberg:
About commention of `LimsHSET`: maybe we should do that in a separate PR? We could then comment out `Lims` from the file where we define limits so that we know that it is not used anywhere.

DanGrayson:
Only the projections of s2 are used, so you might as well change 
```
∏ (s2 : ∑ (y : C),C⟦y,pr1 s1⟧), S (pr1 s2,,pr2 s2 · pr2 s1)
```
to
```
 ∏ y (f:C⟦y,pr1 s1⟧), S (y,, f · pr2 s1)
```

mortberg:
Thanks Dan, I implemented your suggested simplification.

DanGrayson:
You have a conflict now in the list of files with the master branch.

DanGrayson:
lgtm

mortberg:
I know, but I don't want to fix it before https://github.com/UniMath/UniMath/pull/607 has been merged as this PR depends on that one.

PR #633, opened 2017-03-02
improve Makefile message about file ordering

DanGrayson:
Make it clear that we are enforcing the ordering prescribed by
the packages' lists of files, not checking for circularity.


mortberg:
Thanks, LGTM

PR #635, opened 2017-03-02
update my acknowledgements

DanGrayson:


DanGrayson:
I cancelled the build because this changes only my README file

PR #637, opened 2017-03-03
AxiomOfChoice -> LEM

DanGrayson:
Theorem AC_to_LEM : AxiomOfChoice -> LEM.


vladimirias:
In the file hSet there is, I think, a definition of being a base for the axiom of choice that can be rewritten using your definition of axiom of choice. There is also a proof in finitesets that any finite set is a base for the axiom of choice, i.e., that for any surjection onto a finite set there exists a section.

DanGrayson:
Okay, thanks, I reorganized it using `choicebase`.

DanGrayson:
I reorganized this pull request so it doesn't modify any files except those in Ktheory, and so the new material, which I had intended to put in the package "Foundations", goes now into a package called "MoreFoundations".

UniMath/MoreFoundations/AxiomOfChoice.v
```diff
@@ -0,0 +1,104 @@
...
+Local Open Scope logic.
+
+Local Open Scope set.
+
+(** We write these axioms as types rather than as axioms, which would assert them to be true, to
+    force them to be mentioned as explicit hypotheses whenever they are used. *)
+
+Definition AxiomOfChoice : hProp := ∀ (X:hSet), ischoicebase X.
+
+Definition AxiomOfChoice_surj : hProp := ∀ (X:hSet) (Y:UU) (f:Y→X), issurjective f ⇒ ∃ g, ∀ x, f (g x) = x.
```
tpannila: This line is 107 characters long. Break it into 2 lines.

UniMath/MoreFoundations/AxiomOfChoice.v
```diff
@@ -0,0 +1,104 @@
...
+    set (T := (∑ x, P x)%type).
+    set (f := pr1 : T -> X).
+    assert (k : issurjective f).
+    { intros x. simple refine (hinhuniv _ (ne x)); intro p. apply hinhpr. exists (x,,p). reflexivity. }
+    simple refine (hinhuniv _ (AC X T f k)).
+    intro sec. induction sec as [g e]. apply hinhpr. intro x.
+    assert (e' := e x); simpl in e'; clear e.
+    induction (g x) as [x' p]; simpl in e'.
+    induction e'. exact p.
+Defined.
```
tpannila: Why `Defined.`. Why not `Qed.` or `Opaque AC_impl2.`?

UniMath/MoreFoundations/AxiomOfChoice.v
```diff
@@ -0,0 +1,104 @@
...
+        { apply maponpaths, q. }
+        { apply h. } }
+      change (E true false).
+      apply (invmap (weqpathsinsetquot _ _ _)).
+      change (f true = f false).
+      exact r. } }
+  (* the equation is decidable, and thus P is decidable, by the lemma above *)
+  apply (logeq_dec (issymm_logeq _ _ coll)).
+  apply isdeceqbool.
+Defined.
```
tpannila: Why `Defined`? Why not `Qed.` or `Opaque AC_to_LEM.`?

tpannila:
lgtm

vladimirias:
Could you please separate the part of the PR dealing with the AC and LEM from the part dealing with the structural changes to the library? They should not be together.

UniMath/MoreFoundations/AxiomOfChoice.v
DanGrayson: "Qed" prevents unfolding, and thus prevents computation.

UniMath/MoreFoundations/AxiomOfChoice.v
DanGrayson: "Qed" prevents unfolding, and thus prevents computation.

DanGrayson:
What do you mean by the "structural changes to the library?"

tpannila:
Dan, I tried to unfold `AC_impl2` and `AC_to_LEM` and the output was either long 
or unfold was slow. I optimized the proofs so that the output of unfold is shorter 
and unfold is faster. See tpannila@5eeb14e . Is faster unfold  useful for you?

UniMath/MoreFoundations/AxiomOfChoice.v
mortberg: But this is about implications of classical axioms? So what do you expect to compute?

Maybe we can have a constructive proof of `AxiomOfChoice` for some type and then get a constructive proof of `LEM` using this result? I would assume this to be possible if the type is finite for example.

DanGrayson:
I guess that might be useful -- what experiment did you do, exactly?

Meanwhile, I thought of a mathematically sensible way to shorten the proof of the main implication, just pushed, and I have taken your code and abstract the lemma about surjectivity of pr1, whic
could be moved upstream later, thanks.  But I'm not a fan of Qed, so I changed that.

DanGrayson:
Re: "Could you please separate the part of the PR dealing with the AC and LEM from the part dealing with the structural changes to the library? They should not be together."

Maybe you mean I should introduce the new package first?  I don't see how a new package can be judged if it's empty of content.  I'm happy to deal with feedback about the new package in this pull request, which should be unproblematic.  I'm not committed to the name, by the way.

tpannila:
> what experiment did you do, exactly?

```
Lemma AC_impl2_test (a : AxiomOfChoice) (s : AxiomOfChoice_surj) : empty.
Proof.
  set (tmp1 := dirprod_pr1 AC_impl2 a). cbn in tmp1.
  set (tmp2 := dirprod_pr2 AC_impl2 s). cbn in tmp2.
Abort.
```

```
Lemma AC_to_LEM_test (a : AxiomOfChoice) : empty.
Proof.
  set (tmp1 := AC_to_LEM a). unfold AC_to_LEM in tmp1. cbn in tmp1.
Abort.
```

DanGrayson:
Oh, well, I suppose it might be useful in certain circumstances to have proof terms that cbn won't attack seriously, but I generally avoid cbn and simpl, anyway, except on small terms.

tpannila:
Dan, I don't understand. I thought you wanted to compute the map g in `AC_impl2`. 
You used `assert` in the first version to show `issurjective`, which is Opaque (?), 
and now you replaced that with a Lemma closed by `Defined.`, which is Transparent, 
which is worse(?) from performance point of view than the first version. What do you want
to compute using `AC_impl2` (or `AC_to_LEM`)? I'm interested in  why you want/need to 
use `Defined.` like this. Coq is slow if you  don't use `Qed.` or `Opaque`.

Why don't you use cbn or simpl? Others use them, so closing the proofs by `Defined.` 
might generate performance issues for others in future.

DanGrayson:
No, `assert` is not opaque after the proof is over, it just prevents you from seeing the term during the proof.  Perhaps you're thinking of `abstract`, which does make opaque lemmas.

I don't like Qed because it prevents terms from computing all the way.  I don't mind if other people use it, until the day comes when I want to compute a term containing one of their proofs.

DanGrayson:
Re: "Why don't you use cbn or simpl? "

I think it just slows things down, due to unfolding.

benediktahrens:
lgtm

DanGrayson:
Re: "Could you please separate the part of the PR dealing with the AC and LEM from the part dealing with the structural changes to the library? They should not be together."

This has been done.

Issue #544, opened 2017-01-15
definition of abelian category


DanGrayson:
There is a problem with the definition of abelian category, discussed in the comment https://github.com/UniMath/UniMath/pull/534#issuecomment-272638723 and the comments following it: it contains elements of structure that are not uniquely determined up to isomorphism, such as the data proving that every monic arrow is a kernel.

The problem is that we will eventually make the 2-category of abelian categories and wish to prove the 2-category of univalent abelian categories is univalent in an appropriate sense.  For that, we will need isomorphisms to preserve all the structure, since identities preserve all the structure.

DanGrayson:
In general, to replace a type by a proposition, one may seek either a quotient type or a subtype.  Propositional truncation yields a quotient type, and, in the example above, changing " every monic arrow is a kernel " to "  every monic arrow is a kernel of an epi " yields a subtype, which, if the category were univalent, would be a proposition, hence satisfactory for our purposes.

DanGrayson:
Resolved in https://github.com/UniMath/UniMath/pull/534/commits/d24f46d99396d2235037514c72c2e8fb8c666e2c

PR #638, opened 2017-03-05
Fixed abelian categories

tpannila:
This pull request simplifies definition of abelian categories by replacing

- every monic is a kernel of some epi, `AbelianMonicKernelsData`
- every epi is a cokernel of some monic, `AbelianEpiCokernelsData`

by

- every monic is a kernel of its cokernel, `MonicsAreKernels`
- every epi is a cokernel of its kernel, `EpisAreCokernels`

in `Abelian.v`. This change was suggested by Thierry. I guess this PR solves issue #544.


DanGrayson:
You already resolved my issue #544  in https://github.com/UniMath/UniMath/pull/534/commits/d24f46d99396d2235037514c72c2e8fb8c666e2c , which has been merged, by changing the condition from being the kernel of some map to being the kernel of some epi.  (Observe that if f is the kernel of an epi g, then g is the cokernel of f.)

So the current pull request seems to have no purpose to it.  Or does it?

DanGrayson:
@tpannila , what about my question above?

tpannila:
Dan,  I tried to show that if C is a category, then the axioms of an abelian category isaprop. 
I wasn't able to show the monic and epi part of the axioms. More precisely, I wasn't able 
to show 

- is_category C -> isaprop (every monic is a kernel of some epi)

What one should be able to show is 

- is_category C and C is an abelian precategory -> isaprop (every monic is a kernel of some epi)

This PR changes the definition of abelian category so that one can show 
- is_category C -> isaprop (axioms of an abelian category on C).

DanGrayson:
Hmm, did you try to show that if a monic f is the kernel of two epis g and g', then g and g' are isomorphic to each other, and to the cokernel of f?

Is the proof of "is_category C -> isaprop (axioms of an abelian category on C)" in this pull request?

Assuming we could help you prove that, which definition of abelian category would be preferable?  Does it matter to anyone else?

tpannila:
> Hmm, did you try to show that if a monic f is the kernel of two epis g and g', then g and g' are isomorphic to each other, and to the cokernel of f?

Yes. If one only assumes that is_category C , then I don't see why the epis should be cokernels
for the monic.

> Is the proof of "is_category C -> isaprop (axioms of an abelian category on C)" in this pull request?

No

> Assuming we could help you prove that, which definition of abelian category would be preferable? Does it matter to anyone else?

The definition in this PR should be more useful because it allows us to show that

              is_category C -> isaprop (axioms of abelian category on C).

This is more useful because then one can use `squash_to_prop` to prove that C is an abelian 
category. If one only has

     is_category C and C is an abelian category -> isaprop (axioms of abelian category on C),

then one cannot use `squash_to_prop` to prove that C is abelian. 

The first implication is needed to show that heart/core of a t-structure is abelian, because the 
axioms of triangulated category use propositional truncation.

DanGrayson:
If one assumes that is_category C , then it is certainly true that the epis are cokernels for the monic.  Maybe you need that anyway, since it's a common statement.

When answering my question ( "Assuming we could help you prove that, which definition of abelian category would be preferable?" ) you seem to have ignored the assumption, which would also allow to prove that being an abelian category is a proposition.

tpannila:
> If one assumes that is_category C , then it is certainly true that the epis are cokernels for 
> the monic. Maybe you need that anyway, since it's a common statement.

Can you give an informal proof of this or a reference to a proof?

> When answering my question ( "Assuming we could help you prove that, which definition of 
> abelian category would be preferable?" ) you seem to have ignored the assumption, which 
> would also allow one to prove that being an abelian category is a proposition.

Sorry, I don't understand. What is the assumption I have ignored?

vladimirias:
We should have both abelian precategories and abelian categories.

DanGrayson:
Dear Tomi,

Suppose we assume that a monic f : A --> B is a kernel of an epi in two ways, i.e., g : B --> C and g' : B --> C' are epis so that f = ker g and f = ker g'.  We will show that g and g' give isomorphic quotient objects of B.  For that it will suffice to find maps q : C --> C' and r : C' --> C with q g=g' and r g' = g, for then q r g' = q g = g' = 1 g', thus q r = 1 because g' is epi, and r q g = r g' = g = 1  g, thus r q = 1 because g is epi.  Let us show how just to find q; the proof for finding r will be the same.  Since g is an epi, it is the cokernel of some monic h : A' --> B.  Since g = coker h, we know g h = 0.  We also know g f = 0 and f is the kernel of g, so by the definition of kernel, there is a map p : A' --> A with f p = h.  Then g' h = g' f p = 0 p = 0; here we used g' f = 0, which we know because f is the kernel of g'.  Now since g' h = 0 and since g is the cokernel of h, the map q as above exists.  That should do it.

When C is a category, the isomorphism q can be promoted to an equality (C,g) = (C',g') of quotient objects of B.  That shows the existence of an epi that f is the kernel of is a proposition.

Okay, now let's assume that proof works out.  Then that means your original definition of being an abelian category is a proposition.  Being a proposition was the main advantage of the new definition, so now maybe the two definitions are both okay from that point of view.  The question then is whether there is some other reason to prefer one over the other.  Perhaps you're trying to stick close to a textbook and its definitions, and that would give a reason to prefer one to the other.

tpannila:
Line 5 of your proof

> Since g is an epi, it is the cokernel of some monic h : A' --> B.

How do you get this? Remember that I can only assume that the underlying precategory is
a category, not that it is an abelian category.

DanGrayson:
I'm allowed to use that because that's part of the definition of being an abelian precategory.
I am not assuming the precategory is a category.  We are showing that
two proofs that an additive category is abelian are equal.  Being abelian is a conjunction of two conditions: every monic being the kernel of an epi and every epi being the cokernel of a monic.  We have two proofs of the conjunction, i.e., two proofs of the first part and two proofs of the second part.  While showing the two proofs of the first part are equal we can use one of the proofs of the second part, and conversely.

I now see that I should have noticed you were trying to do too much, from one of your previous comments.  It is not true that every monic being the kernel of an epi is a proposition.  We need both parts of the conjunction to get it.

DanGrayson:
I said that wrong and have edited it above: https://github.com/UniMath/UniMath/pull/638#issuecomment-287318000

DanGrayson:
These two lemmas touch upon the issues above:
```

Lemma A P : (P -> isaprop P) -> isaprop P.
Proof.
  intros f.
  apply invproofirrelevance.
  intros p q.
  apply proofirrelevance.
  exact (f p).
Defined.

Lemma B P Q : (P -> isaprop Q) -> (Q -> isaprop P) -> isaprop (P × Q).
Proof.
  intros f g.
  apply invproofirrelevance.
  intros [p q] [p' q'].
  apply total2_paths2.
  - now apply g.
  - now apply f.
Defined.
```

tpannila:
Dan, thank you, now I understand. You are correct. In my reasoning I missed
Lemma A.

The only remaining reason to merge this PR is that it simplifies definition of 
abelian category, as described in my first comment, and simplifies the proof that

- is_category C -> isaprop (axioms of abelian category).

If these simplifications are not what others want, then please say it, and I close
this PR.

DanGrayson:
Okay, I'm glad that's straightened out.

Simplifications are always welcome, so I'm taking a look.

By the way, the proof of CokernelOfDiagonal_is_iso seems too long for something so special.  What about proving something more generally useful, namely: given A, B, C, i : A -> B, p : B -> C, j : C -> B, q : B -> A, such that pj=1, qi=1, iq+jp=1, then p = coker i, i = ker p, q = coker j, and j = ker q?  I.e. each direct sum diagram give two short exact sequences.

DanGrayson:
... oh, never mind, I see that lemma is being proved before you know that an abelian category is an additive category.

It seems that making additivity part of the definition of abelian category would be a simplification, allowing one to eliminate all 800 lines of the file AbelianToAdditive.v.  What are your thoughts on that?  (There don't seem to be any practical examples of abelian categories where "minus" on maps is hard to define.)

DanGrayson:
Another simplification would be to include having homsets in the definition of AbelianPreCat.  Then gazillions of lemmas wouldn't need having homsets as an extra hypothesis.

DanGrayson:
lgtm

DanGrayson:
@benediktahrens -- any comments?

benediktahrens:
lgtm

I have one question: I have noticed that an abelian (pre)category is not literally an additive (pre)category with extra structure. I find that a weird choice. What is the reason for that choice?

Another question that Dan has already commented on: is there anything interesting that can be done without the homsets assumption? If yes, then it is worth keeping this assumption separate. If not, it would be better to assume a `Precategory` instead of a `precategory` everywhere.

tpannila:
Dan and Benedikt, I have followed Borceux Handbook of Categorical algebra 2 
chapter 1 on abelian categories. I don't know what is the reason behind defining 
abelian categories in this way, but I guess it might be the following.

If one looks at the definition of abelian categories (Abelian.v), one can see that
the definition only assumes that abelian category must have some limits and colimits.
There are no conditions imposed on the homsets. Then, from these limits and colimits 
it follows that one has a natural abelian group structure on the homsets, and these 
abelian groups behave nicely with respect to composition of morphisms. I think 
some mathematicians find it beatiful that algebraic structure on homsets naturally 
arises from certain limits and colimits.

To be honest, I don't know any practical benefit of this definition. Hence, it
might be a good idea to include additivity to abelian category for practical reasons.
I don't know any interesting result which would not require homset assumption.

@vladimirias, do you want/need abelian precategories (without homsets)?

mortberg:
Just a thought about the definition of abelian categories: maybe Coq would be faster if they were defined as additive categories with extra structure? If I understand Tomi's comment the abelian group structure on the homsets is now constructed using limits/colimits, this means that they will be represented using concrete (and maybe very big?) terms in Coq. A consequence is that Coq might end up unfolding these when comparing things which leads to the very slow proofs we are seeeing in HomologicalAlgebra. If the additive structure would instead be abstract (i.e. if abelian would just extend the additive category structure) comparing terms could be a lot faster. You could then have a lemma saying that the additive structure can be constructed from certain limits/colimits.

DanGrayson:
Anders' remark about speed is sensible.  Even if abelian categories were defined as additive categories with a certain property, one could still formulate a proof of the beautiful fact, for its own sake.  It just wouldn't be much used by code later on.

DanGrayson:
Vladimir is probably out of the country by now, back after March 25.

A good reason to bury deeply the hypothesis that the categories have homsets is that the same is done in the textbooks you are following.  It would be good to make the presentation of abelian categories to look standard.

tpannila:
@mortberg  and @DanGrayson , I suspect that this is not the underlying 
reason behind slowness in HomologicalAlgebra, but let's hope that the 
change will make HomologicalAlgebra a little faster. My guess is that 
the slowness is due to lack of/"broken" Opaque in UniMath/coq and 
lack of data sharing in  coq when type checking / unifying / ... (what 
ever is the correct term) .  The latter probably leads into copying of data 
in the kernel, which is  slow. (Remember that HomologicalAlgebra uses
Unset Kernel Term Sharing , because the term sharing is broken for 
UniMath). See, for example, @ppedrot 's comment 
https://github.com/coq/coq/pull/448#issuecomment-282662328

PR #639, opened 2017-03-05
Closure of hlevels under products of families of types

AnthonyBordg:
State and prove in PartB.v the theorem 7.1.9 of HoTT book about the closure of hlevels under products of families of types.


DanGrayson:
This is already in PartD:
```
Theorem impred (n : nat) {T : UU} (P : T -> UU) :
  (∏ t : T, isofhlevel n (P t)) -> (isofhlevel n (∏ t : T, P t)).
```

AnthonyBordg:
Ah thanks, I was looking for this theorem, I did not find it, I should have asked ... Why "impred" for the name ? "impred" does not seem very informative.

DanGrayson:
Better to use "Search About" to find theorems, because you'll never guess the name.

AnthonyBordg:
What is this "Search About" ? Is it an emacs/ProofGeneral command ?

vladimirias:
“Impred” because it is a form of impredicativity especially when applied to the hlevel 1. 

It is in PartD because it requires function extensionality. The sequence of parts in Basics is according to the use of axioms.

We have decided not to suggest any changes in Foundations until at least early April, because they should be approved by me and I am very busy until then.

DanGrayson:
Search About is a Coq command.  It has a PG keystroke sequence, too.

mortberg:
To find this I would do:
```
Search (isofhlevel _ (∏ _, _)).
```

AnthonyBordg:
Thank you all.

mortberg:
I would suggest you take a look at the files in the Foundations folder. Especially PartA-PartD, this way you will have some better understanding for where basic things are located and why they are named the way they are. Depending on your background the naming convention can be quite confusing at first, for example `total2` means Sigma type... If you run `make html` some coqdoc html files will be generated in `html/` which might be easier to read than the source files. 

About `impred`: a name that would be easier to guess would be `isofhlevelforall` (this way it would follow the same convention as `isofhleveltotal2`, `isofhleveldirprod`, etc.). This renaming might be something to consider once the lock on Foundations has been lifted.

PR #640, opened 2017-03-06
The syntax of the simply typed lambda calculus as a multisorted binding signature

mortberg:
In this PR I implement the syntax of the STLC as an instance of the framework for constructing a monad on Set/sort from a multisorted signature. I also extract the constructors of the STLC and define a recursor. 

When I did this I noticed that we were composing with both the identity functor and the constant functor in various place in the construction of a functor from a multisorted signature. In order to fix this I changed the definition slightly (essentially by being more careful about what to do for the empty list). Because of this the construction of the strength laws and the proof of omega-cocontinuity had to be changed slightly, but in the end I think we are now getting the right thing for the STLC.


DanGrayson:
lgtm

UniMath/SubstitutionSystems/MultiSorted.v
```diff
@@ -126,22 +145,35 @@ Defined.
...
-eapply functor_composite.
-- apply (pre_composition_functor _ _ _ hs _ (option_list (pr1 lt))).
-- apply post_comp, (proj_functor (pr2 lt)).
+induction lt as [l t].
+(* use list_ind to do a case on whether l is empty or not *)
+use (list_ind _ _ _ l); clear l.
+- exact (post_composition_functor _ _ (proj_functor t)).
+- intros s l _; simpl.
+  eapply functor_composite.
+  + exact (pre_composition_functor hs hs (option_list (cons s l))).
```
rmatthes: A bit unfortunately, the list cons s l has to be decomposed into s and l and then recomposed into cons s l. I do not know better.

rmatthes:
the optimizations look nice and clean; did not yet make up my mind on the new file STLC.v

rmatthes:
lgtm (this is wip but the part that has been formalized looks fine - better have it in UniMath as is, to foster further development)

benediktahrens:
lgtm

UniMath/SubstitutionSystems/MultiSorted.v
mortberg: Yeah, it's not so nice. I just want to do a case-split on whether the list is empty or not and this was the first thing that came to my mind. I guess I could add a isNil function returning a bool and then use an if-then-else if you think that would be easier to read?

benediktahrens:
@mortberg : build failure:
```
COQC UniMath/SubstitutionSystems/STLC.v

File "./UniMath/SubstitutionSystems/STLC.v", line 56, characters 4-7:

Syntax error: 'Fixpoint' or 'CoFixpoint' or [Prim.identref] expected after 'Let' (in [vernac:gallina]).
```

mortberg:
The build failure has now been fixed. Feel free to merge.

PR #642, opened 2017-03-06
Move Elements and ElementsOp to CategoryTheory

mortberg:
This resolves https://github.com/UniMath/UniMath/issues/525. 

I did some minor changes to make the code look more like the rest of CT. I also moved `reflects_isos` to functor_categories.v and put HSET and SET into the cat scope. 

I noticed that there is a file on the category of elements in TypeTheory as well:

https://github.com/UniMath/TypeTheory/blob/master/TypeTheory/Categories/category_of_elements.v

The only mathematical difference, as far as I can tell, is that that file also contains a proof that ∫ F for F : C -> Set is univalent if C is. Maybe the author of that proof could upstream it to UniMath (and port it to ElementsOp.v)?


benediktahrens:
lgtm

DanGrayson:
lgtm

PR #644, opened 2017-03-09
Remove is_isomorphism in favor of is_iso

mortberg:
This resolves https://github.com/UniMath/issues/641.


DanGrayson:
lgtm

mortberg:
I should say that doing this change took me less than 2 minutes to do. Thank you very much Dan for teaching us about TAGS replace!

benediktahrens:
lgtm

Issue #621, opened 2017-02-26
Redefine now


mortberg:
As pointed out in https://github.com/UniMath/UniMath/issues/609 the Coq version of the `now` tactic has various issues. It's doing a bit too much and it's using `case` which is not allowed in UniMath. I think it is nice to have a tactical like now for solving trivial goals generated by a tactic so I think we should have our own version of `now`.

DanGrayson:
Here's the 8.6 source code for it, from Coq's file Tactics.v : 

```
Ltac easy :=
  let rec use_hyp H :=
    match type of H with
    | _ /\ _ => exact H || destruct_hyp H
    | _ => try solve [inversion H]
    end
  with do_intro := let H := fresh in intro H; use_hyp H
  with destruct_hyp H := case H; clear H; do_intro; do_intro in
  let rec use_hyps :=
    match goal with
    | H : _ /\ _ |- _  => exact H || (destruct_hyp H; use_hyps)
    | H : _ |- _ => solve [inversion H]
    | _ => idtac
    end in
  let do_atom :=
    solve [ trivial with eq_true | reflexivity | symmetry; trivial | contradiction ] in
  let rec do_ccl :=
    try do_atom;
    repeat (do_intro; try do_atom);
    solve [ split; do_ccl ] in
  solve [ do_atom | use_hyps; do_ccl ] ||
  fail "Cannot solve this goal".

Tactic Notation "now" tactic(t) := t; easy.
```

DanGrayson:
I'll do this one.

mortberg:
I discussed this with some Coq experts here and apparently the code for `now` above is very slow (probably because it is doing too complicated things). So exchanging it for something very simple should hopefully improve the compilation time of UniMath. 

I thinkt he SSReflect tactic `done` is a good place to start when designing our own `UniMath_easy` tactic:
```
Ltac done :=
  trivial; hnf; intros; solve
   [ do ![solve [trivial | apply: sym_equal; trivial]
         | discriminate | contradiction | split]
   | case not_locked_false_eq_true; assumption
   | match goal with H : ¬ _ |- _ ⇒ solve [case H; trivial] end ].
```

benediktahrens:
On 03/02/2017 02:59 PM, Anders Mörtberg wrote:
> I discussed this with some Coq experts here and apparently the code for `now` above is very slow (probably because it is doing too complicated things). So exchanging it for something very simple should hopefully improve the compilation time of UniMath. 
> 
[...]

Issue #609, opened 2017-02-24
Against usage of `now`.


benediktahrens:
Sometimes, things break at a line starting with `now`. 
That often means that one or several hypotheses, previously successfully applied by `now`, look different now. But it is not recorded which are hypotheses applied by `now`.
That makes it very difficult and annoying to debug such breakage. Please stop using that tactical.
@mortberg

mortberg:
But that's the whole point of using `now`... One wants to quickly find exactly where things break if hypotheses change. Do you have some example?

benediktahrens:
On 02/24/2017 04:02 PM, Anders Mörtberg wrote:
> But that's the whole point of using `now`... One wants to quickly
> find exactly where things break if hypotheses change. Do you have
> some example?

Absolutely: go to branch 'adjunction' and see the 'admit' that I have
temporarily put into the file CT/Adjunctions.v

DanGrayson:
This proposed prohibition of "now" seems to be a complaint that could be leveled against the use any tactic that finds a proof for us.  But the natural progression will be for us to use smarter and smarter tactics, so we don't have to find them.

benediktahrens:
On 02/24/2017 04:02 PM, Anders Mörtberg wrote:
> But that's the whole point of using `now`... One wants to quickly
> find exactly where things break if hypotheses change. Do you have
> some example?

Your remark is a bit misleading, in the following sense:
`now` does two things

1. some proof automation
2. check that the tactic finishes the subgoal

However, we are already using bullets to achieve 2., and we (I?) do not
want 1., for exactly the reason that when things break, it becomes
difficult to debug.

mortberg:
If I understand it correctly the problem would have been the same if I would have used `assumption` as well right?

mortberg:
I guess using `exact bla.` would have been perfect here?

benediktahrens:
On 02/24/2017 04:12 PM, Anders Mörtberg wrote:
> If I understand it correctly the problem would have been the same if
> I would have used `assumption` as well right?

I understand what assumption does, but I don't know what `now` does.
`assumption` did not work in the case where I got stuck.

benediktahrens:
On 02/24/2017 04:13 PM, Anders Mörtberg wrote:
> I guess using `exact bla.` would have been perfect here?

`apply` or `exact` are much more informative than `assumption` or `now`.

vladimirias:
[...]

PR #646, opened 2017-03-09
Simplified now tactic

mortberg:
The goal of this PR is to resolve https://github.com/UniMath/UniMath/issues/621.

Some of us have been using the Coq tactic `now` quite heavily, what we didn't know was that it was doing a lot of complicated things. This tactic is used in just over 1600 times in more or less every package (including Foundations). As pointed out in https://github.com/UniMath/UniMath/issues/609 this has led to some problems with proof maintance.

In this PR I define a very simple tactic called `unimath_easy` which is a replacement for the complicated Coq tactic `easy`, this is then used to define our own `now` tactic which is supposed to be very simple and only try to solve really trivial goals. The new `unimath_easy` tactic works as follows:
1. First try to solve the goal using `trivial` (https://coq.inria.fr/refman/Reference-Manual010.html#hevea_tactic150). This solves things like X=X. 
2. If that didn't work do `intros` and then try to repeatedly decompose the goal using `split` and apply pathsinv0 and try to solve the goal using `trivial`. If that didn't work check if there is a negated type or the empty type in the context and call induction on this hypothesis (this seems better than calling the standard Coq tactic `contradiction` as it is not clear to me if that will use the eliminator or not). 

With this very simple tactic all but 6 uses of `now` still works. The few that failed were very easy to fix. While fixing them I noticed that they were using the Coq tactic `inversion` (https://coq.inria.fr/refman/Reference-Manual010.html#hevea_tactic95) which I think was very bad because this is a very complicated and slow tactic. I have also no idea what kind of proof terms it is generating.

Because `now` is used all over UniMath I had to put this as early as possible in PartA (directly after the definition of `pathsinv0`), so I guess this PR cannot be merged until the lock on the Foundations has been lifted.

### DIFF
--- a/UniMath/Algebra/Apartness.v
+++ b/UniMath/Algebra/Apartness.v
@@ -8,6 +8,8 @@ Unset Kernel Term Sharing.
 Require Export UniMath.Algebra.BinaryOperations.
 Require Import UniMath.Foundations.Propositions.
 
+Require Import UniMath.MoreFoundations.Tactics.
+
 (** ** Additionals theorems about relations *)
 
 Lemma isapropisirrefl {X : UU} (rel : hrel X) :
--- a/UniMath/Algebra/Archimedean.v
+++ b/UniMath/Algebra/Archimedean.v
@@ -17,6 +17,8 @@ Require Import UniMath.Algebra.DivisionRig .
 Require Import UniMath.Algebra.Domains_and_Fields .
 Require Import UniMath.Algebra.ConstructiveStructures.
 
+Require Import UniMath.MoreFoundations.Tactics.
+
 (** ** The standard function from the natural numbers to a monoid *)
 
 Fixpoint natmult {X : monoid} (n : nat) (x : X) : X :=
--- a/UniMath/Algebra/ConstructiveStructures.v
+++ b/UniMath/Algebra/ConstructiveStructures.v
@@ -9,6 +9,8 @@ Require Export UniMath.Algebra.Apartness.
 Require Export UniMath.Algebra.DivisionRig.
 Require Export UniMath.Algebra.Domains_and_Fields.
 
+Require Import UniMath.MoreFoundations.Tactics.
+
 (** ** Predicats in a rig with a tight appartness relation *)
 
 Definition isnonzeroCR (X : rig) (R : tightap X) := R 1%rig 0%rig.
--- a/UniMath/Algebra/DivisionRig.v
+++ b/UniMath/Algebra/DivisionRig.v
@@ -15,6 +15,8 @@ Require Export UniMath.Foundations.Sets.
 Require Export UniMath.Algebra.Rigs_and_Rings.
 Require Export UniMath.Algebra.Domains_and_Fields.
 
+Require Import UniMath.MoreFoundations.Tactics.
+
 (** ** Definition of a DivRig *)
 
 Definition isnonzerorig (X : rig) : UU := (1%rig : X) != 0%rig.
--- a/UniMath/Algebra/IteratedBinaryOperations.v
+++ b/UniMath/Algebra/IteratedBinaryOperations.v
@@ -2,6 +2,9 @@ Require Export UniMath.Combinatorics.Lists.
 Require Export UniMath.Combinatorics.FiniteSequences.
 Require Export UniMath.Algebra.Rigs_and_Rings.
 Require Export UniMath.Foundations.UnivalenceAxiom.
+
+Require Import UniMath.MoreFoundations.Tactics.
+
 Unset Automatic Introduction.
 
 (* move upstream *)
--- a/UniMath/Algebra/Lattice.v
+++ b/UniMath/Algebra/Lattice.v
@@ -31,6 +31,8 @@ Truncated minus is a lattice:
 
 Require Export UniMath.Algebra.Monoids_and_Groups.
 
+Require Import UniMath.MoreFoundations.Tactics.
+
 Unset Automatic Introduction.
 
 (** ** Strong Order *)
--- a/UniMath/CategoryTheory/Adjunctions.v
+++ b/UniMath/CategoryTheory/Adjunctions.v
@@ -24,10 +24,13 @@ Require Import UniMath.Foundations.PartD.
 Require Import UniMath.Foundations.Propositions.
 Require Import UniMath.Foundations.Sets.
 
+Require Import UniMath.MoreFoundations.Tactics.
+
 Require Import UniMath.CategoryTheory.precategories.
 Require Import UniMath.CategoryTheory.functor_categories.
-Local Open Scope cat.
 Require Import UniMath.CategoryTheory.whiskering.
+
+Local Open Scope cat.
 
 (** * Adjunctions *)
 Section adjunctions.
--- a/UniMath/CategoryTheory/CocontFunctors.v
+++ b/UniMath/CategoryTheory/CocontFunctors.v
@@ -60,10 +60,11 @@ Require Import UniMath.Foundations.Propositions.
 Require Import UniMath.Foundations.Sets.
 Require Import UniMath.Foundations.NaturalNumbers.
 
+Require Import UniMath.MoreFoundations.Tactics.
+
 Require Import UniMath.CategoryTheory.total2_paths.
 Require Import UniMath.CategoryTheory.precategories.
 Require Import UniMath.CategoryTheory.functor_categories.
-Local Open Scope cat.
 Require Import UniMath.CategoryTheory.limits.graphs.colimits.
 Require Import UniMath.CategoryTheory.category_hset.
 Require Import UniMath.CategoryTheory.category_hset_structures.
@@ -83,6 +84,8 @@ Require Import UniMath.CategoryTheory.exponentials.
 Require Import UniMath.CategoryTheory.whiskering.
 Require Import UniMath.CategoryTheory.RightKanExtension.
 Require Import UniMath.CategoryTheory.slicecat.
+
+Local Open Scope cat.
 
 (** * Definition of cocontinuous functors *)
 Section cocont.
--- a/UniMath/CategoryTheory/DiscretePrecategory.v
+++ b/UniMath/CategoryTheory/DiscretePrecategory.v
@@ -12,8 +12,11 @@ Require Import UniMath.Foundations.PartD.
 Require Import UniMath.Foundations.Propositions.
 Require Import UniMath.Foundations.Sets.
 
+Require Import UniMath.MoreFoundations.Tactics.
+
 Require Import UniMath.CategoryTheory.precategories.
 Require Import UniMath.CategoryTheory.functor_categories.
+
 Local Open Scope cat.
 
 (** * Discrete precategories *)
--- a/UniMath/CategoryTheory/Elements.v
+++ b/UniMath/CategoryTheory/Elements.v
@@ -19,6 +19,8 @@ Require Import UniMath.Foundations.PartD.
 Require Import UniMath.Foundations.Propositions.
 Require Import UniMath.Foundations.Sets.
 
+Require Import UniMath.MoreFoundations.Tactics.
+
 Require Import UniMath.CategoryTheory.precategories.
 Require Import UniMath.CategoryTheory.functor_categories.
 Require Import UniMath.CategoryTheory.category_hset.
--- a/UniMath/CategoryTheory/ElementsOp.v
+++ b/UniMath/CategoryTheory/ElementsOp.v
@@ -19,6 +19,8 @@ Require Import UniMath.Foundations.PartD.
 Require Import UniMath.Foundations.Propositions.
 Require Import UniMath.Foundations.Sets.
 
+Require Import UniMath.MoreFoundations.Tactics.
+
 Require Import UniMath.CategoryTheory.precategories.
 Require Import UniMath.CategoryTheory.functor_categories.
 Require Import UniMath.CategoryTheory.category_hset.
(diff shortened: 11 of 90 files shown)


DanGrayson:
Why does this bit work?
```
H : ¬ _ |- _ => solve [induction H; trivial]
```
The induction tactic is supposed to be for terms whose type is an induction type.   (I see that it does work.)

DanGrayson:
I think "solve" is redundant when there is only one thing to try, so you can
change `solve [induction H; trivial]` to `induction H; trivial`.

DanGrayson:
The "match" tactic will accept multiple branches, as in 
```

Ltac unimath_easy :=
  trivial; intros; solve
   [ repeat (solve [trivial | apply pathsinv0; trivial] || split)
   | match goal with
       | H : ¬ _ |- _ => solve [induction H; trivial]
       | H : ∅   |- _ => induction H
     end ].
```
, which looks nicer and might be more efficient.

DanGrayson:
We should also be able to solve when something in the context is of type `X -> empty`, so this might be better:
```
Ltac unimath_easy :=
  trivial; intros; solve
   [ repeat (solve [trivial | apply pathsinv0; trivial] || split)
   | match goal with
       | H : ¬ _ |- _ => solve [induction H; trivial]
       | H : _ → ∅ |- _ => solve [induction H; trivial]
       | H : ∅   |- _ => induction H
     end ].
```

mortberg:
I can't find it in the documentation, maybe it's a special case of the
induction tactic? I got this trick from the SSReflect done tactic I posted
before.

--
Anders

On Thu, Mar 9, 2017 at 10:19 PM, Daniel R. Grayson <notifications@github.com

DanGrayson:
Adding a case `H : _ -> _ -> ∅ |- _ => induction H; trivial` to the match would allow
two steps in the proof of `isapropcoprod`

DanGrayson:
This is better:
```
Ltac unimath_easy :=
  trivial; intros; solve
   [ repeat (solve [trivial | apply pathsinv0; trivial] || split)
   | match goal with
       | H : ¬ _ |- _ => induction H; trivial
       | H : _ → ∅ |- _ => induction H; trivial
       | H : _ -> _ -> ∅ |- _ => induction H; trivial
       | H : ∅   |- _ => induction H
     end ].
```

mortberg:
Thanks for the suggestions. I wonder if this leads to some speed-up, the old `now` tactic was doing so many complicated things that I wouldn't be surprised if the new simple one makes things compile faster.

mortberg:
Apparently this modified tactic doesn't solve all goals in RealNumbers. I'll investigate.

mortberg:
I see what happened, the goal was looking like this:
```
  x : Dcuts
  Hx : isboolDcuts x
  Hx0 : ¬ (0%NRat ∈ x)
  r : NonnegativeRationals
  Hr : ∅
  ============================
  r ∈ x
```
So the old tactic would first try `induction` on Hx0, if that didn't solve the goal it would do induction on Hr. With the new tactic it will only call `induction` on Hx0...

I suggest we split the `match` into 4 different matches so that we know that all of them will be tried if possible. The case for `∅` should be the first one as if it applies it is guaranteed to solve the goal.

DanGrayson:
Good point.  I guess there's no way to get it to try induction on all the entries in the context that match a given pattern.  Oh, well.

lgtm

DanGrayson:
@benediktahrens ?   Does this one look okay to you?

benediktahrens:
On 03/12/2017 03:04 PM, Daniel R. Grayson wrote:
> @benediktahrens ?   Does this one look okay to you?

Well, it changes Foundations, which it shouldn't.
The tactic declaration could go into MoreFoundations instead.

Otherwise it looks fine to me.

mortberg:
The `now` tactic is used 94 times in Foundations, which is why I put the new one as early as possible. By putting it in MoreFoundations these 94 occurences will remain as the bad experimental Coq tactic. I'm not sure how to solve this, maybe these 94 occurences should be removed completely once the lock on Foundations has been lifted.

benediktahrens:
On 03/13/2017 10:33 AM, Anders Mörtberg wrote:
> The `now` tactic is used 94 times in Foundations, which is why I put
> the new one as early as possible. By putting it in MoreFoundations
> these 94 occurences will remain as the bad experimental Coq tactic.
> I'm not sure how to solve this, maybe these 94 occurences should be
> removed completely once the lock on Foundations has been lifted.

I could imagine that Vladimir would consider the removal of the `now`
tactical from Foundations a significant improvement.

I think that for the time being, the new `now` should go in MoreFoundations.

mortberg:
Ok, I'll move the tactic. I'll also make an issue about removing the now's from Foundations. I expect this to be easy, but tedious.

mortberg:
At least we now know that none of them need anything more complicated than what `unimath_easy` does

mortberg:
Hmm, a problem with moving this to MoreFoundations is that I need to add an import to every file using `now`. If someone forget to add such a line they will still use the Coq `now` tactic... It would be a lot better if this could appear very early on in the development so that we are sure that when someone uses `now` it means what it should mean. Maybe the best is to wait until the lock on Foundations has been lifted and then merge this? This change is really innocent in the sense that it is not adding any new results to Foundations, it is just fixing a tactic so that it does what we expect it to do. 

Another thought I had is that maybe we should gather all tactics defined in Foundations in a separate file (maybe in MoreFoundations)? At the moment they appear scattered all over the place which means that it is hard to know what there is.

benediktahrens:
On 03/13/2017 02:09 PM, Anders Mörtberg wrote:
> Hmm, a problem with moving this to MoreFoundations is that I need to
> add an import to every file using `now`. If someone forget to add
> such a line they will still use the Coq `now` tactic... It would be a
> lot better if this could appear very early on in the development so
> that we are sure that when someone uses `now` it means what it should
> mean. Maybe the best is to wait until the lock on Foundations has
> been lifted and then merge this? This change is really innocent in
> the sense that it is not adding any new results to Foundations, it is
> just fixing a tactic so that it does what we expect it to do.

VV wants to clean Foundations of "noise". A tactic that is never even
used there (or should never be used there) is certainly noise in that sense.

> Another thought I had is that maybe we should gather all tactics
> defined in Foundations in a separate file (maybe in MoreFoundations)?
> At the moment they appear scattered all over the place which means
> that it is hard to know what there is.

That is a good idea.

mortberg:
So I'll make a file called Tactics in MoreFoundations, in there we will only have `unimath_easy` and `now` for now. Eventually other tactics from Foundations will be moved there to reduce the "noise". 

All of the files using `now` should import this Tactics file. Maybe the easiest way to achieve this is to Require Export this from toplevel files in each package (e.g. precategories.v in CategoryTheory)? Or should I add a Require Import to all of the affected files? Another alternative would be to make all files import MoreFoundations/All.v which will give them all of Foundations and MoreFoundations, the good thing with this is that one doesn't have to be careful about what one imports from (More)Foundations, the drawback is that we get a less granular dependency tree...

vladimirias:
> On Mar 13, 2017, at 9:09 AM, Anders Mörtberg <notifications@github.com> wrote:
> 
> Another thought I had is that maybe we should gather all tactics defined in Foundations in a separate file (maybe in MoreFoundations)? At the moment they appear scattered all over the place which means that it is hard to know what there is.
> 
This is a very good idea.

mortberg:
@DanGrayson and @benediktahrens : If I make a file called Tactics in MoreFoundations with `now` in it, what is the best way to make all the files in UniMath (outside of Foundations) that depend on `now` use this? Adding an import to every file using `now` seems both tedious and not so nice. I think the idea of importing MoreFoundations.All instead of various subsets of the files in Foundations and MoreFoundations in all of the other UniMath packages might be good. The dependency tree will then have all of (More)Foundations as root and there is no risk that any files use the wrong version of `now`, but the price is a loss of granularity in the dependency tree.

DanGrayson:
For K-theory, my plan is to make all the files there import MoreFoundations.All , so you may
free to do that there.  The loss of granularity is of no concern to me, because when working on
K-theory I will not be modifying MoreFoundations.

mortberg:
I decided that the best is to import MoreFoundations/Tactics in all of the files, outside of Foundations, that uses `now` in UniMath. I compiled a complete list of them below:
```
UniMath/Algebra/Apartness.v
UniMath/Algebra/Archimedean.v
UniMath/Algebra/ConstructiveStructures.v
UniMath/Algebra/DivisionRig.v
UniMath/Algebra/IteratedBinaryOperations.v
UniMath/Algebra/Lattice.v
UniMath/CategoryTheory/Adjunctions.v
UniMath/CategoryTheory/bicategories/whiskering.v
UniMath/CategoryTheory/category_hset_structures.v
UniMath/CategoryTheory/CocontFunctors.v
UniMath/CategoryTheory/DiscretePrecategory.v
UniMath/CategoryTheory/ElementsOp.v
UniMath/CategoryTheory/Elements.v
UniMath/CategoryTheory/EndofunctorsMonoidal.v
UniMath/CategoryTheory/EpiFacts.v
UniMath/CategoryTheory/EquivalencesExamples.v
UniMath/CategoryTheory/equivalences.v
UniMath/CategoryTheory/exponentials.v
UniMath/CategoryTheory/FunctorAlgebras.v
UniMath/CategoryTheory/functor_categories.v
UniMath/CategoryTheory/HorizontalComposition.v
UniMath/CategoryTheory/Inductives/LambdaCalculus.v
UniMath/CategoryTheory/Inductives/Lists.v
UniMath/CategoryTheory/Inductives/Trees.v
UniMath/CategoryTheory/limits/bincoproducts.v
UniMath/CategoryTheory/limits/BinDirectSums.v
UniMath/CategoryTheory/limits/binproducts.v
UniMath/CategoryTheory/limits/cats/limits.v
UniMath/CategoryTheory/limits/coproducts.v
UniMath/CategoryTheory/limits/graphs/bincoproducts.v
UniMath/CategoryTheory/limits/graphs/binproducts.v
UniMath/CategoryTheory/limits/graphs/colimits.v
UniMath/CategoryTheory/limits/graphs/eqdiag.v
UniMath/CategoryTheory/limits/graphs/initial.v
UniMath/CategoryTheory/limits/graphs/limits.v
UniMath/CategoryTheory/limits/graphs/terminal.v
UniMath/CategoryTheory/limits/products.v
UniMath/CategoryTheory/limits/pullbacks.v
UniMath/CategoryTheory/limits/pushouts.v
UniMath/CategoryTheory/Monads.v
UniMath/CategoryTheory/opp_precat.v
UniMath/CategoryTheory/precategories.v
UniMath/CategoryTheory/precomp_ess_surj.v
UniMath/CategoryTheory/RightKanExtension.v
UniMath/CategoryTheory/RModules.v
UniMath/CategoryTheory/SetValuedFunctors.v
UniMath/CategoryTheory/ShortExactSequences.v
UniMath/CategoryTheory/SimplicialSets.v
UniMath/CategoryTheory/slicecat.v
UniMath/CategoryTheory/Subobjects.v
UniMath/Combinatorics/FiniteSequences.v
UniMath/Combinatorics/FiniteSets.v
UniMath/Combinatorics/Lists.v
UniMath/Combinatorics/OrderedSets.v
UniMath/Combinatorics/StandardFiniteSets.v
UniMath/HomologicalAlgebra/KA.v
UniMath/Ktheory/AffineLine.v
UniMath/Ktheory/Bifunctor.v
UniMath/Ktheory/Equivalences.v
UniMath/Ktheory/Nat.v
UniMath/Ktheory/Precategories.v
UniMath/Ktheory/Representation.v
UniMath/Ktheory/Utilities.v
UniMath/MoreFoundations/DecidablePropositions.v
UniMath/MoreFoundations/Tactics.v
UniMath/NumberSystems/NaturalNumbersAlgebra.v
UniMath/PAdics/frac.v
UniMath/RealNumbers/DecidableDedekindCuts.v
UniMath/RealNumbers/NonnegativeRationals.v
UniMath/RealNumbers/NonnegativeReals.v
UniMath/RealNumbers/Prelim.v
UniMath/RealNumbers/Reals.v
UniMath/RealNumbers/Sets.v
UniMath/SubstitutionSystems/BindingSigToMonad.v
UniMath/SubstitutionSystems/GenMendlerIteration_alt.v
UniMath/SubstitutionSystems/LamFromBindingSig.v
UniMath/SubstitutionSystems/Lam.v
UniMath/SubstitutionSystems/LiftingInitial_alt.v
UniMath/SubstitutionSystems/LiftingInitial_new.v
UniMath/SubstitutionSystems/LiftingInitial.v
UniMath/SubstitutionSystems/MonadsFromSubstitutionSystems.v
UniMath/SubstitutionSystems/MultiSorted.v
UniMath/SubstitutionSystems/SignatureCategory.v
UniMath/SubstitutionSystems/SignatureExamples.v
UniMath/SubstitutionSystems/STLC.v
UniMath/Tactics/Nat_Tactics.v
UniMath/Topology/Filters.v
UniMath/Topology/Prelim.v
UniMath/Topology/Topology.v
```

@DanGrayson : Do you have an idea how I can script this?

DanGrayson:
You could try something like this:
```
$ for i in a b c d ; do echo hi there > $i ; done
$ for i in a b c d ; do mv $i $i.tmp ; echo "Require something." > $i ; cat $i.tmp >> $i ; rm $i.tmp ; done
$ for i in a b c d ; do mv $i $i.tmp ; echo "Require something else." > $i ; cat $i.tmp >> $i ; rm $i.tmp ; done
$ head *
==> a <==
Require something else.
Require something.
hi there

==> b <==
Require something else.
Require something.
hi there

==> c <==
Require something else.
Require something.
hi there

==> d <==
Require something else.
Require something.
hi there
```

mortberg:
Thanks Dan, I'll see if I can come up with something nice.

mortberg:
I've now included the new Tactics file in all files using `now`. Once Travis is happy I think we should merge as this affects so many files.

Issue #643, opened 2017-03-09
switch total2 to Record to make "apply" totally cool


DanGrayson:
Here is the documentation of the tactic "apply":

"apply term

This tactic applies to any goal. The argument term is a term well-formed in the local context. The
tactic apply tries to match the current goal against the conclusion of the type of term. If it succeeds,
then the tactic returns as many subgoals as the number of non-dependent premises of the type of term.
If the conclusion of the type of term does not match the goal and the conclusion is an inductive type
isomorphic to a tuple type, then each component of the tuple is recursively matched to the goal in the
left-to-right order."

The last sentence would be totally cool for us if only total2 were isomorphic to a tuple type, because then, if we needed some sequence of projections `pr1` and `pr2` applied to a term `w` to fulfill the current goal, then we could just use `apply w` instead of something like `apply (pr1 (pr2 (pr1 (pr2 (pr1 w)))))`.  It would be if only `total2` were defined as a record instead of as an inductive type.  I have tested it, and that change would work for us.  All we have to do is to change the definition of `total2` and to delete the definitions of `pr1` and of `pr2`.

We have to switch `total2` to be a record soon, anyway, once we get primitive projections working, so we can have the benefit of eta-reduction for pairs.  I propose we do it now, so we can get the benefit of this cool feature of `apply`.

Yes, I know Foundations is "locked", but this is such a simple change, which we have to implement eventually anyway, that it should be possible to consider it and approve it in one minute.

benediktahrens:
On 03/09/2017 05:40 PM, Daniel R. Grayson wrote:
> Here is the documentation of the tactic "apply":
> 
> "apply term
> 
> This tactic applies to any goal. The argument term is a term
> well-formed in the local context. The tactic apply tries to match the
> current goal against the conclusion of the type of term. If it
> succeeds, then the tactic returns as many subgoals as the number of
> non-dependent premises of the type of term. If the conclusion of the
> type of term does not match the goal and the conclusion is an
> inductive type isomorphic to a tuple type, then each component of the
> tuple is recursively matched to the goal in the left-to-right
> order."
> 
> The last sentence would be totally cool for us if only total2 were
> isomorphic to a tuple type, because then, if we needed some sequence
> of projections `pr1` and `pr2` applied to a term `w` to fulfill the
> current goal, then we could just use `apply w` instead of something
> like `apply (pr1 (pr2 (pr1 (pr2 (pr1 w)))))`.  It would be if only
[...]

PR #647, opened 2017-03-10
switch to total2 defined as a record

DanGrayson:
This would resolve issue #643, which is still under discussion, if we
decide to make the switch.

### DIFF
--- a/UniMath/Foundations/Preamble.v
+++ b/UniMath/Foundations/Preamble.v
@@ -118,7 +118,7 @@ One can not use a new record each time one needs it because the general theorems
 construction would not apply to new instances of "Record" due to the "generativity" of inductive
 definitions in Coq.
 
-We use "Inductive" instead of "Record" here.
+We use "Record" instead of "Inductive" here.
 
 Using "Record" which is equivalent to "Structure" would allow us later to use the mechanism of
 canonical structures with total2.
@@ -131,32 +131,11 @@ if we used "Record", has a known interpretation in the framework of the univalen
 
 *)
 
-(* two alternatives: *)
-(* total2 as a record with primitive projections: *)
+(* Set Primitive Projections. *)
 
-    (* Set Primitive Projections. *)
+Set Nonrecursive Elimination Schemes.
 
-    (* Set Nonrecursive Elimination Schemes. *)
-
-    (* Record total2 { T: Type } ( P: T -> Type ) := tpair { pr1 : T; pr2 : P pr1 }. *)
-
-(* or total2 as an inductive type:  *)
-
-    Inductive total2 { T: Type } ( P: T -> Type ) := tpair : ∏ (__t__:T) (__p__:P __t__), total2 P.
-
-    (* Do not use "induction" without specifying names; seeing __t__ or __p__ will indicate that you *)
-    (*    did that.  This will prepare for the use of primitive projections, when the names will be pr1 *)
-    (*    and pr2. *)
-
-    Definition pr1 { T : Type } { P : T -> Type } ( t : total2 P ) : T .
-    Proof . intros .  induction t as [ t p ] . exact t . Defined.
-
-    Definition pr2 { T : Type } { P : T -> Type } ( t : total2 P ) : P ( pr1 t ) .
-    Proof . intros .  induction t as [ t p ] . exact p . Defined.
-
-(* end of two alternatives *)
-
-    Print total2.               (* log which definition of total2 is currently in use *)
+Record total2 { T: Type } ( P: T -> Type ) := tpair { pr1 : T; pr2 : P pr1 }.
 
 Arguments tpair {_} _ _ _.
 Arguments pr1 {_ _} _.
@@ -172,8 +151,6 @@ Definition whether_primitive_projections : bool.
 Proof.
   tryif primitive_projections then exact true else exact false.
 Defined.
-
-Print whether_primitive_projections.
 
 Notation "'∑'  x .. y , P" := (total2 (fun x => .. (total2 (fun y => P)) ..))
   (at level 200, x binder, y binder, right associativity) : type_scope.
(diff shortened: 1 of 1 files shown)


vladimirias:
Can we followup on my second suggestion about definition total2_rec for the total2 defined as a record and of a tactic that would use it smartly to obtain proof terms with total2_rec instead of the ones with the match?

The argument that since Coq people checked that match can be translated into an induction principle we can use the match is a very weak one. It has not been done mathematically.

DanGrayson:
Re: "Can we followup on my second suggestion about definition total2_rec for the total2 defined as a record and of a tactic that would use it smartly to obtain proof terms with total2_rec instead of the ones with the match?"

Sure, I've just made a trial PR and I will see if I can add such a tactic to it.  The idea would be that we could use it instead of "induction" when operating on pairs.  We could call it "pair_induction".

DanGrayson:
Sorry, I don't see how to write such a tactic in the Ltac language.  Here are the steps that one can perform manually:
```
Goal ∏ X (Y:X->UU) (w:∑ x, Y x) (u: Y(pr1 w)), Y (pr1 w).
  intros.
  generalize u; clear u.
  simple refine (total2_rect X Y (λ w, Y (pr1 w) -> Y (pr1 w)) _ w).
  clear w; intros x y u.
  change (Y x).
  change (Y x) in u.
  exact y.
Defined.
```
Somehow the tactic has to figure out the same steps, and I don't see how.

DanGrayson:
@benediktahrens , does this PR look okay to you?

vladimirias:
It is *not* a small change. It is a major change and I do want to wait about it. By doing this change we move the UniMath another step away from the Martin-Lof type theory. This will have to be reflected in the papers that explain UniMath as a language. It will make older such papers incorrect in their description of the UniMath. It does in fact mean that we change the underlying type system and therefore move from what UniMath is to another language that we will have to call UniMath.2.0 or something similar.

vladimirias:
Just to make myself clear - I am not against this change. I think it is a progressive change and we should make it. I was just pointing out some of the consequences that we should be ready to deal with.

It does make sense to have versions in the UniMath anyway. A major restructuring of the core with the MoreFoundations and eventually creating EssentialFoundations deserves a new version number as well.

When, hopefully, we’ll be able to fix the universe issues it will be a new version. So I think versions are necessary.

Vladimir.

DanGrayson:
A simple way to add version numbers to the repository would be to use "git tag" to attach names to various commits.  The names could something like "version-n" or "version-m.n", etc.  We have one such tag, https://github.com/UniMath/UniMath/tree/v0.1 , from about a year ago, which could be renamed.  The list of branch names and tag names is visible to all on github.  Git tags can be annotated and cryptographically signed:
```
       A GnuPG signed tag object will be created when -s or -u <keyid> is used. When -u
       <keyid> is not used, the committer identity for the current user is used to find
       the GnuPG key for signing. The configuration variable gpg.program is used to
       specify custom GnuPG binary.

       Tag objects (created with -a, -s, or -u) are called "annotated" tags; they
       contain a creation date, the tagger name and e-mail, a tagging message, and an
       optional GnuPG signature. Whereas a "lightweight" tag is simply a name for an
       object (usually a commit object).

       Annotated tags are meant for release while lightweight tags are meant for private
       or temporary object labels. For this reason, some git commands for naming objects
       (like git describe) will ignore lightweight tags by default.
```

DanGrayson:
Re: "This will have to be reflected in the papers that explain UniMath as a language. It will make older such papers incorrect in their description of the UniMath. It does in fact mean that we change the underlying type system and therefore move from what UniMath is to another language that we will have to call UniMath.2.0 or something similar."

I think you are not right about that.

There is no need to treat "match" mathematically, when used just in the context of induction on pairs, to justify it.  Instead, we may just say that the "match" construction is simply an abbreviation for the eliminator that omits explicit mention of the first two parameters (X:Type) and (Y:X->Type).

I.e., whenever you see
```
match w as t0 return Q with | tpair _ _ x y => body end
```
in a proof term, just mentally replace it by
```
@total2_rect _ _ (λ t0, Q) (λ x y, body) w
```

vladimirias:
It is not about “match”. It is about the fact the Martin-Lof type theory does not have eta for pairs.

DanGrayson:
This pull request doesn't turn on primitive projections, so it doesn't introduce eta for pairs.

vladimirias:
This is a good argument. Have you tried what Jason Gross suggested to make Coq to create the eliminator for the total2 despite it being a record?

mortberg:
I think git tags is a great way to make a snapshot of a development for a
paper. In the paper you want to be able to refer to a development at a
specific point in time, but you still want the development to be able to
evolve without making the references in the paper outdated. By making a git
tag you can refer to the tag in the paper without being afraid of the link
ever getting outdated (unless github shuts down in which case we will have
more serious problems than our papers containing dead links).

--
Anders

On Tue, Mar 14, 2017 at 4:04 PM, Daniel R. Grayson <notifications@github.com

DanGrayson:
Re: "Have you tried what Jason Gross suggested to make Coq to create the eliminator for the total2 despite it being a record?"

That was in the original pull request, so the eliminator, total2_rect, was present.  But "induction" doesn't use it, even if it's there, and I couldn't figure out a way to write a tactic as good as "induction" in the Ltac language for using it.

mortberg:
That is strange, I remember seeing a trick many years ago where someone
defined a type without an induction principle (using CoInductive) and then
defined their own induction principle and named it "mytype_rect" (assuming
that the type was called "mytype") and the standard Coq tactics would then
pick this. I wonder if they did something more... I'll do some digging.


On Tue, Mar 14, 2017 at 9:28 PM, Daniel R. Grayson <notifications@github.com

vladimirias:
@JasonGross do you have any suggestions for us?

JasonGross:
It looks like if your type is a non-recursive record, Coq will always do `destruct`; see https://github.com/coq/coq/blob/a52d06ea16cff00faa7d2f63ad5c1ca0b58e64b4/tactics/tactics.ml#L1480.  If your type is recursive / not a record, Coq will pick up any constant named `mytype_rect` (or `mytype_ind`, or whatever).  You could submit a feature request?

DanGrayson:
This is included in #652, closing.

PR #650, opened 2017-03-14
fix an unused variable name warning

Zimmi48:
This removes a spurious `n` which was not used and warned as such.
This warning is going to be transformed into an error when https://github.com/coq/coq/pull/474 gets merged so this fix is needed.

### DIFF
--- a/UniMath/Foundations/NaturalNumbers.v
+++ b/UniMath/Foundations/NaturalNumbers.v
@@ -852,7 +852,7 @@ Defined.
 
 Lemma natplusr0 (n : nat) : (n + 0) = n.
 Proof.
-  intro. induction n as [ | n IH n ].
+  intro. induction n as [ | n IH].
   - apply idpath.
   - simpl. apply (maponpaths S IH).
 Defined.


mortberg:
How does one get coqc to print this kind of warnings?

When I do `coqc -w all UniMath/Foundations/NaturalNumbers.v` I only get warnings about notations being declared in different scopes. If I load the file in emacs I do get the warning when processing this line, but as it is on line 855 of a file on nat I would never find it.

Zimmi48:
That's puzzling indeed. I even tried to make the warning an error with the following command but it had no effect:

    coqc -w "+unused-intro-pattern +tactics" UniMath/Foundations/NaturalNumbers.v

In CoqIDE though, the same option does make it an error.

@maximedenes @herbelin Do you have an answer? Do you know why this warning only appears in CoqIDE?

maximedenes:
There is a suspicious `Flags.is_verbose ()` test.

Zimmi48:
@maximedenes You're right. This patch

```
diff --git a/tactics/tactics.ml b/tactics/tactics.ml
index 6205bd109..946ce6e7f 100644
--- a/tactics/tactics.ml
+++ b/tactics/tactics.ml
@@ -3010,7 +3010,7 @@ let warn_unused_intro_pattern =
         (fun c -> Printer.pr_constr (fst (run_delayed (Global.env()) Evd.empty c)))) names)
 
 let check_unused_names names =
-  if not (List.is_empty names) && Flags.is_verbose () then
+  if not (List.is_empty names) then
     warn_unused_intro_pattern names
 
 let intropattern_of_name gl avoid = function
```

corrects this behavior (the warning is printed by coqc and can also be made an error). I suppose it makes sense to fix this in 8.6, doesn't it?

maximedenes:
It probably does, yes. Does it show up e.g. on the stdlib? We can anyway discuss what the default value should be.

DanGrayson:
lgtm.  @vladimirias , this is a needed change to the Foundations.

Zimmi48:
@maximedenes It does: twice in `theories/Lists/List.v` and once in `theories/QArith/Qround.v`.

DanGrayson:
We could have ProofGeneral tell Coq to hush warnings such as
```
Toplevel input, characters 166-368:
 Warning: Notation _ ≠ _ was already used in scope nat_scope
 [notation-overridden,parsing]
```
, if people wish it.  Just let me know. ( @benediktahrens  )

vladimirias:
I am fine with this change.

PR #651, opened 2017-03-14
Fix a notation

DanGrayson:
Fix a notation of mine so it is restricted to the use for which it was intended.
(It's annoying to see it on the screen when stepping through a proof unrelated
to standard finite sets.)


benediktahrens:
lgtm

mortberg:
lgtm

Issue #602, opened 2017-02-23
duality in category theory


DanGrayson:
A lot of our category theory code is doubled in size because notions and their duals get separate definitions and proofs.  Is there a way to economize?

More specifically, is ElementsOp needed?  If so, can its statements be corollaries of those in Elements?  The file is used here: https://github.com/mweav/UniMath/blob/elems-slice-equiv/UniMath/CategoryTheory/elems_slice_equiv.v

See Anders' comment here: https://github.com/UniMath/UniMath/pull/590#issuecomment-281938037

mortberg:
An issue that we have is that C^op^op is not judgmentally the same as C. I suspect that this is one of the reasons for the problems we had when dualizing statements about limits. A solution to this problem is to define categories as in the HoTT library:

https://github.com/HoTT/HoTT/blob/master/theories/Categories/Category/Core.v#L58

As you see a category comes with two proofs of associativity and op'ing a category just swaps the two proofs, hence one gets C^op^op=C judgmentally. Mathematically this is not so nice, but maybe it's a trick which is worthwhile to implement?

Note that in cubicaltt we don't have this problem as symmetry is an involution (i.e. `<i> p @ -i @ -i = <i> p @ i` judgmentally) and we get C^op^op=C for the standard definition of category:

https://github.com/mortberg/cubicaltt/blob/master/examples/opposite.ctt

DanGrayson:
Re: "one gets C^op^op=C judgmentally"

That would be wonderful, but it's only with primitive projections turned on, which still don't work for us.  I was working on that yesterday...

benediktahrens:
On 02/23/2017 02:36 PM, Anders Mörtberg wrote:
> An issue that we have is that C^op^op is not judgmentally the same as
> C. I suspect that this is one of the reasons for the problems we had
> when dualizing statements about limits. A solution to this problem is
> to define categories as in the HoTT library:
> 
> https://github.com/HoTT/HoTT/blob/master/theories/Categories/Category/Core.v#L58
>
>  As you see a category comes with two proofs of associativity and
> op'ing a category just swaps the two proofs, hence one gets C^op^op=C
> judgmentally. Mathematically this is not so nice, but maybe it's a
> trick which is worthwhile to implement?
> 
> Note that in cubicaltt we don't have this problem as symmetry is an
> involution (i.e. `<i> p @ -i @ -i = <i> p @ i` judgmentally) and we
> get C^op^op=C for the standard definition of category:
> 
[...]

Issue #77, opened 2015-02-16
UniMath-edu


vladimirias:
I suggest to create a branch of the UniMath called UniMath-edu (or something better) for the version of UniMath and of individual files of UniMath that are oriented towards teaching of UniMath and its style of formalization to mathematicians. There we will keep total2 to be an "Inductive", will avoid destruct (until a good explanation of the "match" to mathematicians can be found) etc.

On the main UniMath we will switch to total2 being a record, use primitive projections and start using canonical structures. Also on the main branch we will start using more advance features of Coq in the newer files (under the condition that the use of any such feature is explicitly described in the README of the corresponding directory!).

DanGrayson:
From "UniMath and its style of formalization" it seems you think there is one style of formalization.  But the two branches will have different styles, and one of them will not be taught.

vladimirias:
I suggest that the earlier files of the UniMath (Foundations, RezkCompletion) will be preserved in the way they are now (modulo the ongoing beautification). The only change will be in uuu.v and it will be with replacing the inductive definition of total2 with the Record definition.

This will enable those who write more advance code to use canonical structures with the total2.

The teaching of the basics of the UniMath will use the version with “Inductive” so that one can “print” various proofs during the demonstrations to show how they are built using the “total2_rect” (whose semantics is easy to explain).

Check the difference here:

****

Set Primitive Projections.

Record test ( T : UU ) ( P : T -> UU ) := test_constr { pr1 : T ; pr2 : P pr1 } .

Arguments pr1 { _ } { _ } _ .
Arguments pr2 { _ } { _ } _ .

Lemma testlemma ( T : UU ) ( P : T -> UU ) ( xp : test T P ) :
  xp = test_constr _ _ ( pr1 xp ) ( pr2 xp ) .
Proof .
  intros .
  induction xp as [ x p ] .
  apply idpath .

Defined.

Print testlemma .

Inductive test' ( T : UU ) ( P : T -> UU ) :=
  test'_constr : forall ( x : T ) ( p : P x ) , test' T P .

Definition pr1' { T : UU } { P : T -> UU } ( xp : test' T P ) : T .
Proof .
  intros . induction xp as [ x p ] . apply x .

Defined.

Definition pr2' { T : UU } { P : T -> UU } ( xp : test' T P ) : P ( pr1' xp ).
Proof .
  intros T P xp . induction xp as [ x p ] . apply p .

Defined.

[...]

PR #652, opened 2017-03-14
Primitive projections

### LINKED ISSUE #111: total2 as record, use of "primitive projections" ?

### LINKED ISSUE #643: switch total2 to Record to make "apply" totally cool

DanGrayson:
I finally got primitive projections with η-reduction for pairs working completely.  I ran a time trial and observed a speed-up of 56%.  That will go a long way to postponing the day when UniMath fissions or we start paying for longer travis builds.

Now, to get C^op^op and C to be judgmentally the same, all we need to do is to have two associativity axioms, f(gh)=(fg)h and (fg)h=f(gh), so the construction of C^op can just interchange them instead of applying pathsinv0.

The only file in the Foundations that is affected is Preamble.v, where total2 is defined. I added my defense of the use of the "match" construction for this one thing, stated in the comment https://github.com/UniMath/UniMath/pull/647#issuecomment-286489637 , to a comment in the file Preamble.v, for future reference

This includes #647.  
This resolves #643.  
This would help with #602.  
This resolves #111.
This affects #77.


DanGrayson:
The comment https://github.com/UniMath/UniMath/pull/647#issuecomment-286529751 presumably applies here, even though it doesn't apply there.

DanGrayson:
I noticed your comment on the google group : 

" the only published Martin-Lof type theory with the eta for products is MLTT79. So UniMath is MLTT79 minus the reflection rule for the equality and plus the resizing rule(s)"

your comment in https://github.com/UniMath/UniMath/pull/647#issuecomment-286441593 is not valid : 

" By doing this change we move the UniMath another step away from the Martin-Lof type theory "

So now I wonder whether the phrase "Martin-Lof type theory" has a definite meaning, with regard to eta reduction.

In any case, perhaps it is time for UniMath-2.0, as you suggest in https://github.com/UniMath/UniMath/pull/647#issuecomment-286441593

vladimirias:
My comment on the google group is about dependent products not dependent sums.

DanGrayson:
Re: "My comment on the google group is about dependent products not dependent sums."

Oh, I misunderstood, thank you for the correction.

mortberg:
So this PR gives us eta for dependent sums and a 56% speedup? That is pretty impressive.

If this is merged I think formalizing mathematics in UniMath will be even more pleasant. The lack of eta for pairs is sometimes annoying me, but the slowness has been really bothering me lately (every time I make some change early on I have to wait for about 1.5 hours on my laptop to compile the entire library, luckily I don't have to do this too often).

vladimirias:
I agree. Should we start UniMath-2.0 then?

DanGrayson:
```
The 10 slowest files:

without records and primitive projections:

    user time 321.70: UniMath/HomologicalAlgebra/TranslationFunctors
    user time 257.96: UniMath/HomologicalAlgebra/MappingCone
    user time 251.80: UniMath/HomologicalAlgebra/KAPreTriangulated
    user time 209.78: UniMath/SubstitutionSystems/LiftingInitial_alt
    user time 190.03: UniMath/HomologicalAlgebra/MappingCylinder
    user time 189.73: UniMath/HomologicalAlgebra/KA
    user time 171.84: UniMath/HomologicalAlgebra/CohomologyComplex
    user time 163.66: UniMath/SubstitutionSystems/MonadsFromSubstitutionSystems
    user time 162.91: UniMath/HomologicalAlgebra/KATriangulated
    user time 138.55: UniMath/SubstitutionSystems/LiftingInitial

with:

    user time 174.42: UniMath/HomologicalAlgebra/TranslationFunctors
    user time 141.92: UniMath/CategoryTheory/limits/graphs/pushouts
    user time 126.28: UniMath/HomologicalAlgebra/MappingCone
    user time 112.85: UniMath/CategoryTheory/limits/graphs/pullbacks
    user time 111.56: UniMath/HomologicalAlgebra/KAPreTriangulated
    user time 101.27: UniMath/HomologicalAlgebra/KA
    user time 80.00: UniMath/HomologicalAlgebra/KATriangulated
    user time 70.32: UniMath/CategoryTheory/LocalizingClass
    user time 64.85: UniMath/HomologicalAlgebra/CohomologyComplex
    user time 49.99: UniMath/CategoryTheory/limits/graphs/equalizers
```

DanGrayson:
Starting UniMath-2.0 sounds like a great idea!

vladimirias:
Thank you for the timing info, but the first and the second list in your timing message are different.

DanGrayson:
Yes, the 10 slowest files from the first run are different from the 10 slowest files of the second run.

mortberg:
Interesting that pushouts and pullbacks got slower. I can definitely spend some time making them faster once this is merged.

DanGrayson:
```
The 10 slowest files before this PR, without records and primitive projections:

    user time 321.70: UniMath/HomologicalAlgebra/TranslationFunctors
    user time 257.96: UniMath/HomologicalAlgebra/MappingCone
    user time 251.80: UniMath/HomologicalAlgebra/KAPreTriangulated
    user time 209.78: UniMath/SubstitutionSystems/LiftingInitial_alt
    user time 190.03: UniMath/HomologicalAlgebra/MappingCylinder
    user time 189.73: UniMath/HomologicalAlgebra/KA
    user time 171.84: UniMath/HomologicalAlgebra/CohomologyComplex
    user time 163.66: UniMath/SubstitutionSystems/MonadsFromSubstitutionSystems
    user time 162.91: UniMath/HomologicalAlgebra/KATriangulated
    user time 138.55: UniMath/SubstitutionSystems/LiftingInitial

The same files, with:

    user time 174.42: UniMath/HomologicalAlgebra/TranslationFunctors
    user time 126.28: UniMath/HomologicalAlgebra/MappingCone
    user time 111.56: UniMath/HomologicalAlgebra/KAPreTriangulated
    user time  21.08: UniMath/SubstitutionSystems/LiftingInitial_alt
    user time  17.48: UniMath/HomologicalAlgebra/MappingCylinder
    user time 101.27: UniMath/HomologicalAlgebra/KA
    user time  64.85: UniMath/HomologicalAlgebra/CohomologyComplex
    user time  12.07: UniMath/SubstitutionSystems/MonadsFromSubstitutionSystems
    user time  80.00: UniMath/HomologicalAlgebra/KATriangulated
    user time  16.23: UniMath/SubstitutionSystems/LiftingInitial
```

mortberg:
Wow, incredible that LiftingInitial and LiftingInitial_alt got that much faster! I spent half a day trying to speed them up by hand, leading only to a few seconds speedup in total...

DanGrayson:
Now I wonder whether Coq was the same Coq in the two runs.  Perhaps I should repeat the timing trials, after adding something to the makefile to display the current sub/coq commit.

vladimirias:
Yes, this looks good.

mortberg:
About UniMath 2.0: This system will be UniMath 1.0 with eta for pairs? (With total2 defined as a Record in the implementation)  So it is MLTT79 without equality reflection and W-types, but with eta for pairs, the univalence axiom and resizing rules?

DanGrayson:
Re: " I have to wait for about 1.5 hours on my laptop to compile the entire library"

Anders, I wonder whether something like "make -j2" or "make -j4" would speed it up for you.  (if you have  multiple cores).

DanGrayson:
Okay, I just ran time trials again on my branch and on the master branch, with the latest Coq v8.6 on both, and got an improvement from 58m20.226s to 26m16.951s for user time, and from 35m9.172s to 18m13.682s for real time (elapsed time).

DanGrayson:
@benediktahrens , if this is okay with you, it seems we can merge.  It's okay with Vladimir, provided we maintain the code just before the merge on its own branch, UniMath-1.0, so there will be something concrete still in existence for demonstration purposes for papers that don't include eta for pairs.  I'll take care of setting that up afterward and make an issue for it now.

I think I'll set it up so the two branches have the exact same Makefile, by putting the list of packages in a separate file.  Both branches will track the latest Coq releases.

benediktahrens:
Why does this PR change the Coq version, and what version does it change to?

DanGrayson:
I thought it might as well change the coq version to the latest v8.6.

benediktahrens:
On 03/15/2017 07:21 PM, Daniel R. Grayson wrote:
> I thought it might as well change the coq version to the latest v8.6.

I thought we had agreed on going with the latest released version?

benediktahrens:
On 03/15/2017 07:04 PM, Daniel R. Grayson wrote:
> Both branches will track the latest Coq releases.

https://github.com/coq/coq/releases ?

benediktahrens:
lgtm except for change of coq

DanGrayson:
Yes.  Is that one not released?

benediktahrens:
On 03/15/2017 08:54 PM, Daniel R. Grayson wrote:
> Yes.  Is that one not released?

I would call 'released' what appears on

[1] https://github.com/coq/coq/releases

and on

[2] https://coq.inria.fr/download

The latest release on [1] is the tag V8.6, which points to 0d14388 ,
which is also the commit currently referred to in our submodule setting.

Maybe we can wait for V8.6pl1 ?


The reason for sticking with released (in the sense above) versions was
to allow others to compile UniMath with a precompiled Coq.
This might fail if we don't test against the latest released Coq.

DanGrayson:
Okay, I've fixed it so sub/coq is unchanged.

benediktahrens:
On 03/15/2017 09:51 PM, Daniel R. Grayson wrote:
> I've made a branch called "UniMath-1.0", see https://github.com/UniMath/UniMath/tree/UniMath-1.0, based on the current "master" -- anticipating this PR will be the next one pulled.
> 
> If we want a branch called "UniMath-2.0", should we rename "master" to that?
> 

I think we should keep the name "master", which should be the branch
against which PRs are filed.

A branch "UniMath-2.0" should be made once we want a version that
diverges from master.

DanGrayson:
Shall we activate travis on the new branch?  Shall we have pull requests, or should I just
maintain it silently?

DanGrayson:
Okay, I'll just take care of it silently.

DanGrayson:
Re: "Is there an option to (de)activate Travis on a per-branch basis? I was
not aware of that."

Well, I know almost nothing about it.  Maybe any pull request to any branch would activate it.

benediktahrens:
On 03/15/2017 10:17 PM, Daniel R. Grayson wrote:
> Re: "Is there an option to (de)activate Travis on a per-branch basis? I was
> not aware of that."
> 
> Well, I know almost nothing about it.  Maybe any pull request to any branch would activate it.  


One can black- and white-list branches:

https://docs.travis-ci.com/user/customizing-the-build/
"Building specific branches"

One can also have different travis.yml files on different branches.

PR #654, opened 2017-03-15
fix the Makefile

DanGrayson:
... so it works when BUILD_COQ is "no".  The problem was reported by Anders.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,13 +27,19 @@ BUILD_COQ ?= yes
 BUILD_COQIDE ?= no
 COQBIN ?=
 ############################################
+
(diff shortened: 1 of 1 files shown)


mortberg:
Thanks for fixing this! 

lgtm

mortberg:
@matej-kosik tested this branch for the benchmarking script he's working on for us and it worked perfectly. The plan is to get UniMath into the automatic benchmarking tests of Coq so that we know as early as possible of changes in performance when Coq gets updated. So the sooner this PR gets merged the sooner UniMath will get in there.

PR #660, opened 2017-03-16
Alternative definition of binary coproduct of functors and new proofs that this is omega cocont

mortberg:
In this PR I define a new version the binary coproduct of two functors (called `BinCoproduct_of_functors_alt`). This is defined in the same way as Peter defined `coproduct_of_functors_alt` (i.e. using `tuple_functor`). This gives a more direct proof that this functor is omega cocontinuous if the input functors are which allows me to remove an assumption of the category having binary products in Theorem 48 of our recent paper (see `InitHSS` in FromBindingSigsToMonads_summary.v). 

Note that I had to keep the old version of `BinCoproduct_of_functors_alt`, it is now called `BinCoproduct_of_functors_alt2` and it is still used in CategoryTheory/Inductives/Lists.v. The reason is that the proof that this functor is omega cocontinuous has better computational behavior in the sense that we can compute the length of lists using `lazy`, if I instead use the new `BinCoproduct_of_functors_alt` this doesn't work anymore.

### DIFF
--- a/UniMath/CategoryTheory/limits/bincoproducts.v
+++ b/UniMath/CategoryTheory/limits/bincoproducts.v
@@ -682,21 +682,22 @@ abstract (split;
   | now intros x y z f g; simpl; rewrite BinCoproductOfArrows_comp ]).
 Defined.
 
+(* Defines the copropuct of two functors *)
+Definition BinCoproduct_of_functors_alt {C D : precategory}
+  (hsD : has_homsets D) (HD : BinCoproducts D) (F G : C ⟶ D) : C ⟶ D :=
+    tuple_functor (λ b, bool_rect (λ _, C ⟶ D) F G b) ∙
+    coproduct_functor bool (CoproductsBool hsD HD).
+
 (* Defines the copropuct of two functors by:
     x -> (x,x) -> (F x,G x) -> F x + G x
 
   For a direct and equal definition see FunctorsPointwiseBinCoproduct.v
 
-  Below follows a slightly simpler definition
+  Above is a slightly simpler definition
 *)
-Definition BinCoproduct_of_functors_alt_old {C D : precategory}
+Definition BinCoproduct_of_functors_alt2 {C D : precategory}
   (HD : BinCoproducts D) (F G : functor C D) : functor C D :=
     bindelta_functor C ∙ (pair_functor F G ∙ bincoproduct_functor HD).
-
-Definition BinCoproduct_of_functors_alt {C D : precategory}
-  (hsD : has_homsets D) (HD : BinCoproducts D) (F G : C ⟶ D) : C ⟶ D :=
-    tuple_functor (λ b, bool_rect (λ _, C ⟶ D) F G b) ∙
-    coproduct_functor bool (CoproductsBool hsD HD).
 
 End functors.
 
@@ -792,8 +793,8 @@ Qed.
 Definition BinCoproduct_of_functors : functor C D :=
   tpair _ _ is_functor_BinCoproduct_of_functors_data.
 
-Lemma BinCoproduct_of_functors_alt_old_eq_BinCoproduct_of_functors :
-  BinCoproduct_of_functors_alt_old HD F G = BinCoproduct_of_functors.
+Lemma BinCoproduct_of_functors_alt2_eq_BinCoproduct_of_functors :
+  BinCoproduct_of_functors_alt2 HD F G = BinCoproduct_of_functors.
 Proof.
 now apply (functor_eq _ _ hsD).
 Defined.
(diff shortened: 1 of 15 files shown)


DanGrayson:
lgtm

UniMath/CategoryTheory/limits/coproducts.v
```diff
@@ -204,13 +205,20 @@ Defined.
 
 End functors.
 
-(* Defines the arbitrary copropuct of a family of functors *)
-Definition coproduct_of_functors_alt (I : UU) {C D : precategory}
+(* The arbitrary copropuct of a family of functors *)
```
benediktahrens: Could you remove the word 'arbitrary' here?

UniMath/CategoryTheory/limits/coproducts.v
```diff
@@ -204,13 +205,20 @@ Defined.
...
+(* This is the old and not so good definition *)
+Definition coproduct_of_functors_alt_old (I : UU) {C D : precategory}
   (HD : Coproducts I D) (F : I -> functor C D) : functor C D :=
   functor_composite (delta_functor I C)
      (functor_composite (family_functor _ F)
                         (coproduct_functor _ HD)).
 
+(* The arbitrary copropuct of a family of functors *)
```
benediktahrens: remove 'arbitrary'.

UniMath/CategoryTheory/limits/coproducts.v
```diff
@@ -204,13 +205,20 @@ Defined.
 
 End functors.
 
-(* Defines the arbitrary copropuct of a family of functors *)
-Definition coproduct_of_functors_alt (I : UU) {C D : precategory}
+(* The arbitrary copropuct of a family of functors *)
+(* This is the old and not so good definition *)
```
benediktahrens: Explain why this def is not so good - we will forget otherwise.

UniMath/CategoryTheory/limits/bincoproducts.v
```diff
@@ -657,16 +682,22 @@ abstract (split;
   | now intros x y z f g; simpl; rewrite BinCoproductOfArrows_comp ]).
 Defined.
 
+(* Defines the copropuct of two functors *)
+Definition BinCoproduct_of_functors_alt {C D : precategory}
```
benediktahrens: I am wondering why using this definition prevents the lazy computations for lists.
Do you think that if instead of instantiating the general functor to the binary case, we would have a separate implementation for the binary coproduct, we could recover computation?

mortberg: Good point, I was thinking the same, but it's a little annoying to implement this as I have to duplicate what Peter did in the binary case. In order to instantiate his work I had to define the following Lemma:

https://github.com/UniMath/UniMath/pull/660/files#diff-e31a44d4c269829c172653b10b5907b8R641

Maybe this indirect definition makes things stuck... 

Do you think it is worth it to duplicate his work just to try this? Note that there is no problem to use the different functors in different files and I did prove that all of the 3 different binary coproduct functors are equal (and these proofs are trivial). If one wants things to compute (as in Lists) one uses the _alt2 version and if one wants to have theorems with few hypotheses one uses the _alt version.

UniMath/CategoryTheory/limits/coproducts.v
mortberg: fixed

UniMath/CategoryTheory/limits/coproducts.v
mortberg: fixed

UniMath/CategoryTheory/limits/coproducts.v
mortberg: fixed

benediktahrens:
On 03/17/2017 02:39 PM, Anders Mörtberg wrote:
> 
> 
> 
>> @@ -657,16 +682,22 @@ abstract (split;
>    | now intros x y z f g; simpl; rewrite BinCoproductOfArrows_comp ]).
>  Defined.
>  
> +(* Defines the copropuct of two functors *)
> +Definition BinCoproduct_of_functors_alt {C D : precategory}
> 
> Good point, I was thinking the same, but it's a little annoying to implement this as I have to duplicate what Peter did in the binary case. In order to instantiate his work I had to define the following Lemma:
> 
> https://github.com/UniMath/UniMath/pull/660/files#diff-e31a44d4c269829c172653b10b5907b8R641
> 
> Maybe this indirect definition makes things stuck... 
> 
> Do you think it is worth it to duplicate his work just to try this? Note that there is no problem to use the different functors in different files and I did prove that all of the 3 different binary coproduct functors are equal (and these proofs are trivial). If one wants things to compute (as in Lists) one uses the _alt2 version and if one wants to have theorems with few hypotheses one uses the _alt version. 

Unfortunately I do think it is worth it - maybe not now, but some time
in the near future.

Could you file an issue on that problem?

mortberg:
Yeah, after some thinking I've concluded that it is probably worth it. There will be some more duplication (I'll have to define a binary version of the tuple functor and prove that it is omega-cocont), but the more direct definition will probably have better computational behavior. If it doesn't have better computational behavior that will also be interesting as it suggests that we were very lucky that our first definition computes using `lazy`. I'll make an issue about this and do it in a future PR.

PR #665, opened 2017-03-25
Well ordering

DanGrayson:
We prove Zermelo's theorem:
```
Theorem ZermeloWellOrdering {X:hSet} : AxiomOfChoice ⇒ ∃ R : hrel X, isWellOrder R.
```
The proof is Zermelo's, from 1904.  There are various expositions of it.  See, for example,
http://www.math.illinois.edu/~dan/ShortProofs/WellOrdering.pdf .


UniMath/Combinatorics/WellOrderedSets.v
```diff
@@ -0,0 +1,1260 @@
...
+  - intros x m. apply (squash_to_hProp m); intros [B [Bx [_ [BT _]]]]; clear m. exact (BT _ Bx).
+Defined.
+
+Local Lemma h1' {X} {S:TOSubset X} {s t u:S} : s ≤ t -> pr1 t = pr1 u -> s ≤ u.
+Proof.
+  intros le p.
+  assert (q : t = u).
+  { now apply subtypeEquality_prop. }
+  induction q. exact le.
+Defined.
```
tpannila:
```
Local Lemma h1' {X : hSet} {S:TOSubset X} {s t u : S} : s ≤ t -> pr1 t = pr1 u -> s ≤ u.
Proof.
  intros le p. induction (subtypeEquality_prop p). exact le.
Defined.
```

UniMath/Combinatorics/WellOrderedSets.v
```diff
@@ -0,0 +1,1260 @@
...
+
+Local Lemma h1'' {X} {S:TOSubset X} {r s t u:S} : (r ≤ t -> pr1 r = pr1 s -> pr1 t = pr1 u -> s ≤ u)%tosubset.
+Proof.
+  intros le p q.
+  assert (p' : r = s).
+  { now apply subtypeEquality_prop. }
+  assert (q' : t = u).
+  { now apply subtypeEquality_prop. }
+  induction p', q'. exact le.
+Defined.
```
tpannila: Small simplification
```
Local Lemma h1'' {X : hSet} {S : TOSubset X} {r s t u : S} :
  (r ≤ t -> pr1 r = pr1 s -> pr1 t = pr1 u -> s ≤ u)%tosubset.
Proof.
  intros le p q. induction (subtypeEquality_prop p). induction (subtypeEquality_prop q). exact le.
Defined.
```

vladimirias:
This can be written as a := definition and it will be much easier for the reader to comprehend.

vladimirias:
> On Mar 25, 2017, at 11:38 PM, Tomi Pannila <notifications@github.com> wrote:
> 
> In UniMath/Combinatorics/OrderedSets.v <https://github.com/UniMath/UniMath/pull/665#discussion_r108048427>:
> 
> > +Definition isTotalOrder {X : hSet} (R : hrel X) : hProp.
> +Proof.
> +  intros. exists (isPartialOrder R × istotal R). apply isapropdirprod.
> +  - apply isaprop_isPartialOrder.
> +  - apply isaprop_istotal.
> +Defined.
> +
> +Section A.
> +
> +  Open Scope logic.
> +
> +  Lemma tot_nge_to_le {X:hSet} (R:hrel X) : istotal R -> ∏ x y, ¬ (R x y) ->  R y x.
> +  Proof.
> +    intros ? ? tot ? ? nle.
> +    now apply (hdisjtoimpl (tot x y)).
> +  Defined.
> If you want to avoid now, you can write this as
> 
> Lemma tot_nge_to_le {X : hSet} (R : hrel X) : istotal R -> ∏ (x y : X), ¬ (R x y) -> R y x.
> Proof.
>   intros X R tot x y nle. exact (hdisjtoimpl (tot x y) nle).
> Defined.

Yes, this is better. One can also write it as a := definition.

vladimirias:
I agree with Tomi.

vladimirias:
Is it really necessary? It is one of the few cases where I would say that the types are obvious.

vladimirias:
@DanGrayson do you need the full univalence axiom for your proof? When you make introductory comments could you specify which of the axiomatic levels (funext, univalence for propositions (UAP), funext+UAP, UA) is used where?

In general, organizing results by the axiom level as it is done in Basics may be a good way to organize things.

DanGrayson:
Re: "You use Notation a lot. I don't like Notation because I find it difficult to
work with them. How can I unfold Notation? How can I easily find, without git grep,
where Notation is defined?"

Don't use grep.  Instead, use M-x tags-search, which is bound to M-s M-s .  The advantage
of tags search is that the files are search in logical order.  To find all the notations for <, for example, do `M-s M-s Notation.*<`

I like notations, and so do mathematicians, who use a lot of them, to improve readability.  The ones I define in Subtypes.v are completely standard:
```
Require Export UniMath.MoreFoundations.Notations.
Notation " s ∈ T " := (subtype_isIn s T) (at level 70) : subtype.
Notation " s ∉ T " := (¬ (subtype_isIn s T) : hProp) (at level 70) : subtype.
Notation " S ⊆ T " := (subtype_containedIn S T) (at level 70) : subtype.
Notation " S ⊈ T " := (subtype_notContainedIn S T) (at level 70) : subtype.
Notation " S ⊊ T " := (subtype_smallerThan S T) (at level 70) : subtype.
Notation " S ≡ T " := (subtype_equal S T) (at level 70) : subtype.
Notation " S ≢ T " := (subtype_notEqual S T) (at level 70) : subtype.
Notation " S - T " := (subtype_difference S T) : subtype.
Notation "⋃ S" := (subtype_union S) (at level 100, no associativity) : subtype.
```

Coq has a command for finding now notations are defined.  For example, `Locate "≤" . ` yields this:
```
Notation            Scope     
" x ≤ y " := natleh_DecidableProposition x y
                      : decidable_nat
                      
" x ≤ y " := natleh x y        : nat_scope
                      (default interpretation)
"m ≤ n" := posetRelation _ m n : poset
```

ProofGeneral provides that command on a key stroke with C-c C-a C-n.

To see the ProofGeneral commands and their keystrokes, do C-h m

DanGrayson:
Re: "Sometimes you use {X:UU} and not {X : UU}. Can you fix these?"

Thanks, but I don't see much point in striving for that kind of uniformity.  Sometimes I like
one and sometimes I like the other.

DanGrayson:
Re: "Can you split the longer proofs into smaller results?"

Yes, but I think I can first shorten the proof by 30%, so I'd like to do that in another pull request.  Meanwhile, someone might want to use it.

DanGrayson:
Re: "I don't believe that you really need Defined in many places, perhaps in a few, and thus I would like to see more Opaque and Qed."

Amusingly, I have the opposite attitude.  I don't believe you really need Qed in many places, and thus I would like to see more Defined.  I think there are probably better ways to get Coq to avoid needless unfolding.

UniMath/CategoryTheory/Inductives/Lists.v
```diff
@@ -301,6 +301,9 @@ apply (foldr A (list (pr1 A),,isaset_list A)).
   apply (tpair _ (S (pr1 L)) (a,,pr2 L)).
 Defined.
 
+Set Printing All.
+Print to_list.
+
```
DanGrayson: oops, thanks.

UniMath/Combinatorics/OrderedSets.v
```diff
@@ -8,6 +8,62 @@ Local Open Scope poset.
 
 (** partially ordered sets and ordered sets *)
 
+Definition isTotalOrder {X : hSet} (R : hrel X) : hProp.
+Proof.
+  intros. exists (isPartialOrder R × istotal R). apply isapropdirprod.
+  - apply isaprop_isPartialOrder.
+  - apply isaprop_istotal.
+Defined.
```
DanGrayson: That's 6 lines instead of 3, and I see no particular reason to prefer "use" to "apply" if "apply" works, nor to use hProppair when it isn't needed.

UniMath/Combinatorics/OrderedSets.v
```diff
@@ -8,6 +8,62 @@ Local Open Scope poset.
...
+
+Section A.
+
+  Open Scope logic.
+
+  Lemma tot_nge_to_le {X:hSet} (R:hrel X) : istotal R -> ∏ x y, ¬ (R x y) ->  R y x.
+  Proof.
+    intros ? ? tot ? ? nle.
+    now apply (hdisjtoimpl (tot x y)).
+  Defined.
```
DanGrayson: That's true, but I like "now".

UniMath/Combinatorics/OrderedSets.v
```diff
@@ -8,6 +8,62 @@ Local Open Scope poset.
...
+    intros X R i.
+    assert (tot := pr2 i); simpl in tot.
+    assert (refl := pr2 (pr1 (pr1 i))); simpl in refl.
+    assert (anti := pr2 (pr1 i)); simpl in anti.
+    split.
+    { intros nle. split.
+      - now apply tot_nge_to_le.
+      - intros ne. induction ne. exact (nle (refl y)). }
+    { intros yltx xley. induction yltx as [ylex neq]. exact (neq (anti _ _ ylex xley)). }
+  Defined.
```
DanGrayson: Yes, it might be good to get rid of that `pr2 (pr1 (pr1`

UniMath/Combinatorics/OrderedSets.v
```diff
@@ -8,6 +8,62 @@ Local Open Scope poset.
...
+
+Definition isBiggest {X : Poset} (x : X) : UU := ∏ y, y ≤ x.
+
+Definition isMinimal {X : Poset} (x : X) : UU := ∏ y, y ≤ x -> x = y.
+(* the definition in Sets.v is wrong *)
+
+Definition isMaximal {X : Poset} (x : X) : UU := ∏ y, x ≤ y -> x = y.
+(* the definition in Sets.v is wrong *)
+
+Definition consecutive {X : Poset} (x y : X) : UU := x < y × ∏ z, ¬ (x < z × z < y).
```
DanGrayson: It seems cleaner and clear enough without adding bulk to the expression.

UniMath/Combinatorics/OrderedSets.v
```diff
@@ -16,65 +72,55 @@ Local Arguments isPosetEquivalence : clear implicits.
...
-  set (f' := weqcomp g h).
-  assert (k : pr1weq f' ~ f).
-  try reflexivity.              (* this doesn't work *)
+  assert (k : pr1weq (Poset_univalence_0 X Y) ~ @Poset_univalence_map X Y).
   { intro e. apply isinj_pr1_PosetEquivalence. induction e. reflexivity. }
-  assert (l : isweq f).
-  { apply (isweqhomot f'). exact k. apply weqproperty. }
-  exact (f,,l).
+  exact (remakeweq k).
 Defined.
```
DanGrayson: The `Abort` is useful for showing the reader that a certain proof doesn't work, and `remakeweq` is a good thing, not a bad thing.

UniMath/Combinatorics/OrderedSets.v
```diff
@@ -216,6 +262,32 @@ Proof.
...
+  intros ? ? ? i j. assert (q := OrderedSet_istotal x y). apply (squash_to_prop q).
+  { apply setproperty. }
+  intro c. induction c as [xley|ylex].
+  - apply OrderedSet_isantisymm.
+    + assumption.
+    + now apply j.
+  - apply OrderedSet_isantisymm.
+    + now apply i.
+    + assumption.
+Defined.
```
DanGrayson: I like both of those tactics.

UniMath/Combinatorics/WellOrderedSets.v
```diff
@@ -0,0 +1,1260 @@
...
+Proof.
+  apply (pr2 S).
+Defined.
+
+Definition TOeq_to_refl {X:hSet} (S : TOSubset X) : ∀ s t : carrier_set S, s = t ⇒ s ≤ t.
+Proof.
+  intros s t e. induction e. apply TOrefl.
+Defined.
+
+Definition TOeq_to_refl_1 {X:hSet} (S : TOSubset X) : ∀ s t : carrier_set S, pr1 s = pr1 t ⇒ s ≤ t.
```
DanGrayson: Thanks, I'll include it!

UniMath/Combinatorics/WellOrderedSets.v
DanGrayson: Thanks, I'll include it

UniMath/Combinatorics/WellOrderedSets.v
```diff
@@ -0,0 +1,1260 @@
...
+Notation "S ≼ T" := (tosub_le _ S T) (at level 70) : tosubset.
+
+Definition sub_initial {X:hSet} (S : hsubtype X) (T : TOSubset X) (le : S ⊆ T) : hProp
+  := ∀ (s t : X) (Ss : S s) (Tt : T t), TOrel T (t,,Tt) (s,,le s Ss) ⇒ S t.
+
+Definition same_induced_ordering {X:hSet} {S T : TOSubset X} {B : hsubtype X} (BS : B ⊆ S) (BT : B ⊆ T)
+  := ∀ x y : B,
+             subtype_inc BS x ≤ subtype_inc BS y
+                         ⇔
+             subtype_inc BT x ≤ subtype_inc BT y.
```
DanGrayson: I think it's clearer if parallel constructions are displayed in parallel lines for ease of visual comparison.

UniMath/Combinatorics/WellOrderedSets.v
DanGrayson: Thanks, but I think I'll eliminate this lemma, anyway.

UniMath/Combinatorics/WellOrderedSets.v
```diff
@@ -0,0 +1,1260 @@
...
+  { intros le. induction x as [x xm], y as [y ym].
+    apply (squash_to_hProp xm); intros [B [Bx [BS [BT [BSi [BTi BST]]]]]].
+    apply (squash_to_hProp ym); intros [C [Cy [CS [CT [CSi [CTi CST]]]]]].
+    assert (Cx : C x).
+    { apply (CTi y x Cy (BT x Bx)). now apply (h1'' le). }
+    assert (Q := pr2 (CST (x,,Cx) (y,,Cy))); simpl in Q.
+    assert (E : subtype_inc CT (x,, Cx) ≤ subtype_inc CT (y,, Cy)).
+    { now apply (h1'' le). }
+    clear le. now apply (h1'' (Q E)). }
+Defined.
```
DanGrayson: That certainly can be cleaned, as part of my later cleaning.  In particular, I hope it's not necessary to have so much duplication of code.  Putting them into separate lemmas, when so much code is duplicated, will just rigidify the situation.

DanGrayson:
Here's what Coq 8.5 says the theorem depends on:
```
univalenceAxiom : univalenceStatement
propositionalUnivalenceAxiom : propositionalUnivalenceStatement
isweqtoforallpathsAxiom : isweqtoforallpathsStatement
funextemptyAxiom : funextemptyStatement
funcontrAxiom : funcontrStatement
```

I believe the use of full univalence could be avoided if desired -- at the
very end I wanted to transport the well ordering
I found along an equivalence to get a well ordering on the set X, and it
was quicker.  I used it in this incarnation:

```
Theorem hSet_univalence (X Y : hSet) : (X = Y) ≃ (X ≃ Y).
```

(At that final stage, I had a well ordering on the carrier W of a subset of
X and I had a proof that every element of
X was in W, so X was equivalent to W.)

UniMath/Combinatorics/WellOrderedSets.v
```diff
@@ -0,0 +1,1260 @@
...
+    { intros [Ss|ezs] [St|ezt] [St'|ezt']. repeat split.
+      - now induction_hProp St St'.
+      - apply fromempty. induction ezt'. exact (nSz St).
+      - apply fromempty. induction ezt. exact (nSz St').
+      - reflexivity.
+      - reflexivity.
+      - apply fromempty. induction ezt'. exact (nSz St).
+      - apply fromempty. induction ezt. exact (nSz St').
+      - reflexivity. } }
+Defined.
```
DanGrayson: This one is actually pretty clean as it stands.  There are unavoidably many cases to consider, but each case has a succinct proof.  

What's happening here is that I add a point to a subset and extend the relation by regarding the point as larger.  But the relation has to be well defined, so there are lots of equations to establish.

One interesting thing about this proof is that it uses the statement that hProp is a set.

UniMath/Combinatorics/WellOrderedSets.v
```diff
@@ -0,0 +1,1260 @@
...
+        change (htrue ∨ hfalse).
+        exact (hinhpr (ii1 tt)).
+    - induction ezx. induction (ishinh_irrel (ii2 (idpath z)) Wx).
+      apply (squash_to_hProp Wy); intros [Sy|ezy].
+      + induction (ishinh_irrel (ii1 Sy) Wy); change (hfalse ∨ htrue).
+        exact (hinhpr (ii2 tt)).
+      + induction ezy.
+        induction (ishinh_irrel (ii2 (idpath z)) Wy); change (htrue ∨ htrue).
+        exact (hinhpr (ii2 tt)). }
+Defined.
```
DanGrayson: I think that would just add clutter, in this case.  The four comments make it clear what the four paragraphs in the proof are about.

UniMath/Combinatorics/WellOrderedSets.v
```diff
@@ -0,0 +1,1260 @@
...
+        exists (t,,St).
+        change (T (t,, j t St)).
+        induction (proofirrelevance_hProp _ S't (j t St)).
+        exact Tt.
+      * induction ezt.
+        (* now show [z ≤ z], by reflexivity *)
+        change (TOrel S' (z,,S'z) (z,,S't)).
+        induction (proofirrelevance_hProp _ S'z S't).
+        exact (TOrefl S' _).
+Defined.
```
DanGrayson: No, this result is already atomic.  It's proving a single property of the construction.

UniMath/Combinatorics/WellOrderedSets.v
```diff
@@ -0,0 +1,1260 @@
...
+      + apply weqimplimpl.
+        { intros k. split ; apply k. }
+        { intros c. split.
+          { intros x. exact (pr11 c x,,pr12 c x). }
+          { exact c. } }
+        { apply propproperty. }
+        { apply propproperty. } }
+  { apply wosub_univalence_map. }
+  { intros e. induction e. reflexivity. }
+Defined.
```
DanGrayson: This result seems atomic to me.

UniMath/Combinatorics/WellOrderedSets.v
```diff
@@ -0,0 +1,1260 @@
...
+    change ((s ≤ s') ⨿ (s' ≤ s) → s ≤ s').
+    intro c. induction c as [c|c].
+    - exact c.
+    - induction le as [le [com ini]].
+      assert (k := com s' s c).
+      assert (k' := Tanti _ _ l k); clear k.
+      assert (p : s = s').
+      { apply subtypeEquality_prop. exact (maponpaths pr1 k'). }
+      induction p. apply (pr2 S). (* refl *) }
+Defined.
```
DanGrayson: good idea, will do

UniMath/Combinatorics/WellOrderedSets.v
```diff
@@ -0,0 +1,1260 @@
...
+    { now apply hsubtype_univalence. }
+    unfold W in *. clear W.
+    induction (!e); clear e.
+    use tpair.
+    { exact WC. }
+    split.
+    { intros x y le. apply (pr2 (WCD x y)). now apply (h1'' le). }
+    exact WCi.
+  Unset Printing Coercions.
+Defined.
```
DanGrayson: This proof can be shortened, instead, which will be better than just splitting it up.  Look, for example, at the four assertions labeled cmax, cmax', dmax, and dmax' -- they share a lot of code.

Another idea I have for shortening all of the proofs in this file, including this one, is this: Often I'm dealing with elements of the carriers of subsets S of X, and the orderings are relations R on those carriers.  But numerous times those elements evidently have the same first component, so, since the second components are elements of a proposition, the pairs are equal.  It's annoying to have to deal with that explicitly.  The idea is to define a relation on elements of X by `R' x y := ∃ (p : S x) (q : S y) R (x,,p) (y,,q)`.  That will satisfy all of the relevant properties except for reflexivity.  Adding one more (semi-reflexivity) property ensures that it captures everything about R, namely : `R' x y -> R' x x ∧ R' y y`.

I prefer to put such effort into a second pull request.

UniMath/Combinatorics/WellOrderedSets.v
```diff
@@ -0,0 +1,1260 @@
...
+  change (WellOrdering X).
+  assert (e : X = carrier_set W).
+  { apply (invmap (hSet_univalence _ _)). apply invweq. apply weqpr1.
+    intros x.
+    simpl in all.
+    apply iscontraprop1.
+    - apply propproperty.
+    - apply all. }
+  induction e. exact R'.
+Defined.
```
DanGrayson: I'll try (later on).

UniMath/Foundations/Preamble.v
```diff
@@ -43,6 +43,7 @@ Hint Resolve identity_refl : core . ]
 
 Inductive paths {A:Type} (a:A) : A -> Type := paths_refl : paths a a.
 Hint Resolve paths_refl : core .
+Delimit Scope type_scope with type.
```
DanGrayson: Oops, I intended that just to be in MoreFoundations.

UniMath/MoreFoundations/Notations.v
```diff
@@ -15,3 +15,26 @@ Local Open Scope logic.
...
+Notation "'pr22' x" := (pr2 (pr2 x)) (at level 8).
+
+Notation "'pr111' x" := (pr1 (pr1 (pr1 x))) (at level 8).
+Notation "'pr112' x" := (pr1 (pr1 (pr2 x))) (at level 8).
+Notation "'pr121' x" := (pr1 (pr2 (pr1 x))) (at level 8).
+Notation "'pr122' x" := (pr1 (pr2 (pr2 x))) (at level 8).
+Notation "'pr211' x" := (pr2 (pr1 (pr1 x))) (at level 8).
+Notation "'pr212' x" := (pr2 (pr1 (pr2 x))) (at level 8).
+Notation "'pr221' x" := (pr2 (pr2 (pr1 x))) (at level 8).
+Notation "'pr222' x" := (pr2 (pr2 (pr2 x))) (at level 8).
```
DanGrayson: Sure, they're useful when defining accessor functions.  E.g.:
```

Definition TOtotal {X:hSet} (S : TOSubset X) : isTotalOrder (TOSrel S) := pr22 S.

Definition TOtot {X:hSet} (S : TOSubset X) : istotal (TOSrel S) := pr222 S.

Definition TOanti {X:hSet} (S : TOSubset X) : isantisymm (TOSrel S) := pr2 (pr122 S).

Definition TOrefl {X:hSet} (S : TOSubset X) : isrefl (TOSrel S) := pr211 (pr22 S).
```

UniMath/MoreFoundations/Subtypes.v
```diff
@@ -0,0 +1,162 @@
...
+        ** apply isapropneg.         (* uses funextemptyAxiom *)
+        ** intros [x [Sx nTx]] e. use nTx; clear nTx. exact (pr1 (e x) Sx).
+      * apply (squash_to_prop c).
+        ** apply isapropneg.         (* uses funextemptyAxiom *)
+        ** intros [x [Tx nSx]] e. use nSx; clear nSx. exact (pr2 (e x) Tx).
+Defined.
+
+Lemma subtype_notEqual_from_negEqual {X:UU} (S T : hsubtype X) : LEM -> (S ≢ T <- ¬ (S ≡ T)).
+Proof.
+  - intros lem ne. unfold subtype_equal in ne.
```
DanGrayson: will do

DanGrayson:
Re: "There are a lot of {X} where the type information is missing. Can you add the type information?"

Actually, there were just 4.  I'll remove them, but I don't regard this as important, since Coq deduces gazillions of more complicated things behind the scenes for us.  Just think of all the coercions and implicit arguments that are not displayed, even in the statements of theorems.

DanGrayson:
Re: "Please add some comments in the files if possible. Table of contents, informal
definitions, ideas of proofs, etc. This makes it easier to study/review your formalization."

My plan is to rewrite my tex file once the proof is completely simplified -- that will be the best way to study the proof.  See http://www.math.illinois.edu/~dan/ShortProofs/WellOrdering.pdf

Also, there is no point in adding comments until after the proof has been polished and simplified, since the changes might invalidate the comments.

UniMath/Combinatorics/OrderedSets.v
```diff
@@ -8,6 +8,62 @@ Local Open Scope poset.
...
 (** partially ordered sets and ordered sets *)
 
+Definition isTotalOrder {X : hSet} (R : hrel X) : hProp.
+Proof.
+  intros. exists (isPartialOrder R × istotal R). apply isapropdirprod.
+  - apply isaprop_isPartialOrder.
+  - apply isaprop_istotal.
+Defined.
+
+Section A.
```
DanGrayson: The open scope gets closed automatically at the end of the section, which is one justification for using sections.

UniMath/Foundations/Preamble.v
DanGrayson: And it doesn't even need to be there, removing...

DanGrayson:
Yes, doing:

Definition isTotalOrder {X : hSet} (R : hrel X) : hProp
  := hProppair (isPartialOrder R × istotal R)
               (isapropdirprod _ _ (isaprop_isPartialOrder R)
(isaprop_istotal R)).

DanGrayson:
The travis failure is probably due to me compiling at home with a slightly different version of Coq.

DanGrayson:
@benediktahrens may wish to offer comments ...

DanGrayson:
PS: I will have a further motivation for simplification of this file, since I want to ensure that the proof of Zorn's lemma doesn't double the size of the file.  The difference between the two setups is this:  For well ordering, one starts with a set X and considers subsets equipped with a well ordering.  For Zorn, one starts with a partially ordered set X and considers subsets well ordered by the induced ordering.  Aside from that the proofs are almost the same!  Indeed, the statement of Zorn's lemma resulting naturally from the proof is this: "A partially ordered set X with upper bounds for its well ordered subsets has a maximal element."  The usual statement considers totally ordered subsets, instead, but the proof needs only the weaker hypothesis.  For well ordering, one can take any proper well ordered subset and adjoin one more element at the top, getting a bigger well ordered subset.  For Zorn's lemma, one can take any subset well ordered by the induced ordering and take any upper bound outside the set, adjoin it at the top, and get a bigger subset well ordered by the induced ordering.

DanGrayson:
Transfinite recursion seems to be a perfectly acceptable substitute for Zorn's lemma, so maybe I'll formalize that first and use it for some applications.

DanGrayson:
I've made many of the suggested changes and hope this will be acceptable now.

( @benediktahrens  )

tpannila:
LGTM

vladimirias:
Strictly speaking the first one is incorrect since someone may add, upstream, X:UU as an identifier.

vladimirias:
> On Mar 26, 2017, at 11:08 AM, Daniel R. Grayson <notifications@github.com> wrote:
> 
> 
> ```
> Theorem hSet_univalence (X Y : hSet) : (X = Y) ≃ (X ≃ Y).
> ```


This must exist in the category theory where the proof of the univalence of HSET is given.

vladimirias:
> On Mar 26, 2017, at 11:08 AM, Daniel R. Grayson <notifications@github.com> wrote:
> 
> (At that final stage, I had a well ordering on the carrier W of a subset of
> X and I had a proof that every element of
> X was in W, so X was equivalent to W.)

Right, this requires UA fort sets. Even showing that the total2 (fun x => unit ) is equal to X in UU requires the UA.

But it would be much more elegant to prove totalorder_fwd and totalorder_bck - the transport forward and backward of the total orders along equivalences and use one of these two here.

vladimirias:
Yes, these are nice. I wonder why we have not thought about them before.

DanGrayson:
Re: "This must exist in the category theory where the proof of the univalence of HSET is given."

Maybe so, but MoreFoundations comes before CategoryTheory in the linear ordering, so it can't be used here.

DanGrayson:
Re: "But it would be much more elegant to prove totalorder_fwd and totalorder_bck - the transport forward and backward of the total orders along equivalences and use one of these two here."

Using univalence seems elegant.  Indeed, transport of structure is part of its appeal.

DanGrayson:
Re: "Yes, these are nice. I wonder why we have not thought about them before."

My background in Lisp, which offers such functions as caadr, caddr, cdadr, etc., gave me the idea.

vladimirias:
Yes, but the corresponding files in CategoryTheory have to be modified. Can we do it?

benediktahrens:
There is the type of bijections between two sets (two maps back and
forth and two families of pointwise identities) and the type of
(categorical) isomorphisms between two sets.

Passing from the former to the latter requires the axiom of function
extensionality.

The equivalence between identities between sets and bijections between
sets should be in (More)Foundations.
The equivalence betweeen bijections and isomorphisms is necessarily in
CategoryTheory.

Benedikt

DanGrayson:
Does anyone know a name for the property of ordered sets that the following theorem
proves well ordered sets have?:
```
Theorem WellOrderedSet_recursion {X:WellOrderedSet} (P : X -> hSet) :
  LEM -> (∏ x:X, (∏ y, y<x -> P y) -> P x) -> (∏ x, P x).
```

(I've just formalized it, and it's a standard fact used in "transfinite induction".)  I think
I can use it to prove Zorn's lemma and various other things now.)

benediktahrens:
On 04/01/2017 01:43 AM, Daniel R. Grayson wrote:
> Does anyone know a name for the property of ordered sets that the following theorem
> proves well ordered sets have?:
> ```
> Theorem WellOrderedSet_recursion {X:WellOrderedSet} (P : X -> hSet) :
>   LEM -> (∏ x:X, (∏ y, y<x -> P y) -> P x) -> (∏ x, P x).
> ```
> 
> (I've just formalized it, and it's a standard fact used in "transfinite induction".)  I think
> I can use it to prove Zorn's lemma and various other things now.)

I think it is "wellfounded induction".
See
https://coq.inria.fr/library/Coq.Init.Wf.html

vladimirias:
Do you need the target of P to hSet? Why not UU?

DanGrayson:
That's a good question, and I've been puzzling about that and am looking
into it.  As the proof
went, at a certain point I needed to prove an equality by contradiction
using LEM, so needed
the equality to be in a set.

On Sat, Apr 1, 2017 at 8:54 AM, Vladimir Voevodsky <notifications@github.com
> wrote:

> Do you need the target of P to hSet? Why not UU?
>
> > On Apr 1, 2017, at 3:27 AM, Benedikt Ahrens <notifications@github.com

DanGrayson:
Here's why the recursion works for defining a function only to sets:

First, one proves easily that a totally ordered set is well ordered iff it satisfies this
well founded *induction* principle for propositions:
```
    ∀ (P : X -> hProp) (rec : ∀ x:X, (∀ y, y < x ⇒ P y) ⇒ P x), (∀ x, P x)
```
One proves it by contradiction, looking for the first place where P x is false.

Now to prove that a well ordered set satisfies this well founded recursion principle:
```
    ∏ (P : X -> hSet) (rec : ∏ x:X, (∏ y, y < x -> P y) -> P x), (∏ x, P x)
```
one introduces P and rec and studies partially defined sections of P that are
"guided" by rec, in the sense that every value is equal to that obtained by applying
rec to the collection of previous values.

Then one wants to show that any two guided partial sections take equal values on points
where both are defined.  This can be done with well founded induction, but only because
the equations are propositions.

Then one gets a unique largest guided partial section, defined on the union of all of them.
But if its domain is not all of X, one takes the smallest element of X not in the domain 
and adds it to the domain, use rec to extend the definition of the section. achieving a
contradiction.

Perhaps one can repeat the argument, incrementing the hlevel each time...

DanGrayson:
Re: "Strictly speaking the first one is incorrect since someone may add, upstream, X:UU as an identifier."

I don't see how that's possible, since ":" has a special meaning in the syntax.

PR #668, opened 2017-03-27
Equivalence for presheaves of the category of elements 

mweav:
A proof that the category of presheaves of the category of elements of a presheaf P is equivalent to the category of presheaves sliced by P (i.e. pshf ∫P ≅ pshf C / P  when P : pshf C).


mweav:
By mistake was changing more files than the actual proof.

Issue #658, opened 2017-03-16
rethink edict against "match" terms


DanGrayson:
The terms produced by the "induction" tactic appeal to eliminators, whereas the
terms produced by "destruct" appeal to "match" or "if".  That's why we have an
edict against using "destruct" (and against using nested brackets with
"induction").

But there is an equivalence between "match" and the various eliminators, which
is bidirectional, so I propose we eliminate the edict, instead explaining the
"match" terms by showing how to translate each use of it into a use of the
corresponding eliminator, as follows.

Whenever you see
```
	match w as t return TYPE with
       	      | tpair _ _ x y => BODY
	      end
```
in a proof term, replace it by
```
	@total2_rect _ _ (λ t, TYPE) (λ x y, BODY) w
```
The same reasoning applies to all the inductive types we use, as follows.

Whenever you see
```
	match p as p' in (_ = b) return TYPE with
                     | idpath _ => BODY
                     end
```
in a context where p is a path in A starting at a, replace it with
```
        @paths_rect A a (λ (b : A) (p' : a = b), TYPE) BODY y p
```

Whenever you see
```
        match e as e' return TYPE with end
```
replace it with
```
	empty_rect (λ e' : empty, TYPE) e
```

Whenever you see
```
	match u as u' return TYPE with | tt => BODY end
```
replace it with
```
	unit_rect (λ u', TYPE) BODY u
```

Whenever you see
```
	if b as b' return TYPE then f else f'
```
replace it with
```
	bool_rect (λ b', TYPE) f f'
```

vladimirias:
In the last rule (for bool) where is “b” in the replacement?

DanGrayson:
Oops, thanks, I fixed it by editing.

benediktahrens:
I think that the statement 'But there is an equivalence between "match" and the various eliminators' is an interesting conjecture, and we should prove that conjecture before permitting `destruct`.

DanGrayson:
I think the proof is the list of replacements above.

DanGrayson:
Here is one slight difference in all of them, looking at bool for convenience.  The type of the eliminator is `∏ P : bool → Type, P true → P false → ∏ b : bool, P b`, which implies that a universe U_n has to be chosen so that P is a function of type bool -> U_n.  But if the match term is used in the form
```
match b as b0 return (P b0) with
| true => f
| false => f0
end
```
[...]

PR #669, opened 2017-03-27
Equivalence for presheaves of the category of elements

mweav:
A proof that the category of presheaves of the category of elements of a presheaf P is equivalent to the category of presheaves sliced by P (i.e. pshf ∫P ≅ pshf C / P when P : pshf C).

### DIFF
--- a/UniMath/CategoryTheory/elems_slice_equiv.v
+++ b/UniMath/CategoryTheory/elems_slice_equiv.v
@@ -285,11 +285,13 @@ Section elems_slice_equiv.
     fun Q =>  slice_to_pshf_ob_funct_data Q ,,  slice_to_pshf_ob_is_funct Q.
 
   Definition slice_to_pshf_ob_nat {X Y : pshf C / P} (F : X --> Y) (e : ∫P^op) :
-    (slice_to_pshf_ob_ob X) e --> (slice_to_pshf_ob_ob Y) e :=
-    match e with
-    | (e ,, Pe) =>
-      fun p => hfibersgftog ((pr1 (pr1 F)) e) ((pr1 (pr2 Y)) e) Pe (transportf (fun x => hfiber (x e) Pe) (base_paths _ _ (pr2 F)) p)
-    end.
+    (slice_to_pshf_ob_ob X) e --> (slice_to_pshf_ob_ob Y) e.
+  Proof.
+    induction e as [e Pe].
+    exact (fun p => hfibersgftog ((pr1 (pr1 F)) e)
+                                 ((pr1 (pr2 Y)) e) Pe
+                                 (transportf (fun x => hfiber (x e) Pe) (base_paths _ _ (pr2 F)) p)).
+  Defined.
 
   Definition slice_to_pshf_ob_is_nat {X Y : pshf C / P} (F : X --> Y) :
     is_nat_trans (slice_to_pshf_ob X : functor _ _) (slice_to_pshf_ob Y : functor _ _) (slice_to_pshf_ob_nat F).
--- a/UniMath/CategoryTheory/elems_slice_equiv.v
+++ b/UniMath/CategoryTheory/elems_slice_equiv.v
@@ -1,0 +1,434 @@
+(** **********************************************************
+Matthew Weaver, 2017
+************************************************************)
+
+
+(** **********************************************************
+Contents : Equivalence of the categories pshf ∫P and
+           pshf C / P for any P in pshf C
+************************************************************)
+
+Require Import UniMath.Foundations.Sets
+        UniMath.CategoryTheory.precategories
+        UniMath.CategoryTheory.functor_categories
+        UniMath.CategoryTheory.equivalences
+        UniMath.CategoryTheory.category_hset
+        UniMath.CategoryTheory.slicecat
+        UniMath.CategoryTheory.opp_precat
+        UniMath.CategoryTheory.yoneda
+        UniMath.CategoryTheory.ElementsOp.
+
+(* Proof that pshf ∫P ≃ pshf C / P *)
+Section elems_slice_equiv.
+
+  Local Open Scope cat.
+
+  Local Definition pshf (C : precategory) : Precategory := [C^op, SET].
+  Local Notation "C / X" := (slice_precat C X (pr2 C)).
+  Local Definition ap_pshf {X : precategory} := fun (P : pshf X) (x : X) => pr1hSet ((pr1 P) x).
+  Local Notation "##" := ap_pshf.
+
+  (* BEGIN: Stuff from Anders *)
+  Lemma transportb_to_transportf (X : UU) (P : X -> UU) (x y : X) (e : x = y) (px : P x) (py : P y) :
+    transportf P e px = py -> px = transportb P e py.
+  Proof.
+    intros H.
+    now rewrite <- H, transportbfinv.
+  Qed.
+
+  Lemma transportf_pshf (C : precategory) (F : pshf C) (x y z : C)
+        (e : x = y) (f : C⟦x,z⟧) (u : ##F z) :
+    transportf (λ x, ##F x) e (# (pr1 F) f u) =
+    # (pr1 F) (transportf (@precategory_morphisms C^op z) e f) u.
+  Proof.
+    now induction e.
+  Qed.
+
+  Lemma pr1_pair_path_in2 (X : UU) (P : X → UU) (x : X) (p q : P x) (e : p = q) :
+    maponpaths pr1 (pair_path_in2 P e) = idpath x.
+  Proof.
+    now induction e.
+  Qed.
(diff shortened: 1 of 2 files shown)


UniMath/CategoryTheory/elems_slice_equiv.v
```diff
@@ -0,0 +1,434 @@
...
+    simpl. unfold hfiber.
+    rewrite transportf_total2; simpl.
+    now rewrite transportf_const.
+  Qed.
+
+  Definition pshf_unit : nat_trans (functor_identity (pshf ∫P)) (pshf_to_slice ∙ slice_to_pshf) :=
+    pshf_unit_fun ,, is_nat_trans_pshf_unit.
+
+  (* maybe move upstream *)
+  Lemma transportf_paths_rewrite {X : UU} {Q : X → UU} {x y : X} {e e' : x = y} {q : Q x} :
```
DanGrayson: Use this instead:
```
Lemma transportf_paths {X : UU} (P : X -> UU) {x1 x2 : X} {e1 e2 : x1 = x2} (e : e1 = e2)
      (p : P x1) : transportf P e1 p = transportf P e2 p.
```

DanGrayson:
Aside from that one remark, it looks fine to me.  @benediktahrens ?

benediktahrens:
lgtm.
could you put in a few comments that show up as HTML section titles?

mortberg:
I didn't have time to take a closer look yet, but maybe it would make sense if I first upstream all the lemmas of mine that you need. Some of them should maybe go in MoreFoundations and some in ElementsOp. Also there was a plan to try to keep Elements and ElementsOp as similar as possible so this would mean that I would have to prove all of the lemmas I put in ElementsOp in Elements as well. 

Another remark is that this file probably should rely on my file about presheaves from https://github.com/UniMath/UniMath/pull/607. Once that PR has been merged I plan to try to figure out if I can define presheaves differently so that we don't need to use pr1 all the time but instead can have a coercion to the underlying functor. If this works out I think parts of this proof will be a bit cleaner.

mortberg:
PS: Congratulations on finishing this! 

Just out of curiosity, did you try to do the rest of the proof using the definition of `pshf_to_slice_ob_funct` that you and Vladimir found?

mweav:
Yeah, the two version of the proof are entirely interchangeable; I just decided to use yours for this PR as it's a more direct proof and personally easier to understand.

UniMath/CategoryTheory/elems_slice_equiv.v
```diff
@@ -0,0 +1,431 @@
...
+    + exact (eqtohomot ((pr1 Qisfunct) x) (pr1 p)).
+    + exact (eqtohomot ((pr2 Qisfunct) x y z f g) (pr1 p)).
+  Qed.
+
+  Definition slice_to_pshf_ob : pshf C / P → pshf ∫P :=
+    fun Q =>  slice_to_pshf_ob_funct_data Q ,,  slice_to_pshf_ob_is_funct Q.
+
+  Definition slice_to_pshf_ob_nat {X Y : pshf C / P} (F : X --> Y) (e : ∫P^op) :
+    (slice_to_pshf_ob_ob X) e --> (slice_to_pshf_ob_ob Y) e :=
+    match e with
```
mortberg: match is not part of UniMath, how about using pr1 and pr2 instead?

UniMath/CategoryTheory/elems_slice_equiv.v
```diff
@@ -0,0 +1,431 @@
...
+Matthew Weaver, 2017
+************************************************************)
+
+
+(** **********************************************************
+Contents : Equivalence of the categories pshf ∫P and
+           pshf C / P for any P in pshf C
+************************************************************)
+
+Require Import UniMath.Foundations.Sets
```
mortberg: Please import UniMath.MoreFoundations.Tactics as well. If you don't do that you are using the bad version of the `now` tactic.

mortberg:
Maybe you can include the other proof as well? If I remember correctly it is shorter than mine and I'd like to understand it. If you want you can then remove all of my proofs and I'll upstream them later. If you want to keep them that's also fine and I'll move them in some future PR.

UniMath/CategoryTheory/elems_slice_equiv.v
DanGrayson: That's going to change soon: https://github.com/UniMath/UniMath/issues/658

DanGrayson:
@mortberg : Re: "Also there was a plan to try to keep Elements and ElementsOp as similar as possible so this would mean that I would have to prove all of the lemmas I put in ElementsOp in Elements as well."

What about instead adding the other associativity to the definition of precategory, so that C^op^op = C judgmentally, and seeing whether we can discard the file ElementsOp?  That would be a big improvement.

UniMath/CategoryTheory/elems_slice_equiv.v
mortberg: Ok, that is great. I think it is often a lot easier to read code written with match than with the eliminators, and if we are going to implement complicated algorithms in UniMath some day (like Gaussian elimination) I think we really need to be able to write readable programs.

mortberg:
About C^op^op=C: This is worth a separate issue and some discussion. The change to precategories will be nontrivial as all of the code where we define a particular precategory will break. The best way is probably to make a constructor function that takes the same data as for our current precategory and build the extra proof for us. This way the changes to existing code will mostly be calling the constructor function instead of mkpair. But like I said, let's discuss this in a separate issue instead.

mortberg:
This PR looks good to me now

UniMath/CategoryTheory/elems_slice_equiv.v
vladimirias: Here is an example of a match that is hard to replace with an eliminator. In the original Foundations have a look at the first match in the proof of theorem onefiber.

vladimirias:
LGTM. Please fix the conflicts and merge.

UniMath/CategoryTheory/elems_slice_equiv.v
DanGrayson: The two matches in `onefiber` can be eliminated as follows:
```
diff --git a/UniMath/Foundations/PartA.v b/UniMath/Foundations/PartA.v
index 90be4143..6fa29247 100644
--- a/UniMath/Foundations/PartA.v
+++ b/UniMath/Foundations/PartA.v
@@ -2867,15 +2867,10 @@ Proof.
       + intro ee. apply ii1; exact (pathscomp0 (pathsinv0 x0) ee).
       + intro phi. apply ii2, phi.
     - intro phi. exact (c x'). }
-
-
-  set (g := fun pp : total2 P =>
-              match (cnew (pr1 pp)) with
-              | ii1 e => transportb P e (pr2 pp)
-              | ii2 phi => fromempty (phi (pr2 pp))
-              end).
-
-
+  transparent assert (g : (total2 P → P x)).
+  { intros pp. induction (cnew (pr1 pp)) as [e|phi].
+    - exact (transportb P e (pr2 pp)).
+    - apply fromempty. exact (phi (pr2 pp)). }
   assert (efg : ∏ pp : total2 P, (f (g pp)) = pp).
   intro. induction pp as [ t x0 ]. set (cnewt := cnew t).
   unfold g. unfold f. simpl. change (cnew t) with cnewt.
@@ -2892,12 +2887,10 @@ Proof.
 
   assert (egf : ∏ p: P x, (g (f p)) = p). intro. simpl in g.
   unfold g. unfold f. simpl.
-
-  set (ff := fun cc:coprod (x = x) (P x -> empty) =>
-               match cc with
-               | ii1 e0 => transportb P e0 p
-               | ii2 phi => fromempty  (phi p)
-               end).
+  transparent assert (ff : ((x = x) ⨿ (P x → ∅) → P x)).
+  { intros cc. induction cc as [e0|phi].
+    - exact (transportb P e0 p).
+    - apply fromempty. exact (phi p). }
   assert (ee : (ff (cnewx)) = (ff (@ii1 (x = x) (P x -> empty) (idpath x)))).
   apply (maponpaths ff e).
   assert (eee : (ff (@ii1 (x = x) (P x -> empty) (idpath x))) = p). apply idpath.
```

DanGrayson: By the way, Anders, I misspoke when I said "That's going to change soon: #658".  I should have said only that when `match` is generated by Coq for us as we use `destruct`, `intros` or `induction`, we know how to replace each such use with an eliminator.  But I don't see how we can allow `match` in our code, as here, because there is no way for us to be sure that each such use can be replaced by eliminators.

DanGrayson:
Shouldn't we get rid of that one use of `match` in a term?  (The `match` tactic is okay, of course.)

Other than that, lgtm.

DanGrayson:
lgtm

PR #675, opened 2017-04-07
 Equivalence of binary products in C/Z to pullbacks of pairs of arrows to Z in C

mweav:
I proved that the type of all binary products in C/Z is equivalent to the type of all pullbacks of pairs of arrows to Z in C for any precategory C with homsets.


UniMath/CategoryTheory/pullbacks_slice_products_equiv.v
```diff
@@ -0,0 +1,187 @@
...
+                    j,, idpath (j · f) × t' · ((r,, req) : C / Z ⟦ (P ,, h), (B ,, g)⟧) = k,, Yeq).
+    {
+      refine (slice_eq _ _ _ ,, slice_eq _ _ _);
+        simpl; [exact tleq | exact treq].
+    }
+    exact (maponpaths pr1 (maponpaths pr1 (pr2 (PisProd (Y ,, j · f) (j ,, idpath _) (k ,, Yeq)) (t' ,, tmors)))).
+  Defined.
+
+   (** ** function taking pullbacks in C to binary products in C / Z *)
+  Definition pullback_to_binprod' {A B : C} {f : A --> Z} {g : B --> Z} :
```
mortberg: Isn't this function part of what I formalized in: https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/slicecat.v#L451

mweav: Yeah, I guess it is contained in the definition of your function, but I don't think I can use your function as it requires all pullbacks in the category to exist, while I am looking at any particular pullback that exists in any category.

mortberg: I think you could split my definition in two. First one that takes the pullback data and construct a binary product in C/Z and then use this to define `BinProducts C/X` from `Pullbacks C`. This way there will be less duplication

vladimirias:
Just a thought - is it related to the last PR by Matthew? I gave him the problem of constructing an equivalence between the type of binary product structures in C/Z and the type of pullback structures for pairs of morphisms with the codomain Z.

mortberg:
Yes, this is a comment on that PR. I noticed that what Matthew does is very
related to my construction of binary products in C/X from pullbacks in C. I
think it would be nice to avoid duplication and only have one formalization
of this. Maybe Matthew could include his formalization in the file on slice
categories and modify my construction to use his construction of binary
product cones in C/X from pullback cones in C.

--
Anders

mortberg:
Nice! lgtm

DanGrayson:
lgtm

benediktahrens:
It happens very often (to me) that given data has to be shown to be a pullback.
For this, I think it would be great to factor out an equivalence between propositions
```
isPullback .... <-> isBinaryProduct ...
```
which I think will be the most useful version of the equivalence in practice.

mweav:
I've defined functions between `isPullback ....` and `isBinaryProduct ...` and proved that both directions are weak equivalences. Is that what you had in mind?

benediktahrens:
On 04/20/2017 07:58 PM, Matthew Weaver wrote:
> I've defined functions between `isPullback ....` and `isBinaryProduct
> ...` and proved that both directions are weak equivalences. Is that
> what you had in mind?

Almost!

To conveniently apply the lemma `isPullback_to_isBinProductCone`, for
instance, it seems better to have only variables in the conclusion.
By this, I mean that `isBinProductCone` should be applied to pure
variables of suitable type, and `isPullback` should take suitable terms
built from those variables as arguments:
```
isPullback <terms built from X Y W a b> →
isBinProductCone (C / Z) X Y W a b
```
In this way, when the goal is `isBinProductCone .....`, and one does
`apply isPullback_to_isBinProductCone.`, then `X Y W a b` are suitably
inferred.

The same should be done for the implication in the other direction.

Does that make sense?

mweav:
It does. Now (I think) it should be what you want.

UniMath/CategoryTheory/pullbacks_slice_products_equiv.v
```diff
@@ -0,0 +1,199 @@
...
+      ++ apply isPull.
+         abstract (simpl; now rewrite <- jeq , keq).
+      ++ abstract (simpl; now rewrite leq, assoc, (pr1 (pr2 (pr1 (isPull _)))), jeq).
+    + abstract (split; apply (eq_mor_slicecat hsC); simpl;
+                [apply (pr1 (pr2 (pr1 (isPull _)))) | apply (pr2 (pr2 (pr1 (isPull _))))]).
+    + abstract (now intros q; apply isapropdirprod; apply (has_homsets_slice_precat hsC)).
+    + abstract (intros q [H1 H2]; apply (eq_mor_slicecat hsC);
+                refine (maponpaths pr1 ((pr2 (isPull _)) ((pr1 q) ,, (_ ,, _))));
+                [apply (maponpaths pr1 H1) | apply (maponpaths pr1 H2)]).
+  Qed.
```
benediktahrens: This should be `Defined.`.

UniMath/CategoryTheory/pullbacks_slice_products_equiv.v
```diff
@@ -0,0 +1,199 @@
...
+    use unique_exists.
+    + exact (pr1 (pr1 (pr1 (PisProd (Y ,, j · f) (j ,, idpath _) (k ,, Yeq))))).
+    + abstract (exact (maponpaths pr1 (pr1 (pr2 (pr1 (PisProd (Y ,, j · f) (j ,, idpath _) (k ,, Yeq))))) ,,
+                                  maponpaths pr1 (pr2 (pr2 (pr1 (PisProd (Y ,, j · f) (j ,, idpath _) (k ,, Yeq))))))).
+    + abstract (intros x; apply isofhleveldirprod; apply (hsC _ _)).
+    + intros t teqs.
+      refine (maponpaths pr1 (maponpaths pr1 (pr2 (PisProd (Y ,, j · f) (j ,, idpath _) (k ,, Yeq))
+                                                  ((t ,, (maponpaths (λ x, x · f) (!(pr1 teqs)) @ !(assoc _ _ _) @ maponpaths (λ x, t · x) (idpath _))) ,, _)))).
+      abstract (split; apply eq_mor_slicecat; [exact (pr1 teqs) | exact (pr2 teqs)]).
+  Qed.
```
benediktahrens: `Defined.`

UniMath/CategoryTheory/pullbacks_slice_products_equiv.v
```diff
@@ -0,0 +1,199 @@
...
+        UniMath.CategoryTheory.limits.binproducts
+        UniMath.CategoryTheory.limits.pullbacks.
+
+Local Open Scope cat.
+
+(** * Proof that the types of binary products in C/Z and pullbacks of pairs of arrows to Z in C are equivalent *)
+Section pullbacks_slice_products_equiv.
+
+  Definition some_pullback (C : precategory) (Z : C) : UU :=
+    ∑ A B (f : A --> Z) (g : B --> Z) , Pullback f g.
```
benediktahrens: What is the mathematical idea behind this definition? This type does not seem useful to me.

UniMath/CategoryTheory/pullbacks_slice_products_equiv.v
```diff
@@ -0,0 +1,199 @@
...
+      rewrite idpath_transportf.
+      use total2_paths2_f.
+    - now apply (eq_mor_slicecat hsC).
+    - rewrite transportf_const.
+      now apply (eq_mor_slicecat hsC).
+  Qed.
+
+  (** ** function taking the type of all binary products in C / Z
+         to the type of all pullbacks of functions to Z in C *)
+  Definition binprod_to_pullback : some_binprod (C / Z) → some_pullback C Z.
```
benediktahrens: This map and the following (and the fact that they form an equivalence) don't seem very useful to me, unlike the ones above.
This is related to those `some_*` types not having any mathematical meaning.

benediktahrens:
On 04/21/2017 07:35 PM, Matthew Weaver wrote:
> It does. Now (I think) it should be what you want.  
> 

Yes, it is. I hope that the maps will be more useful this way.

benediktahrens:
@mweav : could you please change the two Qed that I marked to Defined? We can then merge this PR.

PR #676, opened 2017-04-10
Generalize the proof that the forgetful functor from slice categories is a left adjoint

mortberg:
For some reason I only proved that slicecat_to_cat : C/X -> C is a left adjoint in the case of C=HSET. I realized today that I could generalize this to any C with binary products.


rmatthes:
LGTM - this adds a lot of clarity to the reasoning why each projection functor is omega-cocontinuous

benediktahrens:
lgtm

PR #685, opened 2017-04-14
spell out what monadic bind does in the slice category

rmatthes:
this is all as expected, but it needed to be done
the development is a bit cumbersome and might be improved
the substitution of a single term for an extra variable is also studied and nicely profits from dependent types in the formulation


UniMath/SubstitutionSystems/MultiSorted.v
```diff
@@ -520,6 +522,92 @@ use Monad_from_hss.
 - apply MultiSortedSigToHSS.
 Defined.
 
+
+(** The following definitions do not depend on the monad coming from our construction, only on the
+    slice category we are working in. *)
+
+Definition bind_slice {T:Monad (SET / sort)}{Gamma1 Gamma2 : SET_over_sort} (f : SET_over_sort⟦Gamma1,T Gamma2⟧) :
```
mortberg: Maybe write Gamma1 and Gamma2 using unicode symbols?

UniMath/SubstitutionSystems/MultiSorted.v
```diff
@@ -520,6 +522,92 @@ use Monad_from_hss.
...
+  SET_over_sort⟦T Gamma1,T Gamma2⟧ := bind f.
+
+(* the following would be on the right level of generality but make problems later:
```
mortberg: What problems do they cause?

UniMath/SubstitutionSystems/MultiSorted.v
```diff
@@ -520,6 +522,92 @@ use Monad_from_hss.
...
+  induction a as [a | a].
+  + apply idpath.
+  + simpl.
+    generalize (ii2(A:=unit) a).
+    clear a.
+    apply toforallpaths.
+    simpl.
+    intermediate_path ((BinCoproductArrow HSET (BinCoproductsHSET 1%CS (pr1 Gamma))
+                         (λ _ : unit, N) (pr1 ((Monads.η T) Gamma))) ·  (sort_in T)).
+    * apply idpath.
```
mortberg: It looks like you should be able to achieve this more directy using `change` instead of `intermediate_path` and then `apply idpath`?

UniMath/SubstitutionSystems/MultiSorted.v
```diff
@@ -520,6 +522,92 @@ use Monad_from_hss.
...
+      apply map_on_two_paths.
+      - apply idpath.
+      - set (aux1 := pr2 ((Monads.η T) Gamma)).
+        simpl in aux1.
+        apply pathsinv0.
+        eapply pathscomp0; try eapply aux1.
+        apply idpath.
+Defined.
+
+Lemma subst_slice_ok {T:Monad (SET / sort)}{Gamma:SET_over_sort}(N : wellsorted_in T Gamma)
```
mortberg: Maybe introduce a notation like `M{{N}}` for `subst_in N M`?

mortberg:
Can we prove standard rules for substitution now? For example the "substitution lemma" for permutating substitutions: 

M[N][L] = M[L][N[L]]

rmatthes:
M[x := N][y := L] = M[y := L][x := N[y := L]]
is in fact
(M[N])[L]=(ex(M)[wk(L)])[N[L]]
which starts with an M in a context with two distinguished variables at the end, N one less variable and L with no extra variable. wk (weakening by an extra variable) is an easy instance of bind_slice (better directly define weakening by an extra context), and ex (exchange of the two distinguished variables) should be an instance of bind_slice as well (better directly define exchange of two contexts at the end of a context).
I wonder whether this would be better just imported from the general monad situation.

mortberg:
I think it would be really nice to have this! Could you formalize it and add it to the PR without too much work? 

If you can do it at the generality of monads we should get the substitution lemma for the STLC (and any other language that can be specified as MS signature) by just instantiating the lemma. I think this could be a good selling point for this approach. The framework is quite general and it does not only give you a monad, but also some standard lemmas for a large class of languages.

rmatthes:
Thank you for this encouragement. I would say that weakening and exchange live on the level of abstract monads, as soon as we have binary coproducts in the base category. But one-variable substitution seems to be specific to the slice category. Still, we would have it for all MS signatures. I'll try.

UniMath/SubstitutionSystems/MultiSorted.v
```diff
@@ -520,6 +522,93 @@ use Monad_from_hss.
...
+(* the following would be on the right level of generality but make problems later (the monads are not coerced into their underlying functors):
+Definition wellsorted_in (T:[SET_over_sort,SET_over_sort])(Γ:SET_over_sort): hSet := pr1(pr1 T Γ).
+Definition sort_in (T:[SET_over_sort,SET_over_sort]){Γ:SET_over_sort}(M:wellsorted_in T Γ): sort := pr2 (pr1 T Γ) M.
+ *)
+
+Definition wellsorted_in (T:Monad (SET / sort))(Γ:SET_over_sort): hSet := pr1(pr1 T Γ).
+Definition sort_in (T:Monad (SET / sort)){Γ:SET_over_sort}(M:wellsorted_in T Γ): sort := pr2 (pr1 T Γ) M.
+
+Definition aux_fh {T:Monad (SET / sort)}{A1:hSet}{f1:A1->sort}{Γ2:SET_over_sort}
+   (f : A1->wellsorted_in T Γ2)(H: forall a1:A1, sort_in T (f a1) = f1 a1) : SET_over_sort⟦(A1,,f1),T Γ2⟧.
```
mortberg: If you want to use more unicode you can write \r for ->, \prod for forall and \_1 and \_2 if you want the 1 and 2 as subscripts

UniMath/SubstitutionSystems/MultiSorted.v
```diff
@@ -520,6 +522,93 @@ use Monad_from_hss.
...
+    change ((BinCoproductArrow HSET (BinCoproductsHSET 1%CS (pr1 Γ))
+                                    (λ _ : unit, N) (pr1 ((Monads.η T) Γ))) ·  (sort_in T) =
+                 BinCoproductArrow HSET (BinCoproductsHSET 1%CS (pr1 Γ))
+                                    (λ _ : unit, sort_in T N) (pr2 Γ)).
+    rewrite postcompWithBinCoproductArrow.
+    apply map_on_two_paths.
+    - apply idpath.
+    - set (aux1 := pr2 ((Monads.η T) Γ)).
+      apply pathsinv0.
+      eapply pathscomp0; try eapply aux1.
```
mortberg: We have a tactic called `etrans` for `eapply pathscomp0`

UniMath/SubstitutionSystems/MultiSorted.v
```diff
@@ -520,6 +522,93 @@ use Monad_from_hss.
...
+
+Lemma bind_slice_ok {T:Monad (SET / sort)}{A1:hSet}{f1:A1->sort}{Γ2:SET_over_sort}
+    (f : A1->wellsorted_in T Γ2)(H: forall a1:A1, sort_in T (f a1) = f1 a1)(M: wellsorted_in T (A1,,f1)) :
+    sort_in T (bind_slice f H M) = sort_in T M.
+Proof.
+  assert (H1 := pr2 (bind_instantiated (aux_fh f H))).
+  apply toforallpaths in H1.
+  apply pathsinv0.
+  rewrite H1.
+  apply idpath.
```
mortberg: If you get tired of writing `apply idpath` you can write `now rewrite H1`. When you write `now tac` it applies `tac` and tries to solve all of the generated subgoals using trivial things (like `apply idpath`).

UniMath/SubstitutionSystems/MultiSorted.v
```diff
@@ -520,6 +522,93 @@ use Monad_from_hss.
...
+(** now we only substitute a single sorted variable *)
+
+Definition aux_inject_N {T:Monad (SET / sort)}{Γ:SET_over_sort}(N : wellsorted_in T Γ):
+  SET_over_sort⟦constHSET_slice (sort_in T N),T Γ⟧.
+Proof.
+  mkpair.
+  + exact (fun _=> N).
+  + apply funextsec.
+    intro u.
+    apply idpath.
```
mortberg: This is another case where I would use `now` to avoid typing so much; this subgoal should be solved by `now apply funextsec.`

benediktahrens:
On 04/15/2017 05:19 PM, Ralph Matthes wrote:
> Thank you for this encouragement. I would say that weakening and
> exchange live on the level of abstract monads, as soon as we have
> binary coproducts in the base category. But one-variable substitution
> seems to be specific to the slice category. Still, we would have it
> for all MS signatures. I'll try.
> 

One-variable substitution looks different depending on the base
category, indeed.

The definition monadSubst [1] gives a form that does apply to untyped
syntax (on Set), but not to sorted syntax (on Set/sort or sort->Set).


[1]
https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/Monads.v#L278

UniMath/SubstitutionSystems/MultiSorted.v
rmatthes: I have problems with special symbols on my computer at home and do not follow the otherwise good advice.

rmatthes:
@benediktahrens: yes, this was the problem. In passing, I would also show the untyped case of the interchange law Anders introduced into this thread.

rmatthes:
The commit "wip" did compile on my computer, but there were still Benedikt's experiments in my local copy (entirely my own responsibility), which is why I forgot to commit that file also. The untyped interchange lemma for substitution is now settled in a very abstract setting. This cannot be expected for the multi-sorted one. I'll do this concretely in Set/sort.

rmatthes:
I now have *an* interchange law for one-variable multi-sorted substitution. It is still very categorical, and I did not yet succeed in showing the user-friendly version.

UniMath/CategoryTheory/Monads.v
```diff
@@ -267,15 +267,111 @@ Qed.
...
 (** * Substitution operation for monads *)
 Section MonadSubst.
 
-Context {C : precategory} (T : Monad C) (TC : Terminal C) (BC : BinCoproducts C).
+
+Definition monadSubstGen {b:C} (a : C) (e : C⟦b,T a⟧) : C⟦T (b ⊕ a), T a⟧ :=
+  bind (BinCoproductArrow _ _ e (η T a)).
+
+Lemma subst_interchange_law_gen (c b a : C) (e : C⟦c,T (b ⊕ a)⟧) (f : C⟦b,T a⟧):
+  (monadSubstGen _ e) · (monadSubstGen _ f) = (mexch c b a) · (monadSubstGen _ (f · (mweak c a)))  · (monadSubstGen _ (e · (monadSubstGen _ f))).
```
mortberg: This line is very long and I can't see the end of it in the github interface

UniMath/CategoryTheory/Monads.v
```diff
@@ -267,15 +267,111 @@ Qed.
...
 Local Notation "1" := TC.
-Local Notation "a ⊕ b" := (BinCoproductObject _ (BC a b)) (at level 50).
 
-Definition monadSubst (a : C) (e : C⟦1,T a⟧) : C⟦T (a ⊕ 1), T a⟧ :=
-  bind (BinCoproductArrow _ _ (η T a) e).
+Definition monadSubst (a : C) (e : C⟦1,T a⟧) : C⟦T (1 ⊕ a), T a⟧ :=
+  monadSubstGen a e.
+
+Lemma subst_interchange_law (a : C) (e : C⟦1,T (1 ⊕ a)⟧) (f : C⟦1,T a⟧):
+  (monadSubst _ e) · (monadSubst _ f) = (mexch 1 1 a) · (monadSubst _ (f · (mweak 1 a)))  · (monadSubst _ (e · (monadSubst _ f))).
```
mortberg: another long line

UniMath/CategoryTheory/Monads.v
rmatthes: Sorry, is now hopefully corrected - I do not see it the same way it seems.

PR #691, opened 2017-04-19
Replaced many hSet with UU in BinaryOperations

vladimirias:
Now things like isassoc and iscomm are applicable to binary operations on types.  This should help with Modules.v

The drawback is that the names are misleading now since "isassoc" is a proposition only if the carrier type X is a set. It is an issue. One possibility is to remove the "is" everywhere (except for the isapropisassoc etc. that kept the assumption that X is a set). I create an issue with it.

I want to see how much unnecessary stuff it will help to eliminate in Modules and then probably remove more isaset assumptions introducing typewithbinop instead of setwithbinop etc.


DanGrayson:
Why will it help with Modules.v?  Modules over a ring are sets.

UniMath/Foundations/PartA.v
```diff
@@ -1778,9 +1778,13 @@ Definition weqgradth {X Y : UU} (f : X -> Y) (g : Y -> X)
            (efg: ∏ y : Y, f (g y) = y) : X ≃ Y :=
   weqpair _ (gradth _ _ egf efg).
 
+(* Dubious definition *)
```
DanGrayson: I found the definition useful when constructing representations of functors -- it echoes the traditional approach to proving universal properties in category theory, where first you construct the thing, and then you you prove it's unique.  At least it doesn't start with "is", so it isn't dubious for that reason.

vladimirias:
We are talking about "bijective", right? Fixing a section of a function and asserting that it is surjective (the first half of the definition of being bijective) are very different things. I am not against the lemma as such, but against it's name. Especially, because "bijective" is such a standard thing in set-theoretic math and a word that a newcomer who needs to construct an equivalence is very likely to search for.

So could you please suggest another name and maybe place it where other properties of "injective" maps are studied?

vladimirias:
It will help with modules because of the problem of defining a coercion from a set. Sets in Coq don't have elements because they are not types, something that you have pointed out yourself. Now one can prove all the important properties of the two operations on endomorphisms of abelian groups without having to write the pr1's.

vladimirias:
It is also something that makes no sense not to do - I have changed hSet to UU in about a hundred places and so extended the applicability of about a hundred constructions without having to change anything downstream at all.

DanGrayson:
Indeed, it's not a good name.  I'll come up with one.

DanGrayson:
What about "UniqueConstruction" ?

vladimirias:
Sounds good except, possibly, for capitalizations, but this will have to be discussed separately.

benediktahrens:
lgtm

PR #693, opened 2017-04-22
change bijective to UniqueConstruction

DanGrayson:


benediktahrens:
What is the idea behind the name `UniqueConstruction` ?

DanGrayson:
When wanting to prove, in category theory, that a certain construction has universal property, such as being a product, one may do it two steps: construct the arrow to it from a test object; and show any two such arrows are unique.  I use this definition there for representations of functors.

benediktahrens:
lgtm

Issue #699, opened 2017-05-04
Implement VV's naming suggestion for categories 


benediktahrens:
as per https://github.com/UniMath/UniMath/issues/362#issuecomment-225556634

benediktahrens:
@peterlefanulumsdaine and I would like to move forward on this proposal in the next few days (hours?). I will try to merge PRs before this change, to avoid creating merge conflicts. 
Any comments and suggestions amending the original proposal are welcome.

benediktahrens:
I'll merge 
- https://github.com/UniMath/UniMath/pull/607
- https://github.com/UniMath/UniMath/pull/632
- https://github.com/UniMath/UniMath/pull/679
before starting the renaming. The other PRs are not ready to be merged, or are not affected.

benediktahrens:
Renaming `(is_)category` into `(is_)univalent[_category]` is done in #701 .

I'd be happy to have a short discussion about whether
`Precategory` -> `category`
is the best we can come up with.

DanGrayson:
I think `Precategory` -> `category` or `Category` is pretty good, since it's close to what one would get by translating the mathematical definition to our context.  The only thing omitted is the condition that the objects form a set, which we can see now, from the univalent point of view, would be a mistake to perpetuate, and `setcategory` or `SetCategory` will take care of that.

(I prefer type names to be capitalized.)

peterlefanulumsdaine:
> The only thing omitted is the condition that the objects form a set, which we can see now, from the univalent point of view, would be a mistake to perpetuate

I wouldn’t even feel this is a change/omission.  Objects forming a set (in the h-set sense, not the size sense) isn’t a condition in the standard definition, it’s just a consequence of the classical setting.  Whether it’s omitted or included, this definition becomes equivalent to the standard one under the interpretation of types as sets.

So I also feel `Precategory` -> `category` or `Category` is very good.  (Like Dan, I like capitalisation for type names.)

DanGrayson:
#176 presents a uniform naming proposal

DanGrayson:
I agree with Peter.  One may formulate a criterion for deciding whether a type in a translated definition should be a set: if one wants to consider equality of its elements.  In category theory, it is commonly known that considering equality of objects should be avoided, whereas considering equality of maps is unavoidable, because of associativity.

mortberg:
[...]

PR #701, opened 2017-05-04
Introduce vocabulary "univalence" for categories

benediktahrens:
Rename
- `is_category` -> `is_univalent`
- `category` -> `univalent_category`

### DIFF
--- a/UniMath/CategoryTheory/FunctorAlgebras.v
+++ b/UniMath/CategoryTheory/FunctorAlgebras.v
@@ -143,7 +143,7 @@ Qed.
 
 Section FunctorAlg_saturated.
 
-Hypothesis H : is_category C.
+Hypothesis H : is_univalent C.
 
 Definition algebra_eq_type (X Y : FunctorAlg (pr2 H)) : UU
   := ∑ p : iso (pr1 X) (pr1 Y), pr2 X · p = #F p · pr2 Y.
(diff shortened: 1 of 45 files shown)


mortberg:
lgtm

Will you also rename precategory -> category? Maybe this will go into another PR?

benediktahrens:
On 05/04/2017 10:49 PM, Anders Mörtberg wrote:
> lgtm
> 
> Will you also rename precategory -> category? Maybe this will go into another PR?

You mean Precategory -> category ?

I am slightly weary about that. I am thinking about reserving the word
"category" for informal discussion and comments, i.e., for the union of
all the notions of categories in UniMath.

But I haven't had a good idea for a word to replace "Precategory".

I'd be very grateful for any suggestions.

Renaming "Precategory" should probably go in a different PR. At least to
me, the present one is very uncontroversial, whereas replacing
"Precategory" is more difficult.

mortberg:
Yes, that's what I meant. I updated my comment directly after sending it, but I guess that isn't reflected in the email sent out to everyone watching the library.

I like the name "category" for what we now call "Precategory". I think it is the closest to what people normally mean by the word "category" so it makes perfect sense to me to use it in our formalization. Maybe this should be discussed in a separate issue though and not here, as you say this PR should be very uncontroversial in its current form while renaming Precategory requires some discussion.

DanGrayson:
... *wary* ...

DanGrayson:
this addresses part of #699

benediktahrens:
On 05/04/2017 11:07 PM, Daniel R. Grayson wrote:
> ... *wary* ...
> 

Yes, that's it. Thanks :-)

PR #702, opened 2017-05-05
rename Precategory -> category

benediktahrens:
This PR renames `Precategory` to `category`.

The renaming destroyed some CamlCase where "Precategory" occurred in the middle of a word and I replaced it by "category" for uniformity.

I did not rename section titles of the form "Precategory of foo", since it seemed to me that capitalization here stems from the fact that it is the first word of a section title.

I have not renamed files that carry "Precategory" in their names.


DanGrayson:
lgtm

mortberg:
lgtm

PR #710, opened 2017-05-10
Right modules over monads renamed into Left modules over monads

amblafont:
Rename right modules over monads into left modules over monads


benediktahrens:
lgtm

UniMath/CategoryTheory/LModules.v
```diff
@@ -0,0 +1,296 @@
...
+
+
+
+Definition LModule_data (D:precategory) : UU
+  := ∑ F : functor B D, F □ M ⟹ F.
+
+Coercion functor_from_LModule_data (C : precategory) (F : LModule_data C)
+  : functor B C := pr1 F.
+
+Definition σ {C : precategory} (F : LModule_data C) : F□M ⟹ F := pr2 F.
```
benediktahrens: It is not a good idea, in my opinion, to have a one-letter identifier.
Better: use a longer identifier, like `lmbind` (for Left Module bind), and use `σ` as *local* notation.

benediktahrens:
lgtm

PR #711, opened 2017-05-10
Proto monads and modules

benediktahrens:
as per Ralph's slides

### DIFF
--- a/UniMath/CategoryTheory/Proto.v
+++ b/UniMath/CategoryTheory/Proto.v
@@ -17,7 +17,7 @@ Require Import UniMath.Foundations.Sets.
 
 Require Import UniMath.MoreFoundations.Tactics.
 
-Require Import UniMath.CategoryTheory.precategories.
+Require Import UniMath.CategoryTheory.Categories.
 Require Import UniMath.CategoryTheory.functor_categories.
 
 Local Open Scope cat.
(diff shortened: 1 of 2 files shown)


DanGrayson:
lgtm

UniMath/CategoryTheory/Proto.v
```diff
@@ -0,0 +1,70 @@
+(** **********************************************************
+
+Contents:
+
+        - Definition of proto-monads [proto_monad]
```
rmatthes: one should mention the proto-modules as well
there should be a disclaimer that proto-monad has this name only for systematic purposes: it is a pointed functor in the relative setting

vladimirias:
In the definition of a relative monad the first component of the structure is not a functor, but only a function from Ob C to Ob D. I have not checked whether action on morphisms is important in the proto stuff, but for the connection with the relative monads it should not be.

rmatthes:
@vladimirias : for the proto-stuff, I really need functors since the morphism mapping cannot be constructed with good properties if not enough laws are around; however, thanks to naturality, the morphism mapping enjoys the usual presentation in terms of mbind and eta. I also feel the need for being able to instantiate to the signatures with strength as is.

vladimirias:
The fact that one only needs a function on objects is an important simplifying factor in the relative monads theory. With it one avoids the naturality conditions and the whole structure becomes much more compact. I wonder whether you really need this additional stuff.

rmatthes:
I know that I already needed it for legacy reasons: I want that all signatures with strength yield abstract signatures in the new sense based on proto-modules and premodules. But I'll examine this again.

PR #712, opened 2017-05-10
Definition of relative monads

benediktahrens:


DanGrayson:
We'll soon have at least three files about monads, so maybe a subdirectory devoted to them would be a good idea.

DanGrayson:
lgtm

benediktahrens:
On 05/10/2017 04:46 PM, Daniel R. Grayson wrote:
> We'll soon have at least three files about monads, so maybe a
> subdirectory devoted to them would be a good idea.

Good idea!

PR #714, opened 2017-05-16
stop ignoring all warnings

### LINKED ISSUE #713: Ignoring warnings
It was brought to my attention that we currently ignore all warnings when we compile UniMath. This seems a little dangerous as some warnings can be useful and point to things that should be changed as they might break in the future (an example of this was when we were introducing too many things in some places). In 8.6 there is now support for disabling specific warnings or even families of warnings (when a warning is printed it comes with its name and which family it belongs to), so we can probably come up with a more specific set of warnings that we want to ignore while keeping informative ones. 

Maybe @DanGrayson is up for experimenting with this as it involves modifying the Makefile?

DanGrayson:
Now we ignore just two warnings, and we convert one into an error, in
the makefile.  We also fixed some of the code to remove some warnings.

Note: two non-functional coercions were removed:

```
Coercion carrier_set : hsubtype >-> hSet.

Coercion hnnq_set_to_hq : pr1hSet >-> pr1hSet.
```

Resolves #713.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,7 @@ $(VFILES:.v=.vo) : $(COQBIN)coqc
 endif
 
 OTHERFLAGS += $(MOREFLAGS)
-OTHERFLAGS += -indices-matter -type-in-type -w none
+OTHERFLAGS += -indices-matter -type-in-type -w '-notation-overridden,-local-declaration,+uniform-inheritance'
 ifeq ($(VERBOSE),yes)
 OTHERFLAGS += -verbose
 endif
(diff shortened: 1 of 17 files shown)


mortberg:
lgtm

benediktahrens:
Most changes seem to be about splitting up a `Require Import foo.` into `Require foo.` and `Import foo.`. Could you explain what this is about, so that we know when this is necessary to avoid a warning?

DanGrayson:
The warning message explains that using "Require" inside a module is strongly discouraged.  I don't know why, but it is easy to avoid.

Issue #406, opened 2016-07-09
Moving a file


vladimirias:
HLevel_n_is_of_hlevel_Sn.v should not be in category theory, it should be in Foundations.

benediktahrens:
On 07/10/2016 12:45 AM, Vladimir Voevodsky wrote:
> HLevel_n_is_of_hlevel_Sn.v should not be in category theory, it should be in Foundations. 

Absolutely. However, we should first reorganize Foundations as per issue
#350 , and then move this file to the adequate spot.

benediktahrens:
Duplicate of #523 , closing this one.

PR #715, opened 2017-05-16
Use -arg to pass arguments to COQC

gares:
This patch makes it work with https://github.com/coq/coq/pull/406

BTW, this is the recommended way of passing arguments to COQC (that I discovered not being documented... but is there since, at least, 8.4).

Note that I discourage including a generated Makefile because no matter what we change in such file we break users (i.e. renaming variables for example).  I try to recommend alternatives here https://github.com/coq/coq/pull/406/files#diff-aaaca1f91738f8f7dc4afb4f7e249dbd

In this specific case one can use `-arg` (as in this PR), or do the `+=` magic in the `.local` file (see the link above).
Also note that the `.conf` file (again see the link above) contains all variable (and only that) with names that live in the `COQMF_` name space and is hence safe to include that file in other Makefiles.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,6 +136,7 @@ describe:; git describe --dirty --long --always --abbrev=40 --all
 	echo '# It is made by automatically (by code in Makefile)' ;\
 	echo ;\
 	echo '-Q UniMath UniMath' ;\
+	echo '-arg "$(OTHERFLAGS)"' ;\
 	echo ;\
 	for i in $(PACKAGES) ;\
 	do <UniMath/$$i/.package/files $(FILES_FILTER) |sed "s=^=UniMath/$$i/="  ;\


DanGrayson:
Thank you, I'll take a look at that, as well as at possibly avoiding including the makefile.

maximedenes:
Hello,

Can we make progress on this? I'm going to merge Coq's PR #406 which will break UniMath if this is not merged. Thanks!

benediktahrens:
lgtm

DanGrayson:
Pulling https://github.com/coq/coq/pull/406 into Coq won't break anything for us, until we try to upgrade our Coq submodule, so perhaps I can infer that Coq now is tested automatically to see if it builds UniMath successfully.  Is that right?  (If so, @benediktahrens and @vladimirias, I should handle this quickly, so we don't slow down Coq development -- let me know if you want to be involved.)

Assuming that is right, let me also guess how your patch fixes anything for you: is it because the make variable OTHERFLAGS no longer will occur the in the makefile made by coq_makefile?

Assuming that's right, then we still need a way to recover the previous functionality, which allowed us to adjust the options passed to `coqc` by adding options to the `make` command line, without giving the user the responsibility to force coq_makefile to be run again. We document this in our file INSTALL.md, for example, as a way of timing `coqc`.

Also, once polymorphic universes start working better, we'll want to compile just one file with the option `-type-in-type`.  The current `coq_makefile` allows us to do that by adding the lines
```
TYPE_IN_TYPE_FILES := 				\
	UniMath/Foundations/Resizing2.vo

$(TYPE_IN_TYPE_FILES) : OTHERFLAGS += -type-in-type
```
to our Makefile (see https://github.com/DanGrayson/UniMath/tree/file-by-file-type-in-type).  We will need that functionality, too.  Does your `coq_makefile` provide a way?

I'll take a look at https://github.com/coq/coq/pull/406 and the alternatives you suggest to see if there's a better fix that gives us the functionality we need.

benediktahrens:
@DanGrayson : I don't need to be involved, but at the end, I would like to understand what was changed.

DanGrayson:
I just tried building UniMath with https://github.com/coq/coq/pull/406 and it worked, with no changes on our end.  Moreover, the flag OTHERFLAGS is still in the makefile, so I don't yet see the problem fixed by this pull request.

maximedenes:
@gares can you explain what broke when you tested UniMath on coq/coq#406?

maximedenes:
@DanGrayson if you try with a fresh coq tree, you'll see that this PR is needed for `-type-in-type` to be passed to coq properly.

Can this PR be merged? UniMath is now broken in Coq's test suite. Thanks!

maximedenes:
@gares can you make more explicit the change of behavior of `coq_makefile` in the case of UniMath?

ejgallego:
> I just tried building UniMath with coq/coq#406 and it worked, with no changes on our end.

Well, you really need to be sure your tree is clean and you are using trunk's coq_makefile. I can confirm indeed that UniMath makefile is broken with current Coq trunk.

DanGrayson:
Okay, I found the real problem with the new coq_makefile: it's these lines in the resulting
makefile:
```
all:
	$(HIDE)$(MAKE) --no-print-directory -f "$(SELF)" pre-all
	$(HIDE)$(MAKE) --no-print-directory -f "$(SELF)" real-all
	$(HIDE)$(MAKE) --no-print-directory -f "$(SELF)" post-all
```
They should be replaced by
```
all: pre-all real-all post-all
```
, because they bypass any make variable settings that the caller of `make` might have used.

gares:
My concern with that implementation is that it breaks with -j.  In particular the meaning of `foo: bar baz` is, as far as I can tell, "foo depends on bar and baz" not "foo wants bar to be done before baz".

On a side note, with the new implementation of coq_makefile you can stick variable assignements also in the Makefile.local files (that is included even in sub-calls) or (and in this PR) using `-arg`.

gares:
Ops, I think I've replied to the comment I received by email and you've improved your solution afterwards.
This one looks good and is better than the one I found. I think we should apply it in Coq, since it is more retro-compatible than the current code.

gares:
Sorry, I think I spoke too early.  This would be the resulting code:
```

all: post-all

# Extension points for actions to be performed before/after the all target
pre-all::
        @# Extension point
        $(HIDE)if [ "$(COQMAKEFILE_VERSION)" != "$(COQ_VERSION)" ]; then\
          echo "W: This Makefile was generated by Coq $(COQMAKEFILE_VERSION)";\
          echo "W: while the current Coq version is $(COQ_VERSION)";\
        fi
.PHONY: pre-all

post-all:: real-all
        @# Extension point
.PHONY: post-all

real-all: pre-all $(VOFILES) $(CMOFILES) $(CMAFILES) $(if $(DO_NATDYNLINK),$(CMXSFILES))
.PHONY: real-all
```

That does not guarantee that pre-all is run before any real work is done. Maybe I'm a bit tired, but I don't see how to fix that in this approach.

DanGrayson:
Oh, right, good point.  I suppose you can say something like this:
```
$(VOFILES) $(CMOFILES) $(CMAFILES) $(if $(DO_NATDYNLINK),$(CMXSFILES)) : pre-all
```

DanGrayson:
Oops, no, that's bad, because `pre-all` is phony, so the other files will always be rebuilt.

Do people really care about that warning message?  Even if they do, since it's just a warning message, is it so important that it get displayed at the top?  Most people just look at the bottom
of a build log to see if there was an error, not at the top.

DanGrayson:
Ah, here's the point about that warning message: you can generate it as the Makefile is being read.  There is no reason to think of it as a build step.  Then it will always be done first.
Like this:
```
ifneq "$(COQMAKEFILE_VERSION)" "$(COQ_VERSION)"
   $(warning "warning: This Makefile was generated by Coq $(COQMAKEFILE_VERSION) while the current Coq version is $(COQ_VERSION)")
endif
```

DanGrayson:
Notice that with this approach, the warning will be given, as it should, even if there is no real work to be done!

gares:
The pre/post business is not there for the warning, that happens to have landed there but can as well be put at the top level as you suggest.

The pre/post things are extension points to cover some functionality of coq_makefile1 in a less fragile way.
In particular coq_makefile1 lets people specify directories in the _CoeProject file and the resulting Makefile calls recursively in there.  Unfortunately it did not specify exactly when... The extension points pre-all and post-all lets one specify custom action to be performed before and after the project is built.

I personally think this feature is a bit overkill.  If one calls the generated Makefile (rather than including it) it is quite easy to perform actions like a call into a sub directory before/after the ball is passed to the generated Makefile.  In any case, this is the approach I should recommend in the manual.
 
To sum up, I'm happy to remove the pre/post business, but I think we are a bit lost in a XY problem.
Makefiles don't have a notion of "an interface", still I find the target names and the _CoqProject format is the best approximation of it. I should do my best to document which variables names one can set via command line or via the .local file. But we can't fix/improve coq_makefile without changing how targets are implemented, how variables defined etc.  This is why I'm discouraging the inclusion of the generated Makefile, since it is very very fragile. Eg. any recursive call is going to lose the options you set, the pre/post one may be removed, but we may want to use rec-calls for another reasons and the problem is going to pop up again :-(

DanGrayson:
Yes, it might be good if we didn't include your Makefile, but for us it is the best way.  For
one example, I put your list of files VFILES to many good uses.  I don't want to generate
it myself by a different mechanism and risk it not being the same.  For another example,
eventually we will need to pass type-in-type to coq for just one file, so we will need to
add a dependent rule involving OTHERFLAGS.

We include Coq as a submodule, and submodules point to particular commits, so no change
to coq_makefile will take us be surprise, and we will always be able to adapt.

Perhaps it would be useful for you if I were to produce a list of the make variables from
your makefile that we use : you could document them, and designate their behavior as
worth preserving.

I agree that the pre-all - post-all feature is overkill, and I was going to mention the solution
you propose.  It's also not in line with what "make" is written to do: run the steps
needed in the order needed.

For now it seems to be a good idea for you to preserve upward compatibility of the
behavior of "make all".

As an aside, I noticed that your travis test script tracks the UniMath master branch, instead
of referring to a particular commit of UniMath.  That means changes to UniMath
could, in theory, break your test script.  On the other hand, if one does, it might be best
to learn of it quickly.

DanGrayson:
PS: Another reason to avoid all recursive calls to "make" is that it is a poor substitute for correct accounting of all the prerequisites to a target.  Probably the soundest practice is to let every project be built by a single "make" process, provide all the information about prerequisites, so it can do everything and take maximum advantage of parallelism.

DanGrayson:
@gares , if we are in agreement, shall we close this pull request and leave it to you to restore the former behavior of `make all` in makefiles made by `coq_makefile`?

gares:
> I put your list of files VFILES to many good uses. I don't want to generate
> it myself by a different mechanism and risk it not being the same.

I'm a bit in a hurry, but this is how coq_makefile (2) works: it generates a Makefile.coq (assuming you pass to it -o Makefile.coq) and a Makefile.coq.conf file that is meant to be included.  An excerpt from such file:
```
###############################################################################
#                                                                             #
# Project files.                                                              #
#                                                                             #
###############################################################################

COQMF_VFILES = theories/test.v
COQMF_MLIFILES = src/test.mli src/test_aux.mli
COQMF_MLFILES = src/test_aux.ml
COQMF_ML4FILES = src/test.ml4
COQMF_MLPACKFILES = src/test_plugin.mlpack
COQMF_MLLIBFILES =
```
So if you need the list of .v files you can just use `COQMF_VFILES`. Note the the `.conf` has no rules in it, and all variables names start with `COQMF_`, so there is no risk in including it.

Finally the generated Makefile.coq includes a Makefile.coq.local (if present) and there you are welcome to add custom rules/code, or `+=` variables.

DanGrayson:
We might be able to convert our Makefile over to the new paradigm once it's part of 8.7 and 8.7 is officially released, but we can't do so earlier, because we want to maintain compatibility of UniMath with the current official release of Coq so our users don't have to compile Coq.  Even after the release of 8.7, we may prefer to maintain compatibility of UniMath with 8.6.

Do you see a way to proceed other than by making the makefile produced by coq_makefile respond to `make all` as it used to, without calling `make` recursively?  Backward compatibility is a desirable trait for software, if it's easy to achieve.

gares:
I'm fine to remove the pre/post business if it breaks no other project. But I'll have to try (next week).

Sorry for being stubborn, but I still don't see why this PR should be dropped. It uses a feature that works in 8.6 and will still work in 8.7 to pass the `-type-in-type -indices-matter` flags to `coqc`.  So it is harmless today, and a safeguard tomorrow (in case I can't get rid of the rec call, for example). 

Oh, I forgot to mention, if such `-arg stuff` is present in a _CoqProject that CoqIDE or Proof General can find, then `stuff` is also passed to to coqtop in interactive mode (something you obtain via `.dir-locals.el` that today contains a copy of `stuff`).

DanGrayson:
Okay, that's a good argument, so in the interest of not continuing to break your tests, we'll pull your request.  Thanks for agreeing to take a look at it next week.

DanGrayson:
In case some of your users really need the pre/post feature, which is hard to envision, why not leave `make all` as it used to be, with no recursive call, and introduce a new target named something like `pre-all-post`?

DanGrayson:
Alternatively, if `make all` is to continue using recursive calls, I can make our makefile conditional on the version of coq_makefile encountered, to preserve compatibility with the v8.6 and master, testing that also in our travis script.

DanGrayson:
@benediktahrens -- this pull request will not break things for us, but it might look a bit confusing to see command line options such as type-in-type appearing twice on each command line.

gares:
I gave it a stab:
  https://github.com/gares/coq/commit/b4c8632896ccb56e83f065e1ef8f3b525329e32b

In particular the corresponding travis job is running:
  https://travis-ci.org/gares/coq/builds/238416745?utm_source=github_status&utm_medium=notification

fingers crossed.

benediktahrens:
On 06/01/2017 02:54 PM, Daniel R. Grayson wrote:
> @benediktahrens <https://github.com/benediktahrens> -- this pull request
> will not break things for us, but it might look a bit confusing to see
> command line options such as type-in-type appearing twice on each
> command line.

Thanks for letting me know!

DanGrayson:
Thanks, Enrico!

PR #716, opened 2017-05-18
module transformer from strengthened signatures

amblafont:
Let (H, θ) be a strengthened signature, M an endo-functor.
If M has a structure of (left) module over a monad T, then it can be lifted to endow
H(M) with a structure of module over T.


DanGrayson:
lgtm

rmatthes:
lgtm

benediktahrens:
Great!
Could you put in a reference to H&M's "Initial Semantics for Strengthened Signatures"?
With that reference it is ready to be merged.

PR #717, opened 2017-05-22
Define derivative of an endofuctor

jojhelfer:
Also:
 - distributive law for monads
 - derivative of a monad
 - derivative of a left module over a monad

### DIFF
--- a/UniMath/CategoryTheory/Derivative.v
+++ b/UniMath/CategoryTheory/Derivative.v
@@ -31,11 +31,13 @@ Require Import UniMath.CategoryTheory.LModules.
 
 Local Open Scope cat.
 
+(** * Definition of distributive laws for Monads and composition of Monads
+ cf. Beck "Distributive laws" (1969) *)
 Section comp_def.
 
 Context {C : precategory} {S T : Monad C}.
 
-(* distributivity law for a pair of monads *)
+(** distributivity law for a pair of monads *)
 Definition monad_dist_laws (a : T ∙ S ⟹ S ∙ T) :=
   (((∏ x : C, η S (T x) · a x = #T (η S x)) ×
     (∏ x : C, #S (η T x) · a x = η T (S x))) ×
@@ -47,7 +49,7 @@ Definition monad_dist_law2 {a : T ∙ S ⟹ S ∙ T} (l : monad_dist_laws a) := 
 Definition monad_dist_law3 {a : T ∙ S ⟹ S ∙ T} (l : monad_dist_laws a) := (pr2 (pr1 l)).
 Definition monad_dist_law4 {a : T ∙ S ⟹ S ∙ T} (l : monad_dist_laws a) := pr2 l.
 
-(* composition of monads *)
+(** composition of monads with a distributive law *)
 Definition monad_comp_mu (a : T ∙ S ⟹ S ∙ T) : (S ∙ T ∙ S ∙ T) ⟹ (S ∙ T) :=
   nat_trans_comp _ _ _ (post_whisker (pre_whisker S a) T)
                        (nat_trans_comp _ _ _ (pre_whisker (S ∙ S) (μ T)) (post_whisker (μ S) T)).
@@ -58,23 +60,55 @@ Definition monad_comp_eta (a : T ∙ S ⟹ S ∙ T): functor_identity C ⟹ S �
 Definition monad_comp_data (a : T ∙ S ⟹ S ∙ T) : Monad_data C :=
   (tpair _ (tpair _ (S ∙ T) (monad_comp_mu a)) (monad_comp_eta a)).
 
+(** Below are the proofs of the monad laws for the composition of monads. We prove them as separate
+    lemmas not only because they are somewhat lengthy, but also for the following reason: the μ and
+    η for this monad are defined via operations on natural transformations, rather than their value
+    being given explicitly at each object. However, for the proofs, it is desirable to have these
+    explicit expressions; the easiest way to accomplish this is to write out the statements by hand
+    in the desired form (as ugly as they may be).
+
+    This is also done for the same reason later in the file, where the proofs of the individual
+    lemmas are not as lengthy.
+*)
+
+(**
+Here is the diagram corresponding to this proof. The outside of the diagram represents the equation
+to be proved. The numbers indicate the order in which the sub-diagrams are used.
+TSTSx ---------------> TTSSx ----------->  TSSx  ------------> TSx
+  ^     #T (a (Sx))     ^     μ T (SSx)     ^     #T (μ S x)   ^
+  |                     |                  /                  /
+  |η T (STSx)           |η T (TSSx)       /id                /
+  |                 1   |            3   /                  /
+           a (Sx)                       /                  /
+ STSx  --------------> TSSx ------------         4        /
+  ^                     ^                                /
+  |             2      /#T (η S (Sx))                   /
+  |η S (TSx)          /                                /
+  | _________________/                                /
+  |/                               id                /
+ TSx ------------------------------------------------
+*)
 Local Lemma monad_comp_law1 {a : T ∙ S ⟹ S ∙ T} (l : monad_dist_laws a) : ∏ x : C,
   (η S (T (S x))) · (η T (S (T (S x)))) · (#T (a (S x)) · (μ T (S (S x)) · #T (μ S x))) =
   identity (T (S x)).
 Proof.
   intro x.
+  (* 1 *)
   rewrite <- assoc.
   rewrite !(assoc ((η T) (S (T (S x))))).
   rewrite <- (nat_trans_ax (η T) (S (T (S x)))).
+  (* 2 *)
   simpl.
   rewrite !assoc.
   change ((η S) (T (S x)) · a (S x) · (η T) (T (S (S x))) · (μ T) (S (S x)) · # T ((μ S) x)
           = identity (T (S x))).
   rewrite (monad_dist_law1 l).
+  (* 3 *)
   rewrite <- (assoc (# T ((η S) (S x)))).
   change (# T ((η S) (S x)) · ((η T) (T (S (S x))) · (μ T) (S (S x))) · # T ((μ S) x)
           = identity (T (S x))).
   rewrite Monad_law1.
+  (* 4 *)
   rewrite id_right.
   rewrite <- functor_comp.
   rewrite <- functor_id.
@@ -82,11 +116,28 @@ Proof.
   now rewrite Monad_law1.
 Defined.
 
+(**
+The diagram for this proof (see above for explanation):
+TSTSx ---------------> TTSSx ----------->  TSSx  ------------> TSx
+  ^     #T (a (Sx))     ^     μ T (SSx)     ^     #T (μ S x)   ^
+  |                 1  /             2     /                  /
+  |#T #S (η T (Sx))   /                   /id                /
+  |                  /#T (η T (SSx))     /                  /
+  |  _______________/                   /                  /
+  | /                                  /                  /
+ TSSx  --------------------------------         3        /
+  ^                                                     /
+  |                                                    /
+  |#T #S (η S x)                                      /
+  |                                      id          /
+ TSx ------------------------------------------------
+*)
 Local Lemma monad_comp_law2 {a : T ∙ S ⟹ S ∙ T} (l : monad_dist_laws a) : ∏ x : C,
   #T (#S ((η S x) · (η T (S x)))) · (#T (a (S x)) · (μ T (S (S x)) · #T (μ S x))) =
   identity (T (S x)).
 Proof.
   intro x.
+  (* 1 *)
   rewrite !assoc.
   rewrite <- functor_comp.
   rewrite (functor_comp S).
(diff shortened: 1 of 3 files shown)


UniMath/CategoryTheory/Derivative.v
```diff
@@ -0,0 +1,519 @@
...
+
+Local Lemma monad_comp_law_2 {a : T ∙ S ⟹ S ∙ T} (l : monad_dist_laws a) : ∏ x : C,
+  #T (#S ((η S x) · (η T (S x)))) · (#T (a (S x)) · (μ T (S (S x)) · #T (μ S x))) =
+  identity (T (S x)).
+  intro x.
+  rewrite !assoc.
+  rewrite <- functor_comp.
+  rewrite (functor_comp S).
+  rewrite <- !assoc.
+  transitivity (# T (# S ((η S) x) · (η T (S (S x)))) · ((μ T) (S (S x)) · # T ((μ S) x))).
```
benediktahrens: Could you please use bullets (see 7.2.7  Bullets  of https://coq.inria.fr/refman/Reference-Manual009.html) to structure the proof script?

UniMath/CategoryTheory/Derivative.v
```diff
@@ -0,0 +1,519 @@
...
+  rewrite (assoc (# T ((η T) (S (S x))))).
+  rewrite Monad_law2.
+  rewrite id_left.
+  rewrite <- functor_comp.
+  rewrite <- functor_id.
+  apply (leib (fun arg => (#T arg))).
+  exact (@Monad_law2 C S x).
+Qed.
+
+Local Lemma monad_comp_law_3 {a : T ∙ S ⟹ S ∙ T} (l : monad_dist_laws a) : ∏ x : C,
```
benediktahrens: The proof of this equality is very long. Are there any intermediate equalities that are interesting in their own right, and that you could state and prove as separate lemmas?

UniMath/CategoryTheory/Derivative.v
```diff
@@ -0,0 +1,519 @@
...
+  apply tpair.
+  exact (monad_comp_law_1 l).
+  exact (monad_comp_law_2 l).
+  exact (monad_comp_law_3 l).
+Qed.
+
+Definition monad_comp {a : T ∙ S ⟹ S ∙ T} (l : monad_dist_laws a) : Monad C :=
+  (_,, monad_comp_laws l).
+
+(* morphism from first composand to of composition of monads *)
```
benediktahrens: Comment seems garbled.

UniMath/CategoryTheory/Derivative.v
```diff
@@ -0,0 +1,519 @@
...
+  rewrite <- assoc.
+  rewrite !(assoc ((η T) (S (T (S x))))).
+  rewrite <- (nat_trans_ax (η T) (S (T (S x)))).
+  simpl.
+  rewrite !assoc.
+  transitivity (#T (η S (S x)) · (η T) (T (S (S x))) · (μ T) (S (S x)) · # T ((μ S) x)).
+  apply (leib (fun arg => arg · (η T) (T (S (S x))) · (μ T) (S (S x)) · # T ((μ S) x))).
+  exact (monad_dist_law1 l (S x)).
+  rewrite <- (assoc (# T ((η S) (S x)))).
+  transitivity (# T ((η S) (S x)) · identity (T (S (S x))) · # T ((μ S) x)).
```
benediktahrens: Again, two subgoals are created here.

For documentation of bullets, see Section 7.2.7 on https://coq.inria.fr/refman/Reference-Manual009.html .
Also, take a look at other UniMath files.

UniMath/CategoryTheory/Derivative.v
```diff
@@ -0,0 +1,519 @@
...
+  Monad_Mor T (monad_comp l) := (_,, monad_to_comp_laws l).
+
+End comp_def.
+
+Section maybe_def.
+
+Context {C : precategory} (o : C) (co : BinCoproducts C).
+
+(* maybe monad *)
+Definition maybe_functor : functor C C :=
```
benediktahrens: This functor is already defined under the name of `option_functor` in `limits/bincoproducts.v`. 
Use that definition and remove the definition of `maybe_functor`.

UniMath/CategoryTheory/Derivative.v
```diff
@@ -0,0 +1,519 @@
...
+Definition maybe_eta : functor_identity C ⟹ maybe_functor :=
+  coproduct_nat_trans_in1 C C co (functor_identity C) (constant_functor C C o).
+
+Definition maybe_monad_data : Monad_data C := (maybe_functor,, maybe_mu),, maybe_eta.
+
+Local Lemma maybe_monad_law1 : ∏ c : C,
+BinCoproductIn1 C (co (co c o) o) ·
+BinCoproductArrow C _ (identity (co c o)) (BinCoproductIn2 C (co c o)) =
+identity (co c o).
+  intro c.
```
benediktahrens: Insert `Proof.`

UniMath/CategoryTheory/Derivative.v
```diff
@@ -0,0 +1,519 @@
...
+  rewrite BinCoproductIn1Commutes.
+  reflexivity.
+Qed.
+
+Local Lemma maybe_monad_law2 : ∏ c : C,
+BinCoproductOfArrows C (co c o) (co (co c o) o)
+ (BinCoproductIn1 C (co c o)) (identity o) ·
+BinCoproductArrow C _ (identity (co c o)) (BinCoproductIn2 C (co c o)) =
+identity (co c o).
+  intro c.
```
benediktahrens: Insert `Proof.`

UniMath/CategoryTheory/Derivative.v
```diff
@@ -0,0 +1,519 @@
...
+  rewrite id_right.
+  rewrite postcompWithBinCoproductArrow.
+  rewrite id_left.
+  rewrite id_left.
+  rewrite BinCoproductIn2Commutes.
+  reflexivity.
+Qed.
+
+Lemma maybe_monad_laws : Monad_laws maybe_monad_data.
+  intros.
```
benediktahrens: Insert `Proof.`

UniMath/CategoryTheory/Derivative.v
```diff
@@ -0,0 +1,519 @@
...
+  rewrite BinCoproductIn2Commutes.
+  rewrite precompWithBinCoproductArrow.
+  rewrite id_left.
+  rewrite id_right.
+  rewrite <- (nat_trans_ax (μ T) x).
+  simpl.
+  reflexivity.
+Qed.
+
+Local Lemma deriv_dist_law_4 (T : Monad C) : ∏ x : C,
```
benediktahrens: Could you spell out this equation in prose in a comment? The statement is illegible, and a more mathematical statement at least in comment would be very helpful.

UniMath/CategoryTheory/Derivative.v
```diff
@@ -0,0 +1,519 @@
...
+  rewrite assoc.
+  rewrite BinCoproductIn2Commutes.
+  rewrite postcompWithBinCoproductArrow.
+  rewrite id_left.
+  rewrite BinCoproductIn2Commutes.
+  reflexivity.
+Qed.
+
+Lemma deriv_dist_is_monad_dist (T : Monad C) : monad_dist_laws (deriv_dist T).
+  unfold monad_dist_laws.
```
benediktahrens: Insert `Proof.`

jojhelfer:
Hi Benedikt,

> This lemma already exists, and is called maponpaths. Please delete this
> copy.
>
Yes, I discovered this at some point. But as it turns out, I don't need it
any more anyway since Vladimir alerted me to the "change" tactic. More on
this below.

> Please insert Proof. at the beginning of the proof.
>
Done (and likewise for the other instances).

> +  transitivity (#T (η S (S x)) · (η T) (T (S (S x))) · (μ T) (S (S x)) · # T ((μ S) x)).
> +  apply (leib (fun arg => arg · (η T) (T (S (S x))) · (μ T) (S (S x)) · # T ((μ S) x))).
>
> Here, the goal splits into two subgoals (two equations). Use bullets -
> and indentation to mark the two subgoals.
>
I was actually able to get rid of these "transitivity" "leib" combos. The
reason I needed them was to get around the annoying fact that Coq wouldn't
let me do a "rewrite" since it didn't think the left hand side of the
equation to be rewritten was in the goal (though it visibly was) because of
hidden implicit arguments. However, in all these cases, I was able to just
use "change (entire-text-of-goal)", and then it worked. I am still quite
unhappy with this, as it results in very bulky and ugly code, and would be
very interested if there was some other solution.

> +
> +(* maybe monad *)
> +Definition maybe_functor : functor C C :=
>
> This functor is already defined under the name of option_functor in
> limits/bincoproducts.v.
> Use that definition and remove the definition of maybe_functor.
>
It seems to me to be a little different. The "option_functor" is only
coproduct-with-a-terminal-object whereas this is
coproduct-with-an-arbitrary-object. Of course, what people usually call the
Maybe monad is the one with underlying functor "option_functor" and not
"maybe_functor". Maybe I should use a different name... What do you think?


> +Local Lemma deriv_dist_law_4 (T : Monad C) : ∏ x : C,
>
> Could you spell out this equation in prose in a comment? The statement is
> illegible, and a more mathematical statement at least in comment would be
> very helpful.
> ------------------------------
>
>
> This is actually just the fourth clause of monad_dist_laws (∏ x : C, #S (a
x) · a (S x) · #T (μ S x) = μ S (T x) · a x) specialized to the case of
distributive law being defined. I have added comments explaining why I
spelled out these statements in such gruesome detail. I could still add an
extra explanatory comment to this one, if you think it's a good idea.

Thanks so much for taking the time to look at and comment on my code.

Best,
Joj

rmatthes:
coproduct with an arbitrary object is also available as constcoprod_functor1 - trace sorted_option_functor in Multisorted.v in SubstitutionSystems

benediktahrens:
The following way of "rewriting" might be more to your liking: 
```
diff --git a/UniMath/CategoryTheory/Derivative.v b/UniMath/CategoryTheory/Derivative.v
index 5b01e8f..989ade5 100644
--- a/UniMath/CategoryTheory/Derivative.v
+++ b/UniMath/CategoryTheory/Derivative.v
@@ -100,9 +100,10 @@ Proof.
   (* 2 *)
   simpl.
   rewrite !assoc.
-  change ((η S) (T (S x)) · a (S x) · (η T) (T (S (S x))) · (μ T) (S (S x)) · # T ((μ S) x)
-          = identity (T (S x))).
-  rewrite (monad_dist_law1 l).
+  etrans. (* "transitivity" with a not-yet-determined term in the middle *)
+    { do 3 apply cancel_postcomposition. (* isolating the subterm we want to rewrite *)
+      apply (monad_dist_law1 l). (* this tactic determines the term in the middle *)
+    }
   (* 3 *)
   rewrite <- (assoc (# T ((η S) (S x)))).
   change (# T ((η S) (S x)) · ((η T) (T (S (S x))) · (μ T) (S (S x))) · # T ((μ S) x)
```
I suggest you try these tactics and see how they act on the goal. Afterwards I am happy to explain more.

I have found rewriting in this style to be very robust, but it is also quite verbose, unfortunately.

jojhelfer:
> coproduct with an arbitrary object is also available as constcoprod_functor1 - trace sorted_option_functor in Multisorted.v in SubstitutionSystems

maybe_functor is just a nickname for `constcoprod_functor2 co o`:
`Definition maybe_functor : functor C C :=
  constcoprod_functor2 co o.
`
It's not necessary but it makes the code a little more readable. I assume that's okay?

> The following way of "rewriting" might be more to your liking

Yes, I like it very much! I changed all of them, except one, which I couldn't make any prettier than it already was.

mortberg:
Just so you know, I think we will soon be able to use a much better rewriting tactic as the Coq developers have merged SSReflect in trunk now. I used this for many years before switiching to UniMath and in my experience it is extremely robust and predictable. I hope it will make it into the next release of Coq so that we can stop these very verbose tricks.

mortberg:
About `maybe_functor`: maybe it's better to make it into a notation or Let? This way you don't have to manually unfold it.

vladimirias:
> On Jun 20, 2017, at 4:50 AM, jojhelfer <notifications@github.com> wrote:
> 
> > This functor is already defined under the name of option_functor in
> > limits/bincoproducts.v.
> > Use that definition and remove the definition of maybe_functor.
> >
> It seems to me to be a little different. The "option_functor" is only
> coproduct-with-a-terminal-object whereas this is
> coproduct-with-an-arbitrary-object. Of course, what people usually call the
> Maybe monad is the one with underlying functor "option_functor" and not
> "maybe_functor". Maybe I should use a different name... What do you think?

Let me express my opinion - if it works, and I’d surprised if it did not, we should allow arbitrary objects in option_functor.

Vladimir.

rmatthes:
My concern is more with consistent naming. [option_functor](https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/limits/bincoproducts.v#L1006) uses constcoprod_functor1, [sorted_option_functor](https://github.com/UniMath/UniMath/blob/master/UniMath/SubstitutionSystems/MonadsMultiSorted.v#L55) as well. Why is it now constcoprod_functor2?

UniMath/CategoryTheory/Derivative.v
```diff
@@ -0,0 +1,553 @@
+(** **********************************************************
+Contents:
+        - "maybe" monad (binary coproduct with a fixed object)
```
benediktahrens: Could you add the Coq identifier for that construction, as in
```
        - "maybe" monad (binary coproduct with a fixed object) [maybe_monad]
```
In the HTML doc, `[maybe_monad]` will be turned into a link to the definition, which is very convenient.

Could you do this for all the items in the list?

benediktahrens:
Looks good to me, modulo Ralph's comment [1], which I think should be addressed.
The choice is arbitrary, but since it has been made, we should stick to that choice in order to ease reuse of existing results.


[1] https://github.com/UniMath/UniMath/pull/717#issuecomment-309765404

UniMath/CategoryTheory/Derivative.v
jojhelfer: Done

jojhelfer:
> My concern is more with consistent naming. option_functor uses constcoprod_functor1, sorted_option_functor as well. Why is it now constcoprod_functor2?

I agree. I've changed it to
`Definition maybe_functor : functor C C :=
  constcoprod_functor1 co o.`

PR #720, opened 2017-05-27
Modules

AnthonyBordg:
Hi @vladimirias , I updated my file according to your recent PR, ie I re-wrote the proofs of all of the properties of the operations on monoidfun in terms of the monoidfun.

Also I updated the file according to the recent changes in terminalogy for categories.


DanGrayson:
I think it would be good for modules and homomorphisms, etc., to be defined in Algebra, and to put just the construction of the category into CategoryTheory.

benediktahrens:
I repeat what I have already said: the definition of a ring does not belong in CategoryTheory. You must split the file to separate out the part that does not use category theory.

Please try to address ALL the comments we give you on your proposed code.

AnthonyBordg:
@DanGrayson , Hi Dan, yes I agree, I will do it.

DanGrayson:
The name of this file should be added to `CategoryTheory/.package/files`.

DanGrayson:
Closing.  Reopen when it ready for reconsideration.

PR #721, opened 2017-05-27
Algebra of modules

AnthonyBordg:
The algebra part of my formalization of modules.


DanGrayson:
If you don't also add the name of this package to `Algebra/.package/files`, then it won't get compiled by `make`.

AnthonyBordg:
@DanGrayson ,done. 
Once it has been merged in UniMath, I will recompile my local UniMath library to make the category part of Modules and make a second pull request.

DanGrayson:
The test failed.  If you type `make` now you'll see that it doesn't work, because Algebra files cannot include CategoryTheory files.

AnthonyBordg:
Could someone merge this pull request ? 
I would like to make a second pull request with the categorical part of my file "Modules". 
Also, I have a formalization on algebras over a commutative ring (it relies on the algebra part of "Modules") that is waiting for a pull request.

DanGrayson:
I'd like to take a look at this pull request, but we've been on the highway all day, and will be tomorrow.

benediktahrens:
On 05/29/2017 10:57 PM, Anthony Bordg wrote:
> Could someone merge this pull request ?
> I would like to make a second pull request with the categorical part of
> my file "Modules".
> Also, I have a formalization on algebras over a commutative ring (it
> relies on the algebra part of "Modules") that is waiting for a pull request.

You can increase the speed with which PRs are being looked at by
submitting high-quality code from the beginning on.
The improvements we suggested were quite obvious, and you could have
seen the need for those improvements yourself.
By submitting code for review that is obviously not ready, you consume
our energy and slow us down.

AnthonyBordg:
@DanGrayson , thanks Dan to let me know.

@benediktahrens , I guess you are not talking about the mathematical quality of this code, only about some organizational issues (which are very important I know, and I understand perfectly that the owners have a heavy load of work) like moving things around and put the right definitions in the right files. 
If one is over aggressive in public, one should be careful and very specific. 
Next time I will be more careful about these important and time-consuming organizational issues.

benediktahrens:
On 05/30/2017 03:41 PM, Anthony Bordg wrote:
> 
> @benediktahrens <https://github.com/benediktahrens> , I guess you are
> not talking about the mathematical quality of this code, only about some
> organizational issues (which are very important I know, and I understand
> perfectly that the owners have a heavy load of work) like moving things
> around and put the right definitions in the right files.

I am talking about the problems that I have mentioned in previous
comments. I am not going to repeat them here. But yes, they were of
organizational nature.

Once the organizational issues are fixed, we can look at the mathematics.

UniMath/Algebra/Modules.v
```diff
@@ -0,0 +1,341 @@
...
+
+(** The set of endomorphisms of an abelian group is a ring *)
+
+Definition rngofendabgr (G : abgr) : rng :=
+   @rngpair (setwith2binopofendabgr G) (@isrngops_setofendabgr_op G).
+
+
+(** ** The definition of the small type of (left) R-modules over a ring R *)
+
+Definition module (R : rng) : UU := ∑ G, rngfun R (rngofendabgr G).
```
DanGrayson: The last time you presented a definition of module to us as a ring homomorphism to the ring of endomorphisms, I expressed the preference for a definition of module M over a ring R to be a binary operation R x M --> M satisfying various properties, because it seems more elementary and more in accord with textbook definitions.

I wonder if others share the preference.

UniMath/Algebra/Modules.v
```diff
@@ -0,0 +1,341 @@
...
+
+Definition pr1module {R : rng} (M : module R) : abgr := pr1 M.
+
+Coercion pr1module : module >-> abgr.
+
+Definition module_struct {R : rng} (G : abgr) : UU := rngfun R (rngofendabgr G).
+
+Definition pr2module {R : rng} (M : module R) : rngfun R (rngofendabgr (pr1module M)) := pr2 M.
```
DanGrayson: This definition could use `module_struct`.

UniMath/Algebra/Modules.v
```diff
@@ -0,0 +1,341 @@
...
+
+(** ** The definition of the small type of (left) R-modules over a ring R *)
+
+Definition module (R : rng) : UU := ∑ G, rngfun R (rngofendabgr G).
+
+Definition pr1module {R : rng} (M : module R) : abgr := pr1 M.
+
+Coercion pr1module : module >-> abgr.
+
+Definition module_struct {R : rng} (G : abgr) : UU := rngfun R (rngofendabgr G).
```
DanGrayson: R is not determined by G, so it probably should not be an implicit parameter.

UniMath/Algebra/Modules.v
DanGrayson: This definition could use `module_struct`.

UniMath/Algebra/Modules.v
```diff
@@ -0,0 +1,341 @@
...
+
+Definition pr2module {R : rng} (M : module R) : rngfun R (rngofendabgr (pr1module M)) := pr2 M.
+
+Coercion pr2module : module >-> rngfun.
```
DanGrayson: This will be a confusing coercion, for if M is an R-module and r:R ,then `M r` will denoted left multiplication, contrary to standard mathematical practice.

UniMath/Algebra/Modules.v
```diff
@@ -0,0 +1,341 @@
...
+
+(** The multiplication defined from a module *)
+
+Definition module_mult {R : rng} {M : @module R} : R × M -> M.
```
DanGrayson: `module_mult` would be more usable if its type were `R -> M -> M`.

DanGrayson: After defining `module_mult` it would be good to introduce a notation `r*m` in a scope with a name such as 'module'.

benediktahrens:
On 05/31/2017 04:49 PM, Daniel R. Grayson wrote:
> 
> 
> In UniMath/Algebra/Modules.v
> <https://github.com/UniMath/UniMath/pull/721#discussion_r119376727>:
> 
>> +
> +Definition isdistr_setofendabgr_op {G : abgr} : isdistr (@rngofendabgr_op1 G) (@rngofendabgr_op2 G) :=
> +   dirprodpair isldistr_setofendabgr_op isrdistr_setofendabgr_op.
> +
> +Definition isrngops_setofendabgr_op {G : abgr} : isrngops (@rngofendabgr_op1 G) (@rngofendabgr_op2 G) :=
> +   mk_isrngops isabgrop_rngofendabgr_op1 ismonoidop_rngofendabgr_op2 isdistr_setofendabgr_op.
> +
> +(** The set of endomorphisms of an abelian group is a ring *)
> +
> +Definition rngofendabgr (G : abgr) : rng :=
> +   @rngpair (setwith2binopofendabgr G) (@isrngops_setofendabgr_op G).
> +
> +
> +(** ** The definition of the small type of (left) R-modules over a ring R *)
> +
> +Definition module (R : rng) : UU := ∑ G, rngfun R (rngofendabgr G).
> 
> The last time you presented a definition of module to us as a ring
> homomorphism to the ring of endomorphisms, I expressed the preference
> for a definition of module M over a ring R to be a binary operation R x
> M --> M satisfying various properties, because it seems more elementary
> and more in accord with textbook definitions.
> 
> I wonder if others share the preference.

I do share this preference.

In general, I think that in formalization, giving the most elementary
definition is a good idea.

That being said, it would be good to have several constructor and
accessor functions that allow taking either point of view.

UniMath/Algebra/Modules.v
```diff
@@ -0,0 +1,341 @@
...
+   - exact (rigfun_to_unel_rigaddmonoid M).
+   - rewrite X.
+     reflexivity.
+Defined.
+
+
+(** (left) R-module homomorphism *)
+
+Definition ismodulefun {R : rng} {M N : module R} (f : M -> N) : UU :=
+   (isbinopfun f) × (∏ r : R, ∏ x : M, f (module_mult (r,,x)) = module_mult (r,,(f x))).
```
DanGrayson: This is an example where a notation for `module_mult` would improve appearance.

DanGrayson: Move `is_linear` so you can use it here.

DanGrayson:
Re: "Could someone merge this pull request please ?"

If part of your impatience is due to a desire to continue formalizing proofs, there's nothing preventing you from continuing, by the way.  Just make a new branch based on the branch used in the pull request, and work on that branch.

UniMath/Algebra/Modules.v
DanGrayson: Oh, and actually, I should have said the binary operation would be encoded as a function of type R -> M -> M.

UniMath/Algebra/Modules.v
AnthonyBordg: Ok. Indeed `Definition pr2module {R : rng} (M : module R) : module_struct (pr1module M) := pr2 M.`  is better.

UniMath/Algebra/Modules.v
AnthonyBordg: I agree.

UniMath/Algebra/Modules.v
AnthonyBordg: Yes.

UniMath/Algebra/Modules.v
AnthonyBordg: Ok, I will remove it. 
I will add an `Identity Coercion id_module_struct : module_struct >-> rngfun`

UniMath/Algebra/Modules.v
AnthonyBordg: Yes, indeed. I will rename module_mult to module_mult_uncurry, and I will introduce `module_mult` of type `R -> M -> M`

AnthonyBordg: Ok.

UniMath/Algebra/Modules.v
AnthonyBordg: I agree.

AnthonyBordg: yes.

UniMath/Algebra/Modules.v
```diff
@@ -0,0 +1,341 @@
...
+  isunitalpair setofendabgr_un0 isunit_setofendabgr_un0.
+
+Definition ismonoidop_rngofendabgr_op1 {G : abgr} : ismonoidop (@rngofendabgr_op1 G) :=
+   mk_ismonoidop isassoc_rngofendabgr_op1 isunital_rngofendabgr_op1.
+
+Local Close Scope abgr_scope.
+
+(** rngofendabgr_op1 is a group operation *)
+
+(* The definition below should be moved to Monoids_and_Groups *)
```
AnthonyBordg: I forgot to move it. It will be done in my next commit.

benediktahrens:
On 05/31/2017 06:20 PM, Anthony Bordg wrote:
> *@AnthonyBordg* commented on this pull request.
> 
> 
> In UniMath/Algebra/Modules.v
> <https://github.com/UniMath/UniMath/pull/721#discussion_r119404688>:
> 
>> +
> +Definition module (R : rng) : UU := ∑ G, rngfun R (rngofendabgr G).
> +
> +Definition pr1module {R : rng} (M : module R) : abgr := pr1 M.
> +
> +Coercion pr1module : module >-> abgr.
> +
> +Definition module_struct {R : rng} (G : abgr) : UU := rngfun R (rngofendabgr G).
> +
> +Definition pr2module {R : rng} (M : module R) : rngfun R (rngofendabgr (pr1module M)) := pr2 M.
> +
> +Coercion pr2module : module >-> rngfun.
> +
> +(** The multiplication defined from a module *)
> +
> +Definition module_mult {R : rng} {M : @module R} : R × M -> M.
> 
> Ok.

There is no need in general to say "Ok" when you agree.

UniMath/Algebra/Modules.v
DanGrayson: There is no reason to have two functions, one of type RxM->M, and the other of type R->M->M.

vladimirias:
Agree.

vladimirias:
Agree.

vladimirias:
I would agree, but the point of view R -> End(M) has a number of interesting advantages. In particular it is easy to make R -> M -> M from R -> End(M) but not so easy in the opposite direction. I am unsure.

AnthonyBordg:
@DanGrayson , is it possible to make a pull request on a file (my file on algebras over a commutative rings) that relies on a previous file that is still under review (the algebra of modules actually) ? I can download and install my branch "Algebra of modules" in to order to be able to "require import" the algebra of modules in my file on algebras over a commutative ring, but I'm unsure if this is something reasonable to do.

DanGrayson:
There is no problem with making a pull request that depends on a previous pull request.  You do that by making a new branch based on the branch of the first pull request, and then you submit a pull request for the second branch.  Just say in the pull request that it should not be reviewed until the other pull request has been granted.

Then, suppose you make an update to the first branch, in response to criticism of the first pull request.  Merge the first branch into the second.

On second thought, it would be confusing for us to receive a pull request that tells us not to consider the request yet.  It would be better to just hold the second pull request in abeyance until the first one has been granted.  What's the rush?  Just make a new branch as described above and continue development privately.

UniMath/Algebra/Modules.v
```diff
@@ -0,0 +1,356 @@
...
+
+Definition modulefun_to_binopfun {R : rng} {M N : module R} (f : modulefun M N) : binopfun M N :=
+  binopfunpair (pr1modulefun f) (modulefun_to_isbinopfun f).
+
+Definition modulefun_to_islinear {R : rng} {M N : module R} (f : modulefun M N): islinear (pr1modulefun f) := pr2 (pr2 f).
+
+Definition modulefun_to_linearfun {R : rng} {M N : module R} (f : modulefun M N) : linearfun M N :=
+  linearfunpair f (modulefun_to_islinear f).
+
+Definition modulefun_unel {R : rng} {M N : module R} (f : @modulefun R M N) : f (@unel M) = @unel N.
```
benediktahrens: Should be
```
Definition modulefun_unel {R : rng} {M N : module R} (f : modulefun M N) : f (unel M) = unel N.
```
instead.

UniMath/Algebra/Modules.v
```diff
@@ -0,0 +1,356 @@
...
+Lemma isapropismodulefun {R : rng} {M N : module R} (f : M -> N) : isaprop (ismodulefun f).
+Proof.
+   refine (@isofhleveldirprod 1 (isbinopfun f) (islinear f) _ _).
+   exact (isapropisbinopfun f).
+   apply (impred 1 _). intro r.
+   apply (impred 1 _). intro x.
+   apply (setproperty N).
+Defined.
+
+Definition modulefun {R : rng} (M N : module R) := total2 (λ f : M -> N, ismodulefun f).
```
benediktahrens: Type annotation for result is missing.

UniMath/Algebra/Modules.v
```diff
@@ -0,0 +1,356 @@
...
+
+Definition modulefunpair {R : rng} {M N : module R} (f : M -> N) (is : ismodulefun f) :=
```
benediktahrens: Type annotation for result is missing.

benediktahrens:
Maybe we can agree that the actual definition is not very important as
long as there are suitable functions building such structures, and
functions from such structures?

Maybe @AnthonyBordg can add a function to build a module from a
multiplication satisfying suitable data.

The one thing that will depend on the precise definition of module is
the proof of univalence of the category of modules.
Is it correct that @AnthonyBordg already has such a proof?

UniMath/Algebra/Modules.v
DanGrayson: I wouldn't insist on that - we often define types this way.  But here some unicode notation would look nicer, instead of `total2`.

AnthonyBordg:
@benediktahrens , yes it is a good idea, I will try to add a function to build a module from suitable data, and yes it is correct that I have a proof of the univalence property.

UniMath/Algebra/Modules.v
```diff
@@ -0,0 +1,409 @@
...
+      use total2_paths2_f.
+      apply funextfun. intro x. apply ax2.
+      apply isapropismonoidfun.
+    + use total2_paths2_f.
+      apply funextfun. intro x. change (m rngunel1 x = unel G). apply (grlcan G (m (rngunel1) x)). rewrite runax.
+      rewrite <- (ax2 rngunel1 rngunel1 x). rewrite rngrunax1. apply idpath.
+      apply isapropismonoidfun.
+  -  apply dirprodpair.
+     + intros r s.
+       use total2_paths2_f.
```
benediktahrens: After `total2_paths2_f`, please use bullets. Same a few lines down.

AnthonyBordg:
@benediktahrens : done.

benediktahrens:
@AnthonyBordg : I have merged your PR on modules to avoid further delay. However, the file could use some more comments and a table of contents in a comment at the beginning of the file. Please add these in a subsequent PR.

AnthonyBordg:
@benediktahrens , yes I will add a table of contents in a subsequent PR (I will make a PR soon to add a definition of submodules when my file on groups with operators has been reviewed (my def of submodules uses stable subgroups, but again I could add a second, more common, access function)). Regarding the comments, yes more comments would be useful. I don't know what are the views of the owners but if the library is intended at some point in the future as a useful ressource for the undergraduate/graduate student, not only for the working mathematician, then various and useful comments would be a very nice feature. With this respect, having an undergraduate student hanging around  and trying to read the material, asking questions, could be useful to provide insights where comments would be welcome (since of course for the people behind the formalizations everything is always crystal clear and they may have a hard time to see where comments would be helpful).

benediktahrens:
For the time being, it would be useful to insert section titles using
```
(** * Heading 1 *)

(** ** Heading 2 *)

```

AnthonyBordg:
@benediktahrens , I think `(** ** -----------)` is used for a subsection, not a second section. Am I wrong ? Hence, do you suggest to use one section and a few subsections instead of a few sections (with possibly no subsection) ?

benediktahrens:
On 07/01/2017 07:48 PM, Anthony Bordg wrote:
> @benediktahrens <https://github.com/benediktahrens> , I think |(** **
> -----------)| is used for a subsection, not a second section. Am I wrong
> ? 

No, you are correct.

Hence, do you suggest to use one section and a few subsections instead
> of a few sections (with possibly no subsection) ?

I just wanted to illustrate how to do first and second level section
headers.
You should place those section (and maybe subsection) headers as you prefer.

PR #729, opened 2017-06-05
Types and groups with operators

AnthonyBordg:
This PR contains :

- The definition of (arbitrary) types with operators.

- The definiton of groups with operators.

- The definition of stable subgroups of a group with operators.

### DIFF
--- a/UniMath/Algebra/Types_and_groups_with_operators.v
+++ b/UniMath/Algebra/Types_and_groups_with_operators.v
@@ -1,0 +1,249 @@
+(** Anthony Bordg, June 2017 *)
+
+
+Require Import UniMath.Algebra.Monoids_and_Groups.
+
+(** * Types with operators *)
+
+(** Types with a binary operation *)
+
+Definition typewithbinop : UU := ∑ X : UU, binop X.
+
+Definition pr1typewithbinop (X : typewithbinop) : UU := pr1 X.
+
+Coercion pr1typewithbinop : typewithbinop >-> UU.
+
+Definition op {X : typewithbinop} : binop X := pr2 X.
+
+Definition ishbinopfun {X Y : typewithbinop} (f : X -> Y) : UU :=
+  ∏ x x' : X, paths (f (op x x')) (op (f x) (f x')).
+
+Definition typewithbinopfun (X Y : typewithbinop) : UU := ∑ f : X -> Y, ishbinopfun f.
+
+Definition pr1typewithbinopfun {X Y : typewithbinop} (f : typewithbinopfun X Y) : X -> Y := pr1 f.
+
+Coercion pr1typewithbinopfun : typewithbinopfun >-> Funclass.
+
+(** Types with a left action *)
+
+Definition laction (Ω X : UU) : UU := Ω -> X -> X.
+
+Definition typewithlaction (Ω : UU) : UU := ∑ X : UU, laction Ω X.
+
+Definition typewithlaction_pair {Ω : UU} (X : UU) (ω : laction Ω X) : typewithlaction Ω := tpair _ X ω.
+
+Definition pr1typewithlaction {Ω : UU} (X : typewithlaction Ω) : UU := pr1 X.
+
+Coercion pr1typewithlaction : typewithlaction >-> UU.
+
+Definition pr2typewithlaction {Ω : UU} (X : typewithlaction Ω) : laction Ω X := pr2 X.
+
+Local Notation " α · x " := (pr2typewithlaction _ α x) : typewithlaction_scope.
+
+Delimit Scope typewithlaction_scope with typewithlaction.
+
+Local Open Scope typewithlaction.
+
+(** Types with a binary operation and a left action *)
+
+Definition typewith_binop_laction (Ω : UU) : UU := ∑ X : typewithbinop, laction Ω X.
+
+Definition pr1typewith_binop_laction {Ω : UU} (X : typewith_binop_laction Ω) : typewithbinop := pr1 X.
+
+Coercion pr1typewith_binop_laction : typewith_binop_laction >-> typewithbinop.
+
+Definition typewith_binop_laction_to_binop {Ω : UU} (X : typewith_binop_laction Ω) : binop X := pr2 (pr1 X).
+
+Definition pr2typewith_binop_laction {Ω : UU} (X : typewith_binop_laction Ω) : laction Ω X := pr2 X.
+
+(** Types with operators *)
+
+Definition ishdistr_laction {Ω X : UU} (opp : binop X) (ω : laction Ω X) : UU :=
+  ∏ α : Ω, ∏ x y : X, ω α (opp x y) = opp (ω α x) (ω α y).
+
+Definition typewithoperators (Ω : UU) : UU :=
+  ∑ X : typewith_binop_laction Ω, ishdistr_laction (typewith_binop_laction_to_binop X) (pr2typewith_binop_laction X).
+
+Definition pr1typewithoperators {Ω : UU} (X : typewithoperators Ω) : typewith_binop_laction Ω := pr1 X.
+
+Coercion pr1typewithoperators : typewithoperators >-> typewith_binop_laction.
+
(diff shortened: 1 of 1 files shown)


UniMath/Algebra/Types_and_groups_with_operators.v
```diff
@@ -0,0 +1,251 @@
...
+Definition typewithbinop : UU := ∑ X : UU, binop X.
+
+Definition pr1typewithbinop (X : typewithbinop) : UU := pr1 X.
+
+Coercion pr1typewithbinop : typewithbinop >-> UU.
+
+Definition op {X : typewithbinop} : binop X := pr2 X.
+
+Definition ishbinopfun {X Y : typewithbinop} (f : X -> Y) : UU :=
+  ∏ x x' : X, paths (f (op x x')) (op (f x) (f x')).
```
benediktahrens: We have had this very convenient notation `a = b` for `paths a b`. I strongly encourage you to use this notation.

UniMath/Algebra/Types_and_groups_with_operators.v
```diff
@@ -0,0 +1,251 @@
...
+
+Definition typewithbinopfun (X Y : typewithbinop) : UU := ∑ f : X -> Y, ishbinopfun f.
+
+Definition pr1typewithbinopfun {X Y : typewithbinop} (f : typewithbinopfun X Y) : X -> Y := pr1 f.
+
+Coercion pr1typewithbinopfun : typewithbinopfun >-> Funclass.
+
+(** Types with a left action *)
+
+Definition laction (Ω X : UU) : UU := Ω -> X -> X.
```
benediktahrens: I am a bit puzzled by this definition. An action usually involves a group. Do you really think that this definition and the subsequent ones are useful?

AnthonyBordg: @benediktahrens , here indeed there is a slight departure from the classical definition. The motivation for such a departure arose from  earlier comments from @vladimirias on modules. One could write most of the algebra in UniMath to apply to types with operations, not only to sets with operations, and then one could get more specific definitions by specialization of these operations to sets for instance. This is the point of view chosen here.

Regarding your first comment, I'm following Bourbaki (with the modification motivated above). Groups with operators have many applications. In the short term I used them to formalize submodules, a submodule being a stable subgroup of a module (seen as a group with operators given by the module multiplication). Then with the notion of a submodule I formalized the notion of a subalgebra.

DanGrayson: If you're following Bourbaki, it would be good to give precise references.  For example, for "action" refer to Definition 1, Section 3.1, Algebra I, Bourbaki.

DanGrayson: "laction" isn't a word.  What about "Action" instead?

AnthonyBordg: @DanGrayson , yes you are right, "Action" might be better. I will add comments and references.

AnthonyBordg: @DanGrayson , it is not clear for me if there are definitive rules in UniMath to achieve uniformity of identifiers. Here do you suggest to use "Action" instead of "action" ?

AnthonyBordg:
@benediktahrens , @DanGrayson , I improved the comments and I added a table of contents.

UniMath/Algebra/Types_and_groups_with_operators.v
```diff
@@ -0,0 +1,266 @@
...
+***********************************************)
+
+
+Require Import UniMath.Algebra.Monoids_and_Groups.
+
+(** * Types with operators *)
+
+(** Types with a binary operation *)
+
+Definition typewithbinop : UU := ∑ X : UU, binop X.
```
siddharthist: This could go in `BinaryOperations.v`

siddharthist:
A group with operators can (evidently) be defined as a monoid homomorphism `M -> End(G)` [1]. It seems that this is what is claimed by `ismonoidfun_homothety`. I'm hoping that #796 will unify the following three (currently totally unrelated) definitions, so that they can be studied from a more uniform interface:
 1. The ones in this pull request, namely, groups with operators
 2. The definition of a group action on a set in `Ktheory/GroupAction.v`
 3. The definition of a module as a ring action on an abelian group

[1] Mac Lane, Categories, pp. 41. This is just a small remark.

siddharthist:
@AnthonyBordg See #316 for a discussion of how citation via comments is supposed to work.

DanGrayson:
I wonder what we're waiting for in this PR.  The issue #316 seems to have not resulted in a standard, which would have been recorded in our style guide UniMath/README.md, and anyway, trying to build the latex documentation file results in an error message, anyway.  We do need a workable standard for academic citations, so perhaps discussion should continue in #316.

@benediktahrens - Benedikt, did he add enough comments for you?

UniMath/Algebra/Types_and_groups_with_operators.v
```diff
@@ -0,0 +1,266 @@
...
+
+Definition pr1typewithbinop (X : typewithbinop) : UU := pr1 X.
+
+Coercion pr1typewithbinop : typewithbinop >-> UU.
+
+Definition op {X : typewithbinop} : binop X := pr2 X.
+
+Definition ishbinopfun {X Y : typewithbinop} (f : X -> Y) : UU :=
```
DanGrayson: What does the "h" in "ishbinopfun" denote?

UniMath/Algebra/Types_and_groups_with_operators.v
```diff
@@ -0,0 +1,266 @@
...
+  ∏ x x' : X, f (op x x') = op (f x) (f x').
+
+Definition typewithbinopfun (X Y : typewithbinop) : UU := ∑ f : X -> Y, ishbinopfun f.
```
DanGrayson: Compare:
```
Definition typewithbinop : UU := ∑ X : UU, binop X.
...
Definition typewithbinopfun (X Y : typewithbinop) : UU := ∑ f : X -> Y, ishbinopfun f.
```
We see that an element of `typewithbinop` is a type with some data, but `typewithbinopfun` is not.
Maybe using "typewith" as part of the name `typewithbinopfun` is capable of improvement.

UniMath/Algebra/Types_and_groups_with_operators.v
```diff
@@ -0,0 +1,266 @@
...
+
+Coercion pr1typewith_binop_action : typewith_binop_action >-> typewithbinop.
+
+Definition typewith_binop_action_to_binop {Ω : UU} (X : typewith_binop_action Ω) : binop X := pr2 (pr1 X).
+
+Definition pr2typewith_binop_action {Ω : UU} (X : typewith_binop_action Ω) : action Ω X := pr2 X.
+
+(** Types with operators *)
+
+Definition ishdistr_action {Ω X : UU} (opp : binop X) (ω : action Ω X) : UU :=
```
DanGrayson: What does the "h" in "ishdistr_action" denote?

peterlefanulumsdaine:
I’ve brought this PR up to date with current master.  I’m not sure what to suggest we do with it:

- its mathematical motivation is unclear (which could be fixed by more expository comments, as requested in review at the time)
- it has some duplication with the material of #796 on actions, which should be unified
- several minor improvements suggested in reviews, which would be good to fix

Just in case @AnthonyBordg or anyone else would still be interested in fixing these issues (especially documenting the mathematical motivation more clearly), I think this would be good to merge.  If no-one is motivated to fix these, though, I suggest closing?

(This is a similar situation to #1221 , but with less clear motivation for the material.  It could reasonably either left as “closed (with possibility of reopening if someone is motivated to fix the issues)” or “open (but dormant until someone steps up to fix the issues)”.)

benediktahrens:
I agree with @peterlefanulumsdaine 's list of options. Since there are mostly definitions, and only a few results, closing the PR would be my preferred option.

peterlefanulumsdaine:
Closing this for now, then, and it can be reopened in future if anyone is inspired to champion it.

PR #730, opened 2017-06-09
Upstream code on category of elements and presheaves

mortberg:
This PR moves some code upstream:

- Various things on the category of elements
- Basic lemma about transport and presheaves
- Remove code that was duplicated in `UniMath/CategoryTheory/elems_slice_equiv.v` and use PreShv notation.
- Make parts of `functor_data_eq` transparent (I needed this somewhere else)

This PR also adds a very simple, but sometimes useful, lemma to Foundations. Maybe I should put this in MoreFoundations instead?


UniMath/CategoryTheory/functor_categories.v
```diff
@@ -83,31 +83,39 @@ Proof.
   apply hs.
 Qed.
 
+Lemma functor_data_eq_prf {C C' : precategory_ob_mor} (F F' : functor_data C C')
+      (H : ∏ (c : C), (pr1 F) c = (pr1 F') c)
```
benediktahrens: Are those projections necessary?

UniMath/Foundations/PartA.v
```diff
@@ -844,6 +844,12 @@ Proof.
   intros. now apply maponpaths.
 Defined.
 
+Lemma base_paths_pair_path_in2 {X : UU} (P : X → UU) {x : X} {p q : P x} (e : p = q) :
```
benediktahrens: There is a ban on modifying the files in Foundations, I think. Alternatively, this lemma can go into MoreFoundations.

UniMath/CategoryTheory/ElementsOp.v
```diff
@@ -102,6 +104,69 @@ Definition make_ob (c : C) (x : X c : hSet) : ∫ X := (c,,x).
...
 
+(* Any f : J → I and ρ : X I defines a morphism from (J,,# X ρ) to (I,,ρ) in ∫ X *)
+Definition mor_to_el_mor {I J : C} (f : J --> I) (ρ : pr1 X I : hSet) :
+  ∫ X ⟦ make_ob J (# (pr1 X) f ρ), make_ob I ρ ⟧ :=
+  make_mor (J,,# (pr1 X) f ρ) (I,,ρ) f (idpath (# (pr1 X) f ρ)).
+
+Lemma make_ob_eq {I : C} {x y} (e : x = y) :
+  make_ob I x = make_ob I y.
+Proof.
+apply pair_path_in2, e.
```
benediktahrens: Is this not simply `maponpaths`?
I think you could do without this lemma.

I also suspect that the lemmas below are not necessary.

UniMath/CategoryTheory/functor_categories.v
mortberg: Yeah, they were because the definition appeared before the coercion were defined. I'll move it down and remove some projections.

UniMath/CategoryTheory/ElementsOp.v
mortberg: Yes, this is defined using `maponpaths`. I'll see if I can remove it, this way my addition to Foundations can also be removed.

mortberg:
I removed `make_ob_eq` now which made it possible to remove the lemma I had added to Foundations. 

@benediktahrens : do you think I can remove even more lemmas in `ElementsOp.v`? Some of the equalities there are very easily proved, but it is nice to have them explicitly so that I can rewrite with them.

benediktahrens:
On 06/09/2017 03:35 PM, Anders Mörtberg wrote:
> I removed |make_ob_eq| now which made it possible to remove the lemma I
> had added to Foundations.
> 
> @benediktahrens <https://github.com/benediktahrens> : do you think I can
> remove even more lemmas in |ElementsOp.v|? Some of the equalities there
> are very easily proved, but it is nice to have them explicitly so that I
> can rewrite with them.


Lemma `transportf_make_ob_eq` is only about pairs, and you should be
able to write a more general lemma. This in turn is likely already
available somewhere.

Same for `transportf_make_ob`.

In the end, it is your choice how many auxiliary lemmas you put in - but
you might want to try to do without the aforementioned lemmas.

mortberg:
Both of these are special cases of `functtransportf` and rewriting with it right-to-left proves both of them. However if I try to solve the goals later on where I use these Coq picks the wrong predicate in the rewrite which screws up the goal, giving the predicate by hand is a pain so I think having a special lemma for `transportf_make_ob_eq` makes sense.

vladimirias:
>> In UniMath/Foundations/PartA.v <https://github.com/UniMath/UniMath/pull/730#discussion_r121114110>:
>> 
>> > @@ -844,6 +844,12 @@ Proof.
>>    intros. now apply maponpaths.
>>  Defined.
>> 
>> +Lemma base_paths_pair_path_in2 {X : UU} (P : X → UU) {x : X} {p q : P x} (e : p = q) :
>> There is a ban on modifying the files in Foundations, I think. Alternatively, this lemma can go into MoreFoundations.
>> 
>> 

Yes, there is a ban and this lemma should go into MoreFoundations. That is what MoreFoundations are for.

Vladimir.

mortberg:
I created a filed called "Various.v" in MoreFoundations where I put this result. I also upstreamed some more code to this file.

benediktahrens:
Could you find a more descriptive name than "Various"?

Am 11. Juni 2017 10:21:44 MESZ schrieb "Anders Mörtberg" <notifications@github.com>:
>I created a filed called "Various.v" in MoreFoundations where I put
>this result. I also upstreamed some more code to this file.

mortberg:
How about "FoundationsExtras"? There are various other results that I want to put there, like what is in the beginning of https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/limits/graphs/eqdiag.v. For example the lemma https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/limits/graphs/eqdiag.v#L84 is something I often want to use.

vladimirias:
Foundations2 is also a possibility.

DanGrayson:
"FoundationsExtras" is a synonym of "MoreFoundations", so that won't work.  The definition of `transportf` is in `Foundations/PartA`, and your lemma is about that, so why not name the file `MoreFoundations/PartA.v`?

DanGrayson:
"Foundations2" is also a synonym of "MoreFoundations".

mortberg:
I like Dan's idea of having results that could potentially be moved to Foundations/PartA.v in MoreFoundations/PartA.v (and similarily for the other file names). This way it will be easier to maybe move things even further upstream one day. A potential user who imports MoreFoundations/All.v will not see any difference as all of the results from Foundations and MoreFoundations are exported by this file.

UniMath/MoreFoundations/Various.v
```diff
@@ -0,0 +1,103 @@
...
+Qed.
+
+End various.
+
+(** Various algebraic properties of hProp *)
+Section hProp_logic.
+
+Lemma isassoc_hconj (P Q R : hProp) : ((P ∧ Q) ∧ R) = (P ∧ (Q ∧ R)).
+Proof.
+apply hPropUnivalence.
```
benediktahrens: I think it would be useful to split all these lemmas in two lemmas each: one lemma proving a biimplication, without using univalence for propositions, and one lemma proving equality, using univalence and the previously proved biimplication (`weqimplimpl`, I believe, is the lemma needed).

mortberg: I totally agree. I actually found this https://github.com/UniMath/UniMath/blob/master/UniMath/Foundations/Propositions.v#L453 today. The comment seems to promise a lot more than what is actually there. I'll make a file called MoreFoundations/Propositions.v where I'll prove all of these as biimplication and then as equalities.

DanGrayson: Why would you ever need an equality of propositions.  I recommend settling for the implications.

mortberg:
I'll close this for now and then reopen once I have fixed everything to reduce the amount of noise in your mailboxes.

UniMath/MoreFoundations/Various.v
mortberg: The reason I proved these as equalities is that I wanted to show that hProp is a bounded lattice (and the lattice equations are stated using equality). I think it makes sense to have both versions. Another benefit of equalities is that you can rewrite with them.

mortberg:
I proved all of the bi-implications now and the equalities are now proved directly from these.

UniMath/CategoryTheory/Presheaf.v
```diff
@@ -50,121 +48,22 @@ Require Import UniMath.CategoryTheory.LatticeObject.
...
-Proof.
-use mkbounded_lattice.
-- exact hProp_lattice.
-- exact hfalse.
-- exact htrue.
-- split.
-  + intros P; apply hfalse_hdisj.
-  + intros P; apply htrue_hconj.
-Defined.
+(* TODO: upstream *)
```
benediktahrens: There is an empty section and a TODO remark left.

UniMath/MoreFoundations/PartA.v
```diff
@@ -0,0 +1,22 @@
+(** This file contain various results that could be upstreamed to Foundations/PartA.v *)
+Require Import UniMath.MoreFoundations.Foundations.
+
+Lemma base_paths_pair_path_in2 {X : UU} (P : X → UU) {x : X} {p q : P x} (e : p = q) :
+  base_paths _ _ (pair_path_in2 P e) = idpath x.
+Proof.
+now induction e.
+Qed.
+
+(* stolen from TypeTheory/Display_Cats/Auxiliary.v *)
```
benediktahrens: "stolen"? I'd say "taken"...

UniMath/CategoryTheory/Presheaf.v
mortberg: Thanks, fixed it now.

benediktahrens:
lgtm

UniMath/MoreFoundations/PartA.v
mortberg: I don't know who wrote that comment. I just upstreamed the code with the comment as it was. I'll change it as you suggest.

Issue #733, opened 2017-06-18
Contents of category_hset.v


AnthonyBordg:
Why is it written "Colimits in HSET" in the Contents on top of the category_hset.v file while there is nothing in this file about colimits in HSET ?

mortberg:
I think the code on colimits in HSET was in that file at some point, we then moved it to `category_hset_structures.v` but forgot to remove the comment.

AnthonyBordg:
Thanks Anders, I will have a look at `category_hset_structures.v`.

mortberg:
I think we should either remove this comment, or merge `category_hset.v` with `category_hset_structures.v`. We discussed this before at some point and I'm in favour of merging as it makes it easier to find results. The CT folder has 63 files at the toplevel now making it quite hard to get an overview of what is where. What does @benediktahrens think?

benediktahrens:
On 06/20/2017 05:21 PM, Anders Mörtberg wrote:
> I think we should either remove this comment, or merge |category_hset.v|
> with |category_hset_structures.v|. We discussed this before at some
> point and I'm in favour of merging as it makes it easier to find
> results. The CT folder has 63 files at the toplevel now making it quite
> hard to get an overview of what is where. What does @benediktahrens
> <https://github.com/benediktahrens> think?

I think the same as last time we discussed this issue: it is not
feasible to have the category of sets depend on *every* concept we want
to consider.

The fundamental property of the category of sets is that it is
univalent. This should be in the first file.

Any other structure on the category of sets should go into a subsequent
file.

In short: I suggest to remain with the status quo, and to improve the
documentation instead:
- remove the misleading comment in category_hset.v
- insert a pointer to category_hset_structures.v

A PR to that end would be very welcome!

PR #734, opened 2017-06-21
Better documentation for category HSET

mortberg:
This resolves https://github.com/UniMath/UniMath/issues/733


mortberg:
Maybe we should also move the files on HSET to the subfolder `categories` that contains other examples of categories?

benediktahrens:
On 06/21/2017 09:25 AM, Anders Mörtberg wrote:
> Maybe we should also move the files on HSET to the subfolder
> |categories| that contains other examples of categories?

Yes, that is a good idea!

mortberg:
I moved the files the other day, so this PR should be ready for merge unless there are more comments.

PR #736, opened 2017-06-29
The univalent category of modules over a ring

AnthonyBordg:
This PR contains:

- The precategory of modules over a ring
- The univalent category of modules over a ring


UniMath/CategoryTheory/.package/files
```diff
@@ -116,3 +116,4 @@ categories/intdoms.v
 categories/flds.v
 categories/StandardCategories.v
 Derivative.v
+Modules.v
```
benediktahrens: Here it should be
```
categories/Modules.v
```
instead of 
```
Modules.v
```

mortberg: Isn't the Makefile supposed to complain about this now?

AnthonyBordg:
@benediktahrens : error in .package/files fixed.

benediktahrens:
Your use of sections is strange.
The purpose of sections (better: one of the purposes) is to locally declare section variables, as in
```
Section fix_a_category.
Variable C : category.
...
End fix_a_category.
```
After `End fix_a_category.` any result proved within the section about the category C will be available universally quantified over categories.

You do not introduce any section variables in your sections. The sections hence have no purpose.

On the other hand, you introduce a variable outside any section. I would prefer you wrapped all that part into a section.

mortberg:
I usually put related results in a Section, even if I don't introduce any variables. I think it is a nice way to organize the files, especially if one puts a comment that gets displayed as a header in the documentation before the section starts. Also, I often realize after I'm done developing the section that I could abstract over some argument that appears everywhere so in the end the Section actually ends up having some variables even if it didn't have any from the beginning. However introducing variables outside of a section seems like bad practice and should be avoided.

Btw, I always use `Context` instead of `Variables` these days. The reason is that one can make the variables defined using `Context` implicit by `Context {foo : T}`.

AnthonyBordg:
@mortberg , according to COQ reference manual a Variable declaration  outside a section is equivalent to a Local Parameter declaration. Why does it seem "bad practice" to you ?

mortberg:
I find it confusing to introduce variables/parameters in the middle of the file. If you open a section you put all the variables in the top and you know that they will be there until you close the section. If you just put a variable declaration in the middle of the file things can get quite confusing later on, especially if the code is very long and written by someone else... So I think it is bad practice from the point of view of long term code maintenance.

AnthonyBordg:
@benediktahrens , @mortberg , thanks for the feedback and the explanations. The required changes have been done.

UniMath/CategoryTheory/categories/Modules.v
```diff
@@ -0,0 +1,408 @@
...
+
+End category_modules.
+
+
+(** *** The univalent category of (left) R-modules and R-modules homomorphisms *)
+
+Section univalent_category_modules.
+
+Variable R : rng.
+Notation "R-mod" := (precategory_module R).
```
DanGrayson: The category of modules over a ring is important enough that one could justify renaming `precategory_module` to `Mod`, and then the notation `R-mod` would not be needed.

UniMath/CategoryTheory/categories/Modules.v
```diff
@@ -0,0 +1,408 @@
...
+
+Definition moduleiso_to_modulefun (M N : R-mod) : moduleiso M N -> modulefun M N.
+Proof.
+   intro f.
+   exact (tpair _ (pr1weq (pr1 f)) (pr2 f)).
+Defined.
+
+Coercion moduleiso_to_modulefun : moduleiso >-> modulefun.
+
+Definition pr1moduleiso {M N : R-mod} (f : moduleiso M N) : weq (pr1module M) (pr1module N) := pr1 f.
```
DanGrayson: How about using the notation for `weq` here?

UniMath/CategoryTheory/categories/Modules.v
```diff
@@ -0,0 +1,408 @@
...
+Lemma isaset_rngfun (X Y : rng) : isaset (rngfun X Y).
+Proof.
+   apply (isofhleveltotal2 2).
+   - use impred_isaset. intro x.
+     apply isaset_el_of_setwith2binop.
+   - intro f.
+     apply (isasetaprop (isaprop_isrigfun f)).
+Defined.
+
+Definition modules_univalence_weq1 (M N : R-mod) : (M = N) ≃ (M ╝ N) := total2_paths_equiv _ M N.
```
DanGrayson: Why bother with this definition?  It doesn't shorten anything.

UniMath/CategoryTheory/categories/Modules.v
AnthonyBordg: @DanGrayson , I assume you want me to make the change everywhere for elegance ? For instance, starting on line 24, to replace `precategory_ob_mor_module` by `precategory_ob_mor_Mod` or even before with the name of the section.

AnthonyBordg:
@DanGrayson 's last comments are addressed by the last commit.

AnthonyBordg:
@DanGrayson : it should be fixed now (but can't tell for sure since the Travis build is in progress).

UniMath/CategoryTheory/categories/Modules.v
```diff
@@ -0,0 +1,405 @@
...
+
+Definition Mod : precategory :=
+   mk_precategory (precategory_data_module) (is_precategory_precategory_data_module).
+
+End precategory_modules.
+
+
+(** ** The category of (left) R-modules and R-modules homomorphisms *)
+
+Section category_modules.
```
benediktahrens: There is no reason to close and re-open a section here - simply remove the lines 92 and 97 and 99.

UniMath/CategoryTheory/categories/Modules.v
```diff
@@ -0,0 +1,405 @@
...
+     + exact (p (pr1 f) (pr1 g)).
+   - intro p.
+     assert (q : isaset (ismodulefun (pr1 g))).
+     + exact (isasetaprop (isapropismodulefun (pr1 g))).
+     + apply q.
+Defined.
+
+Definition category_Mod : category := category_pair (Mod R) has_homsets_Mod.
+
+End category_modules.
```
benediktahrens: Again, you can simply keep the section open, no need to close the current one and open a new one.

AnthonyBordg:
@benediktahrens : done.

benediktahrens:
lgtm, can be merged when travis is done

UniMath/CategoryTheory/categories/Modules.v
```diff
@@ -0,0 +1,397 @@
...
+Definition invmoduleiso {M N : Mod} (f : moduleiso M N) : moduleiso N M.
+Proof.
+   use moduleisopair.
+   - exact (invweq f).
+   - apply dirprodpair.
+     + exact (isbinopfuninvmap f).
+     + exact (islinearinvmap f).
+Defined.
+
+Definition isaprop_islinear {M N : Mod} (f : (pr1module M) -> (pr1module N)) : isaprop (islinear f).
```
DanGrayson: This lemma should be moved to Algebra/Modules.v, where it can replace part of the proof of isapropismodulefun, since it has nothing to do with category theory.

UniMath/CategoryTheory/categories/Modules.v
```diff
@@ -0,0 +1,397 @@
...
+   unfold moduleiso'_to_moduleiso, moduleiso_to_moduleiso'. cbn.
+   rewrite (modulefun_unel_uniqueness (dirprod_pr2 (pr2 (pr1 w)))).
+   apply idpath.
+   intro w. apply idpath.
+Defined.
+
+Definition moduleiso'_to_moduleiso_weq (M N : Mod) : (moduleiso' M N) ≃ (moduleiso M N) :=
+   weqpair (moduleiso'_to_moduleiso M N) (moduleiso'_to_moduleiso_isweq M N).
+
+Lemma isaset_el_of_setwith2binop (X : setwith2binop) : isaset X.
```
DanGrayson: This lemma can probably be removed.  Below, where it is used, try just `apply setproperty`.

UniMath/CategoryTheory/categories/Modules.v
```diff
@@ -0,0 +1,397 @@
...
+Proof.
+   intros x y.
+   apply (setproperty X).
+Defined.
+
+Lemma isaprop_isrigfun {X Y : rig} (f : X -> Y) : isaprop (isrigfun f).
```
DanGrayson: This lemma is redundant, since `isapropisrigfun` can be found in Rigs_and_Rings.

UniMath/CategoryTheory/categories/Modules.v
```diff
@@ -0,0 +1,397 @@
...
+Proof.
+  apply (isofhleveldirprod 1).
+  - apply isapropismonoidfun.
+  - apply isapropismonoidfun.
+Defined.
+
+Lemma isaset_rngfun (X Y : rng) : isaset (rngfun X Y).
```
DanGrayson: This can be moved to Algebra.

UniMath/CategoryTheory/categories/Modules.v
```diff
@@ -0,0 +1,397 @@
...
+
+
+Variable R : rng.
+
+Definition precategory_ob_mor_module : precategory_ob_mor :=
+   precategory_ob_mor_pair (module R) (λ M N, modulefun M N).
+
+Local Open Scope Cat.
+
+Definition islinear_id (M : precategory_ob_mor_module) : islinear (idfun (pr1module M)).
```
DanGrayson: This lemma is phrased as a lemma about objects in your category, but it's really just about modules, so belongs in Algebra.

There are several other such statements, too.

AnthonyBordg: @DanGrayson : Indeed, I think that the lines 30 to 41 belong in Algebra. What do you think about the best place for the lines 121 to 300 ? I would say they belong in Algebra too, the rest after line 300 should remain in Category Theory.

DanGrayson: Sure, any definition that really is just about modules and not about the category of modules should go there.

On the other hand, see #757.

siddharthist:
I can take a look at this once #789 is merged and try to get it into a state where it can be merged.

siddharthist:
I'm currently updating this PR.

DanGrayson:
You could just leave that up to the author.

mortberg:
Was this PR fully incorporated in https://github.com/UniMath/UniMath/pull/816 ? If that is the case then we should close it. @siddharthist

siddharthist:
@mortberg Yes.

benediktahrens:
Incorporated in #816.

PR #740, opened 2017-07-07
Capitalize Kleisli

jojhelfer:
As @vladimirias pointed out, there is at least as much reason to capitalize Kleisli as there is to capitalize Monad, given that the former is a proper noun.


vladimirias:
LGTM

PR #745, opened 2017-07-14
check for changes to Foundations

DanGrayson:
Here we arrange for one step of the travis job to fail if any changes
have been made to Foundations.

We also fix some punctuation in Foundations, to demonstrate how the
travis failure appears.

We also suppress some warning messages about deprecated options, such as
"Automatic Introduction".  At some point we may have to fix the files
that use "Unset Automatic Introduction", but that can wait.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,8 +25,9 @@ env:
   - PACKAGES=Tactics BUILD_COQIDE=yes BUILD_ALSO=TAGS
   - PACKAGES=Folds
   - PACKAGES=HomologicalAlgebra
+  - PACKAGES= FOUNDATIONS_CHANGE_ERROR=yes
 # building Coq in a separate stage folds up the output in the log:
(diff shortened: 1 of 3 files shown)


DanGrayson:
Click on "Details" where it says "The Travis CI build failed".  Is it clear enough
that the only problem is a change to Foundations?

benediktahrens:
Quite unconventional, but why not...

DanGrayson:
I should mention that the same check occurs for each of us when we run "make all" or "make" at home, but the resulting error is ignored by "make".

DanGrayson:
I'm a bit unsure what will happen to pending pull requests if a change to Foundations is accepted.  If the travis check is rerun in a pending pull request, it might be that my "diff" command will erroneously report that a change to Foundations has occurred.  So let's keep an eye open for that kind of thing.

PR #746, opened 2017-07-14
Redefine Kleisli in terms of RelMonad

jojhelfer:
The definitions of Kleisli and RelMonad were redundant. Here they are unified.


benediktahrens:
lgtm

PR #749, opened 2017-07-27
Define Kleisli category for a relative monad

jojhelfer:


UniMath/CategoryTheory/Monads/RelativeMonads.v
```diff
@@ -87,3 +88,45 @@ End RMonad_def.
...
+Lemma Kleisli_precat_is_precat : is_precategory Kleisli_precat_data.
+  do 2 try apply tpair;
+    try unfold compose; simpl.
+  - intros a b f.
+    apply (r_eta_r_bind R).
+  - intros a b f.
+    now rewrite (r_bind_r_eta R), id_right.
+  - intros a b c d f g h.
+    now rewrite <- assoc, (r_bind_r_bind R).
+Defined.
```
benediktahrens: It seems to me that you could close this proof off with `Qed.` instead of `Defined.`.

jojhelfer: Oh? I though we never did that (as per UniMath/README.md)

jojhelfer: Done

DanGrayson:
@jojhelfer -- It would be great if you would add your real name and portrait to your github profile!  Then we could get to know you better.

jojhelfer:
@DanGrayson 

> @jojhelfer -- It would be great if you would add your real name and portrait to your github profile! Then we could get to know you better.

Good idea, thanks! :)

PR #750, opened 2017-07-30
add construction of left adjoint from universal data

benediktahrens:


UniMath/CategoryTheory/Adjunctions.v
```diff
@@ -275,145 +275,252 @@ Coercion adjunction_data_from_is_left_adjoint {A B : precategory}
...
-    Defined.
-
-  End postcomp.
+Section right_adjoint_from_partial.
+
+Definition is_universal_arrow_from {D C : precategory}
+           (S : functor D C) (c : C) (r : D) (v : C⟦S r, c⟧) : UU :=
+  ∏ (d : D) (f : C⟦S d,c⟧), ∃! (f' : D⟦d,r⟧), f = # S f' · v.
+
+Variables (X A : precategory) (F : functor X A).
```
mortberg: Maybe this could be a good time to make some of these arguments implicit?

mortberg:
LGTM modulo some arguments that should be implicit (I should have done this when I wrote the original code...)

PR #753, opened 2017-08-25
avoid warnings on ignored implicit arguments

rmatthes:
by no longer declaring them as implicit


DanGrayson:
lgtm

This is the sort of warning that was appearing:
```
Warning: Ignoring implicit status of product binder g and following binders
```

rmatthes:
I ought to have mentioned that in my log message. Thanks.

PR #755, opened 2017-09-18
Update FromBindingSigsToMonads_Summary.v

mortberg:
This is done to reflect final version of paper

### DIFF
--- a/UniMath/SubstitutionSystems/FromBindingSigsToMonads_Summary.v
+++ b/UniMath/SubstitutionSystems/FromBindingSigsToMonads_Summary.v
@@ -357,9 +357,9 @@ Defined.
 Definition DatatypeOfBindingSig :
   ∏ (C : precategory) (hsC : has_homsets C)
     (BPC : BinProducts C) (BCC : BinCoproducts C)
-    (IC : Initial C) (TC : Terminal C)
-    (CLC : Colims_of_shape nat_graph C)
-    (HF : ∏ (F : functor_precategory C C hsC),
+    (_ : Initial C) (TC : Terminal C)
+    (_ : Colims_of_shape nat_graph C)
+    (_ : ∏ (F : functor_precategory C C hsC),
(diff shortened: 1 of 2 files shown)


DanGrayson:
lgtm

What a pain, though!

rmatthes:
@DanGrayson : Are you referring to the work that is needed to provide this interface file consistently for the readers of the article?

DanGrayson:
Yes, changing each of those citation numbers every time the paper changes must be a pain.

mortberg:
Yes, that was a tedious thing to change. But as this will be the final published version of the paper we will not have to do this again.

UniMath/SubstitutionSystems/FromBindingSigsToMonads_Summary.v
```diff
@@ -22,7 +22,6 @@ Require Import UniMath.Combinatorics.Lists.
 Require Import UniMath.CategoryTheory.total2_paths.
```
rmatthes: could this import statement for total2_paths not be dropped?

rmatthes:
lgtm; checked all the numbers (the file typically gives the solution to a problem under the number of the problem, not under the number of the construction solving the problem - understandable since the solution has the problem as its type, and certainly good enough for this extra documentation)

UniMath/SubstitutionSystems/FromBindingSigsToMonads_Summary.v
mortberg: I guess so. I'll prune the list of imports.

mortberg:
Just so everyone knows the best is probably if we leave this PR open until we submit the final version of the paper as we might have to change a few more things. Benedikt can then merge it when we are done.

DanGrayson:
Why leave it open, when you could close it now and reopen it later?

mortberg:
Ahh right, I forgot about that feature @DanGrayson

mortberg:
I changed the names to underscores and I think this is ready to be merged now as the final version of the paper is getting finished.

DanGrayson:
lgtm

PR #756, opened 2017-09-19
Library on displayed cats from UniMath/TypeTheory

benediktahrens:
This PR brings many of the files on displayed categories to UniMath/UniMath.
Two files not included here depend on other files in UniMath/TypeTheory, and are very specialized anyway.

This PR does not pass the Travis test yet, for the following reason: an example - the displayed category of topological spaces - requires the "Topology" package, which CT should not depend on.
It is not clear to me how to solve this issue: neither of Topology or CT should depend on the other package. The only solution to avoid this, though, would be to have a third package depending on both, which could then contain that example. That is quite silly, though... Suggestions welcome!


DanGrayson:
Why not move the example into the Topology package?  That would be better than putting Topology before CategoryTheory in the list of package, because CategoryTheory is more basic.

benediktahrens:
On 09/19/2017 07:12 PM, Daniel R. Grayson wrote:
> Why not move the example into the Topology package?  That would be
> better than putting Topology before CategoryTheory in the list of
> package, because CategoryTheory is more basic.

Yes, that is one possibility. If that is the preferred one, I'll
implement it.

benediktahrens:
I have moved the example of the displayed category of topological spaces to the package Topology. The PR should compile fine on Travis now.

UniMath/CategoryTheory/DisplayedCats/Codomain.v
```diff
@@ -0,0 +1,192 @@
+
+(** * Examples
+
+A typical use for displayed categories is for constructing categories of structured objects, over a given (specific or general) category. We give a few examples here:
```
rmatthes: This comment does not seem to fit the file contents.

benediktahrens:
On 09/20/2017 05:57 PM, Ralph Matthes wrote:
> rmatthes commented on this pull request.
> 
> 
> 
>> @@ -0,0 +1,192 @@
> +
> +(** * Examples
> +
> +A typical use for displayed categories is for constructing categories of structured objects, over a given (specific or general) category. We give a few examples here:
> 
> This comment does not seem to fit the file contents.
> 

Well spotted, thanks a lot!

UniMath/CategoryTheory/DisplayedCats/Auxiliary.v
```diff
@@ -0,0 +1,373 @@
...
+
+(** Compare [pathsdirprod]. *)
+Definition dirprod_paths {A B : UU} {p q : A × B}
+  : pr1 p = pr1 q -> pr2 p = pr2 q -> p = q.
+Proof.
+  destruct p as [a b], q as [a' b']; apply pathsdirprod.
+Defined.
+
+(** Compare [total2asstol]. *)
+Definition dirprod_assoc {C0 C1 C2 : UU}
```
DanGrayson: This would be more accessible to others in MoreFoundations/PartA.v

UniMath/CategoryTheory/DisplayedCats/Auxiliary.v
```diff
@@ -0,0 +1,373 @@
...
+  : (C0 × (C1 × C2)) -> ((C0 × C1) × C2).
+Proof.
+  intros c. exact ((pr1 c ,, (pr1 (pr2 c))) ,, pr2 (pr2 c)).
+Defined.
+
+(** Identical to [dirprodf]. *)
```
DanGrayson: Why make it if it's identical something already present?

UniMath/CategoryTheory/DisplayedCats/Auxiliary.v
DanGrayson: Or eliminate it in favor of one of these definitions in PartA:

```
weqdirprodasstor: ∏ X Y Z : UU, (X × Y) × Z ≃ X × Y × Z
weqdirprodasstol: ∏ X Y Z : UU, X × Y × Z ≃ (X × Y) × Z
```

benediktahrens:
Thanks for pointing out these issues. Indeed, the file Auxiliary.v should be disolved.
I would like to do that in a second PR, if that is acceptable to you, sometime in the near future.

DanGrayson:
okay

UniMath/CategoryTheory/DisplayedCats/Core.v
```diff
@@ -0,0 +1,1912 @@
...
+    ; id_right_disp {x y} {f : x --> y} {xx} {yy} (ff : mor_disp f xx yy)
+        : comp_disp ff (id_disp yy)
+          = transportb (fun g => mor_disp g xx yy) (id_right _) ff
+    ; assoc_disp {x y z w} {f : x --> y} {g : y --> z} {h : z --> w}
+        {xx} {yy} {zz} {ww}
+        (ff : mor_disp f xx yy) (gg : mor_disp g yy zz) (hh : mor_disp h zz ww)
+        : comp_disp ff (comp_disp gg hh)
+          = transportb (fun k => mor_disp k _ _) (assoc _ _ _)
+            (comp_disp (comp_disp ff gg) hh)
+    ; homsets_disp {x y} {f : x --> y} {xx} {yy} : isaset (mor_disp f xx yy)
```
DanGrayson: It might be cleaner to just make mor_disp take values in hSet instead of in UU.

UniMath/CategoryTheory/DisplayedCats/Core.v
```diff
@@ -0,0 +1,1912 @@
...
+
+Local Open Scope type_scope.
+
+(* Undelimit Scope transport. *)
+
+(** * Displayed categories *)
+
+Module Record_Preview.
+
+  Record disp_cat (C : category) : UU :=
```
DanGrayson: The use of `Record` in UniMath is not allowed, but a good case can be made for allowing it, since the developers assert that any divergence of its behavior from an iterated dependent sum type is a bug, and since we now implement `total2` as a record.  One of us should submit an issue, at least, if we're going to accept this pull request.

UniMath/CategoryTheory/DisplayedCats/Core.v
```diff
@@ -0,0 +1,1912 @@
...
+(* Any displayed category has a total category, with a forgetful functor to the base category. *)
+Section Total_Category.
+
+Context {C : category} (D : disp_cat C).
+
+Definition total_category_ob_mor : precategory_ob_mor.
+Proof.
+  exists (∑ x:C, D x).
+  intros xx yy.
+  (* note: we use projections rather than destructing, so that [ xx --> yy ]
```
DanGrayson: Perhaps, since we have eta reduction for pairs by virtue of turning on primitive projections, this comment is no longer relevant.

UniMath/CategoryTheory/DisplayedCats/Core.v
```diff
@@ -0,0 +1,1912 @@
...
+    apply id_left.
+    eapply pathscomp0.
+      apply maponpaths, id_left_disp.
+  (* Note: [transportbfi] is from [UniMath.Ktheory.Utilities].
+  We currently can’t import that, due to notation clashes. *)
+    apply transportfbinv.
+  - intros xx yy ff; cbn.
+    use total2_paths_f; simpl.
+    apply id_right.
+    eapply pathscomp0.
```
DanGrayson: It would be good to fully delimit the parts of a proof that attack separate goals.

DanGrayson:
The word "sigmas" is used in three comments.  It would be better to use "sums", for sigma is just the letter used as notation for sums.

DanGrayson:
Okay, I'll have no further comments.  It's a useful contribution.

benediktahrens:
On 09/21/2017 03:59 PM, Daniel R. Grayson wrote:
> Okay, I'll have no further comments.  It's a useful contribution.

Can we merge then? As I said, I'll be happy to address your comments in
a later PR (and I shall make an issue for that).

DanGrayson:
Yes - I've updated it, and one of us can merge after travis approves.

Issue #763, opened 2017-10-13
rework Foundations


DanGrayson:
Vladimir thought it was important for those not acquainted with UniMath to be able
to learn it well by stepping through the proofs (in ProofGeneral or Coqide), working
their way sequentially through the files from the first to the last, or at least until all 
the files in the first package, UniMath/Foundations, had been scanned.  In early 2017 he told
us that we should stop adding random lemmas to UniMath/Foundations, and that he should
undertake the project of going through it and putting it into good shape.

We should put this project on the list of projects ultimately to do, and hence I
submit this issue.  I believe it should be done by a single person, as uniformity of style
will enhance the pedagogical value, and I'd like to be the one to do it.  

I'd adhere to the following principles:

     -  preserve the spirit of his original Foundations, as in
        https://github.com/UniMath/UniMath/tree/vv-master-for-coq8.4pl3patched
        or https://github.com/vladimirias/Foundations
     -  maintain consistency with published descriptions of Foundations, especially those
        published by Vladimir
     -  adherence to our coding style guidelines

See also the discussion at https://github.com/UniMath/UniMath/issues/77.  We should probably think of this issue as subsuming that one, and try to follow the guidance there.

Items I might address first:

     -  eliminate my stuff based on DecidableProposition or move it to MoreFoundations

See also my branch where I made an attempt to start things afresh: https://github.com/DanGrayson/UniMath/tree/essential-foundations

A good place for others to submit suggestions would be as comments on this issue.

DanGrayson:
I wonder whether we should change this, and would welcome input:
```
Fixpoint natgtb (n m : nat) : bool :=
  match n, m with
  | S n, S m => natgtb n m
  | O, _ => false
  | _, _ => true
  end.

Definition natgth (n m : nat) : hProp := hProppair (natgtb n m = true) (isasetbool _ _).
```
I would prefer defining natgth itself by induction, so it returns either hfalse or htrue, and to prune all such functions to bool from the system.  I never asked Vladimir about that.

benediktahrens:
File PartD could look much better if (unicode) notation were used throughout, e.g., instead of `weq`.
Is there a chance we can fix this before December?

DanGrayson:
I'll try to do it!  Does that comment not apply to PartC?

benediktahrens:
On 10/19/2017 10:06 PM, Daniel R. Grayson wrote:
> I'll try to do it!  Does that comment not apply to PartC?
> 

[...]

PR #764, opened 2017-10-13
make cosmetical changes to PartA.v

rmatthes:
This PR is triggered by issue #763. The commits were made while reading through PartA.v in the manner Vladimir considered appropriate for learning it. I did not expect that touching on Foundations would be on the agenda so soon, and I had intended to go through all of them before filing a PR. But this would now mean that my commits could be easily lost through merge conflicts. None of my proposed changes is a big deal. Please do consider them and decide to which extent you would apply their implicit coding style elsewhere. After all, they are fully compatible with the current style guide.

By the way, this PR does not introduce new definitions or lemmas, the only new item is a tactic notation.

### DIFF
--- a/UniMath/Foundations/PartA.v
+++ b/UniMath/Foundations/PartA.v
@@ -1409,14 +1409,16 @@ Defined.
 Definition isinjinvmap {X Y} (v w:X≃Y) : invmap v ~ invmap w -> v ~ w.
 Proof. intros ? ? ? ? h x.
   intermediate_path (w ((invmap w) (v x))).
-  { apply pathsinv0. apply homotweqinvweq. }
-  rewrite <- h. rewrite homotinvweqweq. apply idpath. Defined.
+       + apply pathsinv0. apply homotweqinvweq.
+       + rewrite <- h. now rewrite homotinvweqweq.
+Defined.
 
 Definition isinjinvmap' {X Y} (v w:X->Y) (v' w':Y->X) : w ∘ w' ~ idfun Y -> v' ∘ v ~ idfun X -> v' ~ w' -> v ~ w.
 Proof. intros ? ? ? ? ? ? p q h x .
   intermediate_path (w (w' (v x))).
-  { apply pathsinv0. apply p. }
-  apply maponpaths. rewrite <- h. apply q. Defined.
+       + apply pathsinv0. apply p.
+       + apply maponpaths. rewrite <- h. apply q.
+Defined.
 
 (** *** Adjointness property of a weak equivalence and its inverse *)
 
@@ -1424,7 +1426,7 @@ Lemma diaglemma2 {X Y : UU} (f : X -> Y) {x x' : X}
       (e1 : x = x') (e2 : f x' = f x)
       (ee : idpath (f x) = maponpaths f e1 @ e2) : maponpaths f (! e1) = e2.
 Proof.
-  intros. induction e1. simpl in *. apply ee.
+  intros. induction e1. simpl in *. exact ee.
 Defined.
 
 (* this is the adjointness relation for w and its homotopy inverse: *)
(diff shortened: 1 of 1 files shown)


DanGrayson:
Our agenda is long ranging, and I think changes to Foundations should be made
by just one person, for uniformity of style.  One aspect of style I would hesitate to adopt
is `nowid`.

What if, instead, we regard this PR as a suggestion for #763?

rmatthes:
I think that issue #763 has more far-reaching aims than what my humble commits would do. I would be happy if they were discussed since they are already available. Elements of this discussion could enter a style guide that also has recommendations that favour uniformity. Of course, I could easily take back some of my suggestions, just by committing inverse changes (this is not the optimal Git style, agreed).

Concerning `nowid`, the proposal tries to be in the style of `now` that I find intuitive and that is already used in UniMath, but wants to be explicit about the little that is required - just reflexivity. Am I right that `reflexivity` itself ought not be used?

DanGrayson:
No, why should reflexivity not be used?  See our coding style guide in 
https://github.com/UniMath/UniMath/blob/master/UniMath/README.md

rmatthes:
It is true that the style guide does not mention it. But I thought that reflexivity is based on the type class mechanism and could potentially do more than just applying idpath.

DanGrayson:
I doubt it's based on type classes, and yes, it doesn't just apply idpath.  For example, if the
goal is `unit`, it will use `tt`; the manual makes this clear.  In general, there are two reasons for the items in the style guide.  One is to use a relatively simply subset of Coq so we can switch to another proof assistant if needed.  Another is to prevent the generation of proof terms that contain things not covered by Vladimir's envisioned proof of soundness (see https://github.com/UniMath/UniMath/issues/658, for example.)

rmatthes:
Could someone please change the Travis checking so that touching on Foundations is no longer considered a failure?

DanGrayson:
Why?  We prefer to receive no suggested changes to Foundations at this time.

DanGrayson:
Thanks, Ralph, for the editing, but as stated in #763, which is assigned to me, I prefer to be the single person reworking Foundations, so we'll keep this in mind as a suggestion for later on, when I get started on that, and close this pull request.

benediktahrens:
I think that many, but not all (e.g., the tactic `nowid`), of the changes of this PR, are improvements to Foundations that do not change its spirit. I strongly urge to integrate those changes.

@DanGrayson : can you ensure that your insistence of being the only person reworking Foundations will not create a bottleneck?

DanGrayson:
I don't see how it could create a bottleneck, since there is no other project that depends on the completion of #763 .

UniMath/Foundations/PartA.v
```diff
@@ -137,6 +137,15 @@ Unset Automatic Introduction.
...
 
+(* in view of maybe too much that is done by the tactic [now], the following tactic
+   had been considered, but not been seen as useful enough to clutter code with an
+   extra such device:
+Tactic Notation "nowid" tactic(t) := t; apply idpath.
+
+sub/coq/tactics/Tactics.ml contains the implementation of reflexivity that refers to
+setoids as well, which is why [apply idpath] is preferred to [reflexivity]
+*)
+
```
benediktahrens: This tactic and comment should be removed.

DanGrayson:
I think it's important to have just one person touching Foundations, for uniformity of style, and I don't find the changes conservative.  In particular, while it might make a proof script look
shorter to have fewer sentences in it (by replacing periods by semicolons), it will reduce the feedback for students stepping through it.  Also, "reflexivity" is a fine tactic, preferable, I think, to "intros; apply idpath".  I strongly recommend closing this pull request and letting me regard it as a set of suggestions.

benediktahrens:
On 10/20/2017 08:38 PM, Daniel R. Grayson wrote:
> I think it's important to have just one person touching Foundations,
> for uniformity of style, and I don't find the changes conservative.
> In particular, while it might make a proof script look shorter to
> have fewer sentences in it (by replacing periods by semicolons), it
> will reduce the feedback for students stepping through it.  Also,
> "reflexivity" is a fine tactic, preferable, I think, to "intros;
> apply idpath".  I strongly recommend closing this pull request and
> letting me regard it as a set of suggestions.
> 


That is fine with me.

I am looking forward to seeing your suggestions soon and thank you in
advance for accepting to do all the work yourself!

PR #766, opened 2017-10-17
The intersection of subgroups is again a subgroup

siddharthist:
Short proofs that the intersection of submonoids is a submonoid, and the same for groups. Let me know if the names need changing to fit in with the library!


benediktahrens:
lgtm

siddharthist:
Thanks!

Issue #772, opened 2017-10-23
make sure Foundations looks a little better for the workshop in December


DanGrayson:
check line length
  formatting style
  indentation
  notations
  bullets
  localize automatic introduction
  move or delete my decidable propostion stuff

DanGrayson:
It's easy to use emacs regular expressions to convert lots of files simultaneously to better use of notations.  Should I do that outside of Foundations, or will there be too many conflicts?  I should be able to do it at least in Algebra, NumberSystems, MoreFoundations, Combinatorics.

benediktahrens:
The only PR I am a bit worried about is Cathy's [1], but that is old and
its status unclear to me.

[1] seems to be at least partially superseded by a PR that was recently
merged, where Cathy gave two traditional definitions of Dedekind cuts.

[1] https://github.com/UniMath/UniMath/pull/536

DanGrayson:
Well, I'll go ahead.  We can always figure that out later.

DanGrayson:
Benedikt, do things look good enough now, in your opinion?

benediktahrens:
On 10/26/2017 03:56 PM, Daniel R. Grayson wrote:
> Benedikt, do things look good enough now, in your opinion?
> 

If I say yes now, does that mean that I won't be able to suggest any
improvements before Dec 15?

DanGrayson:
No, suggestions are always welcome.  Perhaps you close this issue now and reopen it at will when you find anything to suggest.

benediktahrens:
Closing for the time being.

PR #774, opened 2017-10-25
Clean up Foundations

DanGrayson:
We also perform various easily automated conversions to notation in all of UniMath, and fix a few other things.

This is part of fulfilling #772.


UniMath/CategoryTheory/categories/category_hset.v
```diff
@@ -161,15 +161,15 @@ Proof.
...
+  weq ((pr1 A) ≃ (pr1 B))(iso A B).
 Proof.
   exists (hset_equiv_iso A B).
   apply hset_equiv_iso_is_equiv.
 Defined.
 
 (** ** HSET is a univalent_category. *)
 
-Definition univalenceweq (X X' : UU) : weq (X = X') (weq X X') :=
+Definition univalenceweq (X X' : UU) : weq (X = X') (X ≃ X') :=
```
benediktahrens: It looks like the replacement only works on the inner `weq`, in several places in this file.

benediktahrens:
Would it be possible to put the compilation flags in a file that would be read by both the Makefile and the emacs config file?

benediktahrens:
still lgtm

DanGrayson:
At the moment I don't know how to get emacs to read a file of options and put it in the right place.

Merging.

UniMath/Algebra/Apartness.v
```diff
@@ -9,6 +9,7 @@ Require Export UniMath.Algebra.BinaryOperations.
 Require Import UniMath.Foundations.Propositions.
 
 Require Import UniMath.MoreFoundations.Tactics.
+Require Import UniMath.MoreFoundations.DecidablePropositions.
```
siddharthist: If no additional lines in this file are changed, why does it require an additional import?

UniMath/MoreFoundations/Propositions.v
```diff
@@ -55,12 +56,11 @@ Lemma dneg_elim_to_LEM : (∏ P:hProp, ¬ ¬ P -> P) -> LEM.
...
-  - apply isapropdec, P.
-  - simpl. intros n.
-    assert (q : ¬ (P ∨ ¬ P)).
-    { now apply weqnegtonegishinh. }
-    assert (r := fromnegcoprod_prop q).
-    exact (pr2 r (pr1 r)).
+  simpl. intros n.
+  assert (q : ¬ (P ∨ ¬ P)).
+  { now apply weqnegtonegishinh. }
+  assert (r := fromnegcoprod_prop q).
```
siddharthist: What's the difference between using `assert` vs. `pose` here, if any?

benediktahrens:
On 10/26/2017 07:38 PM, Langston Barrett wrote:
> siddharthist commented on this pull request.
> 
> 
> 
>> @@ -55,12 +56,11 @@ Lemma dneg_elim_to_LEM : (∏ P:hProp, ¬ ¬ P -> P) -> LEM.
>  (* a converse for Lemma dneg_LEM *)
>  Proof.
>    intros dne. intros P. simple refine (dne (_,,_) _).
> -  - apply isapropdec, P.
> -  - simpl. intros n.
> -    assert (q : ¬ (P ∨ ¬ P)).
> -    { now apply weqnegtonegishinh. }
> -    assert (r := fromnegcoprod_prop q).
> -    exact (pr2 r (pr1 r)).
> +  simpl. intros n.
> +  assert (q : ¬ (P ∨ ¬ P)).
> +  { now apply weqnegtonegishinh. }
> +  assert (r := fromnegcoprod_prop q).
> 
> What's the difference between using `assert` vs. `pose` here, if any?
> 

From the reference manual
(https://coq.inria.fr/refman/tactics.html#hevea_tactic55):

assert ( ident := term )

This behaves as 'assert (ident : type) by exact term' where type is the
type of term. This is deprecated in favor of pose proof.

So it looks like there is no difference, but we should use `pose` in the
future.

UniMath/Algebra/Apartness.v
DanGrayson: Because a theorem it used was moved.

UniMath/MoreFoundations/Propositions.v
DanGrayson: I never use `pose` and I didn't write that line, sorry.

UniMath/Algebra/Apartness.v
siddharthist: Ah, I see. Thanks!

Issue #63, opened 2014-12-12
make PAdics compile


benediktahrens:
PAdics does not compile at the moment. It should be adapted to compile with latest Coq/UniMath.
Some beautification would also be beneficial.

Anyone starting to work on this should comment in this issue, as to avoid duplication of efforts.

rmatthes:
I have started work on this task on the basis of Benedikt's branch. My first commit makes the file `lemmas.v` compile again. This happens in my branch "repairpadics", see
[https://github.com/rmatthes/UniMath/tree/repairpadics](https://github.com/rmatthes/UniMath/tree/repairpadics)

DanGrayson:
I've assigned Ralph to the issue (that's a good way for others to learn at a glance that someone is already working on an issue).

rmatthes:
Four of the five vernacular files are fine now (in my branch mentioned above).
     spec    proof comments
      185      782       21 UniMath/PAdics/lemmas.v
      177     1774       37 UniMath/PAdics/fps.v
       58      457       12 UniMath/PAdics/frac.v
      219     2049       60 UniMath/PAdics/z_mod_p.v
      639     5062      130 total
But there are real problems with padics.v. Proof checking with `Defined.` takes ages, and at some point, also a `change`, so the development comes to  a halt.
Suggestions welcome.

benediktahrens:
On 10/29/2017 09:36 PM, Ralph Matthes wrote:
> Four of the five vernacular files are fine now (in my branch mentioned above).
>      spec    proof comments
>       185      782       21 UniMath/PAdics/lemmas.v
>       177     1774       37 UniMath/PAdics/fps.v
>        58      457       12 UniMath/PAdics/frac.v
>       219     2049       60 UniMath/PAdics/z_mod_p.v
>       639     5062      130 total
> But there are real problems with padics.v. Proof checking with `Defined.` takes ages, and at some point, also a `change`, so the development comes to  a halt.
> Suggestions welcome.
> 


Here, the following works:

(* Coq hangs on this command on Oct. 29, 2017: apply idpath. Solution: *)
(*      apply hzplusladd.
      Fail (apply hzmultlmul).
      Fail (apply hzmultrmul).
*)
      exact (idpath _ ).

But that does not solve the problem with `Defined.`

benediktahrens:
On 10/29/2017 09:36 PM, Ralph Matthes wrote:
> Four of the five vernacular files are fine now (in my branch mentioned above).
[...]

PR #776, opened 2017-10-30
Repair the contents of the PAdics directory

rmatthes:
Meant to solve most of issue #63 . See the caveats in the [last commit on padics.v](https://github.com/rmatthes/UniMath/commit/fbee2b127978674e08449738d9b077b61a3930c5). There is no problem with the other four vernacular files, though.

If no solution is found to the execution time of proof checking, maybe the directory should be merged and padics.v still not included into the regular make process, which creates a problem since the makefile currently does not accept the presence of .v files in directories which are compiled. On the other hand, padics.v in the proposed form contains `Admitted.` which will not be accepted by the makefile.


rmatthes:
I tested the implementation in the state before the last (merge) commit. On a Core i7-4790 (3.60GHz), after waiting for close to 14 hours, the first four proofs that I close with `Admitted.` were effectively also processed by Coq with a closing `Defined.` instead. The file .padics.aux is not written to concurrently, so I do not have better data (without doing a proper testing setup). At least, the lemmas hzquotientandfpstimesl, precarryandcarry, doublecarry and quotientprecarryplus have to be considered as fully formalized and approved by UniMath.

The ongoing effort concerns the lemmas carryandplus, carryandtimesl and padicintsareintdom.

benediktahrens:
Have you tried cutting the "Admitted" proofs into smaller pieces?

rmatthes:
I did not. But there are also very short ones among them, e.g. the proof of precarryandcarry. What could possibly be gained by taking its proof apart?

rmatthes:
To my mind, the proof of `precarryandcarry_pointwise` looks extremely simple and innocuous. There must be odd things going on somewhere with all the coercions.

benediktahrens:
On 10/31/2017 03:08 PM, Ralph Matthes wrote:
> To my mind, the proof of `precarryandcarry_pointwise` looks extremely
> simple and innocuous. There must be odd things going on somewhere
> with all the coercions.
> 

I suggest removing the `simpl` tactics.

benediktahrens:
A first step is to make this one work:
```
Lemma foo1 (m : hz) (is : hzneq 0 m) (a : fpscommrng hz) n
      : carry m is a (S n) + 0%hz = carry m is a (S n).
Proof.
  intros.
  exact (hzplusr0 _ ).
Qed.
```
I imagine that during `Qed`, the term `carry m is a (S n)` is unfolded, which is a bad idea.

DanGrayson:
It would be better to replace each `Admitted` by `Abort`, as Admitted facts can deceive the user.

rmatthes:
What kind of deception would that be in this case? I bet we would just need stronger hardware to get the verification through.

benediktahrens:
On 11/07/2017 07:47 PM, Ralph Matthes wrote:
> What kind of deception would that be in this case? I bet we would just need stronger hardware to get the verification through.
> 


I think everybody agrees that we want to avoid `Admitted` in UniMath.

The reasons for the need of the Admitted that are still present in
PAdics is mysterious to us. They are not mathematical reasons, at least
not in the sense that we don't know how to prove something. Instead,
they have to do with how things are formulated - an interplay between
the mathematics and the underlying proof assistant.

It would be great if we could fix the remaining `Admitted` soon.
However, I feel that that will require some diving deeply into the code,
and it might not happen for a while.

It would be silly to have the code become incompatible in the meantime.

I think that we should merge the code now, for the sake of keeping it
compatible more easily. But we should be very clear about the fact that
there are lemmas that are admitted in that code.

DanGrayson:
Let's use `Abort` and not use `Admitted` here.  Why make an exception to our policy in this case?  For example, one admitted fact is `padicintsareintdom`.  Just abort it and abort the one proof where it is used (right near the end, so nothing will be lost).

benediktahrens:
On 11/07/2017 09:50 PM, Daniel R. Grayson wrote:
> Let's use `Abort` and not use `Admitted` here.  Why make an exception
> to our policy in this case?  For example, one admitted fact is
> `padicintsareintdom`.  Just abort it and abort the one proof where it
> is used (right near the end, so nothing will be lost).
> 

Oh, right, `Abort` serves almost the same purpose. I had forgotten about
that fact.

Then, yes, aborting instead of admitting.

rmatthes:
If you abort the first lemma, you have to abort all it depends on. What is the information contained in plenty of aborted proofs?
Of course, one can keep track of all the dependencies by hand. Do you want me to do that still?

benediktahrens:
On 11/07/2017 10:29 PM, Ralph Matthes wrote:
> If you abort the first lemma, you have to abort all it depends on.
> What is the information contained in plenty of aborted proofs? Of
> course, one can keep track of all the dependencies by hand. Do you
> want me to do that still?
> 

One way to do it is to add the aborted lemma as a hypothesis to any
result that depends on that. For that, it convenes to give a name to the
type of the lemma, as in

```
Definition statement_of_lemma : UU := ...

Lemma foo : statement_of_lemma.
Proof.
...
Abort.


Lemma bar : statement_of_lemma -> something_else.
Proof.

```
That's cumbersome, but a good way to document things.

rmatthes:
This is what I meant by "by hand".

DanGrayson:
I just tried an experiment: I added `Unset Kernel Term Sharing.` at the top of the file, put back each `Defined.`, and they worked almost instantly.

DanGrayson:
I'm glad that worked out.  Now we should get rid of the notation "~>" for paths and use "=" instead.

rmatthes:
I would be glad to end my work on this package for now. With `Unset Kernel Term Sharing.` in two more of the vernacular files, my machine compiles it in 20s, single-threaded. The other two files did not seem to profit from it, which is why I left them without that option.

DanGrayson:
It would be nice if the tactic `permute` had its code indented readably.

Other than that, it looks good to me.

DanGrayson:
lgtm

rmatthes:
Does `permute` now look reasonable? No point in waiting for compilation results, only whitespace has been modified.

DanGrayson:
Yes, it does.  I'll let Benedikt pull it.

DanGrayson:
Thanks for getting that package to work!

rmatthes:
Thanks for unblocking the proof verification problem - even if that still seems like a hack to me, in particular since the option is not documented for Coq users anywhere.

DanGrayson:
You're right -- it's not documented in the manual, which surprises me.

Here's a discussion where they talk about adding the command but not documenting it: https://coq.inria.fr/bugs/show_bug.cgi?id=3309

rmatthes:
How does this affect independent checking of the proofs? Is coqchk blind to having chosen that option?

DanGrayson:
I don't know.  I can't even get `coqchk` to check Foundations/Preamble.

benediktahrens:
Great, thanks a lot you two for getting this package back in shape.

PR #778, opened 2017-11-07
Add proofs of two monad laws (naturality of mu and eta).

logicosimo:
Two easy fusion laws useful for rewriting (Monad_law4 , Monad_law5).

### DIFF
--- a/UniMath/CategoryTheory/Monads/Monads.v
+++ b/UniMath/CategoryTheory/Monads/Monads.v
@@ -89,6 +89,18 @@ Proof.
 exact (pr2 (pr2 T)).
 Qed.
 
+Lemma Monad_law4 {C : precategory} {T : Monad C} {a b: C} (f: a --> b) : η T a · # T f = f · η T b.
+Proof.
+  apply pathsinv0.
+  apply (nat_trans_ax (η T)_ _ f).
+Qed.
+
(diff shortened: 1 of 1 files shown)


DanGrayson:
Hi, and thank you for your contribution, which we'll take a look at.  Meanwhile, what about adding your name, photo, and homepage link to your github profile, so we can be better acquainted?

benediktahrens:
I don't get the point of this: it is just introducing an additional name (one difficult to memorize for that matter) for something that is already available with a more descriptive name.

mortberg:
I believe this kind of trivial instantiations might be useful as one can directly rewrite with them instead of selecting the subterm by hand and applying `nat_trans_ax (eta T) _ _ f`. But it would be nice with some example of where one actually rewrites with these extra monad laws.

peterlefanulumsdaine:
I agree with Anders about how trivial instantiations can sometimes be useful, as @logicosimo says in the first comment that these are.  The names could be more helpful, but they are following the existing names of the main monad axioms in this file.  So this looks good to me.

benediktahrens:
That the other monad laws are badly named is not a good reason for introducing even more bad names.

`Monad_law4 T` is much less clear than `nat_trans_ax (η T)`, and not much shorter.
If this were part of some considerable contribution, I would not mind it, but by itself it is just noise.

PR #781, opened 2017-11-14
clean up Foundations

DanGrayson:
... in time for the workshop, in line with #763.

Incorporates most of Ralph's suggestions in #764.

I also add this to our style guide:

* Do not use `apply` with a term that needs no additional arguments filled in,
  because using `exact` would be clearer.


benediktahrens:
lgtm

PR #784, opened 2017-11-25
remove exception to enforcing listing

benediktahrens:
following up on #783 , we remove the exception to the rule that *.v files must be listed.


Makefile
```diff
@@ -278,7 +278,7 @@ clean::; rm -f .check-travis.okay
 
 # here we ensure that every *.v file F in each package P is listed in the corresponding file UniMath/P/.package/files
 # except for one, which someone has to look at and fix or eliminate:
```
rmatthes: this comment should also be updated

PR #785, opened 2017-11-27
Change Travis to use OCaml 4.02.3 and Coq 8.7.0

palmskog:
Use an Ubuntu package repo by OPAM maintainers to install OCaml 4.02.3 in Travis. Use OPAM to install GTK bindings and camlp5. Switch to the Coq 8.7.0 commit in the submodule. All Travis tasks pass for me.


benediktahrens:
lgtm in principle.
However, I am thinking that I might prefer delaying the switch to Coq 8.7 until after the UniMath school (e.g., until after Dec 15).

benediktahrens:
On 12/14/2017 12:36 AM, Anders Mörtberg wrote:
> mortberg approved this pull request.
> 
> I can't wait until we are using Coq 8.7. Let's try to switch to it as
> soon as possible after the UniMath school is over.

What exactly will change with v8.7?

mortberg:
SSReflect is shipped with Coq 8.7. I plan to experiment with using their
rewrite tactic instead of the one of Coq 8.6. Hopefully this will lead to
no visible change for the user, except that we will have a rewrite tactic
that much more often succeeds and we won't have to do the "etrans; [apply
maponpaths, maponpaths, H|]" stuff anymore.

--
Anders

benediktahrens:
Thanks for having put this on hold; it would be fine with me now to merge this PR.

benediktahrens:
From a maintainer point of view, the "etrans; [apply maponpaths, 
maponpaths, H|]" is much better than any kind of rewrite!

rewrite does not specify where it rewrites. As a consequence, if 
something breaks, it is much harder for someone who does not know the 
code well to fix it.

So I encourage you to keep up your good work with etrans (which is a 
horrible name for a good device).

mortberg:
One of the many beautiful things with the ssreflect rewrite tactic is that it provides powerful syntax for selecting which subterm to rewrite, so code written using it will be as easy to maintain as code written using etrans. In my experience this works extremely well and the tactic is very predictable and reliable. The main thing I missed when I started working on UniMath was a good rewrite tactic. I'll see if I can rewrite (no pun intended!) one of my proofs that uses a lot of etrans and make a PR so that we can discuss this with some concrete example. 

Another point I want to stress is that the manual subterm selection using etrans that we currently use relies on apply and refine not changing their behavior... If they do change (like they did some versions ago) then we will have to change every proof that uses etrans followed by a sequence of apply/refine. The ssreflect library on the other hand is very stable as the whole mathcomp library relies on it. Furthermore, I think it is especially unlikely that the rewrite tactic changes as this is one of the most used tactics when formalizing algebra (when I formalized a lot of algebra for my PhD about 90% of my proof scripts were rewrites, so if we want to have a good UniMath algebra library I think we will need a good rewrite tactic...)

mortberg:
Anyway, let's postpone this discussion until this PR has been merged and I can try to get SSReflect's rewrite tactic to work and have some concrete example showing the benefits of using that tactic instead of Coq's rewrite or etrans.

benediktahrens:
On 12/16/2017 12:18 PM, Anders Mörtberg wrote:
> One of the many beautiful things with the ssreflect rewrite tactic is that it provides powerful syntax for selecting which subterm to rewrite, so code written using it will be as easy to maintain as code written using etrans. In my experience this works extremely well and the tactic is very predictable and reliable. The main thing I missed when I started working on UniMath was a good rewrite tactic. I'll see if I can rewrite (no pun intended!) one of my proofs that uses a lot of etrans and make a PR so that we can discuss this with some concrete example.
> 
> Another point I want to stress is that the manual subterm selection using etrans that we currently use relies on apply and refine not changing their behavior... If they do change (like they did some versions ago) then we will have to change every proof that uses etrans followed by a sequence of apply/refine. The ssreflect library on the other hand is very stable as the whole mathcomp library relies on it. Furthermore, I think it is especially unlikely that the rewrite tactic changes as this is one of the most used tactics when formalizing algebra (when I formalized a lot of algebra for my PhD about 90% of my proof scripts were rewrites, so if we want to have a good UniMath algebra library I think we will need a good rewrite tactic...)
> 

What is the relationship between the ssreflect library and the ssreflect 
rewrite tactic?

We also have the plan to get rid of Coq's initial state, that is, 
starting Coq with -nois.

mortberg:
The ssreflect library can be found in:

https://github.com/coq/coq/tree/master/plugins/ssr

It is a collection of ocaml files implementing the ssreflect tactic language and three .v files (one about booleans, one about functions and one about general ssreflect stuff).

The rewrite tactic is defined in one of the ml files of this library/plugin:

https://github.com/coq/coq/blob/master/plugins/ssr/ssrequality.ml#L642

I might be able to write a file that only exports this tactic and not the rest of the ssreflect tactics and the three .v files. This is probably ideal for UniMath, the only difference when someone imports this file will be that `rewrite` calls some other ocaml code that works much better than what is provided by the Coq standard library. I believe there won't be any problem with the `-nois` flag, but it is hard to know without trying.

DanGrayson:
Benedikt, it is now after Dec 15.  Can I pull this?

palmskog:
@DanGrayson I believe @benediktahrens already gave his permission:

> Thanks for having put this on hold; it would be fine with me now to merge this PR.

https://github.com/UniMath/UniMath/pull/785#issuecomment-352139774

benediktahrens:
On 12/16/2017 07:39 PM, Karl Palmskog wrote:
> @DanGrayson I believe @benediktahrens already gave his permission:
> 
>> Thanks for having put this on hold; it would be fine with me now to merge this PR.
> 
> https://github.com/UniMath/UniMath/pull/785#issuecomment-352139774
> 

Yes, exactly - feel free to pull whenever is convenient.

PR #787, opened 2017-11-30
Prove an instance of markov's principle

abooij:
If we know that there _exists_ a natural number for which a decidable predicate `P` holds, then we can _find_ a natural number for which `P` holds, by a bounded search.

I couldn't find this anywhere in UniMath yet, although I may have missed something.

I haven't done much with Coq or UniMath yet so I very much welcome criticism at any level of detail. Is this file in the right place? Does any part of the code adhere to coding style standards? (I couldn't find any style guide.)

At the end of the file I have some demo code to test that this actually works, and where I do some computations, to see when things normalise. Possibly this should be documentation rather than code. Or I can take it out entirely, if you want.


abooij:
Oh, already the Makefile complains that the file is not in the right place, as I depend on `Topology.Prelim` for `isaproptotal2'`.

rmatthes:
The style guide is [here](https://github.com/UniMath/UniMath/tree/master/UniMath#unimath-coding-style), one level below one might first expect it.

abooij:
What would be the best solution regarding `isaproptotal2'`? Should my code move, or `isaproptotal2'`? But if my code is supposed to move, then where to? Because, if I understand the `Makefile` correctly, it must go to one of the packages listed after `Topology`, and none of them sound too relevant.

benediktahrens:
On 11/30/2017 11:02 PM, Auke Booij wrote:
> What would be the best solution regarding `isaproptotal2'`? Should my
> code move, or `isaproptotal2'`? But if my code is supposed to move,
> then where to? Because, if I understand the `Makefile` correctly, it
> must go to one of the packages listed after `Topology`, and none of
> them sound too relevant.
> 

It sounds like `isaproptotal2'` should be moved to MoreFoundations.

DanGrayson:
Yes, move it to MoreFoundations/Propositions.

But it has a superfluous hypothesis, so replace it by this version:
```
Lemma isaproptotal2' {X : UU} (P : X → UU) :
  isPredicate P →
  (∏ x y : X, P x → P y → x = y) →
  isaprop (∑ x : X, P x).
Proof.
  intros X P HP Heq.
  apply invproofirrelevance.
  intros [x p] [y q].
  induction (Heq x y p q).
  induction (iscontrpr1 (HP x p q)).
  apply idpath.
Defined.
```

DanGrayson:
And let's rename it from `isaproptotal2'` to `isaproptotal2`.

DanGrayson:
Re: "At the end of the file I have some demo code to test that this actually works, and where I do some computations, to see when things normalise."

A good place for such tests would be in the file `Combinatorics/Tests.v`.  We want to run them, so don't think of them as comments or as documentation.

UniMath/Combinatorics/MarkovPrinciple.v
```diff
@@ -0,0 +1,123 @@
...
+  Proof.
+    refine (@hinhuniv (∑ n : nat, P n) _ _ _).
+    - induction 1 as [n p]. exact (n_to_min_n n p).
+    - exact P_inhab.
+  Defined.
+
+  Definition minimal_n : ∑ n : nat, P n.
+  Proof.
+    induction prop_n_to_min_n as [n pl]. induction pl as [p _].
+    exact (n,,p). (* TODO fix nested brackets *)
```
benediktahrens: What does the TODO mean?

abooij: Oops, this TODO has already been fixed. But somewhere else there's an induction with nested brackets, which I should fix. I'll do it either today or tomorrow.

abooij:
Markov's principle is actually not a good name for this: the Markov's principle gives existence of elements from the inconsistency of their non-existence. But this code does not translate negative information into positive information at all. So I will submit a new PR, renaming this.

mortberg:
Is it a coincidence that @mattam82 seems to have formalized the same thing using the HoTT Coq library yesterday: https://mattam82.github.io/HoTT_Markov/ ?

I like the renaming in https://github.com/UniMath/UniMath/pull/798 as this is indeed not really Markov's principle.

mattam82:
Entirely coincidental on my part, and I did not make anything public either before yesterday, pushed by a colleague, so a true meeting of the minds! Your version is more polished than mine I must say :) Completely agreed that it's not Markov's principle but a way more constructive variant relying on bounded search. This morning I was wondering if we really have to produce the minimal [n] or could actually produce the same that is initially computed in the exists (we certainly can't prove that in the theory though, or we would probably break anonymous existence). Anyway, fun!

mortberg:
This was a funny coincidence!

I'm curious if any of your implementations can be used to compute something or if the axioms get in the way? If they do not compute then we should try this in cubicaltt or cubical Agda :-)

abooij:
Indeed a nice coincidence.

My version computes if all input does; see [Tests.v](https://github.com/abooij/UniMath/blob/577297f1ebf0a413a4f1a66336e72e0fbf995585/UniMath/Combinatorics/Tests.v#L535). But, for example, if the inhabitedness is given axiomatically, then it does not: to see this, execute `Eval compute in new_n'.` after [line 539](https://github.com/abooij/UniMath/blob/577297f1ebf0a413a4f1a66336e72e0fbf995585/UniMath/Combinatorics/Tests.v#L539).

mattam82:
I suppose mine computes but didn't test. Funext is used in the HProp stuff but I suppose it's basically computing on natural numbers like this one.

mattam82:
I'd be curious to know about concrete use cases in math for this theorem as well. Maybe you Auke or Martin have examples in mind?

mortberg:
Ok, nice that they compute! My search for more killer applications of cubicaltt continues... :)

abooij:
@mattam82, I use this theorem in my work on reals with locators. The objective is to do constructive analysis in univalent type theory, in such a way that we can observe discrete information about the real numbers that we calculate, so that we can e.g. compute signed-digit expansions.
I have some old slides [here](http://www.cs.bham.ac.uk/~abb538/posts/2017-09-20-munchen.html). I use the theorem on the slides on the archimedean principle, and on the exact intermediate value theorem.

mattam82:
Really cool! I had the intuition that this could indeed help to "go out of the real quotients" to get kind of "homotopy"-stable approximations, without having to go back to setoids (as T. Streicher thinks if I understand correctly). That's just great!

mattam82:
The constructive indefinite description slide has the diagram I was thinking about actually. What do you think about the minimality condition, is that strictly necessary?

abooij:
@mattam82, given only the truth of a proposition, we necessarily have to produce a constant value, in the sense that the output number cannot depend on the hidden input number. In my test code, the hidden witness is `2`, but the output value is `1`, because `P` already holds for `1`.

The first way to see that the output doesn't have to be the smallest natural satisfying `P`, you could choose a strange ordering on the naturals, and, for example, search backwards between `0` and `100`, and ordinary (increasing) above `100`.

However, the output doesn't have to be the _minimal_ number -- not even in a strange ordering of the naturals. For a (contrived) example, an ill-informed search could choose to always test `P0` through `P5`, and output the biggest `i` between `0` and `4` for which `Pi` and `P(i+1)` are both true, if it exists, and otherwise do ordinary bounded search starting from `0`. This computation always succeeds since, in the worst case, we fall back to the bounded search, which we know works. But the output value is not minimal in any sense, since it may output `4` even if `P0` also holds (but `P1` doesn't), but may also output `0` even if `P4` also holds (but `P5` doesn't).

mattam82:
Yes indeed, I agree. The search just has to be exhaustive and provably try every natural number to be proved terminating.

abooij:
@mortberg, the HoTT version of @mattam82 seems to compute values as well as my version does. In particular
```coq
Eval compute in (markov someseq P_inhab).1.
```
results in the value 1.  (Minor modifications are necessary to make my example work with Matthieu's code.)

PR #788, opened 2017-11-30
Refer to style guide readme in main readme

abooij:


DanGrayson:
Re: "one level below one might first expect it"

I don't see how that's correct.  The style guide addresses the writing of the *.v files, and all the *.v files are in that subdirectory.

rmatthes:
Please note the wording "might first" - the place is logically correct, but I understand that it can be overlooked easily. Moreover, it can happen that one only sees the list of directories and does not scroll further down.

rmatthes:
This does not mean that this comment has to be made in the README file.

abooij:
I agree with your sentiments. I feel that something has to be said because the path `UniMath/README.md` might also be read as "`README.md` in the repositorry called `UniMath`", and people might still not know where to look.

I have now rephrased the parenthetical remark.

benediktahrens:
On 11/30/2017 10:29 PM, Ralph Matthes wrote:
> Please note the wording "might first" - the place is logically
> correct, but I understand that it can be overlooked easily. Moreover,
> it can happen that one only sees the list of directories and does not
> scroll further down.
> 

I remember that it took me a long time to find the style guide, too :-)

DanGrayson:
Yes, it's good to have a pointer to it.

lgtm

PR #789, opened 2017-12-01
Multimodules

siddharthist:
This is a first draft of inclusion of multimodules a la Bourbaki. We end up with a lot of duplicated definitions from plain modules, it might be worth thinking about defining some parts of regular module theory in terms of multimodules where the indexing set is the one-element type (though that seems awkward).


siddharthist:
I've run into some trouble proving that a ring is a module over itself, perhaps someone (@AnthonyBordg?) could lend me some guidance. Here's where I'm at in the proof:
```coq
(** The left action of a ring on itself *)

Definition rng_left_mult {R : rng} (r : R) : rngofendabgr R :=
  (* This is the actual definition of the function *)
  ((λ x : R, (r * x)%rng),,
   (* And this is the justification that it's a monoid morphism *)
   (λ _ _, rigldistr R _ _ _),, rigmultx0 R _).

(** A ring is a left module over itself (that is, its underlying additive
    abelian group) via left multiplication *)

Local Open Scope multmonoid.
Definition ring_is_module (R : rng) : module R.
  (* unfold module, module_struct, rngfun, rigfun. *)
  refine (@rngaddabgr R,, (rng_left_mult,, _)).
  Check isrigfun.

  apply mk_isrigfun; apply mk_ismonoidfun.
  { (* isbinopfun for multiplication *)
    intros x y.
    apply monoidfun_paths, funextfun.
    intros r.
    exact (rigrdistr R x y r).
  }
  { (* rng_left_mult 0 = 0 *)
    apply monoidfun_paths, funextfun.
    intros r.
    exact (rigmult0x R r).
  }
  {
    intros x y.
    apply monoidfun_paths, funextfun.
    intros r.
    simpl.
    unfold funcomp.
    (* This is where I've gotten *)
  }
```
I'm now being asked to prove the goal `(x * y * r) = (y * (x * r))` which is obviously not true in general for noncommutative rings. Can anyone point out what I've done wrong? Thanks in advance for taking a look.

siddharthist:
I've attempted to prove that a ring is a module over itself using the axioms, but I'm also running into trouble there:
```coq
Definition ring_is_module (R : rng) : module R.
  refine (@rngaddabgr R,, _).
  apply (@mult_to_module_struct R R (pr1 ∘ rng_left_mult)%functions);
    unfold funcomp, pr1, rng_left_mult.
  - exact (fun r x y => rngldistr R x y r).
  - exact (fun r x y => rngrdistr R r x y).
  - unfold mult_isrdistr_wrt_rngop2.
```
I'm now being asked to prove 
```coq
∏ r s x : R, r * s * x = (s * (r * x))
```
Which again, is only true for commutative rings. Am I mistaken? Are only commutative rings modules over themselves?

benediktahrens:
On 12/01/2017 08:05 PM, Langston Barrett wrote:
> I've attempted to prove that a ring is a module over itself using the axioms, but I'm also running into trouble there:
> ```coq
> Definition ring_is_module (R : rng) : module R.
>    refine (@rngaddabgr R,, _).
>    apply (@mult_to_module_struct R R (pr1 ∘ rng_left_mult)%functions);
>      unfold funcomp, pr1, rng_left_mult.
>    - exact (fun r x y => rngldistr R x y r).
>    - exact (fun r x y => rngrdistr R r x y).
>    - unfold mult_isrdistr_wrt_rngop2.
> ```
> I'm now being asked to prove
> ```coq
> ∏ r s x : R, r * s * x = (s * (r * x))
> ```
> Which again, is only true for commutative rings. Am I mistaken? Are only commutative rings modules over themselves?
> 

You might have gotten the definition of the multiplication wrong, as in
(a,b) \mapsto b * a
instead of
(a,b) \mapsto a * b
?
Then associativity wouldn't hold, whereas the two distributivity laws 
would still be provable, using the respective other law.

I find the current definition of module not great...

siddharthist:
@benediktahrens You're right, I did mess up the order. What a pedestrian error! Thanks for your insight 😄 

My current plan for this PR (and the rest of this week) is:

 1. Reorganize `Modules.v` into
    * `Modules/Core.v`: what's currently in `Modules.v`
    * `Modules/Multimodules.v`: the content of this PR
    * `Modules/Examples.v`: a ring is a module over itself, `id` is a (multi)module morphism, etc.
 2. Add a general definition of monoid and ring actions as homomorphisms into the endomorphisms and automorphisms of objects in a category (I already have a definition of end-monoids and auto-groups).
 3. Update #736 according to Dan's comments
 4. Define a monoid/group representation as a monoid/group action in the category of modules

siddharthist:
This is now ready for review!

UniMath/Algebra/Rigs_and_Rings.v
```diff
@@ -1425,6 +1485,26 @@ Defined.
 
 Definition rngdirprod (X Y : rng) : rng := @rngpair (setwith2binopdirprod X Y) (isrngdirprod X Y).
 
+(** **** Opposite rings *)
+
+Local Open Scope rig.
+
+(** We just need to reuse and rearrange the opposite rig *)
+Definition opposite_rng (X : rng) : rng.
+  intros X.
```
benediktahrens: `Proof.`

UniMath/Algebra/Rigs_and_Rings.v
```diff
@@ -498,6 +502,43 @@ Defined.
...
+Definition opposite_rig (X : rig) : rig.
+  intros X.
+
+  (* Use the same underlying set and addition, flip the multiplication *)
+  refine (setwith2binoppair (pr1 (pr1rig X))
+                            (dirprodpair (pr1 (pr2 (pr1rig X)))
+                                         (fun x y => y * x)%rig),, _).
+
+  unfold op2; cbn; fold (@op1 X) (@op2 X).
+  Check mk_isrigops.
```
benediktahrens: Please remove this `Check...`.

siddharthist:
@benediktahrens Are there guidelines for when to make things transparent or opaque? I don't see it in `STYLE.md`. Thanks again for your review! I've made the changes.

benediktahrens:
On 12/06/2017 02:49 AM, Langston Barrett wrote:
> @benediktahrens Are there guidelines for when to make things
> transparent or opaque? I don't see it in `STYLE.md`. Thanks again for
> your review! I've made the changes.

The topic is subject to controversy within the UniMath community :-)

My own pov is as follows:

When proving a proposition (such as an equality between arrows in a 
category, or an equality of elements in a group), make it opaque; 
attention, there are important exceptions:

When proving `iscontr X`, one wants to extract the center of X (the 
first component), so the first component and the proof of `iscontr X` 
itself should be transparent. The second component (`forall y, y = ctr`) 
can be opaque.

This extends to other types, such as `isweq f`: when proving `isweq f` 
for some `f`, one wants to be able to extract the inverse. Hence that 
proof should be transparent, or at least the first component of the Sigma.


There might be other exceptions that I have forgotten. Usually one 
notices when one has made too many things opaque, though finding the 
exact definition that needs to be made transparent can be quite challenging.

siddharthist:
@benediktahrens Thanks!! I'm happy to get this merged so as to be able to work on balanced maps and tensor products at the conference.

PR #790, opened 2017-12-01
remove destruct from most of Foundations

rmatthes:
replace the remaining uses of destruct by induction, except for those on natural numbers in NaturalNumbers.v
in PartB.v: remove the comment that the replacement was done until the position of the comment

This touches on Foundations, but is really very shallow.

### DIFF
--- a/UniMath/Foundations/NaturalNumbers.v
+++ b/UniMath/Foundations/NaturalNumbers.v
@@ -179,12 +179,12 @@ Defined.
 Definition isdeceqnat: isdeceq nat.
 Proof.
   unfold isdeceq. intro x. induction x as [ | x IHx ].
-  - intro x'. destruct x'.
+  - intro x'. induction x'.
     + apply (ii1 (idpath O)).
     + apply (ii2 (negpaths0sx x')).
-  - intro x'. destruct x'.
+  - intro x'. induction x'.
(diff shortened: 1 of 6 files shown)


DanGrayson:
Perhaps it's time to revive the discussion at https://github.com/UniMath/UniMath/issues/658, which was aimed a permitting `destruct` in certain cases.

rmatthes:
Sadly, this question is still open one week before the UniMath school. Keeping very few `destruct` in Foundations is no good style, regardless of the outcome of the discussion. One should anyway admit that `induction` on Sigma-types produces `match` constructs, as if one had applied `destruct` (but this might change in the future). Why not `induction` whenever `destruct` and `induction` even have the same proof script? On natural numbers, I agree that `destruct` is more informative.

DanGrayson:
I think there's no rush to resolve the question of relaxing the ban on `destruct`.  Meanwhile, this pull request is welcome and looks good to me.  For beginners, it is important to have not too many equivalent options.

rmatthes:
@benediktahrens : please merge, the Travis failure is immaterial

DanGrayson:
I wonder whether your comment "Keeping very few destruct in Foundations is no good style" should have been "Keeping very few destruct in Foundations is good style".

benediktahrens:
lgtm

PR #791, opened 2017-12-02
Add code style checker

abooij:
I wrote a bash script to check some of the items in the code style guide. It's not perfect and only highlights non-conformant code, without saying what's wrong. But I've tried to minimize the number of false positives.

Usage: 

In the main directory, run
```bash
$ ./checkstyle.sh --all 
```
Or run it on a specific file:
```bash
$ ./checkstyle.sh UniMath/Combinatorics/Lists.v 
```
(In this last case, it would highlight a `destruct`, as well as line 65 which is too long.)

Unfortunately, this also picks up comments, which are very hard to filter out without ruining the output.
Also, I'm not sure I am interpreting the following rule correctly:

> Within the core Foundations package:
> * Do not start lines with : or with :=.


benediktahrens:
Thanks a lot, that looks quite useful. I am leaving it to @DanGrayson to decide if/where to put it.

siddharthist:
On this subject, there's a neat little tool called [pre-commit](http://pre-commit.com/) that automatically runs scripts to determine if code is up-to-snuff before allowing you to commit it. This can decrease maintenance burdens, but it's also a bit of a pain (you have to successfully install it before working on the project). 

One use-case for UniMath would be to run this script, another might be to check that the order of files in the `.package/files` files looks ok.

abooij:
@siddharthist, If you want to check the code style before adding new commits onto the master branch, I think that's a CI job. Otherwise, when a PR comes in, we don't have any guarantees anyway, as the pre-commit hook may not have been run.

And in any case we can't use this script for any strict checking yet, since most files don't adhere to the style guide, and this script does raise false positives (e.g. comments).

But I agree that perhaps some of the ideas in this script can be used to do more style checking than is done currently.

DanGrayson:
@abooij -- If you feel comfortable doing it, I would welcome your putting a photo of yourself on your github profile.

DanGrayson:
I will move the file to the util subdirectory, add a target to the makefile for running it, and then we can reconsider it.  (Maybe I can do it from my command line with git push to the submitter's branch.)

checkstyle.sh
```diff
@@ -0,0 +1,140 @@
...
+check_type() {
+    check_grep "[^[:space:]a-zA-Z.][^a-zA-Z.]+($2)($|[^a-zA-Z_])" "$1"
+}
+
+check_tactic() {
+    check_grep "(^|\.|;)[[:space:]]*($2)[^.;]*(\.|;)" "$1"
+}
+
+check_freestanding() {
+    check_grep "[^[:space:]](.*[^a-zA-Z]|)($2)\." "$1"
```
DanGrayson: `grep` gives an error here:
```
grep: empty (sub)expression
```
That's referring to the emptiness between `|` and `)`.

What did you have in mind there?

abooij: It worked fine for me, so there seems to be disagreement among grep versions. Please try the new version of the style checker.

DanGrayson:
Okay, it runs without error now.  But some of the warnings are inscrutable.  What about this one?:
```
518-       { intros [y [x z]]. reflexivity. }
519-       { intros [x [y z]]. reflexivity. } Defined.
520-
521-(* associativity of ∑ *)
522:**Definition totalAssociativity {X:UU} {Y: ∏ (x:X), UU} (Z: ∏ (x:X) (y:Y x), UU) : (∑ (x:X) (y:Y x), Z** x y) ≃ (∑ (p:∑ (x:X), Y x), Z (pr1 p) (pr2 p)).
523-Proof.                          (* move upstream *)
524-  intros. simple refine (_,,isweq_iso _ _ _ _).
525-  { intros [x [y z]]. exact ((x,,y),,z). }
526-  { intros [[x y] z]. exact (x,,(y,,z)). }
```

DanGrayson:
Ah, the line was too long -- never mind.

abooij:
Feel free to modify as you like.

DanGrayson:
lgtm

@benediktahrens

benediktahrens:
Resolved conflict.

lgtm

PR #793, opened 2017-12-03
more consequent use of Ltac definitions fron Foundations

rmatthes:
I only worked on the packages CategoryTheory and SubstitutionSystems. intermediate_path and use are used in place of pathvia and variations of refine.


DanGrayson:
lgtm

benediktahrens:
lgtm

PR #796, opened 2017-12-04
Actions of monoids, groups, and rings

langston-barrett:
Summary:
 * I define `hasinv`, etc. as propositions expressing "this element has an (left/right) inverse under this binary operation". This is more granular than saying "this operation is (left/right) invertible". We might want to rephrase the latter definitions in terms of the former, but I wanted to get feedback before making those changes.
 * I define the submonoid of elements with inverses, and show that it is actually a group.
 * I define the endomorphism monoid, automorphism group, and endomorphism ring (in an additive category), and define monoid (group) actions and ring actions as homomorphisms into these.

This PR might need a little reorganization (including the first bullet point above). Let me know where things should go!


UniMath/Algebra/BinaryOperations.v
```diff
@@ -312,10 +320,94 @@ Definition allinvvertibleinv {X : hSet} {opp : binop X} (is : ismonoidop opp)
...
+
+(** The inverse of an element's two-sided inverse is just that element *)
+Definition is_inv_inv {X : hSet} opp (is0 : ismonoidop opp)  :
+  ∏ (x x0 : X), (isinvel opp is0 x x0 -> isinvel opp is0 x0 x) :=
+  fun x x0 isinv => (dirprodpair (pr2 isinv) (pr1 isinv)).
+
+(** This is a similar statement to grinvop *)
+Lemma invop {X : hSet} opp (is0 : ismonoidop opp)  :
+  ∏ (x y : X), (hasinv opp is0 x -> hasinv opp is0 y -> hasinv opp is0 (opp x y)).
+  intros X opp is0 x y xhasinv yhasinv.
```
benediktahrens: We have the convention of starting every proof (or construction) with `Proof.`.

UniMath/Algebra/Monoids_and_Groups.v
```diff
@@ -433,6 +434,37 @@ Proof.
   + exact (λ i, pr2 (each_is_submonoid i)).
 Qed.
 
+(** Every monoid has a submonoid which is a group, the collection of elements
+    with inverses. This is used to construct the automorphism group from the
+    endomorphism monoid, for instance. *)
+Local Open Scope multmonoid.
+
+Definition invertible_submonoid (X : monoid) : @submonoid X.
+  intros X.
```
benediktahrens: Insert `Proof.`

UniMath/Algebra/Monoids_and_Groups.v
```diff
@@ -433,6 +434,37 @@ Proof.
   + exact (λ i, pr2 (each_is_submonoid i)).
 Qed.
 
+(** Every monoid has a submonoid which is a group, the collection of elements
+    with inverses. This is used to construct the automorphism group from the
+    endomorphism monoid, for instance. *)
+Local Open Scope multmonoid.
+
+Definition invertible_submonoid (X : monoid) : @submonoid X.
```
benediktahrens: The `@` should not be necessary. If it is, then that is a bug in the definition of `submonoid`, which should be corrected later.

UniMath/Algebra/Monoids_and_Groups.v
```diff
@@ -433,6 +434,37 @@ Proof.
...
+  - apply hinhpr.
+    refine (1,, dirprodpair (lunax _ 1) (lunax _ 1)).
+Defined.
+
+(** This submonoid is closed under inversion *)
+Lemma inverse_in_submonoid (X : monoid) :
+  ∏ (x x0 : X), invertible_elements (@op X) (pr2 X) x ->
+                isinvel (@op X) (pr2 X) x x0 ->
+                invertible_elements (@op X) (pr2 X) x0.
+  intros X x x0 _ x0isxinv.
```
benediktahrens: `Proof.`

UniMath/Algebra/Monoids_and_Groups.v
```diff
@@ -2015,6 +2047,38 @@ Defined.
 Definition grdirprod (X Y : gr) : gr.
 Proof. intros. split with (setwithbinopdirprod X Y). apply isgrdirprod. Defined.
 
+(** **** Group of invertible elements in a monoid *)
+
+Local Open Scope multmonoid.
+
+Definition invertible_submonoid_grop X : isgrop (@op (@invertible_submonoid X)).
+  intros X.
```
benediktahrens: `Proof.`

UniMath/CategoryTheory/Actions.v
```diff
@@ -0,0 +1,106 @@
...
+(** ** Structures on morphisms *)
+
+(** *** Endomorphisms and the endomorphism monoid *)
+
+(** Endomorphisms of X are arrows X --> X *)
+Definition endomorphisms {C : precategory} (X : ob C) : UU := (X --> X).
+
+(** When the hom-types of C are sets, we can form the endomorphism monoid *)
+Definition endomorphism_monoid {C : category} (X : ob C) : monoid.
+  refine ((((X --> X,, pr2 C X X)),, compose),, _).
```
benediktahrens: `Proof.`

UniMath/CategoryTheory/Actions.v
```diff
@@ -0,0 +1,106 @@
...
+Definition automorphism_grp {C : category} (X : ob C) : gr :=
+  gr_invertible_elements (endomorphism_monoid X).
+
+Example symmetric_grp (X : hSet) := @automorphism_grp hset_category X.
+
+(** *** The endomorphism ring in an additive category *)
+
+Definition endomorphism_rng {C : Additive}
+           (_ : ob (categoryWithAbgrops_category C)) : rng.
+
```
benediktahrens: `Proof.`

siddharthist:
I made a git error which caused the Travis build to be canceled. I'm going to close and re-open this to get another build started.

siddharthist:
@benediktahrens Thanks for taking a look! Do you have feedback on whether I should rephrase `islinv`, `isrinv`, `isrinvertible`, etc. in terms of `haslinv` and the others I add in this PR?

My thinking is that `isrinv`,`islinv`, and `isinv` should be rephrased in terms of `isrinvel`, `islinvel` and `isinvel`, but I don't know about `isinvertible`, which is expressed in terms of a weak equivalence instead of paths (is that because the weak equivalence is more useful in some situations?).

benediktahrens:
On 12/06/2017 02:40 AM, Langston Barrett wrote:
> @benediktahrens <https://github.com/benediktahrens> Thanks for taking a 
> look! Do you have feedback on whether I should rephrase |islinv|, 
> |isrinv|, |isrinvertible|, etc. in terms of |haslinv| and the others I 
> add in this PR?
> 
> My thinking is that |isrinv|,|islinv|, and |isinv| should be rephrased 
> in terms of |isrinvel|, |islinvel| and |isinvel|, but I don't know about 
> |isinvertible|, which is expressed in terms of a weak equivalence 
> instead of paths (is that because the weak equivalence is more useful in 
> some situations?).


Reformulating `isinvertible` in terms of `haslinv` etc is a bad idea, 
for the following reason:

The definition of `hexists` involves propositional truncation, which 
impedes computation.
The truncation is necessary because a left (or right) inverse is not 
unique in general.

However, a two-sided inverse is unique when it exists, so there is no 
need for a truncation to turn that into a proposition.

It is hence better to define two-sided inverses not using `haslinv`.


Does that make sense?

benediktahrens:
The PR does not currently compile, see, e.g., https://travis-ci.org/UniMath/UniMath/jobs/312201293 .
Could you fix these errors, please?

siddharthist:
@benediktahrens That does make sense, thank you for explaining. To summarize, I understand you to be saying: when something is not unique (e.g. a left/right inverse), we should use `hexists`, whereas when something is unique (e.g. a two-sided inverse), we should just use `total2`. Is that correct?

Also: I am still left wondering whether or not to rephrase `isrinv`,`islinv`, and `isinv` in terms of `islinvel`, `isrinvel`, and `isinvel`. I guess more generally, what would you like to see changed for this PR to be mergeable?

siddharthist:
@benediktahrens Actually, I just realized that making `hasinv` into a sigma-type makes this whole construction fail. Then we can't define the subtype `invertible_elements` of a monoid, because having an inverse is no longer an `hProp`.

UniMath/Algebra/Monoids_and_Groups.v
siddharthist: @benediktahrens Removing this is what caused the branch to fail to compile, so I guess this is something that should be corrected later.

benediktahrens:
On 12/07/2017 08:18 PM, Langston Barrett wrote:
> @benediktahrens Actually, I just realized that making `hasinv` into a
> sigma-type makes this whole construction fail. Then we can't define
> the subtype `invertible_elements` of a monoid, because having an
> inverse is no longer an `hProp`.
> 

I might be misunderstanding you, but:
in a monoid, you can show that for a given element, its inverse is unique.
That means that in a monoid, `is_invertible a` for some element `a`, is 
a proposition. However, that requires a tiny proof - it is not obvious 
from the definition.

But, the type of left-sided inverses of a fixed element is not unique. 
This is why one needs a truncation to obtain a proposition.

benediktahrens:
I think I understand better now what the issue is. I hadn't looked at the code properly.

There is the following definition:
```
Definition invertible_elements : hsubtype X :=
   fun x => (hasinv x,, isaprophasinv x).
```

That definition is ok when X is a set or a type more generally.

But, when X is a monoid, then a different definition is better.
Namely, in some sort of pseudocode:
```
Definition invertible_elements (X : monoid) :  X -> UU :=
   fun x => total2 (fun y => x * y = 1 \times y * x = 1).
Lemma isaprop_invertible_elements X (x : X) : isaprop (invertible_elements X x)
Proof.
  < insert the well-known proof that in a monoid, an inverse is unique >
Qed.
Definition invertible_elements_hsubtype (X : monoid) : hsubtype X :=
   fun x => <pair of the two above>
```

The advantage of this will be that the inverse is not under a truncation.

UniMath/Algebra/BinaryOperations.v
```diff
@@ -312,10 +320,89 @@ Definition allinvvertibleinv {X : hSet} {opp : binop X} (is : ismonoidop opp)
...
+    apply (@hinhfun2 (∑ x0 : X, isinvel x x0) (∑ y0 : X, isinvel y y0));
+      try assumption.
+    intros [xinv [isxlinv isxrinv]] [yinv [isylinv isyrinv]].
+    refine (opp yinv xinv,, _).
+    pose (assoc := pr1 is0).
+    split; unfold islinvel, isrinvel;
+      rewrite <- assoc;
+      rewrite (assoc4 opp assoc).
+    * rewrite isylinv.
+      rewrite (pr2 (pr2 (pr2 is0))).
```
mortberg: Please use accessor functions

UniMath/Algebra/BinaryOperations.v
```diff
@@ -312,10 +320,89 @@ Definition allinvvertibleinv {X : hSet} {opp : binop X} (is : ismonoidop opp)
...
+      assumption.
+    * rewrite isxrinv.
+      rewrite (pr2 (pr2 (pr2 is0))).
```
mortberg: Please use accessor functions

peterlefanulumsdaine:
This looks good to me, modulo the changes already suggested above.

One minor comment: highly compressed identifiers like `islinvel` are difficult to parse, let alone to remember them for later use, or to find them as a user searching for the concept of “left inverse”, “left invertible” in the library.  I would strongly prefer something like `is_linv_el` to this, and mildly prefer the even more verbose `is_left_inverse_element`.  But a lot of existing identifiers in the library are similarly compressed, so I understand others may prefer keeping consistent with them.

Maintainers @benediktahrens , @DanGrayson : is there any standard guideline/preference on this, for new contributed code?  Or are contributors equally free to use whatever conventions they like (and bikeshedding over it discouraged in reviews)?

DanGrayson:
This is an important issue of style, for which we need a uniform solution, and it goes back to choices Vladimir made in 2010.  He loved even such identifiers as `coconustot`.

benediktahrens:
@siddharthist : could you update this PR, please? There is currently a merge conflict in CT/.package/files.

I suggest we merge this PR and you think about naming conventions afterwards.

DanGrayson:
I've resolved the conflicts, and we can now wait for travis.

siddharthist:
I've updated this and made it conform to the style guide, it should be ready for review.

siddharthist:
We actually might want to define these in further generality: first define the categories corresponding to each monoid, group, and ring, and then define actions as functors. The reason is equivariant maps. The naive definition would look like the following:
```coq
Definition equivariant_map_mon {C : category} {M : monoid} {X Y : ob C}
                               (f : monaction M X) (g : monaction M Y) : UU :=
  ∑ h : X --> Y, h ∘ f = g ∘ h.
```
However, since `h` is a morphism in `C`, it can't be composed with `f` and `g`, which are _functions_, i.e. morphisms in `type_precat`. I don't immediately see a way around this issue, other than generalizing.

siddharthist:
I've now proved an equivalence between monoids and categories with contractible object types, which allowed me to define actions as functors and equivariant maps as natural transformations. 

I haven't established the same correspondence for groups and rings.

siddharthist:
I've also fixed the merge conflict with master. Now, it really should be ready for review!

benediktahrens:
```
COQC UniMath/CategoryTheory/Actions.v
File "./UniMath/CategoryTheory/Actions.v", line 230, characters 55-58:
Error: The reference rng was not found in the current environment.

0.39user 0.09system 0:00.49elapsed 99%CPU (0avgtext+0avgdata 356784maxresident)k
0inputs+64outputs (0major+11270minor)pagefaults 0swaps
build/CoqMakefile.make:645: recipe for target 'UniMath/CategoryTheory/Actions.vo' failed
make[1]: *** [UniMath/CategoryTheory/Actions.vo] Error 1
make[1]: Leaving directory '/UniMath'
```

siddharthist:
Okay, it is finally building! I've expanded the results just a little more: I've proved the equivalence between functors between one-object categories and monoid homomorphisms.

benediktahrens:
@DanGrayson : you had asked for changes here. Have they been addressed?

siddharthist:
I noticed that `Domains_and_Fields` defines a lot of similar ideas about elements with inverses. I've now upstreamed proofs of all the lemmas there to `BinaryOperations`, and a note that they just need to be replaced.

siddharthist:
Fixed merge conflicts!

siddharthist:
Fixed merge conflicts again!

UniMath/Algebra/Domains_and_Fields.v
```diff
@@ -60,6 +60,9 @@ Definition iscancelableif {X : hSet} (opp : binop X) (x : X)
 
 (** To monoids *)
 
+(** TODO: this has now been upstreamed to BinaryOperations.v, these should be
```
benediktahrens: Can this TODO be resolved before merging?

benediktahrens:
See https://travis-ci.org/UniMath/UniMath/jobs/456369958#L1661 for one compilation failure.
It has to do with the addition of the symmetric associativity law to categories, I assume.
```
File "./UniMath/CategoryTheory/Categories.v", line 213, characters 12-40:
Error:
In environment
obj : UU
mor : obj → obj → UU
homsets : ∏ a b : obj, isaset (mor a b)
identity : ∏ i : obj, mor i i
compose : ∏ i j k : obj, mor i j → mor j k → mor i k
right : ∏ (i j : obj) (f : mor i j), compose i i j (identity i) f = f
left : ∏ (i j : obj) (f : mor i j), compose i j j f (identity j) = f
associativity :
∏ (a b c d : obj) (f : mor a b) (g : mor b c) (h : mor c d),
compose a b d f (compose b c d g h) = compose a c d (compose a b c f g) h
associativity' :
∏ (a b c d : obj) (f : mor a b) (g : mor b c) (h : mor c d),
compose a c d (compose a b c f g) h = compose a b d f (compose b c d g h)
The term "(right,, left),, associativity" has type
 "∑
  _ : (∏
       (a
        b : precategory_data_pair
```

peterlefanulumsdaine:
I’ve updated this very old PR to build again (at least locally — hopefully the CI will agree).  It has a lot of good material; almost all changes requested in reviews have been addressed, and as far as I can see, none of this material is obsoleted or duplicated by more recent contributions.

The one outstanding requested change is the “TODO” noting that there’s a treatment of inverses in `Algebra.Domain_and_Fields` that overlaps with the one in `Algebra.BinaryOperations`, which should certainly be unified.  (The duplication is already present before this PR.) However, refactoring that will interact with a fair bit of material downstream (including another large PR in prep by a student of mine) so to avoid entanglement between large PR’s, I suggest leaving that refactor for a separate single-purpose PR, and merging this PR roughly as-is?  I will open an issue for this point.

benediktahrens:
@peterlefanulumsdaine : Thanks for working on this! I have merged current master into this PR. I am in principle happy with merging this PR, in particular since it mostly adds code.

Issue #795, opened 2017-12-04
Reorder travis build jobs to get CI results faster


abooij:
Currently, the biggest build job in Travis CI is to compile the `HomologicalAlgebra` package. It takes around 20 minutes. Luckily, packages are built in parallel. However, the total build time is still around an hour (give or take 30 minutes), because the `HomologicalAlgebra` build gets started very late in the process, and so much of the time is spent waiting for this one package to build.

If the `HomologicalAlgebra` package is built sooner, for example after `MoreFoundations KTheory`, this might reduce the total build time by up to 20 minutes, as the time spent compiling `HomologicalAlgebra` can now also be used to build other packages.

DanGrayson:
Sounds like a good idea.

benediktahrens:
Done

PR #797, opened 2017-12-04
Travis optimization

abooij:
I have done a bunch of experimentation with the travis setup. So I now indeed propose to start building `HomologicalAlgebra` sooner (closing #795). This change takes about 5-10 minutes off of the time from commit to CI completion.

I have tried the new container-based instances. Although they boot faster, they also seem to process noticably slower. So overall it is faster to use full virtualization (the current choice).

I made Travis install the required packages using the `apt` add-on, instead of the `apt-get install` commands. This is not necessarily faster, but it's more elegant, and enables switching to container-based instances.

The next obvious optimization steps would be:

- Optimize / break up `HomologicalAlgebra` (longest component at 19-23 minutes)
- Cache the coq build (this takes 2 minutes for every build that is done)

I have attempted to cache the coq build, but I didn't manage to write something that I felt was reliable. So this PR excludes that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,26 @@
 # validate this file at http://lint.travis-ci.org/
 # ocaml and coq are not on the list of languages:
 language:       generic
-sudo: 		required
+sudo: 		false
 # "trusty" is the most modern release of Ubuntu supported
 dist: 		trusty
 git:
   submodules:   false
-before_install:
-  - sudo apt-get update -qq
-install:
-  - sudo apt-get install -y ocaml ocaml-nox ocaml-native-compilers ocaml-findlib
-  - sudo apt-get install -y camlp5 camlp4-extra time libgtk2.0 libgtksourceview2.0
-  - sudo apt-get install -y liblablgtk-extras-ocaml-dev
-  # get etags from emacs:
-  - sudo apt-get install -y emacs
+cache:
+  apt: true
+addons:
+  apt:
+    packages:
+      - ocaml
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -6,25 +6,32 @@ sudo: 		required
...
-before_install:
-  - sudo apt-get update -qq
-install:
-  - sudo apt-get install -y ocaml ocaml-nox ocaml-native-compilers ocaml-findlib
-  - sudo apt-get install -y camlp5 camlp4-extra time libgtk2.0 libgtksourceview2.0
-  - sudo apt-get install -y liblablgtk-extras-ocaml-dev
-  # get etags from emacs:
-  - sudo apt-get install -y emacs
+cache:
+  apt: true
```
DanGrayson: What does this line do?

abooij: [nothing](https://github.com/travis-ci/travis-ci/issues/5876), apparently.

DanGrayson: How about removing lines 9-10, in that case?

abooij: Done!

DanGrayson:
lgtm

benediktahrens:
lgtm

benediktahrens:
This PR will be in conflict with #785 .

palmskog:
@benediktahrens if you or @DanGrayson merge this PR before #785, I will update #785 to incorporate the changes here. They are quite orthogonal, although some syntax changes overlap.

PR #798, opened 2017-12-05
This is not MP, so don't call it that

abooij:
Markov's principle says that from the double negation of the existence of some element, we can get some element. In other words, if it's impossible that some element _doesn't_ exist, then we can get its existence. It translates negative information into positive information.

What this code proves is that if there _exists_ some element, then we can _find_ it, which is a fundamentally different claim. I have renamed it to BoundedSearch.

I'm sorry for the confusion I may have caused.

Edit: in fact I now realize that even the branch name `rename-markov` is inappropriate, as I am not renaming markov's principle, nor am I renaming something that implements it. Sorry once again. Let's get rid of this quickly.


DanGrayson:
lgtm

Don't worry about names of branches -- they are evanescent.

benediktahrens:
lgtm

PR #799, opened 2017-12-05
rename isconnectedunit to isProofIrrelevantUnit

### LINKED ISSUE #758: isconnectedunit
The name of this lemma should be isProofIrrelevantUnit, instead, since connectedness is a different property.  It's in Foundations.

```
Lemma isconnectedunit : ∏ x x' : unit, x = x'.
Proof.
  intros. induction x. induction x'. apply idpath.
Defined.
```

siddharthist:
Fixes #758


DanGrayson:
Thanks!  We have to start getting people "assigned" to the issues that they work on.

lgtm

the only error comes from the test for not modifying Foundations

PR #801, opened 2017-12-06
Users with their own Coq don't want to set coq-prog-name

siddharthist:
Either we should put this in the documentation, or somehow make it depend on the value of `BUILD_COQ`

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -182,6 +182,12 @@ files of UniMath is opened in emacs; in order to use them more generally, such
 as from the command line,, then add the full path for the directory
 ```./sub/coq/bin``` to your ```PATH``` environment variable, or set the emacs
 variable ```coq-prog-name``` in your emacs initialization file, ```.emacs```.
+If you don't use the version of Coq built by ```make```, then you might wish to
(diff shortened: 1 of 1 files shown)


DanGrayson:
Thanks, that's an important point!

But we need a better way to do it than by having the user edit one of the project files, for then the edited file will eventually be committed accidentally.

So, as you suggest, maybe instead the makefile should create `.dir-locals.el`, with its contents dependent appropriately on the setting of the make variable BUILD_COQ.  And then we have to think about detecting a change in the setting, so the file can be remade.  Perhaps have it dependent on the file `build/Makefile-configuration`?  Except that file, by default, does not exist.  So maybe have the makefile create it, if needed, by copying `build/Makefile-configuration-template`?  

There is also the variable `COQBIN` in the makefile, which should be taken into account.

I'll close this for now until we have a better solution.  It might be good to open an issue.

siddharthist:
@DanGrayson Will do, thanks! I think you're right that we need a more permanent solution. Many of these questions evaporate if we just require users to have Coq 8.6+...

PR #803, opened 2017-12-07
Create CONTENTS.md

benediktahrens:


abooij:
It would be useful to link to this new file from the readme, perhaps after merging #802.

siddharthist:
I feel like we should eventually have this information in the `coqdoc`, and host that online.

abooij:
@siddharthist, Benedikt and I quickly looked at using coqdoc, but we concluded this was by far the easiest solution right now.

PR #804, opened 2017-12-07
Add a link to CONTENTS from README

benediktahrens:
Add a

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,8 @@ Contents
 The [UniMath subdirectory](UniMath/) contains various packages of formalized
 mathematics. For more information about its contents, see the `README` (or `README.md`) in each package.
 
+To get started, see [CONTENTS](CONTENTS.md).
+
 Some scientific articles describing the contents of the UniMath library are listed in the 
 [wiki](https://github.com/UniMath/UniMath/wiki/Documentation:-Articles-describing-UniMath).
 


README.md
```diff
@@ -16,6 +16,8 @@ Contents
 The [UniMath subdirectory](UniMath/) contains various packages of formalized
 mathematics. For more information about its contents, see the `README` (or `README.md`) in each package.
 
+To get started, see [CONTENTS](CONTENTS.md).
```
abooij: "For more information about its contents, see the README in each package. To get started, see CONTENTS."

This reads a bit strange: "For the contents, see the README. To get started, see CONTENTS."

DanGrayson: Yes, the original idea was that each package would display its own contents, as desired.

benediktahrens:
On 12/08/2017 09:45 PM, Daniel R. Grayson wrote:
> DanGrayson commented on this pull request.
> 
> 
> 
>> @@ -16,6 +16,8 @@ Contents
>   The [UniMath subdirectory](UniMath/) contains various packages of formalized
>   mathematics. For more information about its contents, see the `README` (or `README.md`) in each package.
>   
> +To get started, see [CONTENTS](CONTENTS.md).
> 
> Yes, the original idea was that each package would display its own contents, as desired.
> 

The problem with that is that, in the repository, the packages are 
ordered not logically, but alphabetically.

For the uninitiated, there is no clue that they should start by reading 
Foundations.
(I don't believe that we should rely on people following the "Require" 
statements back to Foundations.)

However, I do not like the idea of a manually maintained central 
overview file myself either. So I would be happy about any suggestions.

DanGrayson:
The makefile knows the logical order of the list of packages.  Instead of CONTENTS.md, we could have a file PACKAGES.md, that lists the packages in logical order, with a 1-line comment about the purpose, and a pointer to the README file of the package.  The file could be created automatically, but then checked in to git.

abooij:
@DanGrayson, would you know how to set this up? Otherwise, I think this document is already very helpful.

benediktahrens:
Maybe this page is more suitable for the wiki than for the repository?

DanGrayson:
No, if the table of contents of packages is in the wiki, then someone has to manually update it every time a package is added.

DanGrayson:
This has been subsumed by #861, closing.

PR #806, opened 2017-12-08
update INSTALL instructions for Mac OS X

DanGrayson:
... so that now we depend on opam to get an appropriate version of the
Ocaml suite for use in compiling the version of Coq we have chosen to
build UniMath with.

The problem is that now Homebrew installs a version of Ocaml that is too
modern.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -6,24 +6,50 @@ version of `bash` on your system.
 
 ## Preparing for the installation under Mac OS
 
-Under Mac OS X, the most convenient way to
-do that is with "Homebrew", available from http://brew.sh/, with the following
-command:
+Under Mac OS X, the most convenient way to do that is with "Homebrew",
+available from http://brew.sh/, with the following command:
 
 ```bash
-$ brew install objective-caml camlp5 camlp4 lablgtk bash
+$ brew install bash opam
+$ opam init --no-setup --compiler=4.02.1
+$ opam install --yes lablgtk camlp5 ocamlfind
 ```
 
-Also install "ocamlfind" using "homebrew" with the following commands.
+(We choose version 4.02.1 of ocamlc above, because it can successfully compile
+Coq 8.6.1.)
+
+Now arrange for the programs installed by opam to be available to the currently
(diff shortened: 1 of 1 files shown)


DanGrayson:
I wonder if this looks okay to @palmskog and any other Mac OS X users out there.

palmskog:
@DanGrayson @benediktahrens although I don't use Mac OS X, here are some comments on the instructions (if you want I could submit a PR with changes in this direction):

- I would recommend OCaml 4.02.3 rather than 4.02.1; a [lot](http://caml.inria.fr/pub/distrib/ocaml-4.02/notes/Changes) of bugs were fixed in on the 4.02 branch. In many projects, our baseline is 4.02.3, since it generally compiles newer OCaml programs (with immutable strings) without problems.

- The `lablgtk` package is only necessary if one wants to use `CoqIDE`, and additionally requires development headers for the GTK library version 2.0 - which cannot be installed via OPAM. As it happens, OPAM has a package called `depext`, which uses some heuristics to derive installation instructions for packages with OPAM-external dependencies. For example, on Ubuntu 16.04, `opam depext labgtk` tells me that I need to install `libgtk2.0-dev`.

DanGrayson:
Thanks, I'll take a look.

Some of our users will want to use CoqIDE, so, since we build Coq for them, we also optionally build CoqIDE for them.

palmskog:
@DanGrayson for people who use ProofGeneral, a link to the [`company-coq`](https://github.com/cpitclaudel/company-coq) extensions could be relevant. It is very popular among computer scientists who use Coq, and gives command/tactic completion and other interface productivity features.

DanGrayson:
Okay, thanks, see #807 

CoqIDE compiles fine with opam's lablgtk still.

benediktahrens:
@DanGrayson : the Coq developers are preparing a compatibility commit for Coq 8.6 with ocaml 4.06.
In this case, I would very much prefer the previous installation instructions for Mac OS, since they seemed much simpler.

Can we revert this commit? Or have both ways in the manual?

PR #807, opened 2017-12-09
update INSTALL to use coq 4.02.3

DanGrayson:

PR #808, opened 2017-12-09
Provide two methods of preparing MacOS for install

benediktahrens:
bring back the former installation instructions, as they are *much* simpler.

To be merged when we have upgraded the coq submodule to a commit compatible with ocaml 4.06

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -6,6 +6,33 @@ version of `bash` on your system.
 
 ## Preparing for the installation under Mac OS
 
+### First method (recommended for beginners)
+
+Under Mac OS X, the most convenient way to
+do that is with "Homebrew", available from http://brew.sh/, with the following
+command:
+
+```bash
+$ brew install objective-caml camlp5 camlp4 lablgtk bash
+```
+
+Also install "ocamlfind" using "homebrew" with the following commands.
+
+```bash
+$ brew tap mht208/formal
+$ brew install ocaml-findlib
+```
+
+Under Mac OS X, you may obtain ProofGeneral from http://proofgeneral.inf.ed.ac.uk/.
+It comes with installation instructions.  Your version of emacs determines
+which version of ProofGeneral you need, roughly, so some experimentation may be
+required; you may even need the current development version if your emacs is
+recent.
+
+Now proceed with "Installing UniMath under Mac OS or Linux" below.
+
+### Second method (allows more flexibility, but is more involved than first method)
+
 Under Mac OS X, the most convenient way to do that is with "Homebrew",
 available from http://brew.sh/, with the following command:
 


DanGrayson:
Hmm?  In #807 we switch to having the user use opam to install ocaml, so we can obtain any desired version of it.  This pull request conflicts with that idea.

benediktahrens:
On 12/09/2017 01:44 PM, Daniel R. Grayson wrote:
> Hmm?  In #807 we switch to having the user use opam to install
> ocaml, so we can obtain any desired version of it.  This pull
> request conflicts with that idea.
> 


There is no conflict.

The opam way seems quite complicated to me.

The Coq devs kindly agreed to provide a version of 8.6 that compiles 
with ocaml 4.06.
That allows the use of the previous way of installing UniMath on Mac OS.

We retain the opam way in the manual, and may in the future present this 
as the recommended way to install UniMath.
For the time being, the previous way seems simpler to me.

DanGrayson:
More to the point is that the version of ocaml currently installed by brew will not compile Coq 8.6, because it leads to this type mismatch error:
```
"/usr/local/bin/ocamlfind" opt -rectypes -dtypes -w -3-52-56  -I config -I lib -I kernel -I kernel/byterun -I library -I proofs -I tactics -I pretyping -I interp -I stm -I toplevel -I parsing -I printing -I intf -I engine -I ltac -I tools -I tools/coqdoc -I plugins/omega -I plugins/romega -I plugins/micromega -I plugins/quote -I plugins/setoid_ring -I plugins/extraction -I plugins/fourier -I plugins/cc -I plugins/funind -I plugins/firstorder -I plugins/derive -I plugins/rtauto -I plugins/nsatz -I plugins/syntax -I plugins/decl_mode -I plugins/btauto -I plugins/ssrmatching -I plugins/ltac -I "/usr/local/Cellar/camlp5/7.03_1/lib/ocaml/camlp5" -thread -g    -c lib/pp_control.ml
File "lib/pp_control.ml", line 61, characters 22-33:
Error: This expression has type bytes -> int -> int -> unit
       but an expression was expected of type string -> int -> int -> unit
       Type bytes is not compatible with type string 
```

benediktahrens:
On 12/09/2017 01:53 PM, Daniel R. Grayson wrote:
> More to the point is that the version of ocaml currently installed by brew will not compile Coq 8.6, because it leads to this type mismatch error:
> ```
> "/usr/local/bin/ocamlfind" opt -rectypes -dtypes -w -3-52-56  -I config -I lib -I kernel -I kernel/byterun -I library -I proofs -I tactics -I pretyping -I interp -I stm -I toplevel -I parsing -I printing -I intf -I engine -I ltac -I tools -I tools/coqdoc -I plugins/omega -I plugins/romega -I plugins/micromega -I plugins/quote -I plugins/setoid_ring -I plugins/extraction -I plugins/fourier -I plugins/cc -I plugins/funind -I plugins/firstorder -I plugins/derive -I plugins/rtauto -I plugins/nsatz -I plugins/syntax -I plugins/decl_mode -I plugins/btauto -I plugins/ssrmatching -I plugins/ltac -I "/usr/local/Cellar/camlp5/7.03_1/lib/ocaml/camlp5" -thread -g    -c lib/pp_control.ml
> File "lib/pp_control.ml", line 61, characters 22-33:
> Error: This expression has type bytes -> int -> int -> unit
>         but an expression was expected of type string -> int -> int -> unit
>         Type bytes is not compatible with type string
> ```
> 



"To be merged when we have upgraded the coq submodule to a commit 
compatible with ocaml 4.06" in the description of the PR refers to that 
- see https://github.com/coq/coq/pull/6348.

DanGrayson:
Oh, I didn't notice that!

By the way, it's possible during installation of opam to let opam add lines to the user's .profile so that at the next login, the opam program are found.  But I suppressed that with the option `--no-setup` so the user would have more control.  If simplicity is desired, we could reverse that.

benediktahrens:
On 12/09/2017 02:04 PM, Daniel R. Grayson wrote:
> Oh, I didn't notice that!
> 
> By the way, it's possible during installation of opam to let opam add
> lines to the user's .profile so that at the next login, the opam
> program are found.  But I suppressed that with the option
> `--no-setup` so the user would have more control.  If simplicity is
> desired, we could reverse that.
> 


I don't have an opinion on that. If we have the two different ways of 
installing ocaml on Mac OS, then simplicity will not necessarily be a 
goal for the more advanced of the two ways.


For the time being, I would like something that has proved to work - in 
particular since I have no clue about Mac OS myself.

This is why I don't want to dismiss the previous install method, and 
would be grateful if we kept that.


After the school we might reconsider the installation setup: we might 
make BUILD_COQ=no the default, and leave it up to the Coq community to 
provide good install instructions (which they do, I think!).
After all, that's one of the benefits of using un-patched Coq.

DanGrayson:
On Mac OS X, we can install Coq 8.7.0 most easily using brew.

PR #809, opened 2017-12-09
Add `gtk+` dependency in Install.md

ilyalukinykh:


DanGrayson:
Thank you!  But this is already covered in #807, prompted by questions from @palmskog .

PR #811, opened 2017-12-10
upgrade to coq 8.6 compat commit

benediktahrens:
Upgrade to Coq tip of v8.6 branch, which includes a patch for compilability with ocaml 4.06.

Thanks to Hugo Herbelin, Maxime Denes, and the other Coq developers involved.


DanGrayson:
Now everything works on Mac OS X using the old instructions (no opam).

PR #814, opened 2017-12-11
Define natural numbers objects as an initial algebra

siddharthist:
The natural numbers are initial for the functor X ↦ 1 + X. I _think_ this is somewhat different in focus from the `SubstitutionSystems` code, but someone should double check me on that. This defines a natural numbers object in any category, and derives their usual universal property (just the beta rules so far, no eta yet).

### DIFF
--- a/UniMath/CategoryTheory/FunctorAlgebras.v
+++ b/UniMath/CategoryTheory/FunctorAlgebras.v
@@ -11,18 +11,19 @@ Extended by: Anders Mörtberg. October 2015
 
 Contents :
 
-- category of algebras of an endofunctor
+- Category of algebras of an endofunctor
 
-- Saturated if base precategory is
+- This category is saturated if base precategory is
 
 - Lambek's lemma: if (A,a) is an inital F-algebra then a is an iso
+
+- The natural numbers are intial for X ↦ 1 + X
 
 ******************************************************************)
 
 Require Import UniMath.Foundations.PartD.
 Require Import UniMath.Foundations.Propositions.
 Require Import UniMath.Foundations.Sets.
-
 Require Import UniMath.MoreFoundations.Tactics.
 
 Require Import UniMath.CategoryTheory.Categories.
@@ -30,7 +31,13 @@ Require Import UniMath.CategoryTheory.functor_categories.
 Require Import UniMath.CategoryTheory.whiskering.
 Require Import UniMath.CategoryTheory.limits.initial.
 
+(* The following are used for examples *)
+Require Import UniMath.CategoryTheory.limits.terminal.
+Require Import UniMath.CategoryTheory.limits.bincoproducts.
+
 Local Open Scope cat.
+
+(** ** Category of algebras of an endofunctor *)
 
 Section Algebra_Definition.
 
@@ -46,6 +53,19 @@ Local Coercion alg_carrier : algebra_ob >-> ob.
 
 Definition alg_map (X : algebra_ob) : F X --> X := pr2 X.
 
+(** A morphism of an F-algebras (F X, g : F X --> X) and (F Y, h : F Y --> Y)
+    is a morphism f : X --> Y such that the following diagram commutes:
+    
+    <<
+         F f
+    F x ----> F y
+    |         |
+    | g       | h
+    V         V
+    x ------> y
+         f
+    >>
+ *)
 Definition is_algebra_mor (X Y : algebra_ob) (f : alg_carrier X --> alg_carrier Y) : UU
   := alg_map X · f = #F f · alg_map Y.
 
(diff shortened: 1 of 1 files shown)


UniMath/CategoryTheory/FunctorAlgebras.v
```diff
@@ -335,4 +356,182 @@ Proof.
...
+        diagram commute:
+
+               inlN         identity 1 + nat_ob_rec
+            1 -----> 1 + N -------------------------> 1 + X
+                       |                                |
+             alg_map N |                                | alg_map X
+                       V                                V
+                       N   -------------------------->  X
+                                   nat_ob_rec
+
```
mortberg: Please use verbatim here as well

UniMath/CategoryTheory/FunctorAlgebras.v
```diff
@@ -335,4 +356,182 @@ Proof.
...
+      so by the η-rule (UMP) for the coproduct, we can assume that it
+      arises from a pair of maps [nat_ob_z,nat_ob_s] by composing with
+      coproduct injections.
+
+                  in1         in2
+               1 ----> 1 + N <---- N
+               |         |         |
+      nat_ob_z |         | alg_map | nat_ob_s
+               |         V         |
+               +-------> N <-------+
```
mortberg: Please put this in verbatim

mortberg:
I'm a little confused by this PR. Am I correct in understanding that you're assuming that `nat_ob` is the initial algebra for the functor and then derive some results from that? Wouldn't it be possible to mimic what is being done in the `CategoryTheory/Inductives/` files and actually construct this initial algebra (assuming suitable conditions on C)?

UniMath/CategoryTheory/FunctorAlgebras.v
```diff
@@ -335,4 +356,182 @@ Proof.
...
+
+  (** F on objects: X ↦ 1 + X *)
+  Definition F_compute1 : ∏ c : C, F c = BinCoproductObject _ (bc 1 c) :=
+    fun c => (idpath _).
+
+  (** F on arrows: f ↦ [identity 1, f] *)
+  Definition F_compute2 {x y : C} : ∏ f : x --> y, # F f = (identity 1) + f :=
+    fun c => (idpath _).
+
+  Definition nat_ob : UU := Initial (FunctorAlg F hsC).
```
mortberg: As we discussed, this hypothesis could probably be proved using https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/CocontFunctors.v#L390. You might need additional assumption on C, but I think it would be nice to actually construct this initial algebra and not just assume that it exists.

peterlefanulumsdaine:
This PR looks good to me for merging, modulo the cosmetic addition of verbatim as @mortberg suggests.

Regarding @mortberg’s question about *constructing* NNO’s: I agree that would be nice, but it seems a separate point.  The present PR’s approach — defining an NNO as an initial algebra for the (+ 1) endofunctor, and investigating them when they exist — is completely standard; existence of an NNO is something that’s often taken as a basic assumption (eg in the definition of an elementary topos).  Showing that other natural conditions may imply existence of an NNO is nice, but not necessary to justify/motivate the investigation of NNO’s.

benediktahrens:
Could one avoid the assumption of binary coproducts by not bundling zero and successor?

siddharthist:
@benediktahrens How? Both zero and the successor are defined in terms of the binary coproduct cone. (Honest question, I'm not very familiar with NNOs and would be happy to learn)

siddharthist:
I've made the changes regarding the use of `<<` and `>>`. I'm happy to look into the conditions for when a category has an NNO, perhaps @mortberg could help explain some of that code to me tomorrow.

benediktahrens:
On 12/13/2017 11:19 PM, Langston Barrett wrote:
> @benediktahrens How? Both zero and the successor are defined in terms
> of the binary coproduct cone. (Honest question, I'm not very familiar
> with NNOs and would be happy to learn)
> 

I have the impression that the necessity for binary coproducts comes 
from insisting on formulating the the universal property in terms of 
initial *algebras*, i.e., in insisting on defining a functor X \mapsto X+1.

Why not consider a category an object of which is a triple (X,x : 
1->X,s:X->X)?

But maybe I am missing something obvious.

peterlefanulumsdaine:
@benediktahrens : Yes, NNO’s can be defined without assuming coproducts, just as you describe — this is probably how they’re most commonly defined in the literature (e.g. the definition in Johnstone Sketches of an Elephant, Vol I, A2.5.1).  Eventually I guess it would be good to have that more general definition as our main definition of NNO’s?

On the other hand, it’s also nice (and again quite usual in the literature) that assuming coproducts, NNO’s can be defined more concisely as initial algebras in this way.  So I feel this PR would fine to merge as is, if no-one feels like immediately redoing it with the more general definition, and then the more general definition could be added later.

siddharthist:
Thanks for the review, everyone!

PR #816, opened 2017-12-11
The category of modules over a ring

siddharthist:
This is an update to #736, I believe it corrects the comments there, namely
 * This uses lemmas already in the library where available
 * This places lemmas about algebra in the appropriate package


mortberg:
This should be merged *after* https://github.com/UniMath/UniMath/pull/819. This might need some fixing after that PR  has been merged.

benediktahrens:
Indeed, there is a conflict...

siddharthist:
This is updated, and no longer needs the label!

UniMath/Algebra/Modules/Core.v
```diff
@@ -433,18 +438,35 @@ Defined.
 Definition linearfuncomp {R : rng} {M N P : module R} (f : linearfun M N) (g : linearfun N P) :
   linearfun M P := tpair _ (funcomp f g) (islinearfuncomp f g).
 
-Definition ismodulefun {R : rng} {M N : module R} (f : M -> N) : UU :=
-   (isbinopfun f) × (islinear f).
-
-Lemma isapropismodulefun {R : rng} {M N : module R} (f : M -> N) : isaprop (ismodulefun f).
+Definition isapropislinear {R : rng} {M N : module R} (f : M -> N) :
```
mortberg: I would call this a Lemma

UniMath/Algebra/Modules/Core.v
```diff
@@ -433,18 +438,35 @@ Defined.
...
 Proof.
-   refine (@isofhleveldirprod 1 (isbinopfun f) (islinear f) _ _).
-   exact (isapropisbinopfun f).
    apply (impred 1 _). intro r.
    apply (impred 1 _). intro x.
    apply (setproperty N).
 Defined.
 
+(** The type of linear functions M -> N is a set. *)
+Definition isasetlinearfun {R : rng} (M N : module R) : isaset (linearfun M N).
```
mortberg: I would call this a Lemma as well

UniMath/Algebra/Modules/Core.v
```diff
@@ -433,18 +438,35 @@ Defined.
...
+    intro.
+    apply isapropislinear.
+Defined.
+
+(** **** [modulefun] *)
+
+Definition ismodulefun {R : rng} {M N : module R} (f : M -> N) : UU :=
+   (isbinopfun f) × (islinear f).
+
+Definition isapropismodulefun {R : rng} {M N : module R} (f : M -> N) :
```
mortberg: I would use Lemma and prove this with `exact`

UniMath/Algebra/Modules/Core.v
```diff
@@ -466,6 +488,15 @@ Definition modulefun_to_islinear {R : rng} {M N : module R} (f : modulefun M N)
 Definition modulefun_to_linearfun {R : rng} {M N : module R} (f : modulefun M N) : linearfun M N :=
   linearfunpair f (modulefun_to_islinear f).
 
+
+Definition ismodulefun_comp {R} {M N P : module R} (f : modulefun M N) (g : modulefun N P) :
```
mortberg: Same comment as above

UniMath/Algebra/Modules/Core.v
```diff
@@ -635,3 +669,103 @@ Proof.
...
+Coercion moduleiso_to_modulefun : moduleiso >-> modulefun.
+
+Definition pr1moduleiso {R : rng} {M N : module R} (f : moduleiso M N) :
+  (pr1module M) ≃ (pr1module N) := pr1 f.
+
+Coercion pr1moduleiso : moduleiso >-> weq.
+
+Definition moduleisopair {R} {M N : module R} f is : moduleiso M N := tpair _ f is.
+
+Definition isbinopfuninvmap {R} {M N : module R} (f : moduleiso M N) :
```
mortberg: Lemma

UniMath/Algebra/Modules/Core.v
```diff
@@ -635,3 +669,103 @@ Proof.
...
+   apply (invmaponpathsweq f).
+   rewrite (homotweqinvweq f (op x y)).
+   symmetry.
+   transitivity (op ((pr1moduleiso f) (invmap f x)) ((pr1moduleiso f) (invmap f y))).
+   apply (modulefun_to_isbinopfun f (invmap f x) (invmap f y)).
+   rewrite 2 (homotweqinvweq f).
+   apply idpath.
+Defined.
+
+Definition islinearinvmap {R} {M N : module R} (f : moduleiso M N) :
```
mortberg: Lemma

UniMath/Algebra/Modules/Core.v
```diff
@@ -635,3 +669,103 @@ Proof.
...
+
+Lemma modulefun_unel_uniqueness {R} {M N : module R}
+      {f : pr1module M -> pr1module N} {is: ismodulefun f}
+      (p : f (@unel (pr1module M)) = @unel (pr1module N)) :
+  modulefun_unel (f,,is) = p.
+Proof.
+   apply (setproperty (pr1module N)).
+Defined.
+
+Definition moduleiso'_to_moduleiso_isweq {R} (M N : module R) : isweq (moduleiso'_to_moduleiso M N).
```
mortberg: Lemma

UniMath/Algebra/Modules/Core.v
```diff
@@ -635,3 +669,103 @@ Proof.
...
+Proof.
+   use (gradth _ (moduleiso_to_moduleiso' M N)). intro w.
```
mortberg: I think this should give two subgoals, so please use bullets in the proof script

benediktahrens:
On 12/14/2017 10:20 AM, Anders Mörtberg wrote:
> 
> 
> 
>> +   use tpair.
> +   - exact (pr1 w).
> +   - apply dirprodpair.
> +     + exact (pr1 (pr2 (pr1 w))).
> +     + exact (pr2 w).
> +Defined.
> +
> +Lemma modulefun_unel_uniqueness {R} {M N : module R}
> +      {f : pr1module M -> pr1module N} {is: ismodulefun f}
> +      (p : f (@unel (pr1module M)) = @unel (pr1module N)) :
> +  modulefun_unel (f,,is) = p.
> +Proof.
> +   apply (setproperty (pr1module N)).
> +Defined.
> +
> +Definition moduleiso'_to_moduleiso_isweq {R} (M N : module R) : isweq (moduleiso'_to_moduleiso M N).
> 
> Lemma
> 


Anders, use the "review" mechanism to avoid spamming us with emails!

UniMath/CategoryTheory/categories/modules.v
```diff
@@ -0,0 +1,211 @@
...
+
+Definition mod_precategory_ob_mor : precategory_ob_mor :=
+  precategory_ob_mor_pair (module R) (λ M N, modulefun M N).
+
+Definition mod_precategory_data : precategory_data :=
+  precategory_data_pair
+    mod_precategory_ob_mor (λ (M : module R), (idfun M,, id_modulefun M)) 
+    (fun M N P => @modulefun_comp R M N P).
+
+Definition is_precategory_mod_precategory_data :
```
mortberg: Lemma

siddharthist:
@mortberg Thank you for the review! I agree that using Lemma is better for terms created with tactics, though I'm not convinced of the value of converting one-line terms constructed directly (after a `:=`) to three-line proofs using `exact`, it adds a bit of cruft around the term. 

I very much appreciate the speedy reviews by the UniMath team, but I also wonder if non-codified rules such as when to make proofs opaque, the use of accessor functions, and the use of `Lemma` vs `Definition` would best be put into the style guide, rather than having each contributor learn them through reviews.

UniMath/CategoryTheory/categories/modules.v
```diff
@@ -0,0 +1,211 @@
...
+Definition modules_univalence (M N : mod_precategory) : (M = N) ≃ (moduleiso M N).
+Proof.
+   use weqpair.
+   - exact (modules_univalence_map M N).
+   - exact (modules_univalence_map_isweq M N).
+Defined.
+
+(** Equivalence between isomorphisms and moduleiso in Mod R *)
+
+Lemma iso_isweq {M N : ob mod_precategory} (f : iso M N) : isweq (pr1 (pr1 f)).
```
mortberg: Do you really need both of these pr1's? I would have thought that the coercion from iso to morphism would let you avoid at least one of them?

UniMath/CategoryTheory/categories/modules.v
```diff
@@ -0,0 +1,211 @@
...
+     + apply isapropismodulefun.
+Defined.
+
+Lemma moduleiso_iso (M N : ob mod_precategory) : moduleiso M N -> iso M N.
+Proof.
+   intro f.
+   use isopair.
+   - use tpair.
+     + exact f.
+     + exact (pr2 f).
```
mortberg: Can't you just do `exact f` on line 165 instead of using `tpair` with `f` and `pr2 f`?

UniMath/Algebra/Modules/Core.v
```diff
@@ -635,3 +673,104 @@ Proof.
...
+
+Lemma moduleiso'_to_moduleiso {R} (M N : module R) :
+  moduleiso' M N -> moduleiso M N.
+Proof.
+   intro w.
+   use tpair.
+   - exact (pr1 w).
+   - apply dirprodpair.
+     + exact (pr1 (pr2 (pr1 w))).
+     + exact (pr2 w).
```
mortberg: I think the pr1's and pr2's should either be Coercions (so that you can write `exact w` on line 750) or accessor functions so that you can avoid the chain of `pr1` and `pr2`

siddharthist: I actually think this is okay as it is, this datatype is just defined to show that it is equivalent to the actual `moduleiso`, so I don't think it needs accessors.

siddharthist:
This now defines and uses as many accessors as possible (modulo my comment above), and is updated and merged with the master branch. Let me know if there are any more changes!

UniMath/Algebra/Modules/Core.v
```diff
@@ -635,3 +681,106 @@ Proof.
...
+Defined.
+
+Lemma modulefun_unel_uniqueness {R} {M N : module R}
+      {f : pr1module M -> pr1module N} {is: ismodulefun f}
+      (p : f (@unel (pr1module M)) = @unel (pr1module N)) :
+  modulefun_unel (f,,is) = p.
+Proof.
+   apply (setproperty (pr1module N)).
```
DanGrayson: Can't we do without this trivial lemma?

UniMath/Algebra/Modules/Examples.v
```diff
@@ -73,6 +83,14 @@ Defined.
 Definition ring_is_module (R : rng) : module R :=
   ringfun_module (rigisotorigfun (idrigiso R)).
 
+(** The zero module is the unique R-module structure on the zero group (the
+    group with a single element) *)
+Definition zero_module (R : rng) : module R.
+  refine (unitabgr,, _).
```
DanGrayson: Don't we like to start proofs with `Proof`?

UniMath/Algebra/Modules/Core.v
siddharthist: As you like, I'll remove it.

siddharthist:
I have addressed Dan's comments.

UniMath/Algebra/Modules/Core.v
mortberg: Ok, let's keep it as is

DanGrayson:
lgtm

PR #817, opened 2017-12-12
change definition of module

fpvandoorn:
It was defined as a right-module, but denoted as a left-module. redefined it, so that it is now actually a left-module.


fpvandoorn:
Superseded by #819.

PR #818, opened 2017-12-12
trivial submonoid, total and trivial subgroup

siddharthist:
Construct the trivial submonoid (= {e}), the total subgroup (= G), and the trivial subgroup (= {e}).


UniMath/Algebra/Monoids_and_Groups.v
```diff
@@ -394,11 +394,28 @@ Definition pr1submonoid (X : monoid) : @submonoid X -> hsubtype X := @pr1 _ _.
...
+  split.
+  (* There is probably a way to make this more elegant *)
+  - intros x y.
+    refine (hinhfun2 _ (pr2 x) (pr2 y)).
+    intros x_eq_1 y_eq_1.
+    rewrite x_eq_1.
+    rewrite (@lunax X (pr1 y)).
+    exact y_eq_1.
+  - exact (hinhpr (idpath _)).
+Defined.
```
DanGrayson: Equality is already a proposition in a set, so truncation is not required.  Try this instead:
```

Definition trivialsubmonoid (X : monoid) : @submonoid X.
Proof.
  intros.
  exists (λ x, x = @unel X)%set.
  split.
  - intros b c.
    induction b as [x p], c as [y q].
    simpl in *.
    induction (!p), (!q).
    rewrite lunax.
    apply idpath.
  - apply idpath.
Defined.
```

UniMath/Algebra/Monoids_and_Groups.v
```diff
@@ -1916,6 +1933,28 @@ Definition subgrtosubmonoid (X : gr) : @subgr X -> @submonoid X :=
...
+  intro X.
+  split with (@totalsubtype X).
+  split.
+  - exact (pr2 (totalsubmonoid X)).
+  - exact (fun _ _ => tt).
+Defined.
+
+Definition trivialsubgr (X : gr) : @subgr X.
+  intro X.
+  split with (λ x, ∥x = @unel X∥).
```
DanGrayson: Fix this proof the same way: equality is a proposition in a set.

UniMath/Algebra/Monoids_and_Groups.v
siddharthist: @DanGrayson Thanks! Will do 😄

mortberg:
lgtm

benediktahrens:
How about making the proof parts of those definitions properly opaque using `abstact (...)`?

siddharthist:
@benediktahrens Is there any documentation on the use of `abstract` in UniMath? It seems like there should be some guidelines on what to make opaque and how to do it.

UniMath/Algebra/Monoids_and_Groups.v
```diff
@@ -1921,6 +1937,27 @@ Definition subgrtosubmonoid (X : gr) : @subgr X -> @submonoid X :=
   λ A : _, submonoidpair (pr1 A) (pr1 (pr2 A)).
 Coercion subgrtosubmonoid : subgr >-> submonoid.
 
+Definition totalsubgr (X : gr) : @subgr X.
```
peterlefanulumsdaine: This and the following proof should begin with a `Proof.` line, as described https://github.com/UniMath/UniMath/blob/master/UniMath/README.md.

siddharthist: Yes, thank you!

DanGrayson:
I don't much like `abstract` for simple lemmas such as the one discussed here, because it makes it impossible to step through the proof in a proof assistant.

UniMath/Algebra/Monoids_and_Groups.v
siddharthist: This has been done (see most recent commit)

siddharthist:
I'm unclear on whether I should be using `Opaque`, `abstract`, `Qed`, or none of the above. Should I make further changes to this PR?

DanGrayson:
There seems to be no reason to make `trivialsubmonoid` and `trivialsubgr` opaque, as in the current proposal.  Why not save it for temporary use in proofs where it offers a real time advantage?

By the way, according to the manual, the scope of `Opaque` is limited to the current file.

siddharthist:
Okay, I've removed the `Opaque`.

DanGrayson:
lgtm

peterlefanulumsdaine:
lgtm

PR #819, opened 2017-12-12
Prove that homomorphisms between modules form a module

fpvandoorn:
Supersedes #817.

### DIFF
--- a/UniMath/Algebra/Modules/Core.v
+++ b/UniMath/Algebra/Modules/Core.v
@@ -46,11 +46,6 @@ Defined.
 
 Notation "f + g" := (rngofendabgr_op1 f g) : abgr_scope.
 
-(** the composition below uses the diagrammatic order following the general convention used in
-UniMath *)
-
-Notation "f ∘ g" := (rngofendabgr_op2 g f) : abgr_scope.
-
 (** The underlying set of the ring of endomorphisms of an abelian group *)
 
 Definition setofendabgr (G : abgr) : hSet :=
--- a/UniMath/Algebra/Modules/Core.v
+++ b/UniMath/Algebra/Modules/Core.v
@@ -319,6 +319,21 @@ maponpaths (λ r, pr1setofendabgr r x) (pr1 (pr2 (pr2 (pr2module M))) r s).
 Definition module_mult_1 {R : rng} {M : module R} (r : R) : r * unel M = unel M :=
 pr2 (pr2 (pr2module M r)).
 
+Definition module_mult_unel2 {R : rng} {M : module R} (m : M) : rngunel2 * m = m.
+Admitted.
+
+Definition module_inv_mult {R : rng} {M : module R} (r : R) (x : M) : @grinv _ (r * x) = r * @grinv _ x.
+Admitted.
+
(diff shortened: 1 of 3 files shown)


benediktahrens:
lgtm

UniMath/Algebra/Modules/Core.v
```diff
@@ -41,14 +41,14 @@ Defined.
...
-  apply (monoidfuncomp f g).
+  apply (monoidfuncomp g f).
 Defined.
 
 Notation "f + g" := (rngofendabgr_op1 f g) : abgr_scope.
 
 (** the composition below uses the diagrammatic order following the general convention used in UniMath *)
 
-Notation "f ∘ g" := (rngofendabgr_op2 f g) : abgr_scope.
+Notation "f ∘ g" := (rngofendabgr_op2 g f) : abgr_scope.
```
mortberg: I'm puzzled by this. The comment says that this is in diagrammatic order, but the change breaks this?

UniMath/Algebra/Modules/Core.v
```diff
@@ -41,14 +41,15 @@ Defined.
...
 Defined.
 
 Notation "f + g" := (rngofendabgr_op1 f g) : abgr_scope.
 
-(** the composition below uses the diagrammatic order following the general convention used in UniMath *)
+(** the composition below uses the diagrammatic order following the general convention used in
+UniMath *)
 
-Notation "f ∘ g" := (rngofendabgr_op2 f g) : abgr_scope.
+Notation "f ∘ g" := (rngofendabgr_op2 g f) : abgr_scope.
```
mortberg: I'm puzzled by this, why is this diagrammatic order?

fpvandoorn: This commit also swaps the variables in the definition of `rngofendabgr_op2`, so the composition stays in diagrammatic order (however, this notation is not yet used anywhere, so I could also remove it).

mortberg: I vote for removing it. I'm all for diagrammatic order of composition, but using `o` as the symbol for this seems like a bad idea.

mortberg:
lgtm

siddharthist:
LGTM as well. I have a branch using these lemmas and they are quite helpful!

UniMath/Algebra/Modules/Core.v
```diff
@@ -398,27 +447,191 @@ Defined.
...
+      * intros f. exact (modulebinop_lunax f).
+      * intros f. exact (modulebinop_runax f).
+Defined.
+
+Lemma modulehombinop_inv_ismodulefun {R : rng} {M N : module R} (f : modulefun M N) :
+  ismodulefun (λ m : M, grinv N (pr1 f m)).
+Proof.
+  use tpair.
+  - use mk_isbinopfun. intros x x'. cbn.
+    rewrite (pr1 (pr2 f)). rewrite (pr2 (pr2 (pr1module N))). use (grinvop N).
```
mortberg: It would be nice to not have all these pr1/pr2 and instead have accessor functions

PR #821, opened 2017-12-12
Add full export file for each package (plus maintenance script)

peterlefanulumsdaine:
(This is as much a feature request as a pull request; I’m not sure if this PR is the best way to provide the feature in question, but it seemed clearer to describe it with a proof-of-concept PR than in prose in a standalone issue.)

This PR adds, for each package in the library, a file `Export.v` which re-exports all `.v` files in the package.  It also adds a maintenance script `utils/regenerate-export-files.sh` which auto-regenerates these export files.

The motivation of the export files is so that users working over `UniMath` can easily import a whole package at once using e.g. `Require Import UniMath.CategoryTheory.Export.`, and in particular, can *search* a whole package using Coq’s `Search`, `SearchAbout`, etc., for lemmas whose exact file may not be known in advance.

This PR has several rough edges — listed below.  However, I suggest this PR is already definitely useful enough and hopefully robust enough to be accepted as is, if the maintainers agree.

Rough edges, and possible improvements:

- The `Export.v` files are currently checked in and version-controlled; the regeneration script must be run by hand when necessary.  It would surely better if the files were not version-controlled (but instead git-ignored) and their generation was rolled into the `Makefile`.  This would need to be done by someone who understands the `Makefile` more confidently than I do.

- Each `Export.v` file lists the package files in alphabetical order, not in dependency order.  This is perhaps misleading, though it seems to have no practical consequences.

- The list of files in each package is taken from `git ls-files`, rather than from `.package/files`, since this made the parsing a little easier.  This means the build will break if the export files are regenerated while there are files in a package that are not listed in `.package/files`.

- The list of packages is assumed to be just the subdirectories of `UniMath`, rather than being taken from the list of packages in the Makefile, since this was quite difficult to parse.


peterlefanulumsdaine:
Note: this PR fails Travis checks just because it adds an `Export.v` file within the `Foundations` package.  However, it does not actually touch the code of `Foundations`, so is I think compliant with the intent of the lock on `Foundations`?

mortberg:
This looks good and useful to me.

Ideally there should also be some checks that the Export files are complete, probably by the Makefile, but I think this can be done in a future PR.

JasonGross:
You might want to name them `Exports` rather than `Export`, so as not to clash with the (pseudo?)keyword `Export`.

If you make `util/regenerate-export-files.sh` not touch the file if the one it generates is identical to the existing one, then you might be able to add something like
```Makefile
.PHONY: regenerate-package-exports
regenerate-package-exports:
	bash util/regenerate-export-files.sh

%.d: | regenerate-package-exports
```

You might also need
```Makefile
$(package_export_v_files): | regenerate-package-exports
```

Does that work?

DanGrayson:
The right way to do with be to generate the files from the files `.package/files`, for then `make` would know when to regenerate the files: when those package file list files change.

Notice that I've already started doing this for some packages, but I named the files `All.v`.  That might be a better name, anyway.  In any case, we don't need both.

I'd be happy to take this proposal and adapt it so the files are not committed to github and so the files are generated automatically by `make`.  Then it could use the list of packages in the makefile instead of looking in all the subdirectories.

We don't allow github-committed *.v files in packages that are not on the files list, so that we don't have to worry about a discrepancy between the two concepts.

peterlefanulumsdaine:
@DanGrayson : ah, great — sorry I didn’t find `MoreFoundations.All.v` when preparing this.  Yes, I agree `All.v` is probably a better name.  And I would be very happy for you to take this and adapt it for the files to be autogenerated by `make` from the `.package/files` files — I would try to do that myself, but I’m not experienced enough with makefiles to be confident of getting it right.  So let me know if you would like me to come back to anything else on this PR, otherwise I will leave it for now!

DanGrayson:
I'll do what I said above ( https://github.com/UniMath/UniMath/pull/821#issuecomment-351463339 ) .

DanGrayson:
Supplanted by #878, closing.

PR #826, opened 2017-12-13
Add a comment about limits/cats

mortberg:


peterlefanulumsdaine:
Comment-only, and looks good to me.

PR #828, opened 2017-12-14
Tactic for proving facts about hlevels of types

siddharthist:
This tactic is usful for proving goals of the form `isofhlevel n X`. It dissects commonly used type formers such as direct product, sigma, and pi, and then tries to analyze their components.

It reduces the burden on the programmer to remember the names of such lemmas, and to figure out which ones should be applied, in which branches, how many times, and in what order.

There is plenty of room for improvement, but this is a solid first step. If you all agree, I can add its use to other places in the standard library and/or add more features before merging.

### DIFF
--- a/UniMath/Algebra/Modules/Core.v
+++ b/UniMath/Algebra/Modules/Core.v
@@ -7,6 +7,7 @@ Require Import UniMath.Foundations.PartA.
 Require Import UniMath.Foundations.Preamble.
 Require Import UniMath.Algebra.Domains_and_Fields.
 Require Import UniMath.Foundations.PartD.
+Require Import UniMath.MoreFoundations.Tactics.
 
 (** ** Contents
 - The ring of endomorphisms of an abelian group
@@ -420,8 +421,13 @@ Definition pr1linearfun {R : rng} {M N : module R} (f : linearfun M N) : M -> N 
 
 Coercion pr1linearfun : linearfun >-> Funclass.
 
-Lemma islinearfuncomp {R : rng} {M N P : module R} (f : linearfun M N) (g : linearfun N P) :
-  islinear (funcomp (pr1 f) (pr1 g)).
+Lemma isapropislinear {R : rng} {M N : module R} (f : M -> N) : isaprop (islinear f).
+  unfold islinear.
+  hleveltac.
+  apply (setproperty N).
+Defined.
+
+Definition islinearfuncomp {R : rng} {M N P : module R} (f : linearfun M N) (g : linearfun N P) : islinear (funcomp (pr1 f) (pr1 g)).
 Proof.
   intros r x.
   unfold funcomp.
@@ -438,11 +444,10 @@ Definition ismodulefun {R : rng} {M N : module R} (f : M -> N) : UU :=
 
 Lemma isapropismodulefun {R : rng} {M N : module R} (f : M -> N) : isaprop (ismodulefun f).
 Proof.
-   refine (@isofhleveldirprod 1 (isbinopfun f) (islinear f) _ _).
-   exact (isapropisbinopfun f).
-   apply (impred 1 _). intro r.
-   apply (impred 1 _). intro x.
-   apply (setproperty N).
+  unfold ismodulefun.
+  hleveltac.
+  - exact (isapropisbinopfun f).
+  - exact (isapropislinear f).
 Defined.
 
 Definition modulefun {R : rng} (M N : module R) : UU := ∑ f : M -> N, ismodulefun f.
--- a/UniMath/MoreFoundations/Tactics.v
+++ b/UniMath/MoreFoundations/Tactics.v
@@ -1,8 +1,13 @@
-(************************************************************************
+(** This file contains various useful tactics. *) 
+(** ** Contents:
 
-This file contains various useful tactics
-
-************************************************************************)
+- [unimath_easy]
+- [now]
+- [hSet_induction]
+- [hlevelmatch] - Author: Langston Barrett (@siddharthist)
+- [hleveltac] - Author: Langston Barrett (@siddharthist)
+- [hleveltacS] - Author: Langston Barrett (@siddharthist)
+ *)
 
 Require Import UniMath.MoreFoundations.Foundations.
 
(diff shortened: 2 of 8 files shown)


UniMath/Algebra/Modules/Core.v
```diff
@@ -420,8 +421,13 @@ Definition pr1linearfun {R : rng} {M N : module R} (f : linearfun M N) : M -> N
 
 Coercion pr1linearfun : linearfun >-> Funclass.
 
-Lemma islinearfuncomp {R : rng} {M N P : module R} (f : linearfun M N) (g : linearfun N P) :
-  islinear (funcomp (pr1 f) (pr1 g)).
+Lemma isapropislinear {R : rng} {M N : module R} (f : M -> N) : isaprop (islinear f).
```
siddharthist: Note that this is the only real "content" of this commit, the rest are just using the new tactic in an equivalent way to the original script.

DanGrayson:
The tests should all be put in a separate file named `Tests.v`.

UniMath/MoreFoundations/Tactics.v
```diff
@@ -22,3 +22,161 @@ Tactic Notation "now" tactic(t) := t; unimath_easy.
...
+      <<
+        isofhlevel n (X ⨿ Y)
+      >>
+      unless this is a hypothesis of our context. We try to apply *all*
+      hypotheses in this way, which is not the best, but it works. Someone who
+      knows more about Ltac might want to fix this.
+   *)
+  (*06*)| [ H : (?X -> empty) |- isofhlevel _ ?X ] => apply isapropifnegtrue
+  (*07*)| [ |- isofhlevel _ (?X -> empty) ] => apply isapropneg
+  (*08*)| [ |- isofhlevel _ (?X != ?Y) ] => apply isapropneg
```
DanGrayson: rule 08 will never fire, because its pattern is a specialization of the pattern for rule 07

siddharthist:
@DanGrayson Agreed. Does this tactic seem like generally a decent idea? I didn't want to spend too much more time in case it wasn't worth it. Thanks for taking a look!

DanGrayson:
I've done something similar in `Ktheory/Tactics`, but just for h-level 1:
```
Ltac prop_logic :=
  abstract (intros; simpl;
            repeat (try (apply isapropdirprod);try (apply isapropishinh);apply impred ;intro);
            try (apply isapropiscontr); try assumption) using _L_.
```
I think such tactics are useful, but I would avoid using them in Foundations or MoreFoundations, where it would be better if new users could learn the system by single stepping through the proofs.  Yes, one can print a proof term after the proof is complete, but there are no facilities available for examining the parts of a proof term.

DanGrayson:
By the way, I'm against the automatic use of `isapropneg`, because its proof involves an axiom that impedes computation.  Perhaps remove it from your tactic, to force the user to write it explicitly.

benediktahrens:
Using such automation shifts the burden from the programmer to the reader and the maintainer - whence the reduced burden you perceive. As a maintainer, I do not want to end up with many broken proofs to fix, e.g, upon a Coq upgrade.
I would hence discourage the use of automation.

mortberg:
I'm also skeptical about this kind of automation. It is a bit boring to do these proofs by hand over and over, but as we are doing proof relevant mathematics it is often important to pick a particular proof. If one is relying too much on automation then it is hard to control this.

DanGrayson:
The main obstacle that we have to overcome before there is wide adoption of proof assistants is the tedium of making the proofs.  Given that, we should welcome tactics that make things easier for the user.  

Why would a Coq upgrade break a proof that uses a tactic?  I see no reason.  Admittedly, if the proof does happen to break, it can be harder to debug, because it is hard to guess what the tactic was going to come up with, but one can run the old version and print out the proof term to see what it did, so debugging is not impossible.

benediktahrens:
On 12/16/2017 10:06 PM, Daniel R. Grayson wrote:
> The main obstacle that we have to overcome before there is wide
> adoption of proof assistants is the tedium of making the proofs.
> Given that, we should welcome tactics that make things easier for the
> user.

Whom do you mean when you say "user"? At least three different kinds of
users come to my mind, for two of which the use of automation is an
obstacle rather than a help.

DanGrayson:
The users I have in mind are those mathematicians who publish research papers with proofs in them.  How could automation be an obstacle?

JasonGross:
As another point of reference, in HoTT, we have make use of typeclasses to automatically infer proofs of h-level.  In general, I think that tactics should either be "try this set of things in any order" (e.g., `auto`), or have a clear specification; the specification for automatically inferred h-level proofs that we use is roughly "can be determined by analyzing the type structure".  So we expect that any "clever" proofs of h-level, e.g., those that go via encode-decode or equivalence, will need to be more manual, and any "obvious" proofs of h-level will be picked up automatically by typeclasses.

siddharthist:
I agree that for more complex tactics, one should be careful that their tactics are producing reasonable proof terms. I don't think that tactics are _inherently_ bad for producing consistent and concise proof terms. If you like, I'm happy to test out the proof terms produced by `hleveltac` as opposed to those produced by hand and post the results here.

@mortberg Are there any situations in particular you are worried about? What would it look like for a tactic to choose a "wrong" proof term? 

@benediktahrens How do tactics typically break when Coq is upgraded? 

If we decide not to add tactics to UniMath, there should probably be a thorough discussion as to the tradeoffs. If only certain kinds of tactics are allowed, we should probably make that clear.

mortberg:
Let's say that you're proving that a type is contractible, then there might be many choices for the center of contraction and a tactic might not pick the one you want (in the sense that picking a particular one might simplify other proofs later on). 

One way that we could avoid the behavior I'm concerned about is to only use this tactic in places where we would make the proof term opaque anyway (so ending the proof with `Qed` or wrapping the call to `hleveltac` in `abstract`). This way we would be sure that the automation is not used in any place where we might care about the proof term. If we do it this way then I'm in favour of merging as this tactic seems useful and will relieve the burden of the user in doing boring simple proofs about hlevels.

mortberg:
I would suggest that we have a general discussion about automation in another thread instead of as comments to this PR. I think this is an interesting and important discussion to have and it is definitely something that is worth thinking carefully about before we go further in some direction.

benediktahrens:
On 12/17/2017 09:11 PM, Langston Barrett wrote:
> I agree that for more complex tactics, one should be careful that
> their tactics are producing reasonable proof terms. I don't think
> that tactics are _inherently_ bad for producing consistent and
> concise proof terms. If you like, I'm happy to test out the proof
> terms produced by `hleveltac` as opposed to those produced by hand
> and post the results here.
> 
> @mortberg Are there any situations in particular you are worried
> about? What would it look like for a tactic to choose a "wrong" proof
> term?

I don't think that for this particular kind of goal, the proof terms are 
ever important, except for proofs of `iscontr`.

> @benediktahrens How do tactics typically break when Coq is upgraded?

The more important question here is how Ltac behaviour changes between 
versions of Coq.
I have no experience with that, and I am not eager to find out.

> If we decide not to add tactics to UniMath, there should probably be
> a thorough discussion as to the tradeoffs. If only certain kinds of
> tactics are allowed, we should probably make that clear.

Agreed.
Such a discussion should probably be held on the UniMath mailing list, 
instead of in an issue.

Until then, I'd like to put this PR on hold.

DanGrayson:
Okay, closing.

But why not have this discussion in a github issue?  It seems ideal for that.

PR #832, opened 2017-12-14
Add three points to the style guide

mortberg:
Some additions to the style guide as discussed in https://github.com/UniMath/UniMath/pull/816.

### DIFF
--- a/UniMath/README.md
+++ b/UniMath/README.md
@@ -55,8 +55,9 @@ less fragile and to make the files have a more uniform and pleasing appearance.
 * Start all proofs with `Proof.` on a separate line and end it with
   `Defined.` on a separate line, as this makes it possible for us to generate
   HTML with expansible/collapsible proofs.
-* Use `Definition` for definitions and, `Lemma` or `Theorem` for
-  proofs of properties about the definitions.
+* Use `Lemma`, `Proposition`, or `Theorem` for proofs of propositions;
+  for defining elements of types that are not propositions, use
+  `Definition`.
 * Use Unicode notation freely, but make the parsing conventions uniform across files, and consider
   putting them into a scope.
 * Each line should be limited to at most 100 (unicode) characters.  The
--- a/UniMath/README.md
+++ b/UniMath/README.md
@@ -55,6 +55,8 @@ less fragile and to make the files have a more uniform and pleasing appearance.
 * Start all proofs with `Proof.` on a separate line and end it with
   `Defined.` on a separate line, as this makes it possible for us to generate
   HTML with expansible/collapsible proofs.
+* Use `Definition` for definitions and, `Lemma` or `Theorem` for
+  proofs of properties about the definitions.
 * Use Unicode notation freely, but make the parsing conventions uniform across files, and consider
   putting them into a scope.
 * Each line should be limited to at most 100 (unicode) characters.  The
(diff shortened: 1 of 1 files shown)


siddharthist:
LGTM! Thanks 😄

UniMath/README.md
```diff
@@ -55,6 +55,8 @@ less fragile and to make the files have a more uniform and pleasing appearance.
 * Start all proofs with `Proof.` on a separate line and end it with
   `Defined.` on a separate line, as this makes it possible for us to generate
   HTML with expansible/collapsible proofs.
+* Use `Definition` for definitions and, `Lemma` or `Theorem` for
```
peterlefanulumsdaine: typo: comma should be before “and” (or omitted entirely), not after “and”.

DanGrayson: It's too close to a tautology to say "Use `Definition` for definitions", so let's write things in the other order:

"Use Lemma, Proposition, or Theorem for proofs of propositions; for defining elements of types that are not propositions, use Definition."

UniMath/README.md
```diff
@@ -72,6 +74,15 @@ less fragile and to make the files have a more uniform and pleasing appearance.
   abstracted lemma (whose name typically ends with `_subproof`), because
   its type may vary from one version of Coq to another.  Coq's current behavior is also
   unlikely to be duplicated precisely by a future proof assistant. 
+* Define and use accessor functions for structures instead of chains
+  of `pr1` and `pr2`. This makes the code easier to maintain in the
+  long run (if the structure is rearranged the proofs will still work
+  if the accessor functions are changed accordingly).
+* Define constructor functions starting with `mk_` for structures
```
DanGrayson: Another way to get a flatter proof script with with `repeat split` or `repeat use tpair`, so why is this proposed guideline important?

We currently have an open proposal (#176) to come up with a standard way to construct identifier names,
and this paragraph takes a stance in opposition to the proposal by requiring the form `mk_S`, as opposed to `makeS`.

DanGrayson:
re: "exceptions to this guideline are very widespread (both in colloquial usage and in the library, starting with e.g. the “transport lemmas” in Foundations.PartA)"

Can you point to some examples?

UniMath/README.md
mortberg: I think `use mk_pullback` is more informative than `repeat split`, so I definitely prefer the former. 

I have been consistently using `mk_` for naming constructors in CategoryTheory. If the proposal you refer to is adopted then we should rename all of these. For now I'll just remove "starting with `mk_`" from this sentence. The important point is not what the constructors are called but that they should exist.

UniMath/README.md
mortberg: I like Dan's suggestion so I'll add it. I agree that it might make sense to use Definition for some propositions (like `iscontr`), but I'm not sure if we need to be this detailed in the style guide.

peterlefanulumsdaine:
A few examples of non-propositional “lemmas”, “theorems” etc in the library:

- from `Foundations.PartA`:  `Lemma isrefl_logeq`, `Lemma funcomp_assoc`, and indeed the great majority of “Lemmas” in this file
- from `Ktheory.Representation`:  `Lemma iso_Universal_weq`, `Theorem Representation_to_iso`
- from `CategoryTheory.Adjunctions`: `Lemma are_adjoints_functor_composite`

There are many such examples, in many packages in `UniMath`, from many different authors.  One could argue these are all just lapses of style; but I feel a more convincing explanation is that “use Lemma etc. only for propositions” is only a first approximation to the actual criteria we have in mind when choosing which things to call “Definition” vs “Lemma”/“Theorem”/etc, and that those actual criteria are difficult to fully articulate, and more linguistic/expositional rather than anything as straightforwardly mathematical as “the goal is a proposition”.

Anyhow, I don’t mean to block this PR, just to argue in favour of keeping this point a guideline rather than a strict rule.

DanGrayson:
Yes, `funcomp_assoc` is a good example of that:
```
Lemma funcomp_assoc {X Y Z W : UU} (f : X -> Y) (g : Y -> Z) (h : Z -> W)
: h ∘ (g ∘ f) = (h ∘ g) ∘ f.
```
But I think it shows that, instead having an intuitive criterion that's hard to express, we are writing deceptive mathematics just because it's reminiscent of how mathematics used to look, and we should reform.

PR #831, opened 2017-12-14
Submodule

fpvandoorn:
Define submodules and start on kernels and images.

Also did some cleanup in `Monoids_and_Groups.v`, such as removing `Unset Automatic Introduction.` and making the unique argument of `submonoid` (and similar) explicit.

This pr has conflicts with #816, which I resolved in the branch https://github.com/fpvandoorn/UniMath/tree/submodule-langston.

PR #833, opened 2017-12-14
Module quotients

abooij:
Given a submodule of a module over a ring, construct its quotient module.

This is work in progress.

### DIFF
--- a/UniMath/Algebra/Modules/Quotient.v
+++ b/UniMath/Algebra/Modules/Quotient.v
@@ -49,6 +49,8 @@ Section quotmod_submodule.
 
   Local Notation "x + y" := (@op _ x y).
   Local Notation "x - y" := (@op _ x (grinv _ y)).
+  Local Notation "  - y" := (grinv _ y).
+  Local Notation "0" := (unel _).
   Local Open Scope module_scope.
 
   Definition eqrelsubmodule : eqrel M.
@@ -56,14 +58,20 @@ Section quotmod_submodule.
     use eqrelconstr.
     - exact (λ m m', A (m - m')).
     - intros x y z xy yz.
-      generalize (submoduleadd A (x - y) (y - z) xy yz).
-      now rewrite (assocax M), <- (assocax M (grinv _ y) y), grlinvax, lunax.
+      assert (K := submoduleadd A (x - y) (y - z) xy yz).
+      rewrite (assocax M) in K.
+      rewrite <- (assocax M (grinv _ y) y) in K.
+      rewrite grlinvax in K.
+      rewrite lunax in K.
+      exact K.
     - intros x.
       rewrite (grrinvax M x).
       exact (submodule0 A).
     - intros x y axy.
-      generalize (submoduleinv A (x - y) axy).
-      now rewrite grinvop, grinvinv.
+      assert (K := submoduleinv A (x - y) axy).
+      rewrite grinvop in K.
+      rewrite grinvinv in K.
+      exact K.
   Defined.
 
   Definition module_eqrelsubmodule : module_eqrel M.
--- a/UniMath/Algebra/Modules/Quotient.v
+++ b/UniMath/Algebra/Modules/Quotient.v
@@ -9,8 +9,7 @@ Section bla.
 
   Context {R : rng}
           (M : module R)
-          (A : hsubtype M)
-          {subm : issubmodule A}.
+          (A : submodule M).
 
   Local Notation "x + y" := (@op _ x y).
   Local Notation "x - y" := (@op _ x (grinv _ y)).
@@ -22,15 +21,15 @@ Section bla.
     split.
     - unfold ispreorder. split.
       + unfold istrans. intros x y z xy yz.
-        assert (xyyz := submoduleadd A subm (x - y) (y - z) xy yz).
+        assert (xyyz := submoduleadd A (x - y) (y - z) xy yz).
         rewrite (assocax M) in xyyz.
         rewrite <- (assocax M (grinv _ y) y) in xyyz.
         rewrite grlinvax in xyyz.
         now rewrite lunax in xyyz.
-      + unfold isrefl. intros x. assert (a0 := submodule0 A subm).
+      + unfold isrefl. intros x. assert (a0 := submodule0 A).
         now rewrite <- (grrinvax M x) in a0.
     - unfold issymm. intros x y axy.
-      assert (ainv := submoduleinv A subm (x - y) axy).
+      assert (ainv := submoduleinv A (x - y) axy).
       rewrite grinvop in ainv.
       now rewrite grinvinv in ainv.
   Defined.
@@ -44,7 +43,7 @@ Section bla.
     intros m m' n n' Hmm' Hnn'.
     apply weqpathsinsetquot.
     unfold quotrel in Hmm', Hnn'.
-    generalize (submoduleadd A subm _ _ Hmm' Hnn').
+    generalize (submoduleadd A _ _ Hmm' Hnn').
     assert (H : m + n - (m' + n') = (m - m') + (n - n')).
     {
       rewrite assocax, assocax; apply maponpaths.
@@ -73,7 +72,7 @@ Section bla.
     simpl in H.
     rewrite H.
     assert (H'mm' : A (grinv M (m - m'))).
-    apply (submoduleinv A subm _ Hmm').
+    apply (submoduleinv A _ Hmm').
     rewrite grinvop in H'mm'.
     now rewrite grinvinv in H'mm'.
   Defined.
@@ -183,7 +182,7 @@ Section bla.
       unfold quotrel. simpl.
       rewrite module_inv_mult.
       rewrite <- (module_mult_is_ldistr r m).
-      use (submodulemult A subm). assumption.
+      use (submodulemult A). assumption.
     + use mk_ismonoidfun.
       * use mk_isbinopfun.
         simple refine (setquotuniv2prop quotrel (λ a b, hProppair _ _) _); [use isasetsetquot|].
--- a/UniMath/Algebra/Modules/Quotient.v
+++ b/UniMath/Algebra/Modules/Quotient.v
@@ -59,8 +59,8 @@ Section quotmod_submodule.
       generalize (submoduleadd A (x - y) (y - z) xy yz).
       now rewrite (assocax M), <- (assocax M (grinv _ y) y), grlinvax, lunax.
     - intros x.
-      generalize (submodule0 A).
-      now rewrite <- (grrinvax M x).
+      rewrite (grrinvax M x).
+      exact (submodule0 A).
     - intros x y axy.
       generalize (submoduleinv A (x - y) axy).
       now rewrite grinvop, grinvinv.
--- a/UniMath/Algebra/Modules/Quotient.v
+++ b/UniMath/Algebra/Modules/Quotient.v
@@ -161,13 +161,13 @@ Section bla.
       + unfold iscomm.
         simple refine (setquotuniv2prop quotrel (λ a b, hProppair (quotmod_op a b = quotmod_op b a) _) _).
         * use isasetsetquot.
-        *        intros m m'.
-                 unfold quotmod_op.
-                 etrans.
-                 apply setquotuniv2comm.
-                 rewrite setquotuniv2comm.
-                 apply maponpaths.
-                 apply (commax M).
+        * intros m m'.
+          unfold quotmod_op.
+          etrans.
+          apply setquotuniv2comm.
+          rewrite setquotuniv2comm.
+          apply maponpaths.
+          apply (commax M).
   Defined.
 
   Local Open Scope module_scope.
@@ -263,10 +263,65 @@ Section bla.
           use module_mult_unel2.
   Defined.
 
-  Definition quotmod_module : module_struct R quotmod_abgr.
+  Definition quotmod_mod_struct : module_struct R quotmod_abgr.
   Proof.
     unfold module_struct.
     exact quotmod_rngfun.
+  Defined.
+
+  Definition quotmod : module R.
+  Proof.
+    use modulepair.
+    - exact quotmod_abgr.
+    - exact quotmod_mod_struct.
+  Defined.
+
+  Definition quotmod_quotmap : modulefun M quotmod.
+  Proof.
+    use modulefunpair.
+    - exact (setquotpr quotrel).
+    - split.
+      + use mk_isbinopfun.
+        intros m m'.
+        apply idpath.
+      + intros r m.
+        apply idpath.
+  Defined.
+
+  Definition quotmoduniv
+             (N : module R)
+             (f : modulefun M N)
+             (is : iscomprelfun quotrel f) :
+    modulefun quotmod N.
+  Proof.
+    use modulefunpair.
+    - use setquotuniv.
+      + exact f.
+      + assumption.
+    - split.
+      + use mk_isbinopfun.
+        simple refine (setquotuniv2prop quotrel (λ m n, hProppair _ _) _).
+        * apply (pr2 (pr1 (pr1 (pr1 N)))).
+        * intros m m'.
+          simpl.
+          unfold quotmod_op.
+          rewrite setquotuniv2comm.
+          do 3 rewrite (setquotunivcomm quotrel).
+          apply modulefun_to_isbinopfun.
+      + intros r.
+        simple refine (setquotunivprop quotrel (λ m, hProppair _ _) _).
+        * apply (pr2 (pr1 (pr1 (pr1 N)))).
+        * intros m.
+          simpl.
+          assert (H : r * quotmod_quotmap m = quotmod_quotmap (r * m)).
+          {
+            rewrite <- modulefun_to_islinear.
+            apply idpath.
+          }
+          simpl in H.
+          rewrite H.
+          do 2 rewrite (setquotunivcomm quotrel).
+          apply modulefun_to_islinear.
   Defined.
 
 End bla.
--- a/UniMath/Algebra/Modules/Quotient.v
+++ b/UniMath/Algebra/Modules/Quotient.v
@@ -232,3 +232,23 @@ Section quotmod_def.
   Defined.
 
 End quotmod_def.
+
+(** * Universal property in terms of submodules
+*)
+Section from_submodule.
+
+  Context {R : rng}
+          (M : module R)
+          (A : submodule M).
+
+  Notation "R-mod( M , N )" := (modulefun M N) : module_scope.
+  Local Open Scope module_scope.
+
+  Definition quotmoduniv_submodule
+             (N : module R)
+             (f : R-mod(M, N))
+             (is : iscomprelfun (module_eqrelsubmodule M A) f) :
+    R-mod(quotmod M (module_eqrelsubmodule M A), N) :=
+    quotmoduniv M (module_eqrelsubmodule M A) N f is.
+
+End from_submodule.
--- a/UniMath/Algebra/Modules/Quotient.v
+++ b/UniMath/Algebra/Modules/Quotient.v
@@ -1,0 +1,272 @@
+Require Import UniMath.Foundations.Sets.
+Require Import UniMath.MoreFoundations.All.
+Require Import UniMath.Algebra.Rigs_and_Rings.
+Require Import UniMath.Algebra.Monoids_and_Groups.
+Require Import UniMath.Algebra.Modules.Core.
+Require Import UniMath.Algebra.Modules.Submodule.
+
+Section bla.
+
+  Context {R : rng}
+          (M : module R)
+          (A : hsubtype M)
+          {subm : issubmodule A}.
+
+  Local Notation "x + y" := (@op _ x y).
+  Local Notation "x - y" := (@op _ x (grinv _ y)).
+
+  Definition quotrel : eqrel M.
+  Proof.
+    unfold eqrel. exists (λ x y, A (x - y)).
+    unfold iseqrel.
+    split.
+    - unfold ispreorder. split.
+      + unfold istrans. intros x y z xy yz.
+        assert (xyyz := submoduleadd A subm (x - y) (y - z) xy yz).
+        rewrite (assocax M) in xyyz.
+        rewrite <- (assocax M (grinv _ y) y) in xyyz.
+        rewrite grlinvax in xyyz.
+        now rewrite lunax in xyyz.
+      + unfold isrefl. intros x. assert (a0 := submodule0 A subm).
+        now rewrite <- (grrinvax M x) in a0.
+    - unfold issymm. intros x y axy.
+      assert (ainv := submoduleinv A subm (x - y) axy).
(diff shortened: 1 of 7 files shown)


siddharthist:
It looks like this should be merged after #831, is that right?

abooij:
Yes, this is based on #831.

Currently this PR adds the construction of a quotient module. I will also add the universal property.

The code is somewhat repetitive right now.

abooij:
@fpvandoorn I added the quotient map and the universal property.

UniMath/.dir-locals.el
```diff
@@ -1,13 +1,13 @@
...
+  . ((eval .
 	   (let ((unimath-topdir (expand-file-name (locate-dominating-file buffer-file-name "UniMath"))))
 	     (setq fill-column 100)
 	     (make-local-variable 'coq-use-project-file)
 	     (setq coq-use-project-file nil)
 	     (make-local-variable 'coq-prog-args)
 	     (setq coq-prog-args `("-emacs" "-indices-matter" "-type-in-type" "-w" "-notation-overridden,-local-declaration,+uniform-inheritance,-deprecated-option" "-Q" ,(concat unimath-topdir "UniMath") "UniMath" ))
 	     (make-local-variable 'coq-prog-name)
-	     (setq coq-prog-name (concat unimath-topdir "sub/coq/bin/coqtop"))
+;	     (setq coq-prog-name (concat unimath-topdir "sub/coq/bin/coqtop"))
```
siddharthist: These changes should not be checked in.

abooij: Oops, thanks.

DanGrayson:
Ideally, a PR that depends on another PR would be submitted only after the first one has been accepted, for otherwise we end up reviewing the changes of the first PR twice.

abooij:
@DanGrayson, I submitted this PR so that people know where to find this work. Once it is ready for review (indeed, ideally after merging of the PR that this depends on) I will update this PR.

DanGrayson:
It's easy to find closed unmerged pull requests by a given author, e.g.: https://github.com/UniMath/UniMath/pulls?q=is%3Apr+is%3Aclosed+is%3Aunmerged+author%3ADanGrayson

So shall we close this until it is ready for review?

abooij:
The only work remaining for this is rebasing on top of #831.

abooij:
Rebased, this is now ready for review.

UniMath/Algebra/Modules/Quotient.v
```diff
@@ -0,0 +1,205 @@
+Require Import UniMath.Foundations.Sets.
```
mortberg: Could you please add a header to the file saying what it contains, who wrote it and when. You can take a look at some of the other files for inspiration.

UniMath/Algebra/Modules/Quotient.v
```diff
@@ -0,0 +1,205 @@
...
+Require Import UniMath.Algebra.Modules.Submodule.
+
+Section quotmod_rel.
+
+  Context {R : rng}
+          (M : module R).
+
+
+  Local Open Scope module_scope.
+  Definition isacthrel (E : hrel M) : UU :=
```
mortberg: I guess this is short for `isactionhrel`? I don't think it would hurt to spell it out or even call it `islactionhrel` to clarify that the action is on the left. I know that many of the identifiers in Algebra are extremely short and that there is no global naming convention yet, but I would personally prefer a slightly longer name here so that I don't have to guess what it means.

UniMath/Algebra/Modules/Quotient.v
```diff
@@ -0,0 +1,205 @@
...
+        simpl in H; rewrite H.
+        use eqrelrefl.
+      * unfold unel, quotmod_rngact; simpl; rewrite (setquotunivcomm E).
+        apply maponpaths.
+        apply module_mult_1.
+  Defined.
+
+  Definition quotmod_rngfun : rngfun R (rngofendabgr quotmod_abgr).
+  Proof.
+    unfold rngfun. unfold rigfun.
```
mortberg: You can merge this into one unfold by `unfold rngfin, rigfun.`

UniMath/Algebra/Modules/Quotient.v
```diff
@@ -0,0 +1,205 @@
...
+    use rigfunconstr.
+    - use quotmod_rngmap.
+    - use mk_isrigfun;
+        [ use mk_ismonoidfun | use mk_ismonoidfun ];
+        [ use mk_isbinopfun; intros r r' | | use mk_isbinopfun; intros r r' | ].
```
mortberg: I'm not sure exactly what you're constructing here, but it seems like you have to have to dig very deep to get the goals that you need. I have a feeling that there is a constructor lemma missing?

UniMath/Algebra/Modules/Quotient.v
```diff
@@ -0,0 +1,205 @@
...
+      + use module_mult_0_to_0.
+      + use module_mult_assoc.
+      + use module_mult_unel2.
+  Defined.
+
+  Definition quotmod_mod_struct : module_struct R quotmod_abgr.
+  Proof.
+    unfold module_struct.
+    exact quotmod_rngfun.
+  Defined.
```
mortberg: Wouldn't it be clearer to write:
```
Definition quotmod_mod_struct : module_struct R quotmod_abgr := quotmod_rngfun.
```
?

UniMath/Algebra/Modules/Quotient.v
```diff
@@ -0,0 +1,205 @@
...
+
+  Definition quotmod : module R.
+  Proof.
+    use modulepair.
+    - exact quotmod_abgr.
+    - exact quotmod_mod_struct.
+  Defined.
```
mortberg: How about having a notation for this like `M / A`? (or if that doesn't work `M %/ A`)

UniMath/Algebra/Modules/Quotient.v
```diff
@@ -0,0 +1,205 @@
...
+
+  Definition quotmod_quotmap : modulefun M quotmod.
```
mortberg: It would be nice to have a notation for `modulefun` as well. Maybe something like `Mod(M,N)` could work? One could alternatively use some fancy arrow, but I think I prefer to see which category we are in just by seeing the notation.

UniMath/Algebra/Modules/Quotient.v
```diff
@@ -0,0 +1,205 @@
...
+      + use module_mult_unel2.
+  Defined.
+
+  Definition quotmod_mod_struct : module_struct R quotmod_abgr.
+  Proof.
+    unfold module_struct.
+    exact quotmod_rngfun.
+  Defined.
+
+  Definition quotmod : module R.
```
mortberg: Having the notation `R-mod` for `module R` could look nice

UniMath/Algebra/Modules/Quotient.v
```diff
@@ -0,0 +1,205 @@
+Require Import UniMath.Foundations.Sets.
+Require Import UniMath.MoreFoundations.All.
+Require Import UniMath.Algebra.Rigs_and_Rings.
+Require Import UniMath.Algebra.Monoids_and_Groups.
+Require Import UniMath.Algebra.Modules.Core.
+Require Import UniMath.Algebra.Modules.Submodule.
+
+Section quotmod_rel.
```
mortberg: Could you add a comment starting with `(** * ` before the `Section` saying what it contains? By formatting the comment like this it will display as a section header in the HTML documentation

UniMath/Algebra/Modules/Quotient.v
```diff
@@ -0,0 +1,205 @@
...
+
+  Definition isacthrelmonoideqrel (E : monoideqrel) : isacthrel E := pr2 (pr2 E).
+  Coercion isacthrelmonoideqrel : monoideqrel >-> isacthrel.
+
+  Definition mk_monoideqrel (E : eqrel M) : isbinophrel E -> isacthrel E -> monoideqrel :=
+    λ H0 H1, (E,,(H0,,H1)).
+
+End quotmod_rel.
+
+Section quotmod_submodule.
```
mortberg: Please add a comment here as well saying what the section contains

UniMath/Algebra/Modules/Quotient.v
```diff
@@ -0,0 +1,205 @@
...
+      assert (H : (r * m) - (r * m') = r * (m - m')) by
+          now rewrite module_inv_mult, module_mult_is_ldistr.
+      generalize (submodulemult A r (m - m')).
+      simpl in *.
+      now rewrite H.
+  Defined.
+
+End quotmod_submodule.
+
+Section quotmod_def.
```
mortberg: And a comment here as well please

UniMath/Algebra/Modules/Quotient.v
```diff
@@ -0,0 +1,205 @@
...
+  Context {R : rng}
+          (M : module R).
+
+
+  Local Open Scope module_scope.
+  Definition isacthrel (E : hrel M) : UU :=
+    ∏ r a b, E a b -> E (r * a) (r * b).
+
+  Definition monoideqrel : UU :=
+    total2 (λ E : eqrel M, isbinophrel E × isacthrel E).
```
mortberg: You can use `\sum` instead of `total2` here in order for it to display nicely as a Sigma

UniMath/Algebra/Modules/Quotient.v
mortberg: Thinking a bit more about it I realized that one might want to see the ring R in the notation as well. So maybe `R-mod(M,N)` or `Mod_R(M,N)` could work better

UniMath/Algebra/Modules/Quotient.v
abooij: Are you talking specifcally about this constructor unfolding here, or also the lines starting with `all:` following it?

UniMath/Algebra/Modules/Quotient.v
abooij: I can add that. However, could you point me to a good example of how and where notation should be specified? I'm a little bit worried I will cause trouble for other people down the road. (I am also thinking about the discussion we had in Birmingham in December.)

UniMath/Algebra/Modules/Quotient.v
mortberg: I'm talking about all of the uses of `use`

UniMath/Algebra/Modules/Quotient.v
mortberg: Something along the lines of:

```
Local Notation "M / A" := (quotmod M A) : module_scope.
```

after closing the section should do the trick.  You might have to specify level and associativity as well (they are Reserved at level 40 with left associativity in the Coq standard library).  Every file that depends on this file in which you want to use this notation will have to redefine this notation if you do it this way, this is a bit verbose but it makes it possible for us to avoid the kind of issues that we had in Birmingham.

UniMath/Algebra/Modules/Quotient.v
abooij: I have written this ever so slightly differently, and added explanations in English. Does the new version still need a new constructor lemma?

UniMath/Algebra/Modules/Quotient.v
abooij: But if every file that depends on this file needs to re-introduce the notation, then adding it after the end of this section won't do much, as that's (currently) the end of the file. Or am I missing something?

mortberg: Well, now that you have the definition there should ideally be some theory about the construction in a section at the end of the file. This theory could use this notation. Furthermore, if you have the notation declared it will be easier for someone else to copy-paste it. :)

UniMath/Algebra/Modules/Quotient.v
mortberg: No, I like the new version!

UniMath/Algebra/Modules/Quotient.v
abooij: @mortberg, I have added some notation commits. Is this what you had in mind?

mortberg: Yes, excellent!

DanGrayson:
@abooij : re: " I submitted this PR so that people know where to find this work. "

The ideal way to advertise work is as a link in a comment on the issue that solicits the work.  Also, being assigned to an issue is a good way to advertise who is working on the issue.

abooij:
@DanGrayson This work was mergable, compiled, and was beneficial to the project, at PR time. So, following the style of many other open source projects, I submitted this work as a PR with a WIP tag, to both indicate that the work was at an advanced state, and to allow others to voice their opinion. I am sorry if I broke any of the contribution guidelines of this project.

benediktahrens:
On 01/05/2018 11:51 AM, Daniel R. Grayson wrote:
> @abooij : re: " I submitted this PR so that people know where to find
> this work. "
> 
> The ideal way to advertise work is as a link in a comment on the
> issue that solicits the work.  Also, being assigned to an issue is a
> good way to advertise who is working on the issue.
> 

Not every work is solicited, or announced as an issue, so what you 
consider ideal is not always possible.

I would say that issues announcing future work are much more of a 
nuisance than a PR that is opened not to be merged, but to be discussed.
The reason is that announcing or soliciting work in an issue might 
happen without any work ever being done, that is, it is noise without 
benefit.
In a PR, as @abooij did, some work has been done already, so even if the 
noise is a nuisance, at least it comes with evidence of some achieved work.
In summary, I would like to discourage people from announcing their 
plans in issues.


One facility that would help is allowing PR submitters to add tags to 
their PRs. Unfortunately, only contributors with write access can modify 
tags.
Until that changes, I think that @abooij's way of announcing their work 
is the best one available.

DanGrayson:
The file `Algebra/.package/files` should list Modules/Submodule.v only once.

DanGrayson:
Why not make the notations global?  That was others can use them without copying the code -- all they have to do is to open the appropriate scope.  It won't be annoying to those who don't want to have the notations, as the scope will not be open.

DanGrayson:
A minor comment: It seems a bit unusual to adopt category theory notation such as 
```
Local Notation "R-mod( M , N )" := (modulefun M N) : module_scope.
```
without actually having any categories around, and it doesn't save much clutter, either.

DanGrayson:
Introducing that notation also will tend to make the origin of lemma names more opaque to the reader, such as this one:
``` 
-Lemma isasetmodulefun {R : rng} (M N : module R) : isaset (modulefun M N).
+Lemma isasetmodulefun {R : rng} (M N : module R) : isaset (R-mod(M, N)).
```

DanGrayson:
"Quotient" is a noun, not a verb, so this comment has a grammatical error:
```
Quotienting a module over a ring by a submodule
```

DanGrayson:
Some of the arguments seem a bit convoluted, because of the frequent use of `generalize`.  Compare the way the proof steps go, for example, before and after this change:
```
diff --git a/UniMath/Algebra/Modules/Quotient.v b/UniMath/Algebra/Modules/Quotient.v
index 74714bb..683a32d 100644
--- a/UniMath/Algebra/Modules/Quotient.v
+++ b/UniMath/Algebra/Modules/Quotient.v
@@ -59,8 +59,8 @@ Section quotmod_submodule.
       generalize (submoduleadd A (x - y) (y - z) xy yz).
       now rewrite (assocax M), <- (assocax M (grinv _ y) y), grlinvax, lunax.
     - intros x.
-      generalize (submodule0 A).
-      now rewrite <- (grrinvax M x).
+      rewrite grrinvax.
+      apply submodule0.
     - intros x y axy.
       generalize (submoduleinv A (x - y) axy).
       now rewrite grinvop, grinvinv.
```
Doesn't that make it clearer?

DanGrayson:
It might be a bit clearer to use `exact` instead of `use` when there are no arguments to be filled in, as here:
```
diff --git a/UniMath/Algebra/Modules/Quotient.v b/UniMath/Algebra/Modules/Quotient.v
index 74714bb..5a6f5b1 100644
--- a/UniMath/Algebra/Modules/Quotient.v
+++ b/UniMath/Algebra/Modules/Quotient.v
@@ -149,7 +149,7 @@ Section quotmod_def.
   Proof.
     unfold rngfun, rigfun.
     use rigfunconstr.
-    - use quotmod_rngmap.
+    - exact quotmod_rngmap.
     - use mk_isrigfun.
       (* To show that quotmod_rngmap is a ring action, we show it is a monoid homomorphism with
       respect to both monoids on R. *)
```

DanGrayson:
I have no further comments.

abooij:
Re: applications of `generalize`

In fact, it may not be trivial to get rid of some of these, because Coq won't be able to do the relevant `rewrite`s otherwise. I've gotten rid of the one you mentioned, but am having more trouble with the other ones. This is the technique that we worked out in Birmingham, and that seemd to be most reliable for doing `rewrite`-based proofs.

abooij:
I believe I have processed your comments.

UniMath/Algebra/Modules/Quotient.v
```diff
@@ -0,0 +1,226 @@
...
+
+  Context {R : rng}
+          (M : module R).
+
+
+  Local Open Scope module_scope.
+  Definition isactionhrel (E : hrel M) : UU :=
+    ∏ r a b, E a b -> E (r * a) (r * b).
+
+  Definition monoideqrel : UU :=
```
DanGrayson: The name `monoideqrel` suggests that you are defining an equivalence relation on a monoid, but
actually, you have a relation on a module that is compatible with addition and with scalar multiplication.
Could you choose a better name?  What about `module_eqrel`?

UniMath/Algebra/Modules/Quotient.v
```diff
@@ -0,0 +1,226 @@
...
+  Local Notation "x + y" := (@op _ x y).
+  Local Notation "x - y" := (@op _ x (grinv _ y)).
+  Local Open Scope module_scope.
+
+  Definition eqrelsubmodule : eqrel M.
+  Proof.
+    use eqrelconstr.
+    - exact (λ m m', A (m - m')).
+    - intros x y z xy yz.
+      generalize (submoduleadd A (x - y) (y - z) xy yz).
```
DanGrayson: To remove this use of "generalize", resulting in something more understandable, one may do this:
```
     use eqrelconstr.
     - exact (λ m m', A (m - m')).
     - intros x y z xy yz.
-      generalize (submoduleadd A (x - y) (y - z) xy yz).
-      now rewrite (assocax M), <- (assocax M (grinv _ y) y), grlinvax, lunax.
+      assert (K := submoduleadd A (x - y) (y - z) xy yz); clear xy yz.
+      rewrite (assocax M) in K.
+      rewrite <- (assocax M _ y) in K.
+      rewrite grlinvax in K.
+      rewrite lunax in K.
+      exact K.
     - intros x.
       rewrite (grrinvax M x).
       exact (submodule0 A).
```
It is also a bit more instructive to do one rewrite at a time, so the student can see what's happening.

Other uses of "generalize" can be handled the same way, as well as the places where you introduce all but one variable.

DanGrayson: And the proof above looks better when stepping through it if some more notations are added:
```
 
   Local Notation "x + y" := (@op _ x y).
   Local Notation "x - y" := (@op _ x (grinv _ y)).
+  Local Notation "  - y" := (grinv _ y).
+  Local Notation "0" := (unel _).
   Local Open Scope module_scope.
 
   Definition eqrelsubmodule : eqrel M.
```

UniMath/Algebra/Modules/Quotient.v
```diff
@@ -0,0 +1,226 @@
...
+    use modulefunpair.
+    - exact (setquotpr E).
+    - now use (ismodulefunpair (mk_isbinopfun _)).
+  Defined.
+
+  Definition quotmoduniv
+             (N : module R)
+             (f : R-mod(M, N))
+             (is : iscomprelfun E f) :
+    R-mod(quotmod, N).
```
DanGrayson: This is a fine way to phrase the universal property of the quotient module, but what about including also the standard version, where the quotient module is formed from a module and a submodule?

UniMath/Algebra/Modules/Submodule.v
```diff
@@ -124,3 +124,28 @@ Defined.
...
+          {M : module R}
+          (A : submodule M).
+
+  Local Notation "x + y" := (@op _ x y).
+  Local Notation "x - y" := (@op _ x (grinv _ y)).
+
+  Definition submoduleadd (x y : M) : A x -> A y -> A (x + y).
+  Proof.
+    intros ax ay.
+    exact (pr1 (pr1 (pr1 (pr2 A))) (carrierpair A x ax) (carrierpair A y ay)).
```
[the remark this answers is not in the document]
abooij: These are the accessor functions.

UniMath/Algebra/Modules/Quotient.v
abooij: Ah yeah, that's a typo, I'll fix it.

UniMath/Algebra/Modules/Quotient.v
abooij: Done.

UniMath/Algebra/Modules/Submodule.v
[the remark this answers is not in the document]
DanGrayson: Not exactly -- an accessor function would consist of just a string of projections.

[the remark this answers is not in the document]
abooij: Wait, what is the difference between a "chain" of projections and a "string" of projections?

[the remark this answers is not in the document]
DanGrayson: No difference.  But you also have "carrierpair" getting used in this function, which is not a projection.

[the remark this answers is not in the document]
abooij: Okay. Feel free to take this PR and fix it. I am tired of guessing what the exact code style guide of this project is.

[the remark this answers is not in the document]
DanGrayson: I'm sorry -- perhaps we should loosen the style guide at that point.  The current statement is
```
* Define and use accessor functions for structures instead of chains
  of `pr1` and `pr2`. This makes the code easier to maintain in the
  long run (if the structure is rearranged the proofs will still work
  if the accessor functions are changed accordingly).
```
Your function doesn't deviate much from that...

DanGrayson:
It looks good to me.  The only conflicts are of the following form:
```
<<<<<<< HEAD
  Context {R : ring} {M N : module R} (f : modulefun M N).
||||||| merged common ancestors
  Context {R : rng} {M N : module R} (f : modulefun M N).
=======
  Context {R : rng} {M N : module R} (f : R-mod(M, N)).
>>>>>>> 8e15408f
```
so they should be easy to fix.  Can you do that, Auke?

abooij:
The conflict is a result of the request to change notation in a [previous comment](https://github.com/UniMath/UniMath/pull/833#issuecomment-356444959).

I attempted to rebase, but likely there have been other changes since that I've missed. Let's see what CI says.

abooij:
The CI now passes.

Issue #830, opened 2017-12-14
Definition of `hasProducts` etc are incorrect (or at least mismatched)


peterlefanulumsdaine:
It seems that the definitions of `hasProducts`, `hasFinOrdProducts`, and a few others are incorrect.  They are at least mismatched with the definitions of `hasLims`, `hasEqualizers`, etc.

Specifically, `hasEqualizers` is defined in `CategoryTheory/limits/equalizers.v` as
```
  Definition hasEqualizers : UU := ∏ (y z : C) (f g : y --> z),
      ishinh (Equalizer f g).
```
i.e. “for every pair of arrows, there exists some equalizer”.

On the other hand, `hasProducts` is defined in `CategoryTheory/limits/products.v` by
```
Definition Products := ∏ (ci : ∏ i, C), ProductCone ci.
Definition hasProducts := ishinh Products.
```
i.e. “there exists some function, which for every suitable family of objects gives a product”.

Of these two forms, `hasEqualizers` seems the “correct” one — that is, it corresponds to the standard sense of having (non-chosen) limits of some shape: “for every suitable diagram, there exists some limit”.   (The other form “there exists some function, giving for every suitable diagram a limit” is in general strictly stronger.)

If the maintainers agree that this is not intentional, and that `hasProducts` etc should be fixed to agree with `hasEqualizers` etc, then I will prepare a pull request.

benediktahrens:
Agreed!

Am 14.12.2017 02:26 schrieb "Peter LeFanu Lumsdaine" <
notifications@github.com>:

siddharthist:
At least `hasBinCoproducts` is similarly mismatched, we should probably check all such definitions about having various limits and colimits.

mortberg:
Good catch! I'll fix this

peterlefanulumsdaine:
@mortberg : I had already started preparing a fix for this before I saw your comment above — apologies if you had also been working on this and we have duplicated work!

mortberg:
I'll rebase my PR on top of yours and make a separate one with the cleaning

DanGrayson:
Let's use the "assign" feature of github to advertise the person working to implement the change, to prevent duplication of work.

benediktahrens:
This seems to be resolved, closing.

PR #835, opened 2017-12-14
Fix defs of hasProducts etc, resolving #830

peterlefanulumsdaine:
Mostly just implementing the fix described in #830.  A couple of notes:

- In the fixes, I’ve used the UniCode notation for truncation, i.e. ` ∥ ProductCone ci ∥` not `ishinh (ProductCone ci)` — it seems much more reasonable.  I’ve not however updated the definitions that were already correct (e.g. `hasLims`) to also use the notation.  I can update those as well for consistency if desired.

- I’m think I found all the definitions that need fixing, but it would be good if someone else (Anders?) could also search in case I missed any.


DanGrayson:
lgtm

benediktahrens:
lgtm, to be merged when travis agrees

PR #838, opened 2017-12-14
Clean limits

mortberg:
This is based on https://github.com/UniMath/UniMath/pull/835 so please merge that first.

This PR fixes:
- Remove the prefix (Co)Cone from particular limits ((Bin)(Co)products, etc.) in order to make naming of lemmas and definitions more consistent.
- Add some implicit arguments.
- Use more unicode


peterlefanulumsdaine:
Looks great to me!

DanGrayson:
lgtm

DanGrayson:
Ideally, a PR based on another would not be submitted until the other has been pulled, so the diffs will be relevant.

benediktahrens:
lgtm, to be merged after #835

PR #840, opened 2017-12-15
Poset gives a univalent category.

Chgrau:
If given a preorder over a hSet that is antisymmetric, i.e. a poset, the generated preorder category is univalent.


UniMath/CategoryTheory/categories/preorder_categories.v
```diff
@@ -45,4 +45,46 @@ Defined.
 Definition po_category : category :=
    tpair _ po_precategory po_precategory_has_homsets.
 
-End po_category_def.
\ No newline at end of file
+End po_category_def.
+
+Section poset_univalent_category_def.
+(* A poset is a univalent category *)
```
siddharthist: It is generally a good idea to start comments with `(**` rather than `(*`, because then they show up in the documentation when it is built with `make`.

UniMath/CategoryTheory/categories/preorder_categories.v
```diff
@@ -45,4 +45,46 @@ Defined.
...
+
+Context (POS : Poset).
+
+Definition poset_category : category :=
+  po_category (popair (posetRelation POS) (pr1 (pr2 (pr2 POS)))).
```
benediktahrens: This definition, is it not the same as `po_category`?

Chgrau: I'm not sure I understand the question. The poset_category is defined in terms of po_category, and is the used to refer to that, to avoid write the category in term of po_category each time.

siddharthist: Many structures have what are called "accessor" functions with nicer names so we don't have to write `pr1 (pr2 (pr1 (pr1 ...`. In this case, it looks like there isn't such a function, though, see #841.

UniMath/CategoryTheory/categories/preorder_categories.v
```diff
@@ -45,4 +45,46 @@ Defined.
...
+Lemma poset_category_iso_is_eq {A B : poset_category} (isoAB : iso A B) :
+  A = B.
+Proof.
+  apply (isantisymm_posetRelation POS).
+  apply (morphism_from_iso poset_category A B isoAB).
+  apply (inv_from_iso isoAB).
+Defined.
+
+Lemma poset_category_isweq {A B : poset_category} :
+  isweq (λ p : A = B, idtoiso p).
```
siddharthist: One can eta-reduce this to simply `isweq idtoiso`.

UniMath/CategoryTheory/categories/preorder_categories.v
```diff
@@ -45,4 +45,46 @@ Defined.
...
+    apply setproperty.
+  - intro iso.
+    use eq_iso.
+    apply proofirrelevance.
+    apply propproperty.
+Defined.
+
+Lemma poset_category_is_univalent : is_univalent poset_category.
+  use mk_is_univalent.
+  - intros a b.
```
siddharthist: You can use `intros ? ?.` if you don't use the names again.

benediktahrens:
On 12/15/2017 12:26 PM, Christian Graulund wrote:
> Chgrau commented on this pull request.
> 
> 
> 
>> \ No newline at end of file
> +End po_category_def.
> +
> +Section poset_univalent_category_def.
> +(* A poset is a univalent category *)
> +
> +Context (POS : Poset).
> +
> +Definition poset_category : category :=
> +  po_category (popair (posetRelation POS) (pr1 (pr2 (pr2 POS)))).
> 
> I'm not sure I understand the question. The poset_category is defined in terms of po_category, and is the used to refer to that, to avoid write the category in term of po_category each time.
> 


I am suggesting proving the lemma that I wrote down, and use that to 
give the map from posets to univalent categories.

The lemma says a bit more than what there is right now.

Chgrau:
One comment to the above is that we also require that the relation is over a hSet. So it is not enough that G is antisymm. Using Poset captures both of these notion. I am not sure if:

    is_univalent (po_category (hrel X) ) -> is_antisymmetric (hrel X) /\ isaset(X).

is true.

UniMath/CategoryTheory/categories/preorder_categories.v
Chgrau: Coq cant infer the category if I do this. Is there someway around this?

siddharthist: It might be the case that idtoiso takes implicit arguments, in which case you can prefix it with an `@` and provide them. However, I also think this is an OK way to do it.

PR #842, opened 2017-12-16
Accessor functions for relations

### LINKED ISSUE #841: Accessor functions for posets
Currently [the definition of a poset](https://github.com/UniMath/UniMath/blob/9747e00e833367369ce5b17b8f55155c65882d91/UniMath/Foundations/Sets.v#L797-L802) does not allow for extracting the second projection (the underlying `PartialOrder`), with leads to a lot of `(pr1 (pr2 (pr2...` (see `Sets.v` and #840). These should be replaced by accessor functions.

Chgrau:
Added several accessor functions for standard properties of preorders and partial orders. This should also resolve #841 . I've worked from the notation the you should never write more than one pr1/pr2. The rest should follow from previously defined accessor functions.


UniMath/Foundations/Sets.v
```diff
@@ -779,13 +779,51 @@ Definition popair {X : UU} (R : hrel X) (is : ispreorder R) : po X
 Definition carrierofpo (X : UU) : po X -> (X -> X -> hProp) := @pr1 _ ispreorder.
 Coercion carrierofpo : po >-> Funclass.
 
+Lemma ispreorder_po {X : UU} (PO : po X) : ispreorder PO.
+Proof.
+  intros X PO. exact (pr2 PO).
+Defined.
```
benediktahrens: Maybe this could be given more concisely as
```
Lemma ispreorder_po {X : UU} (PO : po X) : ispreorder PO := pr2 PO.
```
thus using one (or maybe two) line(s) instead of four
Same comment for the other definitions of that kind.

mortberg:
I'm not sure if this is the place to say this, but the name `po` is very short and not very informative... Spelling it out as `preorder` would probably be a good idea.

Chgrau:
I agree that `po` is a bad name. The question is then if the should be `preorder` or `PreOrder`. Is there a naming convention?

Chgrau:
Similarly, there `carrierofpo` is the relation and `carrierofPreorderedSet` is the underlying set. The same is true for `carrierofposet` and `carrierofPartialOrder`

mortberg:
I don't think there is any consensus on what the global naming convention should be for UniMath yet. But it would be good to strive for at least having the same naming convention inside packages, and definitely to have it in individual files. However this is not the case for Foundations/Sets.v as there is for example `ispreorder` and `isPartialOrder`... So if it would be up to me I would expand `po` to `PreOrder` to match `PartialOrder` and hence make the naming a little more consistent in this file.

DanGrayson:
This is a change to Foundations, so it should be left to me.

DanGrayson:
I'll take care of the issues resolved by this pull request later on, when Foundations is quiescent.  (Currently I'm trying to get the use of type-in-type working.)

Closing.

Issue #792, opened 2017-12-03
Rename gradth


mortberg:
There is a very useful lemma called `gradth` in PartA:

https://github.com/UniMath/UniMath/blob/master/UniMath/Foundations/PartA.v#L1772

The name is not very good as it doesn't give any idea about what it actually proves. How about renaming it to `iso_to_weq`?

rmatthes:
I thought this was rather an insider joke which therefore survived for so long. The instance for hSet used to be called `iso_set_isweq`, as seen [here](https://github.com/UniMath/UniMath/commit/e204c4).

DanGrayson:
Good idea. Probably `iso_isweq` is a better name.

To maintain compatibility with publications, we should also have something like
```
Definition gradth := iso_isweq.
```
just afterward.

siddharthist:
If you have [`ag`](https://github.com/ggreer/the_silver_searcher) installed, here's a little `bash` function to find-and-replace over a whole project:
```bash
sed_recurse() { ag -g '.*' -0 | xargs -0 sed -i "$1" }
```
You just feed it a `sed` regex like so: `sed_recurse 's/foo/bar/'`. Just make sure not to use it in the wrong directory 😄

DanGrayson:
I prefer using this function in emacs, since it will run over just the relevant project files, and it will encounter the files in the logical order:
```
M-R runs the command tags-query-replace (found in global-map), which is an
interactive autoloaded compiled Lisp function in ‘etags.el’.

It is bound to M-R, <menu-bar> <edit> <replace> <tags-repl>.

(tags-query-replace FROM TO &optional DELIMITED FILE-LIST-FORM)

Do ‘query-replace-regexp’ of FROM with TO on all files listed in tags table.
Third arg DELIMITED (prefix arg) means replace only word-delimited matches.
If you exit (C-g, RET or q), you can resume the query replace
with the command M-,.
Fourth arg FILE-LIST-FORM non-nil means initialize the replacement loop.

If FILE-LIST-FORM is non-nil, it is a form to evaluate to
produce the list of files to search.

See also the documentation of the variable ‘tags-file-name’.
```

siddharthist:
@DanGrayson That's a great trick! Emacs is a wonderful piece of software.

PR #848, opened 2017-12-18
Rename gradth to iso_isweq

mortberg:
This resolves https://github.com/UniMath/UniMath/issues/792.

I also renamed `weqgradth` to `iso_weq` and added:

```
(** This is kept to keep compatibility with publications that uses the name "gradth" for the "grad
    theorem". *)
Definition gradth {X Y : UU} (f : X -> Y) (g : Y -> X)
        (egf: ∏ x : X, g (f x) = x)
        (efg: ∏ y : Y, f (g y) = y) : isweq f := iso_isweq f g egf efg.
```


benediktahrens:
The name `isweq_iso` seems to me more in line with other namings than `iso_isweq`.

mortberg:
I agree with you. If @DanGrayson agrees I will change it (Dan was the one
who suggested `iso_isweq`)?


--
Anders

mortberg:
By the way, travis keep failing because of changes to Foundations... Maybe it is time to remove this check?

benediktahrens:
On 12/18/2017 02:27 PM, Anders Mörtberg wrote:
> By the way, travis keep failing because of changes to Foundations...
> Maybe it is time to remove this check?
> 

No, the check is useful. More often than not we do not want any changes 
to Foundations.

mortberg:
Ok, just so you know this PR only fails because of changes to Foundations and not anything serious.

DanGrayson:
lgtm

I slightly prefer `iso_isweq` to `isweq_iso` because one may "read" it as saying "an isomorphism is a weak equivalence".  Do you two have a reason for preferring the other one?

mortberg:
I like `isweq_iso` as the conclusion of the lemma is `isweq f` where f is an iso. But I also like your reading of `iso_isweq`. I don't really have any strong opinion in favor of any of the two. What about @benediktahrens ?

As has been pointed out multiple times during the last week we really need to decide on a naming convention in order to solve this kind of issues in a consistent manner, but it might not be worth it to put this PR on hold until we have reached a decision for a naming convention.

benediktahrens:
On 12/18/2017 05:22 PM, Anders Mörtberg wrote:
> I like `isweq_iso` as the conclusion of the lemma is `isweq f` where
> f is an iso. But I also like your reading of `iso_isweq`. I don't
> really have any strong opinion in favor of any of the two. What about
> @benediktahrens ?

As I have already said, I prefer `isweq_iso`, since that is the shape of 
the conclusion the construction applies to (which is the same reason you 
mention).
This is the way I have chosen most of the identifiers in CT.

DanGrayson:
Okay, I defer.  You can change it to `isweq_iso`.

DanGrayson:
... and `weq_iso`.

mortberg:
Renamed `iso_isweq` to `isweq_iso` and `iso_weq` to `weq_iso` now.

DanGrayson:
lgtm

UniMath/Foundations/PartA.v
```diff
@@ -1786,10 +1786,16 @@ Proof.
   apply X0.
 Defined.
 
-Definition weqgradth {X Y : UU} (f : X -> Y) (g : Y -> X)
+(** This is kept to keep compatibility with publications that uses the
```
benediktahrens: "keep" -> "preserve", "uses" -> "use"

mortberg:
Fixed Benedikt's last comment now. I think this is ready to be merged now.

PR #850, opened 2017-12-18
Add formalization of Girard's paradox

jojhelfer:


UniMath/MoreFoundations/GirardsParadox.v
```diff
@@ -0,0 +1,196 @@
+(* This file provides a direct formalizatoin of Girard's paradox, as explained in Martin-Lof's 1972
```
mortberg: "formalizatoin"

UniMath/MoreFoundations/GirardsParadox.v
```diff
@@ -0,0 +1,196 @@
...
+"An intuitionistic theory of types". It can serve as a test as to whether the version of Coq being
+used is (in the most obvious way) inconsistent. *)
+
+(* This section has an arbitrary type instead of False *)
+Section girard.
+
+Variable Flse : Type.
+
+(* an "ordering without infinite descending chains" (wf is for "well-founded") *)
+Definition wf (T : Type) := sigT (fun (lt : T -> T -> Type) =>
```
mortberg: I think you should use `total2` instead of `sigT`?

UniMath/MoreFoundations/GirardsParadox.v
```diff
@@ -0,0 +1,196 @@
...
+                                    prod (forall x y z: T, lt x y -> lt y z -> lt x z)
```
mortberg: Please use the \prod symbol or the forall keyword (prod is defined in the Coq standard library)

UniMath/MoreFoundations/GirardsParadox.v
mortberg: You probably want to import parts (or all) of Foundations so that you get the UniMath versions of all the basic type formers (like `total2`)

UniMath/MoreFoundations/GirardsParadox.v
```diff
@@ -0,0 +1,196 @@
...
+                                         (forall h : (nat -> T),
+                                             (forall n : nat, lt (h (S n)) (h n)) -> Flse)).
```
mortberg: Please use the UniMath version of natural numbers and the comparison operators defined in Foundations/NaturalNumbers (I think the version of `lt` that you use here comes from the standard library and lands in `Prop`...)

UniMath/MoreFoundations/GirardsParadox.v
```diff
@@ -0,0 +1,196 @@
...
+
+(* the type of such orderings *)
+Definition wfs := sigT wf.
```
mortberg: `total2`

UniMath/MoreFoundations/GirardsParadox.v
```diff
@@ -0,0 +1,196 @@
...
+
+(* the underyling set *)
+Definition uset (w : wfs) : Type := projT1 w.
+(* the underlying order *)
+Definition uord (w : wfs) : uset w -> uset w -> Type := projT1 (projT2 w).
+(* the transitivity property *)
+Definition trans (w : wfs) {x y z : uset w} := fst (projT2 (projT2 w)) x y z.
+(* the well-foundedness property *)
+Definition wfp (w : wfs) := snd (projT2 (projT2 w)).
```
mortberg: Please use `pr1` and `pr2` to access the fields

UniMath/MoreFoundations/GirardsParadox.v
```diff
@@ -0,0 +1,196 @@
...
+
+(* the following three definitions and lemmas are for showing that wfs_wf_uord is well founded *)
+
+(* given a descending sequence f : nat -> wfs, map each f(n) to f(0) by composing all the maps *)
+
+Definition wfs_wf_wfp_shift (f : nat -> wfs) (b : forall n : nat, wfs_wf_uord (f (S n)) (f n)) :
+  forall (n : nat), (uset (f n) -> uset (f 0)).
+Proof.
+  intro n.
+  induction n.
```
mortberg: Please use bullets after `induction n`

UniMath/MoreFoundations/GirardsParadox.v
```diff
@@ -0,0 +1,196 @@
...
+  exact (wfs_wf_wfp_shift f b n (domi (b n))).
+Defined.
+
+(* obtain comparisons between the shifted elements *)
+Definition wfs_wf_wfp_compshift (f : nat -> wfs) (b : forall n : nat, wfs_wf_uord (f (S n)) (f n)) :
+  forall (n : nat) {x y : uset (f n)},
+    uord (f n) x y -> uord (f 0) (wfs_wf_wfp_shift f b n x) (wfs_wf_wfp_shift f b n y).
+Proof.
+  intros n.
+  induction n.
```
mortberg: Bullets

UniMath/MoreFoundations/GirardsParadox.v
```diff
@@ -0,0 +1,196 @@
...
+(* thus obtain a sequence in (f 0) *)
+Definition wfs_wf_wfp_seq (f : nat -> wfs) (b : forall n : nat, wfs_wf_uord (f (S n)) (f n)) :
+  nat -> uset (f 0).
+Proof.
+  intro n.
+  exact (wfs_wf_wfp_shift f b n (domi (b n))).
+Defined.
+
+(* obtain comparisons between the shifted elements *)
+Definition wfs_wf_wfp_compshift (f : nat -> wfs) (b : forall n : nat, wfs_wf_uord (f (S n)) (f n)) :
```
mortberg: I would use Lemma instead of Definition, especially as this is Qed'd afterwards

UniMath/MoreFoundations/GirardsParadox.v
```diff
@@ -0,0 +1,196 @@
...
+Proof.
+  intro n.
+  exact (wfs_wf_wfp_compshift f b n (domicom (b n) (domi (b (S n))))).
+Defined.
+
+(* the wf on wfs *)
+Definition wfs_wf : wf wfs.
+Proof.
+  exists wfs_wf_uord.
+  split.
```
mortberg: bullets

UniMath/MoreFoundations/GirardsParadox.v
```diff
@@ -0,0 +1,196 @@
...
+  exact wfs_wf_trans.
+  intro h.
+  intro b.
```
mortberg: You can merge these to `intro` into `intros h b.`

UniMath/MoreFoundations/GirardsParadox.v
```diff
@@ -0,0 +1,196 @@
...
+(* but wfs are irreflexive *)
+Proposition irref (w : wfs) : forall (x : uset w), (uord w) x x -> Flse.
+Proof.
+  intro x.
+  intro p.
+  exact (wfp w (fun n => x) (fun n => p)).
+Defined.
+
+(* therefore the world explodes *)
+Proposition the_world_explodes : Flse.
```
mortberg: Haha, while this name is funny it isn't very informative. How about "GirardsParadox" or something instead?

abooij:
Is Girard's paradox really considered part of the "foundations" of UniMath? This might raise some eyebrows along the way...

UniMath/MoreFoundations/GirardsParadox.v
```diff
@@ -0,0 +1,196 @@
+(* This file provides a direct formalizatoin of Girard's paradox, as explained in Martin-Lof's 1972
+"An intuitionistic theory of types". It can serve as a test as to whether the version of Coq being
+used is (in the most obvious way) inconsistent. *)
+
+(* This section has an arbitrary type instead of False *)
```
abooij: Why?

benediktahrens:
On 12/18/2017 03:36 PM, Auke Booij wrote:
> Is Girard's paradox really considered part of the "foundations" of
> UniMath? This might raise some eyebrows along the way...
> 

I think that the file should go in a package by itself.

DanGrayson:
Won't the paradox no longer be provable after we stop using type-in-type?

benediktahrens:
On 12/18/2017 04:45 PM, Daniel R. Grayson wrote:
> Won't the paradox no longer be provable after we stop using
> type-in-type?
> 

Hopefully!

Until then, it will provide a warning to the users.

After the switch, it can be adapted to show where the proof of the
paradox fails.


Both of these will be very helpful.

DanGrayson:
... and until then, it will provide to users a way to prove any theorem they want to prove.

benediktahrens:
On 12/18/2017 04:51 PM, Daniel R. Grayson wrote:
> ... and until then, it will provide to users a way to prove any
> theorem they want to prove.
> 

What's the difference between us providing the proof and users 
implementing their own version of the paradox in UniMath?

The inconsistency is already there, it is not being introduced by this PR.

mortberg:
I agree with Benedikt. This is an interesting warning to users and once type-in-type is gone it will be fun to see exactly where it fails. Furthermore the formalization is clean and fairly well-documented and can be read as an explanation of this contradiction (which is more subtle than Russel's paradox).

jojhelfer:
@benediktahrens 
> I think that the file should go in a package by itself.

Done

jojhelfer:
@mortberg 
> I think this is a nice result to have, but the code is using a lot of things from the Coq standard library and not the UniMath definitions (like total2). I only commented some of these, but please fix all of them.

My original intention was to formalize this in pure Coq, so that it's clear that it doesn't depend on any particularities of UniMath. However, I am also happy to modify it so as to use only UniMath conventions.

DanGrayson:
Also add the name of the package to the end of the list in `Makefile` and append it
to one of the build steps in `.travis.yml`.

DanGrayson:
Re: "My original intention was to formalize this in pure Coq, so that it's clear that it doesn't depend on any particularities of UniMath."

UniMath offers a consistent view of how to formalize mathematics, and all its components should follow it.

If you want it to be in pure Coq, where it will be accessible to more people, why put it in UniMath?

jojhelfer:
@DanGrayson 
> Also add the name of the package to the end of the list in Makefile and append it
> to one of the build steps in .travis.yml.

Ah yes, I meant to do that. Thanks.

DanGrayson:
Now that the master branch has a version of UniMath that doesn't load Coq.Init.Datatypes, you should be able to update your code to follow the UniMath style guidelines, if you wish.

PR #852, opened 2017-12-19
Making the equality (C^op)^op ≡ C hold judgementally

siddharthist:
This change should be unnoticed by users, because one can still construct a precategory using the same data with `mk_isprecategory` (which automatically symmetrizes the proof of associativity). 

I'm not sure if this can be used to do things like easily prove all the dual facts about epis once we've proved them about monos. The way I've seen the duality principle expressed formally is talking about "the first-order language of categories" or something, and I'm not sure it's easy to express "take this same statement, but change domain for codomain, the order of composition, ..." in Coq. However, I still think it's worth it for the extra judgemental equality. 

This PR is just a proof-of-concept, I didn't want to spend time changing the whole `CategoryTheory` package before asking whether the maintainers were interested. Currently, all the files I made changes in compile (`Categories.v`, `functor_categories.v`, `opp_precategory.v`). 

What do you think?

See #602


benediktahrens:
Can you mention an interesting result which this modification would be useful for?

siddharthist:
@benediktahrens I'm not sure about _interesting_, but the following (and facts like it) seems like it would be easier to prove: 

A monic arrow in C is a monic arrow in C^op^op. 
 * With this change, this proof is absolutely trivial (`assumption.`). 
 * Without it, one must use `transport`.

DanGrayson:
I'm highly in favor of this change.  (But lots of checks fail.)

benediktahrens:
On 12/19/2017 04:47 PM, Daniel R. Grayson wrote:
> I'm highly in favor of this change.  
> 

Why?

DanGrayson:
It would make lots of things simpler.  All concepts in category theory come in dual pairs, and we're probably providing all such dual pairs with dual proofs, doubling the size of our formalizations.  For example, why are there two files, Elements.v and ElementsOp.v, when one should suffice?

benediktahrens:
On 12/19/2017 04:53 PM, Daniel R. Grayson wrote:
> It would make lots of things simpler.  All concepts in category theory come in dual pairs, and we're probably providing all such dual pairs with dual proofs, doubling the size of our formalizations.  For example, why are there two files, Elements.v and ElementsOp.v, when one should suffice?
> 


The suggested change is not about opp-ing, it is about opp-ing twice. It 
does not seem to me to simplify anything involving one opp.

I would first like to see something that is significantly simplified 
with that change.

siddharthist:
@DanGrayson Do you have any ideas for where this might be particularly useful? I'm happy to rewrite some stuff, but I can't think of anywhere off the top of my head that this would "significantly" simplify proofs. Also, on the subject of checks failing, see my opening comment. 

@benediktahrens What about the proof that C^op^op = C? It's one line now 😄 More seriously, I think that changing data structures to support more judgemental equalities is a good idea if it doesn't make life more difficult for users (which, for the reasons I outlined above, I don't think this does) for the reason I outlined above: one doesn't need to use `transport` and keep track of the (necessarily trivial) equality. Is there any reason we might _not_ want this?

benediktahrens:
On 12/19/2017 05:07 PM, Langston Barrett wrote:
> @DanGrayson Do you have any ideas for where this might be particularly useful? I'm happy to rewrite some stuff, but I can't think of anywhere off the top of my head that this would "significantly" simplify proofs. Also, on the subject of checks failing, see my opening comment.
> 
> @benediktahrens What about the proof that C^op^op = C? It's one line now 😄 More seriously, I think that changing data structures to support more judgemental equalities is a good idea if it doesn't make life more difficult for users (which, for the reasons I outlined above, I don't think this does) for the reason I outlined above: one doesn't need to use `transport` and keep track of the (necessarily trivial) equality. Is there any reason we might _not_ want this?
> 


Not having a negative effect is not good enough - please provide an 
example of where this change is helpful.

There are many things to be done in UniMath that are interesting by 
themselves or helpful for others doing interesting things.

Currently, I don't see how this suggested change is either - but I am 
looking forward to being convinced otherwise.

DanGrayson:
Re: "The suggested change is not about opp-ing, it is about opp-ing twice. It does not seem to me to simplify anything involving one opp."

Suppose we define an initial object of C to be an object of that is terminal when viewed in C^op.  Now try to prove that a terminal object c of C becomes initial when viewed in C^op.  The statement involves only single opps.  But to prove it, you have to show that c is terminal in C^op^op.  That's trivial if C^op^op and C are the same by definition.  I offer a non-trivial proof `term_to_opp` in `Ktheory/Representation.v`.

DanGrayson:
Re: " I can't think of anywhere off the top of my head that this would "significantly" simplify proofs."

We could replace the two files `Elements` and `ElementsOp` by one.  That would be significant.

benediktahrens:
On 12/19/2017 06:00 PM, Daniel R. Grayson wrote:
> Re: " I can't think of anywhere off the top of my head that this would 
> "significantly" simplify proofs."
> 
> We could replace the two files |Elements| and |ElementsOp| by one. 
How? Why?

mortberg:
In principle I like the idea of having more judgmental equalities, but if the aim is to have C^op^op=C in order to get rid of duplication in CategoryTheory we have to be a bit careful. The difficulty is not to obtain this equation judgmentally, but to keep the library usable. If we define all limits as colimits on the opposite category (or vice versa) then we have to be sure that we can keep the user from seeing this. When Benedikt and I tried this we often had the problem that `simpl` or `cbn` unfolds things too much and the goal gets completely cluttered with op's. I don't want to have to think about pullbacks in the opposite category when developing the theory of pushouts. 

So I think the question is how to make the implementation details completely opaque to the user while still retaining good computational properties? I tried using `Opaque` and `Arguments foo : simpl never`, but I was not satisfied with the results so the best seemed to be to duplicate the theory. Something that could work better is SSReflect style locking using the `Module` system, but this prohibits computation so it is probably not what we want...

DanGrayson:
Re: "
   We could replace the two files |Elements| and |ElementsOp| by one.
How? Why?"

Remove ElementsOp.v and replace any use of it by Elements.v, but instead of calling the routines with C, call them C^op.  All the proofs should still work trivially.

siddharthist:
@mortberg Maybe a compromise is to not define things in terms of their opposites (e.g. keep current definitions of mono and epi) but to use duality to reduce duplication in proofs (e.g. replace proofs of properties of epis by a combination of duality and the corresponding theorem for monks)? Taking a step further, we could even say one can only use duality in opaque proofs? Or perhaps you're saying you already tried these approaches?

DanGrayson:
Even if none of the proofs in CategoryTheory get simplified by taking advantage of the pull request, I'd like to simplify my proofs in Ktheory.  That's why I'm highly in favor of it.

DanGrayson:
PS: I've organized the proofs about category theory in `Ktheory` to avoid ever writing out a proof and its dual proof.  That's why this pull request would help me.

mortberg:
I'm in favor of merging this PR as well. It will give us more judgmental equalitites which is a good thing, even if we cannot find a satisfactory solution to reducing duplication in CT. I vote for keeping the changes in this PR minimal and then experiment with reducing duplication in CT in future PR's. 

Before merging though, can someone check that it doesn't substantially affect compilation time of CT and the packages that depends on it? Putting more data in the category structure could potentially lead to a decrease in performance because it gets more expensive for Coq to compare terms involving categories, hopefully it doesn't affect anything but I think it would be good to double check this before merging.

benediktahrens:
I think it is valuable to distinguish between the definiton of a concept on the one hand, and a convenient characterization of that concept on the other hand.

Certainly, the fact that colimits in C are limits in C^op is a very convenient *characterization*, but I am not convinced that this is a good *definition*. In the same vein, defining a contravariant functor from C to D to be a functor from C^op to D does not seem a good idea to me - I agree with [1].

Relying on definitional equalities makes the mathematics less modular and portable. One of the fundamental ideas of univalent foundations is that structures can be transported along equivalences. I suggest we take this idea seriously.

[1] https://github.com/UniMath/UniMath/issues/602#issuecomment-282052045

DanGrayson:
Re: "Certainly, the fact that colimits in C are limits in C^op is a very convenient characterization, but I am not convinced that this is a good definition."

The goal of this pull request is not to convince you of that.

Every time we've added definitional equalities, we've benefited.  Look at eta for functions and eta for pairs.  There's no good reason to reject a definitional equality that's easily achievable and has no disadvantages.  Transporting along propositional identities is a pain.

benediktahrens:
On 12/21/2017 06:29 PM, Daniel R. Grayson wrote:
> Re: "Certainly, the fact that colimits in C are limits in C^op is a
> very convenient characterization, but I am not convinced that this is
> a good definition."
> 
> The goal of this pull request is not to convince you of that.


I am confused:
When I asked for uses of that additional definitional equality, the
answers were in that vein, see, e.g., your comment
https://github.com/UniMath/UniMath/pull/852#issuecomment-352837500 .

But maybe I have misunderstood something?

I am asking again: where will this definitional equality be useful?

> 
> Every time we've added definitional equalities, we've benefited.
> Look at eta for functions and eta for pairs.  There's no good reason
> to reject a definitional equality that's easily achievable and has no
> disadvantages.  Transporting along propositional identities is a
> pain.
> 

Note that the eta equalities are not usually part of MLTT.

Disadvantages of additional conversions such as the eta laws is that the 
meta-theory of the type theory becomes more difficult to study, since 
more details need to be checked.


I will stop discussing this proposal now. I am against this change, and 
I won't merge this PR. If you want to proceed with this PR, feel free to 
do so.

DanGrayson:
On Thu, Dec 21, 2017 at 12:47 PM, Benedikt Ahrens <notifications@github.com>
wrote:

>
>
> On 12/21/2017 06:29 PM, Daniel R. Grayson wrote:
> > Re: "Certainly, the fact that colimits in C are limits in C^op is a
> > very convenient characterization, but I am not convinced that this is
> > a good definition."
> >
> > The goal of this pull request is not to convince you of that.
>
>
> I am confused:
> When I asked for uses of that additional definitional equality, the
> answers were in that vein, see, e.g., your comment
> https://github.com/UniMath/UniMath/pull/852#issuecomment-352837500 .
>
> But maybe I have misunderstood something?
>
> I am asking again: where will this definitional equality be useful?
>


Oh, well, then I was confused  --  I thought you were feeling pressure to
rewrite
`CategoryTheory`.  My previous answers stand.  Defining colimits in terms of
limits is a great way to reduce code bloat.  That's the approach I adopt in
`Ktheory` for
representability and limits.


>
> >
> > Every time we've added definitional equalities, we've benefited.
> > Look at eta for functions and eta for pairs. There's no good reason
> > to reject a definitional equality that's easily achievable and has no
> > disadvantages. Transporting along propositional identities is a
> > pain.
> >
>
> Note that the eta equalities are not usually part of MLTT.
>
> Disadvantages of additional conversions such as the eta laws is that the
> meta-theory of the type theory becomes more difficult to study, since
> more details need to be checked.
>

That's true, but speeding up compilation of UniMath by a factor of 2 offered
an irresistible benefit.


>
>
> I will stop discussing this proposal now. I am against this change, and
> I won't merge this PR. If you want to proceed with this PR, feel free to
> do so.


I think we need unanimity in our committee of two to accept a PR, so it's
doomed,
but I don't understand where your opposition really comes from.  It's a
minimal
change, not affecting much at all in `CategoryTheory`.

DanGrayson:
Re: "Certainly, the fact that colimits in C are limits in C^op is a very convenient characterization, but I am not convinced that this is a good definition."

Why must you be convinced of that now?  Isn't it sufficient that some people (such as I) believe it, and would like to use UniMath as a platform for implementing proofs based on that belief?  All this PR does is to provide support for that option, in the spirit of experimentation and research.  It doesn't force colimits to be defined in terms of limits.

benediktahrens:
I have changed my mind on this PR.

The reason I do agree with the proposal to add the second associativity law is to make the definition of `precategory` less arbitrary: currently, one direction of associativity is primitive, whereas the other is derived.

I still disagree with the reason for introducing the second associativity mentioned in the title of this PR, and with the uses of the resulting judgmental equalities mentioned in this discussion. I will strive to keep out of CT any definitions (not: characterizations) given in terms of op'ed categories. In particular, I am still thinking about introducing a type of contravariant functors.

DanGrayson:
Benedikt, I'm grateful to you for reconsidering.  Thanks!

DanGrayson:
lgtm

DanGrayson:
Re: "The reason I do agree with the proposal to add the second associativity law is to make the definition of precategory less arbitrary: currently, one direction of associativity is primitive, whereas the other is derived."

I like this justification.  And since the definition of precategory does not include the hypothesis that the Hom's are sets, this justification may turn out to be central in the exploration of examples where the Hom's are not sets, for correctly and symmetrically stating the higher coherence conditions that will be required.

siddharthist:
Okay, I'll try to update the rest of the files soon. Thanks for the extended discussion, all. It's valuable to think critically about these things.

DanGrayson:
@siddharthist -- if you update your pull request now, we can reconsider it soon.

DanGrayson:
@siddharthist -- maybe there are no updates needed, and we can pull this, assuming travis approves.  In any case, I don't see that any updates were requested by us.

DanGrayson:
Oh, thanks.

The first error seems to be here:
```
COQC UniMath/CategoryTheory/Categories.v
File "./UniMath/CategoryTheory/Categories.v", line 193, characters 66-68:
Error: In environment
C : precategory_data
hs : has_homsets C
Unable to unify "has_homsets C" with
 "isofhlevel 1
    ((∏ (a b : C) (f : C ⟦ a, b ⟧), identity a · f = f)
     × (∏ (a b : C) (f : C ⟦ a, b ⟧), f · identity b = f))".
```

DanGrayson:
It would be nice to get this updated so we can pull it.

DanGrayson:
I'll fix this ...

siddharthist:
Thanks, Dan. Sorry I never got around to it! I really need to revisit a few of my pull requests.

DanGrayson:
I'm making good progress...

DanGrayson:
This works now:
```
Goal ∏ (C D:precategory), (C × D)^op = (C^op × D^op).
  reflexivity.
Qed.
```

PR #856, opened 2017-12-21
install instructions for Arch Linux

DanGrayson:


abooij:
Can we please recommend users to use an AUR tool like `cower` or `yaourt`? It's best to keep all installed software managed by the package manager of the linux distribution. Using a tool like `cower` or `yaourt` will keep the packages under control, rather than having random files and binaries lying around your OS. You can link to the wiki page that I sent earlier. In any case, many Arch users will be aware of this system, and these tools.

DanGrayson:
Done.

INSTALL.md
```diff
@@ -78,21 +81,35 @@ $ camlp5 -v
...
-It comes with installation instructions.  Your version of emacs determines
-which version of ProofGeneral you need, roughly, so some experimentation may be
-required; you may even need the current development version if your emacs is
-recent.
-
-
 ## Preparing for the installation under Linux
 
-Under Ubuntu or Debian, you may install ocaml (and ProofGeneral) with
+uUnder Ubuntu or Debian, you may install ocaml (and ProofGeneral) with
```
mortberg: "uUnder"?

INSTALL.md
```diff
@@ -78,21 +81,35 @@ $ camlp5 -v
...
 $ sudo apt-get install build-essential git ocaml ocaml-nox ocaml-native-compilers camlp4-extra camlp5 proofgeneral proofgeneral-doc libgtk2.0 libgtksourceview2.0 liblablgtk-extras-ocaml-dev ocaml-findlib
 ```
 
+Arch Linux and Manjaro Linux:
+
+    To prepare for compiling it yourself:
+
+      sudo pacman --sync --needed archlinux-keyring
+      sudo pacman-key --populate archlinux
+      sudo pacman --sync --needed ocaml camlp5 ocaml-findlib
```
mortberg: On my Arch system I installed `ocamlfind` using opam by typing:

```
opam install ocamlfind
```

This might be a better instruction as it applies to any system that has opam installed. Some Arch users might prefer using pacman for this, but it is most likely a good idea to use opam for everything ocaml related. 

I'm not sure if one can install `camlp5` through opam as well?

INSTALL.md
```diff
@@ -78,21 +81,35 @@ $ camlp5 -v
...
+
+    Install ProofGeneral from its website as described above.  Or,
+    alternatively, install it from AUR (
+    https://aur.archlinux.org/packages/proofgeneral/ ) using an auxiliary
+    package manager such as "cower" ( https://aur.archlinux.org/packages/cower/
+    ) or yaourt ( https://aur.archlinux.org/packages/yaourt/ ).  That approach
+    will be especially convenient for expert Arch Linux users who find other
+    packages at AUR worth installing.
```
mortberg: I believe most Arch users are "experts" and already use cower or yaourt as a lot of packages can only be found in the AUR

INSTALL.md
```diff
@@ -78,21 +81,35 @@ $ camlp5 -v
...
+
+    To upgrade the system:
+
+      sudo pacman-key --refresh-keys
+      sudo pacman --sync --refresh --sysupgrade
```
mortberg: I usually just type `sudo pacman -Syu` or even better `yaourt -Syu`

INSTALL.md
DanGrayson: I'm an arch linux user who is not expert.

INSTALL.md
DanGrayson: Non-expert users like me would find such abbreviations confusing.  Expert users know the abbreviations and can use them.

INSTALL.md
DanGrayson: My Control key is no longer reliable, so more and more and find cruft like that is produced while I'm editing a file in emacs.

INSTALL.md
mortberg: Fair enough. I guess it cannot hurt to detailed instructions

INSTALL.md
mortberg: Maybe... But if you read the `man` page for pacman the example for how to update the package list and upgrade all packages is simply `pacman -Syu`, so I think that is the standard way to do things. I've been using Arch for about 10 years and I very rarely (like 2 times per year) have to run `pacman-key`.

DanGrayson: I had to run `pacman-key` today, for the first time, hence the instruction about that.  If it saves someone (such as me in the future) some trouble once, then it's worth it.

INSTALL.md
mortberg: Hmm, so there are two ways to deal with the issue of having two package managers like pacman and opam. Either you try to do everything using pacman or you try to do everything ocaml related using opam and the rest using pacman. Haskell users experience the same issue with cabal/stack vs pacman, in my experience you end up with a lot more stable system if you do everything related to a specific programming language (ocaml or Haskell) using their own package manager and do the rest using pacman. So I would advice users to only install ocaml and opam using pacman and then do the rest using opam.

INSTALL.md
mortberg: Strange. Have you installed archlinux-keyring (https://www.archlinux.org/packages/core/any/archlinux-keyring/) ?

INSTALL.md
mortberg: Thinking more about this I think this is fine as it is. It doesn't hurt to give instructions on how to get everything to work only using pacman and experienced users will easily figure out how to set things up using opam on their own.

INSTALL.md
DanGrayson: Yes.  Indeed, that's also in the instructions.

DanGrayson: ... it kept downloading package files and claiming that the PGP key was untrusted, so it couldn't use the files.

mortberg: Ok, very strange. That never happened to me...

DanGrayson:
So, are we good to go on this?

INSTALL.md
```diff
@@ -93,6 +89,27 @@ Under Ubuntu or Debian, you may install ocaml (and ProofGeneral) with
 $ sudo apt-get install build-essential git ocaml ocaml-nox ocaml-native-compilers camlp4-extra camlp5 proofgeneral proofgeneral-doc libgtk2.0 libgtksourceview2.0 liblablgtk-extras-ocaml-dev ocaml-findlib
 ```
 
+Arch Linux and Manjaro Linux:
+
+    To prepare for compiling it yourself:
```
benediktahrens: Not clear what "it" refers to.

INSTALL.md
```diff
@@ -93,6 +89,27 @@ Under Ubuntu or Debian, you may install ocaml (and ProofGeneral) with
...
+      sudo pacman-key --populate archlinux
+      sudo pacman --sync --needed ocaml camlp5 ocaml-findlib
+
+    Install ProofGeneral from its website as described above.  Or,
+    alternatively, install it from AUR (
+    https://aur.archlinux.org/packages/proofgeneral/ ) using an auxiliary
+    package manager such as "cower" ( https://aur.archlinux.org/packages/cower/
+    ) or yaourt ( https://aur.archlinux.org/packages/yaourt/ ).  That approach
+    will be especially convenient for expert Arch Linux users who find other
+    packages at AUR worth installing.
```
benediktahrens: Can we shorten this paragraph? We are not intending to give an introduction to the use of Arch. In particular, I don't feel we need to advertise package managers for Arch in the UniMath install manual.

Every non-essential information we give in this manual is counterproductive, as it makes the manual harder to use.

mortberg:
I agree with Benedikt about the Linux instructions being weird. I don't think it makes much sense to have detailed instructions for how to use the package manager for all different Linux distributions. Anyone using Linux should know how to install packages for their favorite flavour of Linux. I think it would make a lot more sense just to say that they should install `ocaml` and `opam` using their package manager and then give instructions on how to set things up using `opam`. These instructions should apply to all Linux distributions (and hopefully Mac and Windows as well?) and if some power users prefer to install the ocaml specific packages using their favorite package manager then they can just ignore our instructions and set things up themselves.

siddharthist:
@mortberg Maybe the wiki is a good place for this kind of documentation?

DanGrayson:
Okay, I've take your suggestions into account in the current version -- tell me what you think.

PR #857, opened 2017-12-22
Kleisli Triples and cat isomorphism with Monads

logicosimo:
This is our first non-trivial attempt to contribute to UniMath.
Please revise: any comments or suggestions for improvement are welcome.


UniMath/CategoryTheory/Monads/KTriples.v
```diff
@@ -0,0 +1,367 @@
...
+  - apply (invmap (Kleisli_Mor_equiv hs _ _ )).
+    apply funextsec. intros x. apply assoc.
+Defined.
+
+Definition precategory_Kleisli (C : precategory) (hs : has_homsets C) : precategory :=
+  precategory_Kleisli_Data C,, precategory_Kleisli_axioms C hs.
+
+Lemma has_homsets_Kleisli (C : category) :
+  has_homsets (precategory_Kleisli C (homset_property C)).
+Proof.
```
mortberg: Please use bullets in this proof

UniMath/CategoryTheory/Monads/KTriplesEquiv.v
```diff
@@ -0,0 +1,491 @@
...
+Definition unkleislify_mor {C : precategory} {T T' : Kleisli C}
+           (α : Kleisli_Mor T T') :
+  Monad_Mor (unkleislify T) (unkleislify T') :=
+  nat_trans_kleisli_mor α ,, unkleislify_mor_laws α.
+
+(* ----- Functor from Kleisli to Monads. ----- *)
+
+Definition functor_data_unkleislify (C : precategory) (hs : has_homsets C) :
+  functor_data (precategory_Kleisli C hs) (precategory_Monad C hs).
+Proof.
```
mortberg: Please use bullets

UniMath/CategoryTheory/Monads/KTriplesEquiv.v
```diff
@@ -0,0 +1,491 @@
...
+Defined.
+
+Definition kleislify_mor {C : precategory} {M M' : Monad C}
+           (α : Monad_Mor M M') :
+  Kleisli_Mor (kleislify M) (kleislify M') :=
+  (λ x:C, α x) ,, kleislify_mor_law α.
+
+Definition functor_data_kleislify (C : precategory) (hs : has_homsets C) :
+  functor_data (precategory_Monad C hs) (precategory_Kleisli C hs).
+Proof.
```
mortberg: Please use bullets

UniMath/CategoryTheory/Monads/KTriplesEquiv.v
```diff
@@ -0,0 +1,491 @@
...
+  Defined.
+
+  Lemma kleislify_unkleislify (T : Kleisli C) :
+    kleislify (unkleislify T) = T.
+  Proof.
+    unfold unkleislify, kleislify. simpl.
+    destruct T as (F , (K , L)). simpl.
+    change (λ x : C, F x) with F.
+    apply (pair_path_in2). apply subtypeEquality'.
+    2: apply (isaprop_Kleisli_Laws hs K).
```
mortberg: What is `2: ...`?

UniMath/CategoryTheory/Monads/KTriplesEquiv.v
```diff
@@ -0,0 +1,491 @@
...
+    simpl. apply (unkleislify_data_eq (F,, K,, L)).
+  Defined.
+
+  Lemma unkleislify_kleislify (M : Monad C) :
+    unkleislify (kleislify M) = M.
+  Proof.
+    apply (Monad_eq_raw_data hs).
+    unfold Monad_to_raw_data. simpl.
+    change (λ x:C, M x) with (M: ob C → ob C).
+    apply (pair_path_in2). apply total2_paths2.
```
mortberg: Please use bullets

UniMath/CategoryTheory/Monads/KTriplesEquiv.v
```diff
@@ -0,0 +1,491 @@
...
+  Defined.
+
+  Definition etainv_morph (T : Monad C) :
+    Monad_Mor T (unkleislify (kleislify T)) :=
+    etainv_data T ,, etainv_data_laws T.
+
+  Lemma etainv_morph_law (M : Monad C) :
+    Monad_Mor_laws (T := M) (T' := (unkleislify (kleislify M))) (etainv_data M).
+  Proof.
+    split; simpl; intros.
```
mortberg: I think you are using the names introduced here below? (where does `a` come from?) If so please explicitly name them here

UniMath/CategoryTheory/Monads/KTriplesEquiv.v
logicosimo: @mortberg We often used this command to apply the tactic to the second subgoal, in order to avoid proliferating of levels of bulleting that can be easily solved (it is something similar to what you suggested in your first comment).

UniMath/CategoryTheory/Monads/KTriplesEquiv.v
logicosimo: Using the command `2: apply idpath` the second subgoal generated by `total2_paths2` is immediately solved, for this we did not introduced bullets here.

UniMath/CategoryTheory/Monads/KTriplesEquiv.v
logicosimo: I'll change the `intros` in the first line with `intro a`, so it will feel more clear: thank you for your suggestions!

UniMath/CategoryTheory/Monads/KTriples.v
logicosimo: Done, thanks!

UniMath/CategoryTheory/Monads/KTriplesEquiv.v
logicosimo: Done, thanks!

UniMath/CategoryTheory/Monads/KTriplesEquiv.v
logicosimo: Done, thanks!

UniMath/CategoryTheory/Monads/KTriplesEquiv.v
mortberg: I see, I never used this

DanGrayson: Using bullets and braces is better than using the goal selecting colon, because with them, after the clause is over if, if the goal has not been solved, an error is raised.  With the colon, however, if the tactic fails to solve the goal, the proof continues, and the error message is raised at a location distant from the relevant location -- likely at the end of the proof.

Another advantage of the bullets or braces, for someone stepping through the proof one step at a time, is that each time a tactic is applied, there is just one goal visible on the screen, which makes things simpler.

Issue #858, opened 2017-12-24
Travis frequently returns false positives due to difficulties installing opam


benediktahrens:
This error frequently occurs:
```
$ bash -c 'echo $BASH_VERSION'
4.3.11(1)-release
29.66s$ opam init --yes --no-setup
default    Downloading https://opam.ocaml.org/urls.txt
default    Downloading https://opam.ocaml.org/index.tar.gz
[ERROR] No OCaml compiler found in path. You should use:
    opam init --comp=VERSION
'opam init --yes --no-setup' failed.
The command "opam init --yes --no-setup" failed and exited with 66 during .
```
This is annoying since consequently Travis signals a build failure, requiring manual restart of the build.
Is there a way to make the Travis build more resistant to this, e.g., by having it attempt to install opam more than once?

benediktahrens:
As an example, see https://travis-ci.org/UniMath/UniMath/jobs/321055588 - I am purposefully not restarting the build.

DanGrayson:
I'll try to figure this out.

DanGrayson:
It's possible that the old opam 1.1.1 that is available on Travis' old ubuntu 14.04 is too old, and the way around this is to use opam 1.2.2, which is easy to download and install from the web.  I'll submit a pull request to test the theory.

palmskog:
The problem seems to be that Travis sometimes cannot properly access the Ubuntu PPA repositories:
```
W: Failed to fetch http://ppa.launchpad.net/pollinate/ppa/ubuntu/dists/trusty/InRelease  Could not connect to ppa.launchpad.net:80 (91.189.95.83), connection timed out
...
W: Some index files failed to download. They have been ignored, or old ones used instead.
```
For example, see [this build](https://travis-ci.org/UniMath/UniMath/jobs/320394439).

Since OPAM version 1.2.2 is in one of these PPA repositories, it doesn't get properly installed. And when it doesn't get properly installed, the built-in, the obsolete OPAM version 1.1 gets used. The OPAM package repository for this OPAM stopped being maintained years ago.

I don't think this problem can be helped by switching compilers in OPAM, since I don't believe recent compilers are available in OPAM 1.1. And if OPAM 1.2.2 gets properly installed, everything works out anyway.

palmskog:
I believe this is a Travis-related issue, see for example some recent [error descriptions](https://github.com/travis-ci/travis-ci/issues/8507) of this kind in the Travis issue tracker.

[...]

PR #859, opened 2017-12-24
improve travis script ...

DanGrayson:
... so it uses opam 1.2.2 instead of opam 1.1.1, which might
be leading to run time install errors

This tries to address #858.


DanGrayson:
This could be fixed by using opam to install num, but #860 is a more conservative modification.  Closing.

PR #860, opened 2017-12-24
improve travis script

DanGrayson:
by specifying ocaml compiler version to opam

This is probably better than #859 at dealing with #858

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ env:
   - FOUNDATIONS_CHANGE_ERROR=yes BUILD_ALSO="check-for-change-to-Foundations enforce-listing-of-proof-files"
 # building Coq in a separate stage folds up the output in the log:
 before_script:
-  - opam init --yes --no-setup
+  - opam init --yes --no-setup --compiler=4.06.0
   - eval $(opam config env)
   - opam install lablgtk camlp5 --yes --verbose
   - time make build-coq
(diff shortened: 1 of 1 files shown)


DanGrayson:
lgtm

It might even resolve #858

benediktahrens:
Is there a reason why you think that specifying a particular version of ocaml will reduce the likelihood of failure? 

lgtm

DanGrayson:
Maybe no precise reason, but it would be more predictable, and I did pay a little attention to the original error message that you reported:
```
[ERROR] No OCaml compiler found in path. You should use:
    opam init --comp=VERSION
```

DanGrayson:
We had some reports of the library num not being available.  That's an example of an incompatibility caused by the use of different versions of ocaml, for starting with 4.06, num wasn't included.  And old versions of opam may deliver old versions of ocaml, by default.

palmskog:
@DanGrayson These changes don't seem to have fixed the problem, as seen in this [build](https://travis-ci.org/UniMath/UniMath/jobs/322615798). One avenue worth trying out is to use explicit `apt-get` commands rather than the `apt` add-on in the Travis script. Unless you object, I will submit a pull request to try this out.

DanGrayson:
Sure, let's go back to using apt-get, as we did before, if it will help.

By the way, it's good that specifying ocaml 4.06.0 caused the job to stop early.

PR #861, opened 2017-12-27
Make table of contents automatically

DanGrayson:
This supplants #804 and removes the file CONTENTS.md, replacing it by a file UniMath/CONTENTS.md, which contains links to all the packages' README files and to all of their files, in logical order.

See https://github.com/DanGrayson/UniMath/blob/make-table-of-contents/UniMath/CONTENTS.md


DanGrayson:
By the way, a possible future improvement to the table of contents would be to have a brief phrase added to each line describing the contents of the package or file.  These "headlines" could be extracted from specially formatted comment lines in the various files and from a file */.package/headline for the various packages.

benediktahrens:
On 12/28/2017 12:33 PM, Daniel R. Grayson wrote:
> By the way, a possible future improvement to the table of contents
> would be to have a brief phrase added to each line describing the
> contents of the package or file.  These "headlines" could be
> extracted from specially formatted comment lines in the various files
> and from a file */.package/headline for the various packages.
> 


I think that that would be a very useful addition. Why not have a 
specially marked text in .package/files for this?


I'll merge this PR now.

DanGrayson:
That might run quicker.  It could be something like this:
```
-- basic algebra
BinaryOperations.v -- binary operations
Monoids_and_Groups.v -- monoids and groups
Rigs_and_Rings.v -- rigs and rings
...
```
The entry at the top would apply to the whole package.

PR #863, opened 2017-12-28
explicit apt-get commands in Travis script

palmskog:
Experiment to try to solve the problem with intermittently failing Travis builds by using explicit apt-get commands.

Note that as far as I can tell, failing builds will fail quickly no matter whether `--compiler=4.06.0` is used or not. Without the option, the error message will be `[ERROR] No OCaml compiler found in path.` and with the option, it will be `opam: unknown option --compiler`. Compare [failing build a](https://travis-ci.org/UniMath/UniMath/jobs/320394439) and [failing build b](https://travis-ci.org/UniMath/UniMath/jobs/322615798).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,17 +6,10 @@ sudo: 		required
 dist: 		trusty
 git:
   submodules:   false
-addons:
-  apt:
-    sources:
-      - avsm
-    packages:
-      - opam
-      - aspcud
-      - time
-      - libgtk2.0-dev
-      - libgtksourceview2.0-dev
-      - emacs
+install:
+  - sudo add-apt-repository -y ppa:avsm
+  - sudo apt-get update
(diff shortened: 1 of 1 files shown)


palmskog:
Also note that `opam init --yes --no-setup --compiler=4.06.0` will add a few minutes to each build, since OPAM will actually compile the new compiler from scratch. Therefore, I think it is best avoided.

DanGrayson:
lgtm!  It looks like that will do it.

DanGrayson:
Thanks!

palmskog:
@DanGrayson Since the problem manifests intermittently (as discussed in the Travis [issue](https://github.com/travis-ci/travis-ci/issues/8507)), just passing one Travis run doesn't necessarily indicate a permanent fix. However, please feel free to merge this pull request, and I will try to monitor new Travis builds that use the updated script from here.

PR #864, opened 2018-01-05
Fix issue #794

### LINKED ISSUE #794: finish a proof in StandardFiniteSets.v that has not been completed for unknown reasons
[Lemma partial_sum_prop in StandardFiniteSets.v](https://github.com/UniMath/UniMath/blob/master/UniMath/Combinatorics/StandardFiniteSets.v#L1104) is aborted (used to be admitted before the work in PR #786).

Lemma partial_sum_slot should then be freed of the extra assumption of the statement of the former lemma.

tonyxty:
Fix #794 

A proof of the lemma `partial_sum_prop` is added.  I changed the statement to a slightly stronger form (removed an unused hypothesis) and cleaned up the bits in the proof of `partial_sum_slot`.

### DIFF
--- a/UniMath/Combinatorics/StandardFiniteSets.v
+++ b/UniMath/Combinatorics/StandardFiniteSets.v
@@ -1101,22 +1101,77 @@ Proof.
   rewrite weqfromcoprodofstn_eq1. apply idpath.
 Defined.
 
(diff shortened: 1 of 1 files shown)


DanGrayson:
lgtm

PR #866, opened 2018-01-09
change Travis to use Xenial Docker image with OPAM and OCaml 4.06.0

palmskog:
This is an experimental PR to figure out stable Travis builds. Currently using an Ubuntu Xenial image produced from [this Dockerfile](https://github.com/proofengineering/coq-docker/blob/master/Dockerfile-xenial-unimath).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,6 +56,7 @@ script:
          cd $THIS_REPO
          git checkout -qf $TRAVIS_COMMIT
        fi &&
+       eval $(opam config env) &&
        time make build-coq &&
        time make TIMECMD=time $PACKAGES $BUILD_ALSO FOUNDATIONS_CHANGE_ERROR=$FOUNDATIONS_CHANGE_ERROR"
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,32 +1,63 @@
-# validate this file at http://lint.travis-ci.org/
-# ocaml and coq are not on the list of languages:
-language:       generic
-sudo: 		required
-# "trusty" is the most modern release of Ubuntu supported
-dist: 		trusty
-git:
-  submodules:   false
-install:
-  - sudo add-apt-repository -y ppa:avsm
-  - sudo apt-get update
-  - sudo apt-get install -y opam aspcud time libgtk2.0-dev libgtksourceview2.0-dev emacs
-# build coqide along with the Tactics package, just because it's a short package
+language: generic
+sudo: false
+
+services:
+  - docker
+
 env:
-  - PACKAGES="Foundations Combinatorics Algebra NumberSystems PAdics"
-  - PACKAGES=CategoryTheory
-  - PACKAGES="MoreFoundations Ktheory"
-  - PACKAGES=HomologicalAlgebra
-  - PACKAGES=Topology
-  - PACKAGES=RealNumbers
-  - PACKAGES=SubstitutionSystems
-  - PACKAGES=Tactics BUILD_COQIDE=yes BUILD_ALSO=TAGS
-  - PACKAGES=Folds
-  - FOUNDATIONS_CHANGE_ERROR=yes BUILD_ALSO="check-for-change-to-Foundations enforce-listing-of-proof-files"
-# building Coq in a separate stage folds up the output in the log:
+  global:
+    - THIS_REPO=UniMath
+  matrix:
+    - PACKAGES="Foundations Combinatorics Algebra NumberSystems PAdics"
+    - PACKAGES=CategoryTheory
+    - PACKAGES="MoreFoundations Ktheory"
+    - PACKAGES=HomologicalAlgebra
+    - PACKAGES=Topology
+    - PACKAGES=RealNumbers
+    - PACKAGES=SubstitutionSystems
+    - PACKAGES=Tactics BUILD_COQIDE=yes BUILD_ALSO=TAGS
+    - PACKAGES=Folds
+    - FOUNDATIONS_CHANGE_ERROR=yes BUILD_ALSO="check-for-change-to-Foundations enforce-listing-of-proof-files"
+
 before_script:
-  - opam init --yes --no-setup
-  - eval $(opam config env)
-  - opam install lablgtk camlp5 num --yes --verbose
-  - time make build-coq
+  - docker pull palmskog/xenial-unimath
+
+# Using travis_wait here seems to cause the job to terminate after 1 minute
+# with no error (!).
+# The fcntl line works around a bug where Travis truncates logs and fails.
 script:
-  - time make TIMECMD=time $PACKAGES $BUILD_ALSO FOUNDATIONS_CHANGE_ERROR=$FOUNDATIONS_CHANGE_ERROR
+  - python -c "import fcntl; fcntl.fcntl(1, fcntl.F_SETFL, 0)"
+  - REMOTE_ORIGIN_URL=`git config --get remote.origin.url`
+  - echo "THIS_REPO=${THIS_REPO}"
+  - echo "PACKAGES=${PACKAGES}"
+  - echo "BUILD_ALSO=${BUILD_ALSO}"
+  - echo "FOUNDATIONS_CHANGE_ERROR=${FOUNDATIONS_CHANGE_ERROR}"
+  - echo "TRAVIS_BRANCH=${TRAVIS_BRANCH}"
+  - echo "REMOTE_ORIGIN_URL=${REMOTE_ORIGIN_URL}"
+  - echo "TRAVIS_EVENT_TYPE=${TRAVIS_EVENT_TYPE}"
+  - echo "TRAVIS_PULL_REQUEST=${TRAVIS_PULL_REQUEST}"
+  - echo "TRAVIS_PULL_REQUEST_BRANCH=${TRAVIS_PULL_REQUEST_BRANCH}"
+  - echo "TRAVIS_COMMIT=${TRAVIS_COMMIT}"
+  - echo "TRAVIS_PULL_REQUEST_SHA=${TRAVIS_PULL_REQUEST_SHA}"
+  - echo "TRAVIS_REPO_SLUG=${TRAVIS_REPO_SLUG}"
+  - >-
+    docker run palmskog/xenial-unimath /bin/bash -c "true &&
+       if [ $TRAVIS_EVENT_TYPE = pull_request ] ; then
+         git clone --quiet --depth 9 $REMOTE_ORIGIN_URL $THIS_REPO || git clone --quiet --depth 9 $REMOTE_ORIGIN_URL $THIS_REPO
+         cd $THIS_REPO
+         git fetch origin +refs/pull/$TRAVIS_PULL_REQUEST/merge
+         git checkout -qf $TRAVIS_PULL_REQUEST_SHA
+         git config user.email noone@cares.com
+         git config user.name Noone Cares
+         git remote add theupstream https://github.com/$TRAVIS_REPO_SLUG.git
+         git pull --depth 9 theupstream $TRAVIS_BRANCH || git pull --depth 9 theupstream $TRAVIS_BRANCH
+       else
+         git clone --quiet --depth 9 -b $TRAVIS_BRANCH $REMOTE_ORIGIN_URL $THIS_REPO || git clone --quiet --depth 9 -b $TRAVIS_BRANCH $REMOTE_ORIGIN_URL $THIS_REPO
+         cd $THIS_REPO
+         git checkout -qf $TRAVIS_COMMIT
+       fi &&
+       time make build-coq &&
+       time make TIMECMD=time $PACKAGES $BUILD_ALSO FOUNDATIONS_CHANGE_ERROR=$FOUNDATIONS_CHANGE_ERROR"
+
+git:
+  depth: 9
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ env:
 # The "docker run" command will pull if needed.
 # Running this first gives two tries in case of network lossage.
 before_script:
-  - timeout 5m docker pull palmskog/xenial-unimath || true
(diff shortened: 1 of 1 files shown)


palmskog:
Now things look good to me. Comments and feedback appreciated on whether this can be merged. For example, I'd be happy to put the Docker image in another Docker Hub repository that can be pushed to by UniMath developers (rather than my personal one).

abooij:
There is an easier way to obtain the correct commit, as you can mount directories into a docker image directly, without re-executing the git pull. See e.g. my attempt [here](https://github.com/abooij/UniMath/commit/a5af0580af69181bd2c12fa6f90d6d83b2ef74cc) (and the corresponding [travis page](https://travis-ci.org/abooij/UniMath/builds/312146795?utm_source=github_status&utm_medium=notification)).

palmskog:
The script I'm proposing has been hardened in various ways to overcome network connectivity issues. For example, it tries git commands several times. I don't see how that could be done when mounting a directory into the Docker image. Any optimizations are probably best left for when we know this approach fits the project's needs.

benediktahrens:
I am out of my depth here. @DanGrayson , can this be merged?

palmskog:
To possibly give some more confidence in the pull request, this is a direct adaptation of a Travis [script](https://github.com/typetools/checker-framework/blob/master/.travis.yml) and Docker images by Prof. Mike Ernst at the University of Washington, used in the [Checker Framework](https://checkerframework.org) project.

I would be happy to include the Dockerfile that generates the Docker image in this pull request, along with instructions how to generate an image locally. Let me know.

DanGrayson:
I'll take a look ...

.travis.yml
```diff
@@ -1,32 +1,69 @@
...
+    - FOUNDATIONS_CHANGE_ERROR=yes BUILD_ALSO="check-for-change-to-Foundations enforce-listing-of-proof-files"
+
+# The "docker run" command will pull if needed.
+# Running this first gives two tries in case of network lossage.
 before_script:
-  - opam init --yes --no-setup
-  - eval $(opam config env)
-  - opam install lablgtk camlp5 num --yes --verbose
-  - time make build-coq
+  - timeout 5m docker pull palmskog/xenial-unimath || true
```
DanGrayson: Where does docker find "palmskog/xenial-unimath" ?

If the command fails, why ignore the failure?

.travis.yml
```diff
@@ -1,32 +1,69 @@
...
+  - echo "TRAVIS_BRANCH=${TRAVIS_BRANCH}"
+  - echo "REMOTE_ORIGIN_URL=${REMOTE_ORIGIN_URL}"
+  - echo "TRAVIS_EVENT_TYPE=${TRAVIS_EVENT_TYPE}"
+  - echo "TRAVIS_COMMIT=${TRAVIS_COMMIT}"
+  - echo "TRAVIS_PULL_REQUEST=${TRAVIS_PULL_REQUEST}"
+  - echo "TRAVIS_PULL_REQUEST_BRANCH=${TRAVIS_PULL_REQUEST_BRANCH}"
+  - echo "TRAVIS_PULL_REQUEST_SHA=${TRAVIS_PULL_REQUEST_SHA}"
+  - echo "TRAVIS_REPO_SLUG=${TRAVIS_REPO_SLUG}"
+  - >-
+    docker run palmskog/xenial-unimath /bin/bash -c "true &&
```
DanGrayson: What's the use of "true &&" ?

.travis.yml
```diff
@@ -1,32 +1,69 @@
...
+       if [ $TRAVIS_EVENT_TYPE = pull_request ] ; then
+         git clone --quiet --depth 9 $REMOTE_ORIGIN_URL $THIS_REPO || git clone --quiet --depth 9 $REMOTE_ORIGIN_URL $THIS_REPO
+         cd $THIS_REPO
+         git fetch origin +refs/pull/$TRAVIS_PULL_REQUEST/merge
+         git checkout -qf $TRAVIS_PULL_REQUEST_SHA
+         git config user.email noone@cares.com
+         git config user.name Noone Cares
+         git remote add theupstream https://github.com/$TRAVIS_REPO_SLUG.git
+         git pull --depth 9 theupstream $TRAVIS_BRANCH || git pull --depth 9 theupstream $TRAVIS_BRANCH
+       else
```
DanGrayson: In the else clause, what type of event is it?

.travis.yml
```diff
@@ -1,32 +1,69 @@
...
+  - echo "TRAVIS_EVENT_TYPE=${TRAVIS_EVENT_TYPE}"
+  - echo "TRAVIS_COMMIT=${TRAVIS_COMMIT}"
+  - echo "TRAVIS_PULL_REQUEST=${TRAVIS_PULL_REQUEST}"
+  - echo "TRAVIS_PULL_REQUEST_BRANCH=${TRAVIS_PULL_REQUEST_BRANCH}"
+  - echo "TRAVIS_PULL_REQUEST_SHA=${TRAVIS_PULL_REQUEST_SHA}"
+  - echo "TRAVIS_REPO_SLUG=${TRAVIS_REPO_SLUG}"
+  - >-
+    docker run palmskog/xenial-unimath /bin/bash -c "true &&
+       if [ $TRAVIS_EVENT_TYPE = pull_request ] ; then
+         git clone --quiet --depth 9 $REMOTE_ORIGIN_URL $THIS_REPO || git clone --quiet --depth 9 $REMOTE_ORIGIN_URL $THIS_REPO
```
DanGrayson: It would be good to precede this section with a "set -x", so we can see the commands being executed.
Similarly for the else clause.

.travis.yml
DanGrayson: Oh, I guess it finds it as one of your public repositories at docker.com.  How can I get a web page that displays your public repositories?

.travis.yml
```diff
@@ -1,32 +1,69 @@
...
+  - echo "FOUNDATIONS_CHANGE_ERROR=${FOUNDATIONS_CHANGE_ERROR}"
+  - echo "TRAVIS_BRANCH=${TRAVIS_BRANCH}"
+  - echo "REMOTE_ORIGIN_URL=${REMOTE_ORIGIN_URL}"
+  - echo "TRAVIS_EVENT_TYPE=${TRAVIS_EVENT_TYPE}"
+  - echo "TRAVIS_COMMIT=${TRAVIS_COMMIT}"
+  - echo "TRAVIS_PULL_REQUEST=${TRAVIS_PULL_REQUEST}"
+  - echo "TRAVIS_PULL_REQUEST_BRANCH=${TRAVIS_PULL_REQUEST_BRANCH}"
+  - echo "TRAVIS_PULL_REQUEST_SHA=${TRAVIS_PULL_REQUEST_SHA}"
+  - echo "TRAVIS_REPO_SLUG=${TRAVIS_REPO_SLUG}"
+  - >-
```
DanGrayson: Do you really intend to redirect stdio to a file named "-" with ">-" here?

.travis.yml
palmskog: As described [here](https://docs.travis-ci.com/user/environment-variables/) the remaining events are `push`, `api`, `cron`. In practice, the only one this project will see is probably `push`, i.e., regular non-pull-request commits.

.travis.yml
palmskog: Regarding ignoring a failed `docker pull` command, note that the `docker run` command later in the script will also download the Docker image. In this way, there will be two attempts to download the image if network connectivity is bad.

.travis.yml
palmskog: This was a convention used in the Checker Framework script, and I believe it is just convenient way of ignoring the output of the `docker run` command itself, which we are completely uninterested in.

.travis.yml
palmskog: It establishes the pattern of `<command> &&` which is used throughout, but it's not strictly necessary. Mostly for readability.

.travis.yml
palmskog: I'll pass the `-x` option to bash itself instead and see what happens. I'm not sure this will make Travis output more readable though.

.travis.yml
DanGrayson: Okay, I've created a docker organization called UniMath, so we can have our own docker image.  See https://hub.docker.com/u/UniMath/.

@benediktahrens -- make a docker.com account and tell me the user name, and I'll add you as an owner.  (The github notions of user, organization, owner, and team seem the same on docker.)

.travis.yml
DanGrayson: Ah, I did not know travis would run a test after a push.

.travis.yml
DanGrayson: In the travis raw logs of the test the ">-" doesn't appear, and the output from "docker run" is visible.

DanGrayson:
Could you post your Dockerfile here?

DanGrayson:
Never mind, I see the link to your Dockerfile now.

DanGrayson:
The slowest travis build step, with this pull request, takes 10 minutes longer to run than before.  We may not be able to afford that extra run time during tests, because there's a limit of about 45 minutes, and as proofs are added, we'll use more of that.

palmskog:
@DanGrayson Docker image download times are unpredictable. Most of the time, it takes 70 seconds or less, but this time it took more. If you look in the build history, the latest build is not typical.

One option is to precompile Coq in the Docker image. This saves about 2-3 minutes I believe.

DanGrayson:
Ubuntu 17.10 offers Coq 8.6, so maybe we could just use that.  But we'll need 8.7 soon.

You're right, this build was unusually slow.  But it might be due to this same old problem:
```
W: Failed to fetch http://ppa.launchpad.net/couchdb/stable/ubuntu/dists/trusty/InRelease  Could not connect to ppa.launchpad.net:80 (91.189.95.83), connection timed out
```

The point of using docker was to avoid the ppa, right?  And you have removed the use of `ppa:avsm`.  What could explain it trying to fetch from launchpad.net again?

DanGrayson:
If there's nothing fetched from launchpad.net with docker files, why do we get the message 
```
W: Failed to fetch http://ppa.launchpad.net/couchdb/stable/ubuntu/dists/trusty/InRelease  Could not connect to ppa.launchpad.net:80 (91.189.95.83), connection timed out
```
in the build at
```
https://travis-ci.org/UniMath/UniMath/jobs/328938218
```
?

palmskog:
Travis has a bunch of PPAs in their default APT list of repositories, and the "initial" job (before the Docker image is fetched) fails to update the APT metadata in this list. However, this has nothing to do with the actual job that is run once the Docker image is fetched.

Basically: Travis runs job using Ubuntu 14.04 -> Travis job fetches and runs Ubuntu 16.04 Docker image -> Coq files are checked inside 16.04 Docker image

DanGrayson:
Okay, but how do determine the reason the build at https://travis-ci.org/UniMath/UniMath/jobs/328938218 was slow?  It might have been because of launchpad.net not responding, or it might have been because of the docker image download.

palmskog:
There are two pieces of evidence that the Docker image download was slow (and this was what caused the slow build).

1. The command `$ timeout 5m docker pull palmskog/xenial-unimath || true` (line 466) is listed as taking 300 seconds (i.e., until it timed out).
2. Starting on line 562, we have `Unable to find image 'palmskog/xenial-unimath:latest' locally` followed by `latest: Pulling from palmskog/xenial-unimath`. This indicates the first download attempt failed, and the second download attempt started.

palmskog:
For the record, another Docker image variant I have tried that typically speeds up proof checking by several minutes: include a pre-compiled version of Coq built in 32-bit mode. For some reason, 32-bit Coq is much faster, possibly due to faster arithmetic operations on smaller words.

DanGrayson:
I pulled `palmskog/xenial-unimath` to my laptop -- where is the resulting file?

I think the file size displayed during download was less than 600MB, and then it spent a lot of time decompressing it.

Maybe we should have something like
```
timeout 1m docker pull palmskog/xenial-unimath || true
timeout 2m docker pull palmskog/xenial-unimath || true
timeout 3m docker pull palmskog/xenial-unimath || true
```
instead of
```
timeout 5m docker pull palmskog/xenial-unimath || true
```
so we can see what's going on with the download.

DanGrayson:
... or better:
```
timeout 1m docker pull palmskog/xenial-unimath || \
timeout 2m docker pull palmskog/xenial-unimath || \
timeout 3m docker pull palmskog/xenial-unimath
```
I.e., let it fail if it doesn't come in 3 tries.

palmskog:
I believe Docker puts fetched images in some directory in the user home directory. For example, if you want to launch a shell that runs my image, you would run: `sudo docker run -i -t palmskog/xenial-unimath /bin/bash`

palmskog:
As per what I wrote above, a timeout of 1 minute is not generally enough to download a Docker image. If we use this, there is a real chance of wasting lots of time re-downloading. The 5-minute timeout is a tradeoff based on experience in other projects.

DanGrayson:
Then what about
```
set -x ; 
timeout 5m docker pull palmskog/xenial-unimath || \
timeout 5m docker pull palmskog/xenial-unimath || \
timeout 5m docker pull palmskog/xenial-unimath
```

palmskog:
First, I'm not sure I see the point of having `set -x` when there's basically just three commands, and all of them give expicit output directly when run (`latest: Pulling from palmskog/xenial-unimath`).

Second, I believe this change gives almost the same behavior as the current one, which tries downloading for 5 minutes, then tries again for 10 minutes (Travis output timeout). The maximum time limit will be 15 minutes for both the current and this hypothetical case.

DanGrayson:
I don't see a timeout of 10 minutes anywhere.  My idea is that it would be good to get the docker image downloaded before the `docker run` command, so we get an accurate timing for the downloading.  If we wait for the `docker run` command, then other things add to the time of the command.

palmskog:
If there is no output for 10 minutes, Travis kills a build, see [here](https://docs.travis-ci.com/user/common-build-problems/#My-builds-are-timing-out).

DanGrayson:
Thanks.

I think we can pull this now and work toward sufficient docker expertise at this end so we can switch later to using our own docker image.  The travis build has succeeded many times, and I'm convinced now that there is not much delay in switching to this.

@benediktahrens : Benedikt, should I merge?

PR #869, opened 2018-01-12
Define graphs and construct equivalence closure

fpvandoorn:
The construction of the equivalence closure is not impredicative, but uses the truncation of paths in a graph.

I wrote this because I expected I needed it for #868, but in the end I used impredicative definitions of the generated binopeqrel. But maybe it's still useful.

`contents.md` didn't update when I executed `make`, maybe I did something wrong there.


fpvandoorn:
> Actually, since it uses propositional truncation, it is impredicative.

Ah yes, I forgot that the propositional truncation is impredicative in UniMath.

PR #874, opened 2018-01-20
Expunge [cancel_precomposition] and [cancel_postcomposition] from library

peterlefanulumsdaine:
This pull request expunges the lemmas `cancel_precomposition` and `cancel_postcomposition` from the library.  Instructions for removing them from dependent libraries follow below.

# Motivation

These lemmas were the facts `f = f' -> f ;; g = f' ;; g`, and `g = g' -> f ;; g = f ;; g'`.  On the one hand, their names are misleading; on the other hand, they are entirely redundant cases of the general lemmas `maponpaths`, `maponpaths_2`.

Regarding the names, these are not cancellation principles.  Cancellation principles are facts of the form “F a = F b -> a = b”.  These lemmas are the converse implication: facts of the form “a = b -> F a = F b”.

As such, they are just instances of the general principle that a function respects equality in its arguments, i.e. `maponpaths` and `maponpaths_2`.  Using the more general lemmas makes code more uniform, with less for newcomers to learn.  Special case lemmas may often be justified, if they are easier to use than the general lemmas (e.g. if the general lemmas often fail to unify or infer arguments).  In this case, though, the general lemmas are just as easy to use, as the ease of expunction shows.

Therefore (with the approval of @benediktahrens ) I suggest either removing these lemmas entirely from the library (as this PR currently does), or alternatively, if we don’t want to force users to follow suit, leaving the definitions but removing all uses of them and considering them as deprecated (for that, merge just the first commit of this PR).

# Instructions for updating code for compatibility with this PR

This PR will break all occurrences of `cancel_precomposition` and `cancel_postcomposition`.  Almost all occurrences can be fixed just by replacing them with `maponpaths` and `maponpaths_2` respectively.  E.g. run the following commandline search-and-replace, adapting the file pattern to your library’s layout:

```
perl -pi -e "s/cancel_precomposition/maponpaths/" */*.v
perl -pi -e "s/cancel_postcomposition/maponpaths_2/" */*.v
```

and then add the following import line as required to make `maponpaths_2` available:

```
Require Import UniMath.MoreFoundations.PartA.
```

Almost all occurrences in `UniMath` worked fine straight after the search-and-replace plus imports as above.  The remaining broken instances (<5% of occurrences of the `cancel_` lemmas) were fixed by:

- where arguments had been explicitly given to the `cancel_` lemmas, reorder and/or remove the arguments appropriately;

- expand `maponpaths` to `maponpaths (compose _)` (replacing `cancel_precomposition`);

- expand `maponpaths_2` to `maponpaths_2 compose` (replacing `cancel_postcomposition`);

- replace `apply` with `eapply` when invoking the general lemmas.


UniMath/CategoryTheory/limits/graphs/pushouts.v
```diff
@@ -6,6 +6,8 @@
 Require Import UniMath.Foundations.PartD.
 Require Import UniMath.Foundations.Propositions.
 Require Import UniMath.Foundations.Sets.
+Require Import UniMath.MoreFoundations.PartA.
+Require Import UniMath.MoreFoundations.Tactics.
```
siddharthist: Why is this line necessary (the import of `Tactics`)?

siddharthist:
Do we want to Coq submodule updated as part of this PR? LGTM otherwise, and I agree with the motivation.

DanGrayson:
Right, this PR should not change `sub/coq`.

DanGrayson:
Other than fixing it so it doesn't change `sub/coq`, it looks good to me.

peterlefanulumsdaine:
Thanks very much @siddharthist for catching those two errors!  Instead of reverting them with a further commit here, I’ve redone these commits with those errors removed, to keep the repo history cleaner, and submitted that as a new PR #875 .  `git diff` between the two versions confirms that I haven’t introduced further accidental errors: the only difference between these branches is the two points you flagged up.

siddharthist:
@peterlefanulumsdaine While you certainly can open new PRs and your caution is laudible, it also might be easier to `--amend` the commits and `git push --force` occasionally 😉

peterlefanulumsdaine:
@siddharthist : I did indeed `--amend`, intending to `push --force`, and then got paranoid and decided to go with a second PR instead — [Github has a couple of vaguely open-ended warnings](https://help.github.com/articles/about-pull-requests/) against force pushing.  If you’ve had successful experience of doing so, though, that’s good to know!

DanGrayson:
It would be straightforward to add one more commit that undoes the change to the submodule, and then just push it.

peterlefanulumsdaine:
@DanGrayson : Yes, I considered that too, but went for `--amend` + new PR instead, for the sake of keeping the repo history cleaner.  But if you would prefer, I can add a correction commit to this PR instead.

DanGrayson:
No, I have no preference.

I haven't spent any time keeping the repo history clean.  Two days ago I tried rebase and forced push for the first time.  One scary thing about forcing a push is that for an instant, the commits being rebased exist only on your computer, so if it dies before github gets the new commit, you'll have lost your work.

benediktahrens:
I have already explained to Peter why pushing a new commit on top of the 
existing ones is better for reviewers than pushing a rebased branch.

DanGrayson:
Probably you said that each reviewer might like to look at a single new commit that makes just the changes suggested by that reviewer, to see if they were made correctly.

benediktahrens:
Yes, that is what I said. How did you know?

DanGrayson:
A little birdie told me.

peterlefanulumsdaine:
@benediktahrens : in that case, I misunderstood you!  I thought you just wanted to check with a diff that the corrections had been made correctly, and as we checked together at the time, `git diff 3c4a68fb e945a25e` confirms this (or `git diff expunge-cancel-comps-orig expunge-cancel-comps-fixed`, before the latter was merged with updates to master).

Anyhow, in case you (the maintainers) feel the “add separate commits” approach is better, I’ve added fix commits to this PR, and merged in updates from `master` as @DanGrayson did on #875 ; so both this and #875 should be ready to merge, whichever you feel is better.

DanGrayson:
Isn't it time in Sweden to be asleep?

benediktahrens:
On 21/01/18 18:58, Peter LeFanu Lumsdaine wrote:
> @benediktahrens <https://github.com/benediktahrens> : in that case, I 
> misunderstood you! I thought you just wanted to check with a diff that 
> the corrections had been made correctly, 

I have to review many PRs and many updates to PRs, without getting any 
money for it. If every time I need to check an update I first have to 
read a lengthy explanation of how to see the diff, I get annoyed.
Best to just follow best practice of pushing a "correction" commit.

peterlefanulumsdaine:
OK, good to know for the future that that’s the preferred approach here!

DanGrayson:
This is ready to go, but we'll wait a bit so some other PR's involving CategoryTheory can be pulled without conflict.

benediktahrens:
After making this same change in a different repository, I am less convinced that it is a good idea.
More specifically, I agree that
- the current names are bad; in particular, shorter names would be good, since they are often used.
- the lemmas should be proved using maponpaths and maponpaths_2.
I disagree with replacing all uses of (the new versions of) these lemmas by maponpaths{_2}. The reason is that the more specialized lemma provides more clarity about what is going on, in particular when reading the proof without playing it.

DanGrayson:
Can anyone understand a proof without playing it?

benediktahrens:
On 24/01/18 10:29, Daniel R. Grayson wrote:
> Can anyone understand a proof without playing it?

The additional information contained in the more specialized lemma will 
also help repairing proofs when something goes wrong.


To summarize my opinion: my preferred solution is
a. changing the names of the lemmas
b. changing their proofs
c. continuing using the specialized lemmas whenever possible.

Obviously, c. will not be enforced, but I will advertise c., and do c. 
myself.

peterlefanulumsdaine:
I agree there may be some benefit to showing that the `maponpaths` is acting on composition (especially when reading without stepping through).

On the other hand, I think it’s also beneficial to have names which makes it clear that they’re just the very general principle “functions respect equality” — I feel the old names were misleading in giving the impression that some more specific/non-trivial principle is being used.

So if we just rename and reprove the lemmas, how about something like `maponpaths_compose_l” and "maponpaths_compose_r”?  These are still a bit long, but very clear. 

Whatever we do here (both in terms of the arguments for/against having specialised versions, and the choice of names), it seems like the same issues apply for algebraic proofs with any other kind of binary relation — e.g. doing algebraic proofs in rings.  As far as I can see, though, there aren’t currently as many proofs using these as using composition, and the proofs with these that we do have make heavy use of `rewrite`, which avoids using these lemmas explicitly at all.

benediktahrens:
On 24/01/18 11:01, Peter LeFanu Lumsdaine wrote:
> I agree there may be some benefit to showing that the |maponpaths| is 
> acting on composition (especially when reading without stepping through).
> 
> On the other hand, I think it’s also beneficial to have names which 
> makes it clear that they’re just the very general principle “functions 
> respect equality” — 

I feel that this can be made clear in the proof of the specialized lemma 
(which I said in a previous message: the proofs should be using 
`maponpaths{_2}`), but does not need to be clear from its name.

> I feel the old names were misleading in giving the 
> impression that some more specific/non-trivial principle is being used.

As mentioned in previous messages, I fully agree with this. It will be 
very good to get rid of those misleading names.

> So if we just rename and reprove the lemmas, how about something like 
> `maponpaths_compose_l” and "maponpaths_compose_r”? These are still a bit 
> long, but very clear.

I suggest `compose_l` and `compose_r` (see above), but am open to other 
suggestions, provided they are short.

PR #875, opened 2018-01-20
Expunge [cancel_precomposition] and [cancel_postcomposition] from library (take 2)

peterlefanulumsdaine:
This pull request expunges the lemmas `cancel_precomposition` and `cancel_postcomposition` from the library.  Instructions for removing them from dependent libraries follow below.  This is a fixed version of #874 — thanks @siddharthist for catching the errors there!

# Motivation

These lemmas were the facts `f = f' -> f ;; g = f' ;; g`, and `g = g' -> f ;; g = f ;; g'`.  On the one hand, their names are misleading; on the other hand, they are entirely redundant cases of the general lemmas `maponpaths`, `maponpaths_2`.

Regarding the names, these are not cancellation principles.  Cancellation principles are facts of the form “F a = F b -> a = b”.  These lemmas are the converse implication: facts of the form “a = b -> F a = F b”.

As such, they are just instances of the general principle that a function respects equality in its arguments, i.e. `maponpaths` and `maponpaths_2`.  Using the more general lemmas makes code more uniform, with less for newcomers to learn.  Special case lemmas may often be justified, if they are easier to use than the general lemmas (e.g. if the general lemmas often fail to unify or infer arguments).  In this case, though, the general lemmas are just as easy to use, as the ease of expunction shows.

Therefore (with the approval of @benediktahrens ) I suggest either removing these lemmas entirely from the library (as this PR currently does), or alternatively, if we don’t want to force users to follow suit, leaving the definitions but removing all uses of them and considering them as deprecated (for that, merge just the first commit of this PR).

# Instructions for updating code for compatibility with this PR

This PR will break all occurrences of `cancel_precomposition` and `cancel_postcomposition`.  Almost all occurrences can be fixed just by replacing them with `maponpaths` and `maponpaths_2` respectively.  E.g. run the following commandline search-and-replace, adapting the file pattern to your library’s layout:

```
perl -pi -e "s/cancel_precomposition/maponpaths/" */*.v
perl -pi -e "s/cancel_postcomposition/maponpaths_2/" */*.v
```

and then add the following import line as required to make `maponpaths_2` available:

```
Require Import UniMath.MoreFoundations.PartA.
```

Almost all occurrences in `UniMath` worked fine straight after the search-and-replace plus imports as above.  The remaining broken instances (<5% of occurrences of the `cancel_` lemmas) were fixed by:

- where arguments had been explicitly given to the `cancel_` lemmas, reorder and/or remove the arguments appropriately;

- expand `maponpaths` to `maponpaths (compose _)` (replacing `cancel_precomposition`);

- expand `maponpaths_2` to `maponpaths_2 compose` (replacing `cancel_postcomposition`);

- replace `apply` with `eapply` when invoking the general lemmas.

### DIFF
--- a/UniMath/CategoryTheory/Abelian.v
+++ b/UniMath/CategoryTheory/Abelian.v
@@ -24,6 +24,7 @@
 Require Import UniMath.Foundations.PartD.
 Require Import UniMath.Foundations.Propositions.
 Require Import UniMath.Foundations.Sets.
+Require Import UniMath.MoreFoundations.PartA.
 
 Require Import UniMath.Algebra.Monoids_and_Groups.
 
@@ -602,7 +603,7 @@ Section abelian_equalizers.
     use (pathscomp0 (! (maponpaths (λ h : _, PullbackPr1 Pb · h) H1))).
     use (pathscomp0 _ ((id_right (PullbackPr2 Pb)))).
     use (pathscomp0 _ (maponpaths (λ h : _, PullbackPr2 Pb · h) H2)).
-    rewrite assoc. rewrite assoc. apply cancel_postcomposition.
+    rewrite assoc. rewrite assoc. apply maponpaths_2.
     apply PullbackSqrCommutes.
   Qed.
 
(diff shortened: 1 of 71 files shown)


benediktahrens:
Oh no, merge conflicts... @peterlefanulumsdaine can you fix them?

DanGrayson:
Let me fix them, since they're caused by my PR getting pulled.

DanGrayson:
lgtm

DanGrayson:
Closing in favor of #874, which introduces the same changes.

Issue #827, opened 2017-12-13
Do without `Init.Logic`


benediktahrens:
People get massively confused over types from the standard library, such as `prod` (instead of UniMath's `dirprod`). This causes frustration over weird error messages at `Qed` time.

We should not load Coq's init, providing the few items we take from there ourselves instead.

DanGrayson:
That's a good idea.   I'll do it.

abooij:
See also #461

DanGrayson:
Thanks, I've closed #461, since it is identical to this one.

mortberg:
Will the Coq `now` tactic also go away? That would be great because then we know that if someone uses `now` then it is the simple one from MoreFoundations.Tactics

benediktahrens:
On 18/01/18 14:27, Daniel R. Grayson wrote:
> One of things about to go away, since it comes from Coq's init, is the 
> tactic "easy".

I am very much looking forward to that!

DanGrayson:
Yes, Coq's "now" will also go away.

DanGrayson:
I've gotten through much of UniMath without Coq's init, but there is something that breaks a lot of our uses of the "rewrite" tactic, which I've reported as a bug in Coq here: https://github.com/coq/coq/issues/6623 .

DanGrayson:
By the way, in the process of removing Coq's init, I've found several proofs that depend on lemmas about natural numbers proved there!

benediktahrens:
Done in #876, closing.

DanGrayson:
Let's leave this open, until the Init.Logic is no longer used, as a reminder to work on getting rewrite to work our way.  (It uses Prop currently.)

DanGrayson:
About natural number parsing, Hugo says:

The path for O and S is hard-wired in the nat_syntax_plugin, so, for
it to work, they should have global paths Coq.Init.Datatypes.O and
Coq.Init.Datatypes.S. So you need to put your definition of nat in a
file which has this absolute logical name (Coq.Init.Datatypes).

DanGrayson:
In a future version of Coq (8.10) we can parse natural numbers easily with a new command:
```
Numeral Notation mynat of_uint1 to_uint1 : my1_scope.
```
Then we can incorporate the following code to get numeral parsing back again:
```
Require Import Coq.Init.Datatypes. (* to get "option" *)
Require Import Coq.Init.Decimal.   (* decimal numbers *)
[...]

PR #876, opened 2018-01-20
stop using Coq's initial prelude

DanGrayson:
This resolves issue #827 partially.  We still load Init.Logic, because
that extends the capability of the tactic "rewrite" in useful ways that
we use, but it introduces terms that use Prop, which we don't want.  (I
spent some time replacing failing uses of "rewrite" by other things, but
tired of that.)

We also still load Init.Notations, but that is innocent, because we want
to use the standard Coq notations reserved there.

I'm currently puzzled about how to get parsing of numerals, such as "4",
to work, so as a stop-gap measure, I've defined notations for the
numerals less than 25.

I've also renamed the tactic "unimath_easy" to "easy", since Coq's
"easy" is no longer present.

### DIFF
--- a/UniMath/CategoryTheory/DiscreteCategory.v
+++ b/UniMath/CategoryTheory/DiscreteCategory.v
@@ -64,10 +64,24 @@ Definition is_nat_trans_discrete_precategory {D : precategory} (Dhom : has_homse
   : is_nat_trans (pr1 f) (pr1 g) F.
 Proof.
   intros x y h.
-  rewrite h.
-  rewrite (pr1 (pr2 f)) , (pr1 (pr2 g)).
-  rewrite (id_left (F y)) , (id_right (F y)).
-  reflexivity.
+  unfold discrete_precategory in h. simpl in h.
+  induction h.
+  change (idpath x) with (identity x).
+
+  Check (pr1 f).
+  assert (k := ! functor_id f x).
+  (* The type of k is displayed as [ identity (f x) = # f (identity x) ],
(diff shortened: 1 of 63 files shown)


DanGrayson:
The only error from travis is that this changes Foundations.

UniMath/CategoryTheory/DiscreteCategory.v
```diff
@@ -64,10 +64,24 @@ Definition is_nat_trans_discrete_precategory {D : precategory} (Dhom : has_homse
...
   intros x y h.
-  rewrite h.
-  rewrite (pr1 (pr2 f)) , (pr1 (pr2 g)).
-  rewrite (id_left (F y)) , (id_right (F y)).
-  reflexivity.
+  unfold discrete_precategory in h. simpl in h.
+  induction h.
+  change (idpath x) with (identity x).
+
+  Check (pr1 f).
```
benediktahrens: Spurious `Check...`.

UniMath/CategoryTheory/catiso.v
```diff
@@ -123,6 +132,14 @@ Proof.
   reflexivity.
 Defined.
 
+(* Lemma three_arg_paths {A B:UU} {C:A -> B -> UU} {D:UU} {f : ∏ (a:A) (b:B), C a b -> D} {a1 b1 c1 a2 b2 c2} : *)
```
benediktahrens: Is this commented out on purpose?

PR #878, opened 2018-01-23
Add full export file for each package

DanGrayson:
... using the Makefile.

This supplants pull request #821.

All the code is in the Makefile.


DanGrayson:
The only travis error comes from the check for changes to Foundations.

PR #889, opened 2018-02-02
remove two extraneous hypotheses from Ktheory/Utilities

siddharthist:
The points in X are never used. Also, use (idfun UU) rather than (λ T, T), it is more grep-able.


DanGrayson:
lgtm

PR #894, opened 2018-02-09
fix init

DanGrayson:
We address issue #827.  This pull request introduces a disadvantage as it
removes a disadvantage, so we may not want to merge it.  I put it up for discussion,
of it and of alternatives.

By the way, this commit also fixes the Makefile in an important way: we no longer
"include" Coq's makefile -- we call it instead.  That's cleaner, because Coq's makefile
calls itself now, anyway, because of a kludge.

This commit adds a file Coq/Init/Datatypes.v to UniMath in order to make
the natural number syntax plugin work once again -- it hard codes access
to the fully qualified identifiers Coq.Init.Datatypes.O and
Coq.Init.Datatypes.S, and can't work without those.

But I see a problem: after "make install" our copy of Coq.Init.Datatypes
will be preferred to Coq's copy, whenever using the same Coq
installation:

... if I use "-noinit":

    Datatypes.vo has been found in
    [ /Users/dan/src/ProofChecking/UniMath-master/sub/coq/user-contrib/Coq/Init ;
      /Users/dan/src/ProofChecking/UniMath-master/sub/coq/theories/Init ];
    loading /Users/dan/src/ProofChecking/UniMath-master/sub/coq/user-contrib/Coq/Init/Datatypes.vo
    [ambiguous-file-name,filesystem]</warning>
    <infomsg>[Loading ML file ltac_plugin.cmxs ...
    [Loading ML file nat_syntax_plugin.cmxs ...
    </infomsg>

It loads our version and it works.

... if I don't use "-noinit"

    Welcome to Coq 8.7.1 (December 2017)
    Error during initialization:
    Error:
    Compiled library Coq.Init.Prelude (in file /Users/dan/src/ProofChecking/UniMath-master/sub/coq/theories/Init/Prelude.vo) makes inconsistent assumptions over library Coq.Init.Datatypes

So installation will spoil the use of that Coq for anything but UniMath.

Alternatively, if we copy the file nat_syntax.ml and adapt it, we can
fix that problem, but then installation of UniMath on a system that
already has Coq will also require ocaml to be there, as well as a step
where nat_syntax.ml is compiled and the compiled file is placed
somewhere findable (?).

I think these limitations are inherent in the current situation.


benediktahrens:
Let me summarize to see whether I understand correctly:
- If we do not load init, the syntax for natural numbers breaks. Does anything else break?
- To reinstate syntax for natural numbers, our own Coq.Init.Datatypes is added, shadowing the original one. This breaks use of Coq for non-UniMath purposes.

Besides these changes, this PR also improves the Makefile setup.

I think that I do like the changes to the Makefile. On the other hand, I think that the inconvenience of having our own Coq.Init.Datatypes outweighs the benefits of having the syntax for nats work. But then, others might need nats more than I do.

DanGrayson:
I agree.  I'll close this and isolate the changes to the makefile in a separate PR.

PR #895, opened 2018-02-10
add a few notations for bigger natural numbers

DanGrayson:
... so we can reinstate two former examples.

If we don't restore natural number syntax, as in #894, we should add
these simple notations for 100 and 1000, so small natural numbers can be
written as arithmetic expressions.  Numbers bigger than 5000 result in a
warning from Coq, anyway, since they aren't really usable, so we might
not worry about natural number parsing until Coq offers a way to tailor
it to our situation.


benediktahrens:
lgtm

PR #896, opened 2018-02-10
ZF Structures initial commit

dimtsem:
This file begins the formalization of a usable "h-set of sets" in UniMath based on the approach of interpreting sets as trees with extra structure.

The file ZFStructures.v includes the definition of type ZFS of ZF structures as (super)rigid (bi)wellfounded trees with unique representatives, a proof that ZFS is an h-set, and a definition of the elementhood relation in terms of which one gets a model of set theory.

This is the largest amount of useful theory for "ZF structures" that can be built along those lines without using LEM, as far as I can tell.

I have realized that proving even the axiom of extensionality will require an assumption of decidable equality, so it seems to me reasonable to have this file separate from other developments.

I have also formalized some basic auxiliary general lemmas (e.g. [iscontr_to_iscontrid], [substructure_univalence]) that may be of more general interest, but which I found useful stating and proving for this file. These can perhaps be moved.

Finally, I have placed this in MoreFoundations because it seems to me the most appropriate place given what is already in there.


DanGrayson:
```
error: *** file UniMath/MoreFoundations/ZFstructures.v not listed in appropriate file UniMath/*/.package/files
```

DanGrayson:
This is a useful contribution.

I think this belongs in the Combinatorics package, though, along with the other combinatorial structures on sets.

dimtsem:
Is the error in the Travis check due to the fact that the filename for ZFStructures.v has not been added to the "files" file?

DanGrayson:
Yes, but you fixed it.

DanGrayson:
Re: "This would be especially useful e.g. in situations when one wants presheaf categories to have an underlying h-set of objects."

There are easier ways to do that.  In general, if C is a category with ob C not a set, you can use AC to split the map ob C ⟶ π₀ ob C and take the full subcategory on π₀ ob C; it will be equivalent to C.  In special cases, you can do something special: for the category of sets, replace the objects by the well-ordered sets.  For the category of finite dimensional vector spaces, add an ordered basis to each one.  And so on.

mortberg:
Nice Dimitrios! I'll take a closer look at the code later, but I just want to check that I understand everything correctly: for the definition of ZF-structures, the elementhood relation and the proof that they form a set you do not need classical logic? But for everything else (proving that they form a category with good properties) you will need it? 

On Dan's remark about easier ways to do it in special cases: I have a proof that well-founded sets form a category. I expect it to be straightforward to prove that this category has the same good properties of HSET with liberal use of Zeremelo's well-ordering theorem (and hence AC).  I'll clean my code and make a PR later.

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,1019 @@
...
+
+(*** Auxilliary Lemmas ***)
+
+(** Contents **)
+(*
+-Some auxilliary lemmas which may have already been proven, or which might be obtained as easy corollaries of things that have already been proven, but which I could not find, and which are useful in this form for this file.
+*)
+
+
+Lemma iscontr_to_iscontrid (X : UU) : iscontr X → ∏ (x y : X), iscontr (x = y).
```
benediktahrens: This is an instance of
```
hlevelntosn: ∏ (n : nat) (T : UU), isofhlevel n T → isofhlevel (S n) T
```

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,1019 @@
...
+Qed.
+
+Lemma iscontr_propid (P : UU) (is : isaprop P) : iscontr (P = P).
+Proof.
+  assert (π : isaprop (P=P)).
+  apply (isofhlevelpathspace 1 P P is is).
+  apply (iscontraprop1 π (idpath P)).
+Qed.
+
+Lemma iseqset_iseqhset (X Y : hSet) (p : pr1 X = pr1 Y) : X = Y.
```
benediktahrens: You can use one of
```
subtypeInjectivity: ∏ (A : UU) (B : A → UU), isPredicate B → ∏ x y : ∑ y, B y, x = y ≃ pr1 x = pr1 y
total2_paths_hProp_equiv:
  ∏ (A : UU) (B : A → hProp) (x : ∑ x : A, B x) (y : ∑ x0 : A, B x0), x = y ≃ pr1 x = pr1 y
subtypeInjectivity_prop:
  ∏ (A : UU) (B : A → hProp) (x : ∑ x : A, B x) (y : ∑ x0 : A, B x0), x = y ≃ pr1 x = pr1 y
```

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,1019 @@
...
+Definition ext (X : UU) (P : X → hProp) (T : ∑ x, P x) : (pr1 T = pr1 T) → T ╝ T.
+Proof.
+  intros t.
+  destruct T as [x π].
+  simpl in t.
+  exists t.
+  apply propproperty.
+Defined.
+
+Lemma contr_to_pr1contr (X : UU) (P : X → hProp) (T : ∑ x, P x) (C : iscontr (T = T)) : iscontr ((pr1 T) = (pr1 T)).
```
benediktahrens: This is much easier using preservation of hlevels by equivalence, and
```
subtypeInjectivity: ∏ (A : UU) (B : A → UU), isPredicate B → ∏ x y : ∑ y, B y, x = y ≃ pr1 x = pr1 y
total2_paths_hProp_equiv:
  ∏ (A : UU) (B : A → hProp) (x : ∑ x : A, B x) (y : ∑ x0 : A, B x0), x = y ≃ pr1 x = pr1 y
subtypeInjectivity_prop:
  ∏ (A : UU) (B : A → hProp) (x : ∑ x : A, B x) (y : ∑ x0 : A, B x0), x = y ≃ pr1 x = pr1 y
```

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,1019 @@
...
+    rewrite -> l. reflexivity.
+  }
+  set (Θ := (! h1 @ h3 @ h2)).
+  apply total2_paths_equiv in Θ.
+  set (θ := pr1 Θ).
+  apply θ.
+Qed.
+
+
+Lemma iscontrid (X : UU) (P : X → UU) (x : X) (π : P x) (is : isaprop (P x)) : iscontr (x = x) → iscontr (P x = P x) → iscontr ( x ,, π = x ,, π ).
```
benediktahrens: It might be that some of the hypotheses here are redundant. Again, I feel that the proof could be simplified and generalized by using preservation of hlevels under equivalence, and use of the equivalences mentioned before.

DanGrayson: Use this instead:
```
Corollary isasethProp : isaset hProp.
```
Then we know `P x=P x` is a nonempty proposition, so it's contractible.

Try to eliminate the lemma.

dimtsem:
@mortberg 

> for the definition of ZF-structures, the elementhood relation and the proof that they form a set you do not need classical logic?

No, nothing in this file uses classical logic. To achieve this required a few tricks, in the sense that some conditions provable classically from weaker axioms were added as extra axioms. In particular, both [issuperrigid] and [hasuniquerepbranch] follow classically (or with decidable equality) from [isrigid], but had to be asserted as extra conditions here instead. 

> But for everything else (proving that they form a category with good properties) you will need it?

It will definitely be needed in order to prove that ZF structures satisfy the axioms of BZC (ZF set theory with bounded comprehension and choice), and are therefore a model of "traditional" set theory that is itself a set.

For extracting a category with good properties out of ZF structures, I am not yet sure. I have some ideas about how to define such a good category while at the same time avoiding LEM (and this is part of the reason why I wanted the files separate), but it is not entirely clear yet whether they will succeed. (In any case, I expect it to be doable with LEM, of course.)

Very much looking forward to seeing the woset approach.

dimtsem:
I have made the requested changes.

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,909 @@
...
+-Tree [Tree]
+-Up and Down Well-founded tree [Tree_isWellFounded]
+-Rigid tree (no non-trivial automorphisms) [isrigid]
+-ZF structures as rigid well-founded trees [ZFStructures]
+-Proof that ZFStructures is an h-set [isaset_ZFStruc]
+*)
+
+(* The Definition of transitive reflexive rooted graphs [TRRGraph] *)
+
+Definition RootedGraph := ∑ (V : hSet) (E : V → V → hProp), V.
```
mortberg: You can write `E : hrel V`

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,909 @@
...
+
+Definition isreflexive {V : hSet} (E : V → V → hProp) := (∏ x : V, E x x).
```
mortberg: This is `isrefl` in Foundations.Sets

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,909 @@
...
+
+Lemma isaprop_isreflexive {V : hSet} (E : V → V → hProp) : isaprop (isreflexive E).
```
mortberg: This is `isaprop_isrefl`in Foundations.Sets

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,909 @@
...
+Proof.
+  apply impred. intros. apply propproperty.
+Qed.
+
+Definition istransitive {V : hSet} (E : V → V → hProp) := (∏ (x y z : V), E x y → E y z → E x z).
```
mortberg: `istrans`

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,909 @@
...
+
+Lemma isaprop_istransitive {V : hSet} (E : V → V → hProp) : isaprop (istransitive E).
```
mortberg: `isaprop_istrans`

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,909 @@
...
+Proof.
+  repeat (apply impred ; intros). apply propproperty.
+Qed.
+
+Definition isaroot {V : hSet} (E : V → V → hProp) (r : V) := (∏ w : V, E r w).
```
mortberg: Please add which type this lands in (I think it should be `UU`). Same comment applies to many of the other definitions

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,909 @@
...
+
+
+Definition TRRGraph := ∑ (V : hSet), TRRGraphData V.
+
+Definition TRRG_transitivity (G : TRRGraph) := pr12 (pr222 G).
+
+(* Definition of [TRRGraph] homomorphisms [isTRRGhomo], isomorphisms [TRRGraphiso], and a proof that isomorphisms are equivalent to identities [TRRGraph_univalence] *)
+
+
+Definition isTRRGhomo {G H : TRRGraph} (f : pr1 G → pr1 H) := (∏ (x y : pr1 G), ((pr12 G) x y) <-> (pr12 H) (f x) (f y)) × ((f (pr122 G)) = pr122 H).
```
mortberg: Please try to keep lines shorter than 100 characters

mortberg: This Definition has a lot of explicit projections, it is better to define them as separate definitions above so that one doesn't have to remember which combination of 1's and 2's to use

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,909 @@
...
+  apply (isofhleveltotal2 2).
+  - apply isaset_hrel.
+  -  intros x.
+     apply (isofhleveltotal2 2).
+     -- apply setproperty.
+     -- intros.  apply hlevelntosn. apply isaprop_isTRR.
+Qed.
+
+
+Definition TRRGraph := ∑ (V : hSet), TRRGraphData V.
```
mortberg: Maybe you can turn `pr1` for TRRGraph into a Coercion, this should make it possible for you to eliminate all of the `pr1`'s below

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,909 @@
...
+  simpl.
+  induction p.
+  simpl.
+  unfold isTRRGhomo.
+  split.
+  - intros. simpl in X0. rewrite -> idpath_transportf in X0. rewrite <- X0. split. repeat (intros P ; apply P). repeat (intros P ; apply P).
+  - rewrite -> idpath_transportf in X0. induction X0. reflexivity.
+Qed.
+
+(* The lemma [helper] below is needed because of some buggy Coq behaviour - perhaps it can be eliminated with a different proof *)
```
mortberg: It seems to me that you need this lemma because your goal below is not generalized enough. Sometimes one can do this kind of proofs by using combinations of `generalize` and `specialize`, but it is often easier to prove it as a separate lemma with everything generalized appropriately. 

Please put `Local` in front of the `Lemma` so that this isn't exported (alternatively if you think this is a useful lemma then you can give it a better name than `helper`)

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,909 @@
...
+  exists σ.
+  simpl.
+  apply (isaprop_isTRR X E r').
+Qed.
+
+Lemma TRRGhomo_topath (X Y : hSet) (G : TRRGraphData X) (H : TRRGraphData Y) (p : X = Y) : @isTRRGhomo (X ,, G) (Y ,, H) (pr1weq (hSet_univalence _ _ p)) → transportf (λ x : hSet, TRRGraphData x) p G = H.
+Proof.
+  induction p.
+  rewrite -> idpath_transportf.
+  intros.
```
mortberg: Please give names to things you introduce. I believe this intro will give the `X0` that you use below. By doing this proofs get a lot more robust as they are not depending on Coq's automatic naming mechanisms

mortberg: It seems like you are referring to automatically named things in other proofs as well, maybe `grep` for X0, X1... and make sure that you give the hypotheses these names everywhere

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,909 @@
...
+  unfold isTRRGhomo in X0. simpl in X0.
+  destruct X0 as [π σ].
+  set (τ := λ x y, hPropUnivalence (pr1 G x y) (pr1 H x y) (pr1 (π x y)) (pr2 (π x y))).
```
mortberg: Why set tau when you only use it in one place in the rest of the proof? I would just apply this lemma below...

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,909 @@
...
+Qed.
+
+Definition selfedge (G : TRRGraph) (x : pr1 G) : pr1 (pr2 G) x x
+  := (pr1 (pr2 (pr2 (pr2 G))) x).
+
+(* Definition of trees as a subtype [Tree] of [TRRGraph] and proof of univalence for trees [Tree_univalence] *)
+
+Definition DownwardClosure {G : TRRGraph} (x : pr1 G) := ∑ y : pr1 G, pr1 (pr2 G) y x.
+
+Definition antisymmetric {G :TRRGraph} (y z : pr1 G) := ((pr1 (pr2 G) y z) × (pr1 (pr2 G) z y)) → (z = y).
```
mortberg: This is called `isantisymm` in the library

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,909 @@
...
+
+Definition total {G : TRRGraph} (y z : pr1 G) := (pr1 (pr2 G) y z) ∨ (pr1 (pr2 G) z y).
```
mortberg: `istotal`

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,909 @@
...
+Definition Tree :=  ∑ G : TRRGraph, isatree G.
+
+Definition Tree_iso (T1 T2 : Tree) := pr1 T1 ≅ pr1 T2.
+
+Theorem Tree_univalence (T1 T2 : Tree) : (T1 = T2) ≃ (Tree_iso T1 T2).
+Proof.
+  apply (substructure_univalence _ _ (λ G, λ H, TRRGraph_univalence G H) (λ G, (isatree G ,, isaprop_isatree G)) T1 T2).
+Qed.
+
+(* Definition of branches ("upward closures") for [Tree]s as a tree [Up x] for a point x of T *)
```
mortberg: If you put `(**` then this comment will be displayed in the generated documentation

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,909 @@
...
+Lemma Upw_transitive (T : Tree) (x : pr11 T) : ∏ (y z w : pr1 (Upw_to_RootedGraph T x)), pr1 (pr2 (Upw_to_RootedGraph T x)) y z → pr1 (pr2 (Upw_to_RootedGraph T x)) z w → pr1 (pr2 (Upw_to_RootedGraph T x)) y w.
+Proof.
+  intros.
+  simpl.
+  simpl in X.
+  simpl in X0.
+  unfold Upw_E.
+  unfold Upw_E in X.
+  unfold Upw_E in X0.
+  apply (pr1 (pr2 (pr2 (pr2 (pr2 (pr1 T))))) (pr1 y) (pr1 z) (pr1 w) X X0).
```
mortberg: Please define accessor functions instead of using chains of pr1 and pr2

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,909 @@
...
+Lemma UpwardClosure_transitive (T : preZFS) (x : Ve T) : ∏ (y z w : pr1 (UpwardClosure_to_RootedGraph T x)), pr1 (pr2 (UpwardClosure_to_RootedGraph T x)) y z → pr1 (pr2 (UpwardClosure_to_RootedGraph T x)) z w → pr1 (pr2 (UpwardClosure_to_RootedGraph T x)) y w.
+Proof.
+  intros.
+  simpl.
+  simpl in X.
+  simpl in X0.
+  unfold UpwardClosure_E.
+  unfold UpwardClosure_E in X.
+  unfold UpwardClosure_E in X0.
+  apply (pr1 (pr2 (pr2 (pr2 (pr2 (pr1 (pr1 T)))))) (pr1 y) (pr1 z) (pr1 w) X X0).
```
mortberg: accessor functions

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,909 @@
...
+
+
+Definition Up_to_Up (T : Tree) (x : pr11 T) (y : pr11 (Up x)) : pr1 (Upw_to_TRRGraph T (pr1 y)) → pr1 (Upw_to_TRRGraph (Up x) y).
+Proof.
+    simpl.
+    intros.
+    destruct X as [t π].
+    destruct y as [y σ].
+    exact ((t ,, ((pr1 (pr2 (pr2 (pr2 (pr2 (pr1 T)))))) x y t σ π)) ,, π).
+Defined.
```
mortberg: Don't use `destruct` in things you `Define`, use `induction` instead (this will generate an application of the eliminator instead of a `match`)

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,909 @@
...
+  apply (isofhleveltotal2 2).
+  - apply isaset_preZFS.
+  - intros x.
+     apply hlevelntosn.
+     apply isaprop_hasuniquerepbranch.
+Qed.
+
+(*** End of Definition of ZFS ***)
+
+(*** The Elementhood Relation ***)
```
mortberg: I would put a Section here and declare `T` (and maybe some more things) as section variables

mortberg: This comment might apply to the code above as well...

mortberg:
PS: Like I already said I like the math in this PR. I think it would be interesting to try to organize this as a category and see which properties you can prove (preferrably without resorting to classical logic whenever possible).

dimtsem:
@mortberg 

Thanks a lot for these comments. 

I have pushed a lot of changes along the lines you suggested, which resulted in a lot of cleaning up and improvements (including removing a large amount of leftover redundant code).

DanGrayson:
Please remove the file `sub/.DS_store`.

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,953 @@
...
+
+(*** Auxilliary Lemmas ***)
+
+(** Contents **)
+(**
+-Some auxilliary lemmas which which are useful stated in this form for this file.
+**)
+
+
+Lemma iscontr_propid (P : UU) (is : isaprop P) : iscontr (P = P).
```
DanGrayson: Lemma iscontr_propid is unused.

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,953 @@
...
+     }
+     apply L.
+Qed.
+
+Definition Up {T : Tree} (x : pr11 T) : Tree := (Upw_to_TRRGraph T x ,, isatree_Upw T x).
+
+(** Definition of rigid [isrigid] and superrigid [issuperrigid] trees **)
+
+Definition isrigid (T : Tree) : UU
+  := iscontr (T = T).
```
DanGrayson: If you change the definition of `isrigid` to `iscontr (T ≃ T)`, can you avoid some appeals to univalence?

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,953 @@
...
+      { use weqbandf.
+        { apply (hSet_univalence (pr1 G) (pr1 H)). }
+        { intro p. simpl. apply weqimplimpl.
+           - intros q. apply (TRRGhomo_frompath (pr1 G) (pr1 H) (pr2 G) (pr2 H) p q).
+           - intros q. apply (TRRGhomo_topath (pr1 G) (pr1 H) (pr2 G) (pr2 H) p q).
+           - apply (isaset_TRRGraphData (pr1 H)).
+           - apply (isaprop_isTRRGhomo).
+         } } }
+Qed.
+
```
DanGrayson:
```

Definition id_TRRGraphiso (G : TRRGraph) : TRRGraphiso G G.
Proof.
  exists (idweq (pr1 G)).
  split.
  - intros. use isrefl_logeq.
  - reflexivity.
Defined.

Definition TRRGraph_univalence_map (G H : TRRGraph) : (G = H) → (G ≅ H).
Proof.
  intro p.
  induction p.
  use id_TRRGraphiso.
Defined.

Lemma TRRGraph_univalence_compute (G H : TRRGraph) :
  pr1 (TRRGraph_univalence G H) = TRRGraph_univalence_map G H.
Proof.
  Fail reflexivity.
  use funextsec.
  intro p.
  induction p.
  Fail reflexivity.
Abort.

(*

  It's usually important for univalence equivalences to be implemented with the expected
  underlying map, as in the failed lemma above.  After proving the lemma, one could use

     remakeweq: ∏ (X Y : UU) (f : X ≃ Y) (g : X → Y), f ~ g → X ≃ Y

  to fix up the equivalence. *)
```

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,953 @@
...
+-Rigid tree (no non-trivial automorphisms) [isrigid]
+-ZF structures as rigid well-founded trees [ZFStructures]
+-Proof that ZFStructures is an h-set [isaset_ZFStruc]
+**)
+
+(** The definition of transitive reflexive rooted graphs [TRRGraph] **)
+
+Definition RootedGraph := ∑ (V : hSet) (E : hrel V), V.
+
+Definition isaroot {V : hSet} (E : hrel V) (r : V) : UU := (∏ w : V, E r w).
```
DanGrayson: It would be more standard to simply have the root of the graph be provided as part of the definition of RootedGraph.

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,953 @@
...
+-Tree [Tree]
+-Up and Down Well-founded tree [Tree_isWellFounded]
+-Rigid tree (no non-trivial automorphisms) [isrigid]
+-ZF structures as rigid well-founded trees [ZFStructures]
+-Proof that ZFStructures is an h-set [isaset_ZFStruc]
+**)
+
+(** The definition of transitive reflexive rooted graphs [TRRGraph] **)
+
+Definition RootedGraph := ∑ (V : hSet) (E : hrel V), V.
```
DanGrayson: `RootedGraph` is an inappropriate name, since there is no root specified.

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,953 @@
...
+
+Definition isaroot {V : hSet} (E : hrel V) (r : V) : UU := (∏ w : V, E r w).
+
+Lemma isaprop_isaroot {V : hSet} (E : hrel V) (r : V) : isaprop (isaroot E r).
```
DanGrayson: You could have defined the property this way:
```
Definition isaroot {V : hSet} (E : hrel V) (r : V) : hProp := ∀ w : V, E r w.
```

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,953 @@
...
+Proof.
+  repeat (apply impred ; intros). apply propproperty.
+Qed.
+
+Definition isTRR (V : hSet) (E : hrel V) (r : V) : UU :=
+  isrefl E × istrans E × isaroot E r.
+
+Lemma isaprop_isTRR (V : hSet) (E : hrel V) (r : V) : isaprop (isTRR V E r).
```
DanGrayson: One may elimiante `isaprop_isTRR `  by doing this:
```
Definition isrefl {X : UU} (R : hrel X) : hProp
  := ∀ x : X, R x x.

Definition istrans {X : UU} (R : hrel X) : hProp
  := ∀ (x1 x2 x3 : X), R x1 x2 ⇒ R x2 x3 ⇒ R x1 x3.

Definition isTRR (V : hSet) (E : hrel V) (r : V) : hProp :=
  isrefl E ∧ istrans E ∧ isaroot E r.
```
We may want those first two definitions upstream, anyway.

UniMath/Combinatorics/ZFstructures.v
```diff
@@ -0,0 +1,953 @@
...
+  apply isapropdirprod.
+  - apply (isaprop_isrefl E).
+  - apply isapropdirprod.
+    -- apply (isaprop_istrans E).
+    -- apply (isaprop_isaroot E r).
+Qed.
+
+Definition TRRGraphData (V : hSet) : UU := ∑ (E : hrel V) (r : V), isTRR V E r.
+
+Lemma isaset_TRRGraphData (V : hSet) : isaset (TRRGraphData V).
```
DanGrayson: Can eliminate this lemma by writing this:
```
Require Export UniMath.MoreFoundations.Propositions.

Definition TRRGraphData (V : hSet) : hSet := (∑ (E : hrel_set V) (r : V), isTRR V E r) %set.
```

Probably all "isaprop" and "isaset" lemmas could be eliminated this way.

UniMath/Combinatorics/ZFstructures.v
dimtsem: This breaks a lot of things without changing the content and would require a major re-write. So I have not made this change.

Is it crucial? Is it in general preferable in UniMath?

I on purpose wanted to avoid having truncated definitions, because I am not sure to what extent they are necessary for some variations of ZF structures that I would like to consider at some point.

UniMath/Combinatorics/ZFstructures.v
dimtsem: Same comment as above for this

UniMath/Combinatorics/ZFstructures.v
dimtsem: I did not see any real benefits from doing so, and this would require a pretty big re-write.

UniMath/Combinatorics/ZFstructures.v
DanGrayson: Hmm?  I'm not proposing to insert any propositional truncations, just to make the proofs of something being a proposition painless.

What do you mean by breaking a lot of things?

UniMath/Combinatorics/ZFstructures.v
DanGrayson: Okay.

UniMath/Combinatorics/ZFstructures.v
dimtsem: > Hmm? I'm not proposing to insert any propositional truncations ...

Yes, sorry, by "truncated definitions" I meant definitions using the hProp connectives, like the ones you propose. 

> ... just to make the proofs of something being a proposition painless.

Perhaps it makes `isaprop_XXX` proofs less painful, but it also requires changing many minor steps in many proofs (this is what I meant by breaking things).

For example, if as you suggest I use the following definitions:

```coq
Definition isrefl {X : UU} (R : hrel X) : hProp := ∀ x : X, R x x.
 
Definition istrans {X : UU} (R : hrel X) : hProp := ∀ (x1 x2 x3 : X), R x1 x2 ⇒ R x2 x3 ⇒ R x1 x3.
Definition isaroot {V : hSet} (E : hrel V) (r : V) : hProp := (∀ w : V, E r w).

Definition isTRR (V : hSet) (E : hrel V) (r : V) : hProp :=
  isrefl E ∧ istrans E ∧ isaroot E r.

Definition TRRGraphData (V : hSet) : hSet := (∑ (E : hrel_set V) (r : V), isTRR V E r) %set.
```
then the following  kind of replacement would be required

```coq
Definition TRRGraph_univalence_weq2 (G H : TRRGraph) : G ╝ H ≃ G ≅ H.
Proof.
  use weqbandf.
  - exact (hSet_univalence (pr1 G) (pr1 H)).
  - intro p. simpl. use weqimplimpl.
    + intros q. exact (TRRGhomo_frompath (pr1 G) (pr1 H) (pr2 G) (pr2 H) p q).
    + intros q. exact (TRRGhomo_topath (pr1 G) (pr1 H) (pr2 G) (pr2 H) p q).
    + exact (isaset_TRRGraphData (pr1 H) ((p # pr2 G)%transport) (pr2 H)).
      (* The above line doesn't work, but will work if replaced by:
      exact (setproperty (TRRGraphData (pr1 H)) _ _).
      *)
    + exact (isaprop_isTRRGhomo (hSet_univalence_map (pr1 G) (pr1 H) p)).
Defined.
```

etc. 

I started trying to make these changes along the lines you suggested, but then gave up when I realized it was taking a very long time and that I was mainly just re-expressing the same proofs (and also faced extra difficulties with the definition of certain functions). 

Is there a fundamental advantage to adopting this change that is not clear to me?

DanGrayson: Well, it does shorten the code somewhat and make it more readable, but since your code currently works, perhaps that's not worth it to you.   The decision is yours.

dimtsem: OK, got it, thanks. I will keep it as is.

dimtsem:
@DanGrayson @mortberg @siddharthist @benediktahrens Thanks a lot for the suggestions and comments. 

At this point I do not intend to make any more changes so as far as I am concerned it is ready to go.

PR #899, opened 2018-02-12
Polynomial functors and (co)inductive types

siddharthist:
This is what I've formalized so far for [my undergraduate thesis](https://github.com/siddharthist/reed-thesis)!

 - Definition of polynomial functors, definition of W- and M-types as initial (resp. final) algebras (resp. coalgebras) for such functors
 - Proofs of uniqueness for W- and M-types
 - Initial definitions of fibered algebras

Further goals: 
 - Prove the "induction principle" of Sojakova et al.
 - Show that the type `M A B` is contractible, that is, (internal) M-types exist

Many thanks to @pcapriotti and @kristinas for helping me with my thesis thus far.

### DIFF
--- a/UniMath/Induction/M/Uniqueness.v
+++ b/UniMath/Induction/M/Uniqueness.v
@@ -1,0 +1,153 @@
+(** ** Uniqueness of M-types
+
+    M-types are unique up to propositional equality.
+
+    Author: Langston Barrett (@siddharthist)
+ *)
+
+Require Import UniMath.Foundations.PartD.
+Require Import UniMath.CategoryTheory.Categories.
+Require Import UniMath.CategoryTheory.functor_categories.
+Require Import UniMath.CategoryTheory.FunctorCoalgebras.
+Require Import UniMath.CategoryTheory.categories.StandardCategories.
+Require Import UniMath.Ktheory.Utilities.
+Require Import UniMath.Induction.PolynomialFunctors.
+Require Import UniMath.Induction.M.Core.
+
+Section Uniqueness.
+
+  Local Open Scope functions.
+  Local Open Scope cat.
+
+  Context (A : UU).
+  Context (B : A → UU).
+
+  (** Following the paper, we have [X ⇒ Y] = Hom(X, Y) *)
+  Local Notation F := (polynomial_functor A B). (* can be coerced to a function *)
+  Local Notation "F*" := (polynomial_functor_arr A B).
+  Local Notation "X ⇒ Y" := (coalgebra_homo F X Y) (at level 20).
+
+  (** Since we can't use the standard categorical proof, we must re-prove that
+      final coalgebras are unique up to isomorphism.
+
+      (Lemma 5 in Ahrens, Capriotti, and Spadotti)
+   *)
+
+  (** We prove that their carriers (first projections) are isomorphic, and hence
+      equal (by univalence).
+
+      This is standard categorical reasoning: each has exactly one arrow to the
+      other, which, composing, gives an endormorphism. However, each has exactly
+      one endomorphism, the identity map. Therefore, they are isomorphic. *)
+
+  Lemma M_carriers_iso : ∏ X Y : M B, (coalgebra_ob _ X) ≃
+                                      (coalgebra_ob _ Y).
+  Proof.
+    intros X Y.
+
+    (** Get the coalgebra morphisms X → Y and Y → X via finality *)
+    pose (X_mor_Y := iscontrpr1 (pr2 Y X)).
+    pose (Y_mor_X := iscontrpr1 (pr2 X Y)).
+
+    apply (weq_iso
+             (mor_from_coalgebra_homo _ _ _ X_mor_Y)
+             (mor_from_coalgebra_homo _ _ _ Y_mor_X)).
+    - intro x.
+      apply (@eqtohomot _ _ (Y_mor_X ∘ X_mor_Y) (idfun _)).
+      refine (base_paths (coalgebra_homo_comp _ _ _ _ X_mor_Y Y_mor_X)
+                         (coalgebra_homo_id F X) _).
+      apply (proofirrelevancecontr (pr2 X X)).
+    - intro y.
+      apply (@eqtohomot _ _ (X_mor_Y ∘ Y_mor_X) (idfun _)).
+      refine (base_paths (coalgebra_homo_comp _ _ _ _ Y_mor_X X_mor_Y)
+                         (coalgebra_homo_id F Y) _).
+      apply (proofirrelevancecontr (pr2 Y Y)).
+  Defined.
+
+  (** Note the crucial use of univalence *)
+  Lemma M_carriers_eq : ∏ X Y : M B, (coalgebra_ob _ X) = (coalgebra_ob _ Y).
+  Proof.
+    exact (fun X Y => weqtopaths (M_carriers_iso X Y)).
+  Defined.
+
+  (** Now we must prove that the coalgebra morphisms, when transported along
+      the path M_carriers_eq, will be equal.
+
+      (≅⇒≡ in HoTT/M-types)
+   *)
+
+  (* We use "refine (x @ _)" instead of "rewrite" for more predictable proof terms. *)
+  Lemma M_coalg_eq : ∏ X Y : M B, M_coalgebra B X = M_coalgebra B Y.
+  Proof.
+    intros X Y.
+    pose (π1eq := (M_carriers_eq X Y)).
+    pose (f := pr1 ((pr2 Y) (M_coalgebra B X))).
+    apply (total2_paths_f π1eq).
+
+    (** Some shorthands for items we'll use *)
+    pose (is_final_X := pr2 X).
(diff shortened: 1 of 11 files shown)


UniMath/Induction/W/Core.v
```diff
@@ -0,0 +1,39 @@
+Require Import UniMath.Foundations.PartD.
+Require Import UniMath.CategoryTheory.FunctorAlgebras.
+Require Import UniMath.Induction.PolynomialFunctors.
+
+(** The W-type associated to (A, B) is the final coalgebra of the associated
```
mortberg: You want to say "initial algebra" here

DanGrayson:
The copyright notice in Fibered.v contradicts our copyright notice in https://github.com/UniMath/UniMath/blob/master/LICENSE.md , so something has to be done.

siddharthist:
@DanGrayson I believe they are both MIT licensed, could you expand a bit more on the conflict?

UniMath/Induction/W/Core.v
siddharthist: Yes, that's right! Thank you.

siddharthist:
My impression is that we just need to include the original notice (https://tldrlegal.com/license/mit-license). The author of that code has been contacted as well, they licensed it under these terms so that we could use it.

benediktahrens:
On 12/02/18 12:31, Langston Barrett wrote:
> My impression is that we just need to include the original notice 
> (https://tldrlegal.com/license/mit-license). The author of that code has 
> been contacted as well, they licensed it under these terms so that we 
> could use it.


My impression is that you are not obliged to include the original 
copyright notice, since

1. you are taking the original idea, but not the original code
2. the portion of code/ideas you are taking is not "substantial"


I suggest that you
- maintain the attribution of ideas; and
- remove the copyright notice.

DanGrayson:
Re: " I believe they are both MIT licensed, could you expand a bit more on the conflict? "

The licenses are compatible, but look at what we say about the copyright:

```
UniMath consists of the files in the UniMath github repository at
https://github.com/UniMath/UniMath .

UniMath is copyright 2015 by the UniMath Development Team, whose members are
listed at https://github.com/UniMath/UniMath in the file README.md .
```

We can't claim copyright on a file if we don't have the copyright.  Probably we should say something else in LICENSE.md, to allow for individual files that are copyright by someone else, not by us.

siddharthist:
@benediktahrens I am planning to add more of their results, if that makes a difference.

@DanGrayson Thanks for clarifying, agreed!

DanGrayson:
Re: "I am planning to add more of their results, if that makes a difference".

Yes, but will you be providing your own statements and proofs of their results?  If so, you're not copying their code.

siddharthist:
Okay, done! The adjustment of the `LICENSE` file to allow for compatibly-licensed individual files should probably be moved to a separate issue, it seems like a good idea regardless.

UniMath/Induction/W/Fibered.v
```diff
@@ -0,0 +1,127 @@
...
+  Proof.
+    induction X as [C c]; induction Y as [E e].
+    intros i j; induction i as [f f0]; induction j as [g g0].
+    refine (∑ (n : f ~ g), ∏ x, f0 x @ maponpaths (e x) _ = n (c x) @ g0 x).
+    exact (funextsec _ _ _ (@funhomotsec _ _ _ (pr2 x) f g n)).
+  Defined.
+
+  Definition algebra_mor_to_algsec {X Y : algebra_ob F}
+            (f : X ⇒ Y) : @algebra_section X (alg2fibalg Y).
+    unfold algebra_section; cbn.
```
benediktahrens: Missing `Proof.`

UniMath/Induction/W/Fibered.v
```diff
@@ -0,0 +1,127 @@
...
+    intros ? ? ? ? h.
+    exact (!homotinvweqweq0 (weqtoforallpaths _ _ _) h).
+  Qed.
+
+  (** The identity homotopy on a P-algebra section. *)
+  Definition algebra_sec_homot_id {X : algebra_ob F} {Y : FibAlg X}
+             {i : algebra_section Y} : algebra_sec_homot i i.
+  Proof.
+    induction X as [C c]; induction Y as [D d];
+      induction i as [f f0]; split with (fun _ => idpath _).
```
benediktahrens: Do you really need to destruct these pairs? Wouldn't it be cleaner to project out the things that you want to use?

UniMath/Induction/W/Fibered.v
```diff
@@ -0,0 +1,127 @@
...
+    induction X as [C c]; induction Y as [E e].
+    intros i j; induction i as [f f0]; induction j as [g g0].
+    refine (∑ (n : f ~ g), ∏ x, f0 x @ maponpaths (e x) _ = n (c x) @ g0 x).
+    exact (funextsec _ _ _ (@funhomotsec _ _ _ (pr2 x) f g n)).
+  Defined.
+
+  Definition algebra_mor_to_algsec {X Y : algebra_ob F}
+            (f : X ⇒ Y) : @algebra_section X (alg2fibalg Y).
+    unfold algebra_section; cbn.
+    induction f as [f is_mor].
```
benediktahrens: Use of projections results in nicer proof terms than destructing pairs.

UniMath/Induction/W/Fibered.v
```diff
@@ -0,0 +1,127 @@
...
+  Defined.
+
+  (** A P-algebra section.
+
+      (Definition 4.6 in Awodey, Gambino, and Sojakova)
+    *)
+  Definition algebra_section {X : algebra_ob F} :
+    forall (Y : FibAlg X), UU.
+  Proof.
+    induction X as [C c]; intro Y; induction Y as [E e].
```
benediktahrens: Prefer projections over destruction of pairs.

UniMath/Induction/M/Uniqueness.v
```diff
@@ -0,0 +1,153 @@
...
+      equal (by univalence).
+
+      This is standard categorical reasoning: each has exactly one arrow to the
+      other, which, composing, gives an endormorphism. However, each has exactly
+      one endomorphism, the identity map. Therefore, they are isomorphic. *)
+
+  Lemma M_carriers_iso : ∏ X Y : M B, (coalgebra_ob _ X) ≃
+                                      (coalgebra_ob _ Y).
+  Proof.
+    induction X as [X is_final_X]; induction Y as [Y is_final_Y]; cbn.
```
benediktahrens: projections instead

siddharthist:
@benediktahrens I've expunged all uses of `induction`. I can clean up the commit history just before merging!

UniMath/Induction/README.md
```diff
@@ -0,0 +1,17 @@
+# (Co)inductive types
+
```
DanGrayson: I suggest you identify yourself as the author of this *package* (not directory), to take credit for it, and to be notified if changes are proposed.  If you wish ...

siddharthist: @DanGrayson Sure, would you suggest removing my name/username from the individual file headers and just adding it to the README? I'm not super concerned about if/where it is listed.

UniMath/Induction/M/Uniqueness.v
```diff
@@ -0,0 +1,159 @@
...
+  (** Since we can't use the standard categorical proof, we must re-prove that
+      final coalgebras are unique up to isomorphism.
+
+      (Lemma 5 in Ahrens, Capriotti, and Spadotti)
+   *)
+
+  (** TODO: If this isn't in UniMath already, it should be.
+      appfun in Utilities?
+      (app= in HoTT/M-types) *)
+  Definition app_eq {X Y : UU} {f g : X → Y} : forall x, f = g → f x = g x :=
```
siddharthist: Do any of the maintainers have suggestions on where this should go? Or if it already exists somewhere? It's a specialization of `Ktheory.Utilities.apfun`, but that and this should probably go somewhere more like `MoreFoundations`, since neither is specific to inductive types or K-theory.

UniMath/Induction/README.md
DanGrayson: No, leave it in the individual files.  The README file is the most visible introduction to the package, so having it there (also) seems good.

UniMath/Induction/M/Uniqueness.v
DanGrayson: It's not needed, because we have these in PartA:
```
Definition toforallpaths {T:UU} (P:T->UU) (f g:∏ t:T, P t) : f = g -> f ~ g.
...
Definition eqtohomot     {T:UU} {P:T->UU} {f g:∏ t:T, P t} : f = g -> f ~ g.
```

siddharthist: Ah, duh! Thanks for the reminder 😄 I'll remove it.

UniMath/Induction/README.md
siddharthist: Done!

siddharthist:
Thanks for the thorough review, everyone! I've addressed all comments so far and rebased onto the current `master` branch. Let me know if there are more improvements to be made.

Issue #897, opened 2018-02-10
a weaker AC that doesn't imply LEM


DanGrayson:
I asked Martín Escardó: "what if we replace AC by a new axiom AC' that splits only surjections X ↠ Y of types where Y has decidable equality? Then AC' does not imply LEM (I hope), and accepting it would be more comfortable. Moreover, AC' + LEM ⇒ AC, we get all the theorems we had before, but we have to cite two axioms instead of 1.  What about the Well Ordering Axiom? If every set can be well ordered, does that imply LEM?"

His reply: "Fred Richman mentioned this AC' axiom to me. He asked whether this already implies LEM (doubting it). The well ordering theorem should be equivalent to AC."

I intend to add this weaker axiom and see what happens.

dimtsem:
I was wondering about *exactly* the principle which you call AC’ yesterday before my pull request when I realised that I will have to work with types with decidable equality and therefore asked myself whether imposing the split surjection form of the axiom of choice for only those types would suffice to prove the axiom of choice for ZF structures, without implying LEM. 

So this makes a lot of sense to me, and would be great to have. I also believe it should not imply LEM, though I have not thought about it much.

(I am interested in contributing to this if you need or want.)

PR #900, opened 2018-02-12
A weaker Axiom of Choice

DanGrayson:
This addresses #897.

 Having proved above that the Axiom of Choice implies the Law of the Excluded Middle, we would
    like to formulate a weaker axiom of choice that would be usable in formalization, but without
    implying the Law of the Excluded Middle, thus making it a more acceptable omniscience principle.
    Our idea here is to add the hypothesis that the base set have decidable equality.  Classically,
    there is no difference between the two axioms.  Recall from [isasetifdeceq] that a type with
    decidable equality is a set, so we don't include being a set explicitly in the statement of the
    axiom.

Definition AxiomOfDecidableChoice : hProp := ∀ (X:UU), isdeceq X ⇒ ischoicebase X.

Theorem AC_iff_ADC_and_LEM : AxiomOfChoice ⇔ AxiomOfDecidableChoice ∧ LEM.

### DIFF
--- a/UniMath/MoreFoundations/AxiomOfChoice.v
+++ b/UniMath/MoreFoundations/AxiomOfChoice.v
@@ -80,33 +80,6 @@ Proof.
     induction (pr2 sec x). exact (pr2 (pr1 sec x)).
 Defined.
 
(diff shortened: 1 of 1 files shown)


benediktahrens:
lgtm

PR #901, opened 2018-02-13
remove extraneous lemma from Ktheory/Utilities

siddharthist:
I found the lemma in Foundations when grepping for "inhabited" and "prop", so I added a comment explaining the lemma with those words.


DanGrayson:
lgtm

Adding the comment to Foundations is okay.

PR #902, opened 2018-02-18
make data of total cat depend only on data

benediktahrens:
For reuse of the definition when defining total bicategory...


peterlefanulumsdaine:
Belatedly: yes, I also am in favour of this change, for the reason @benediktahrens states!

PR #909, opened 2018-02-20
Generalize curry and uncurry

siddharthist:
I acknowledge this is a change to `Foundations`, so I leave it totally to your discretion @DanGrayson. 

I see no harm in generalizing these definitions, though it also might be helpful to provide (several) non-dependent versions for simplicity. Some other changes would have to be made around the codebase before merging, either changing proofs to use a new non-dependent version or providing implicit arguments.

### DIFF
--- a/UniMath/Algebra/IteratedBinaryOperations.v
+++ b/UniMath/Algebra/IteratedBinaryOperations.v
@@ -457,7 +457,7 @@ Section NatCard.
   Defined.
 
   Corollary nat_plus_associativity' n (m:stn n->nat) (k:∏ i, stn (m i) -> nat) :
-    stnsum (λ i, stnsum (k i)) = stnsum (uncurry k ∘ lexicalEnumeration m).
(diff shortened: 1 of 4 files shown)


DanGrayson:
I like the change, but it breaks some proofs, according to travis, such as this one:
```
COQC UniMath/Algebra/IteratedBinaryOperations.v
File "./UniMath/Algebra/IteratedBinaryOperations.v", line 460, characters 42-49:
Error:
In environment
n : nat
m : ⟦ n ⟧ → nat
k : ∏ i : ⟦ n ⟧, ⟦ m i ⟧ → nat
p : ∑ x : ⟦ n ⟧, ⟦ m x ⟧
Unable to unify "?Z0" with "?Z".
```

siddharthist:
@DanGrayson Right, that's what I was asking about. As I see it, we have two options:
 1. Introduce another version of `curry` and `uncurry` for non-dependent functions and use this where the former versions used to be used.
 2. Go through the proofs and explicitly add the `Z` parameter (which can't always be inferred, as seen in the error you posted). 

What do you think?

siddharthist:
Great, thanks for the patch! I've applied it and re-committed.

PR #911, opened 2018-02-21
Categories of well-ordered sets

mortberg:
This PR defines two category structures on well-ordered sets:

1. The first where the morphisms are functions preserving ordering and initial segments ([wosetfuncat]).
2. The second with arbitrary functions as morphisms ([WOSET]).

Both of these have initial and terminal objects. The former doesn't seem to have binary products (see
discussion in the file), but using Zermelo's well-ordering theorem (and hence the axiom of choice) I have proved that the latter merely has binary products. I believe that the proofs that WOSET has all limits and colimits carry over exactly like the proof for binary products, but because the category only **merely** has binary products I ran into all kinds of problems when trying to prove that it merely has exponentials, see discussion at the end of the file.

In the process of writing this code I added various things to the file on well-ordered sets and cleaned the category_hset_structures.v file (in particular I renamed has_exponentials to Exponentials as it is a structure, not a property).


DanGrayson:
That's a strange travis error.  
```
+ git pull --depth 9 theupstream master
From https://github.com/UniMath/UniMath
 * branch            master     -> FETCH_HEAD
 * [new branch]      master     -> theupstream/master
Auto-merging sub/coq
CONFLICT (submodule): Merge conflict in sub/coq
```

@palmskog -- what is the purpose of these lines in our travis script?:
```
git remote add theupstream https://github.com/UniMath/UniMath.git
git pull --depth 9 theupstream master
```
I tried the travis sequence at home and got this error message:
```
fatal: refusing to merge unrelated histories
```
, not the same as in the jobs reported above, which also may indicate that the `--depth 9` options can cause problems.

dimtsem:
Just a couple of comments that I already made to @mortberg regarding the content:

1. The category `wosetfuncat` should be a poset, provable using the unproven `transfiniteRecursion` in WellOrderedSets.v and this may not be a bad result to have in there since there might be uses for such a poset. (If it is not included in this PR I will do it at some point.)

2. It is is not entirely clear to me whether `isaset WellOrderedSet` can be proven without LEM. (In the  following argument sketched by @DanGrayson in WellOrderedSets.v

```
Lemma isaset_WellOrderedSet : isaset WellOrderedSet.
Proof.
  (*
     First show that an order preserving equivalence f : X ≃ Y of well
     ordered sets has the property that for any x, f x is least element
     greater than f y for every y < x.  Now given also an order
     preserving f' : X ≃ Y, take the smallest x where f x ≠ f y.  Since
     they agree on previous values, they agree at x, too, using the previous
     statement.
   *)
```

don't you need that for any `x : X` either `f x = f' x` or `f x ≠ f' x`?)

palmskog:
@DanGrayson I'm investigating these problems now, I will probably have a revised version of the Travis script by tomorrow as a pull request. I think it's a question of inheriting some script parameters that were appropriate for non-Coq projects.

DanGrayson:
Actually, the unproven `transfiniteRecursion` is for families of types.  The classical theorem for families of sets would be easily proven the standard way.

PR #915, opened 2018-02-23
Filenames

### LINKED ISSUE #686: Capitalization
I suggest that we rename all the .v files in UniMath/UniMath into starting with a small letter and subdirectories into starting with a capital. This way files, as seen in the GitHub, will be ordered alphabetically.

The same probably applies to UniMath/TypeTheory.

siddharthist:
Fixes #686. See commit messages for scripts/regexes used.


UniMath/CategoryTheory/Limits/Equalizers.v
```diff
@@ -67,7 +67,7 @@ Section def_equalizers.
     apply (base_paths _ _ T').
   Defined.
 
-  (** Definition and construction of equalizers. *)
+  (** Definition and construction of Equalizers. *)
```
DanGrayson: oops, the search and replace went into comments, too.  Maybe you should have replaced ".equalizers." by ".Equalizers.", instead.

UniMath/CategoryTheory/Limits/Coproducts.v
```diff
@@ -353,7 +353,7 @@ Defined.
 End def_functor_pointwise_coprod.
 
 (** * Coproducts from colimits *)
-Section coproducts_from_colimits.
```
DanGrayson: Here's a non-file name getting changed.

siddharthist:
If I redo this, I'll do it on a per-package basis; it's too easy for this to fall out-of-date and into conflict. If someone else is interested in taking this up, the scripts I wrote are in the above commits.

PR #918, opened 2018-02-26
Split CocontFunctors into a package "Chains"

siddharthist:
This splits the file `CocontFunctors` into a package.

* `Chains.v` contains the definition of a chain diagram
* `Adamek.v` contains Adamek's theorem
* `OmegaCocontFunctors.v` contains results about omega-cocontinuous functors
* `Cochains.v` is brand new, and contains the definition of a cochain diagram (this will be useful for defining M-types)

Having separate packages is nice, since most modules that import `CocontFunctors` only needed a subset of its functionality. Relatedly, Adamek's theorem and the results about cocontinuous functors aren't interdependent, so I think it makes a bit more sense to separate them.


DanGrayson:
By the way, it's not a "package", it's a "subdirectory".  Our "packages" are the subdirectories of `UniMath`.

siddharthist:
It is worth noting that current terminology is dual to that I linked in the paper above. There, Ahrens et al. call diagrams of the form `X0 <-- X1 <-- ...` chains and the others cochains.

mortberg:
Do you know of https://www.ps.uni-saarland.de/~rech/containers/ ? This gives a construction of W-types in UF by carving them out as the well-founded M-types, so once you have formalized the construction of M-types you should be able to get the W-types as well.

I think Benedikt and some other people were working on formalizing this at the UniMath school in December. I never saw any PR about this, but @benediktahrens probably knows more than I do about how far they got.

DanGrayson:
Re: "There, Ahrens et al. call diagrams of the form `X0 <-- X1 <-- ...` chains and the others cochains."

Ahrens' way makes a little more mathematical sense, since it's reminiscent of a (positive) chain complex.

On the other hand, since a "chain" is a totally ordered set, there's not much to recommend this terminology be used at all.  The terms "tower" and "cotower" are more traditional.  See, for example, https://ncatlab.org/nlab/show/homotopy+limit#SequentialHocolims .

siddharthist:
@mortberg Yes, I learned of that at the UniMath conference :) Thanks for the pointer! I got a link to some of their work: https://github.com/LuisScoccola/UniMath/tree/inductives. 

@DanGrayson That seems like a good change to me. Perhaps we can make an issue after this PR?

DanGrayson:
@siddharthist Re: " That seems like a good change to me. Perhaps we can make an issue after this PR? "

Feel free to make an issue out of it for later, if you agree with the proposal.

benediktahrens:
@siddharthist : there is now a conflict in this PR. Could you fix it?

siddharthist:
Okay, fixed!

siddharthist:
I've just redone cochains to avoid using `match` and simplify some proofs.

PR #919, opened 2018-02-26
Rename "rng" to "ring"

### LINKED ISSUE #822: Rename rng to ring
Some use the term "rng" to refer to [non-unital rings](https://en.wikipedia.org/wiki/Rng_(algebra)). Shortening to rng would be confusing without this fact, and is misleading considering it.

varkor:
The current use of "rng" instead of "ring" to mean the latter is misleading, as "rng" often refers to a ring without a multiplicative identity (hence the missing "i"). This fixes #822.


DanGrayson:
Some filenames have also changed, and we have lost their history, as here:
https://github.com/varkor/UniMath/commits/392d3ce4d7b8a9ce90de81ef2f36bd4dfc8e8e59/UniMath/CategoryTheory/categories/commrings.v

I suspect that splitting the change into two commits will preserve the history: in one commit the filenames are changed but not the contents; in the other the contents are changed.  I think git and github have an easier time detecting file renaming when reconstructing history if the contents don't also change (much?).  (File renaming is not tracked in git repositories, by design.)

There is some discussion of that issue at https://stackoverflow.com/questions/2314652/is-it-possible-to-move-rename-files-in-git-and-maintain-their-history , with differing opinions.

Discussion?

varkor:
This is a good point: I've separated the changes into two commits, but now, although GitHub recognises the second commit simply renames the files, the history is still broken. (I feel this is a feature GitHub should provide to work around git's deficiencies, but for the moment, this doesn't seem to the the case.)

DanGrayson:
Well, maybe git or gitk will do a better job than github when reconstructing the history.  It just stands to reason that someone will take advantage of it, so thanks!

DanGrayson:
Yes, I just tested git on your renaming -- and "git log --follow" does track the renaming now, so that's good.  I don't know whether it tracked it before.

DanGrayson:
At least github provides a clear picture of the commit itself, at https://github.com/UniMath/UniMath/pull/919/commits/9a1208dddd379a7c9129826943ecc6012bda7a63 .

varkor:
@siddharthist: this particular renaming is convenient in that the character combination "rng" is not common (in fact, I can't think of any words or common abbreviations making use of it), so replacing all occurrences (with a "Replace in all files" command), and scanning through the replacements to make sure they seemed sensible, sufficed, along with renaming the relevant files.

To tackle something like #915, I'd probably resort to more complicated regular expressions too, though it's probably inevitable that you'll get false positives/negatives regardless, so I imagine it won't be possible to avoid some manual changes, unfortunately. Until Coq has some dedicated refactoring tools, I don't think this sort of change is ever really going to be pleasant.

DanGrayson:
Also, our makefile will build a TAGS file for you, making it easy to do global and replace operations throughout the entire set of *.v files with emacs, in the prescribed (dependency) ordering.

DanGrayson:
And for changing file names with such things as search and replace, nothing beats M-x dired-toggle-read-only in emacs.

siddharthist:
This should probably be merged before other PRs that touch rings (e.g. #796) so that it doesn't continually fall out of date.

DanGrayson:
The travis error was this:
```
COQC UniMath/CategoryTheory/All.v
File "./UniMath/CategoryTheory/All.v", line 124, characters 15-53:
Error: Unable to locate library UniMath.CategoryTheory.categories.rngs.
```

varkor:
@DanGrayson: I've rebased and fixed the errors. Should be good to try again now!

PR #920, opened 2018-02-26
Suppress build output

### LINKED ISSUE #773: some files print needlessly
```
COQC UniMath/CategoryTheory/Inductives/Lists.v
     = 0
     : nat
     = 2
     : nat
     = 2
     : nat
     = 7
     : nat
     = 9
     : nat
     = 4
     : nat
     = 16
     : nat
     = 2,, 5,, 2,, tt
     : list (pr1 natHSET)
COQC UniMath/CategoryTheory/Inductives/Trees.v
     = 11
     : nat
     = 11
     : nat
```

varkor:
When building, a number of files print output (some seemingly unnecessarily, and some to previously ensure that Travis would not fail spuriously while trying to detect a lack of progress). I think the output in `CategoryTheory/Inductives` can definitely be removed, but the easiest way to tell whether Travis still has issues with `SubstitutionSystems` is to try it.

If Travis times out with this change, perhaps the best solution is to use `travis_wait` (https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received) rather than printing needlessly?

Fixes #773.

### DIFF
--- a/UniMath/CategoryTheory/Inductives/Lists.v
+++ b/UniMath/CategoryTheory/Inductives/Lists.v
@@ -240,13 +240,13 @@ Definition sum : List natHSET -> nat :=
 (* Eval vm_compute in sum testlistS. *)
 
 (* All of these compute *)
-Eval lazy in length _ (nil natHSET).
-Eval lazy in length _ testlist.
-Eval lazy in length _ testlistS.
-Eval lazy in sum testlist.
-Eval lazy in sum testlistS.
-Eval lazy in length _ (concatenate _ testlist testlistS).
-Eval lazy in sum (concatenate _ testlist testlistS).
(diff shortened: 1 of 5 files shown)


DanGrayson:
Good idea. 

About the commented out Eval statements: it might be good to revive them as tests, possibly in a new file called `Tests.v` if the tests are slow, removing the comments.

For example, revive
```
Eval lazy in length _ testlist.
```
as
```
Goal length _ testlist = 2.
  reflexivity.
  Qed.
```

varkor:
I've replaced the `Eval` statements with `Goal`s, as you suggested: is this the sort of thing you had in mind?

DanGrayson:
Yes, exactly, thank you!

PR #924, opened 2018-02-26
remove duplicate definitions of `flip`

siddharthist:
Also, generalizes it and shows that it's a weak equivalence.

### DIFF
--- a/UniMath/MoreFoundations/PartA.v
+++ b/UniMath/MoreFoundations/PartA.v
@@ -164,3 +164,17 @@ Proof.
     apply transitive_paths_weq.
     apply (toforallpaths _ _ _ (transportf_const p Y) y1).
 Defined.
+
+(** Flip the arguments of a function *)
+Definition flipsec {A B : UU} {C : A -> B -> UU} (f : ∏ a b, C a b) : ∏ b a, C a b :=
+  λ x y, f y x.
+Notation flip := flipsec.
+
+(** Flip is a weak equivalence (in fact, it is an involution) *)
+Lemma isweq_flipsec {A B : UU} {C : A -> B -> UU} : isweq (@flipsec A B C).
+Proof.
+  apply (isweq_iso _ flipsec); reflexivity.
+Defined.
+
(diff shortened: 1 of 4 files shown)


DanGrayson:
This is okay with me, but is it going to be useful for anything?

siddharthist:
It is used twice already, and is a common pattern in functional programming. I don't have much grounds to say it might be used more, but it's certainly worth not defining it twice.

DanGrayson:
Oh, yes, I agree.  But I was actually wondering if you have an application in mind for ` flipsec_weq `.

UniMath/MoreFoundations/PartA.v
```diff
@@ -138,3 +138,16 @@ Proof.
...
   apply idpath.
 Defined.
+
+(** Flip the arguments of a function *)
+Definition flipsec {A B : UU} {C : A -> B -> UU} (f : ∏ a b, C a b) : ∏ b a, C a b :=
+  λ x y, f y x.
+Notation flip := flipsec.
+
+(** Flip is a weak equivalence (in fact, it is an involution) *)
+Lemma isweq_flipsec {A B : UU} {C : A -> B -> UU} : isweq (@flipsec A B C).
```
mortberg: `Proof.` missing

siddharthist:
@DanGrayson The one that comes to mind is using `intermediate_weq` with something of the form `∏ x y, z` to turn it into `∏ y x, z`. I was fiddling with a proof that did just that, so that I could then `use weqonsecfibers`.

siddharthist:
Perhaps it's because I've been thinking about it recently, but when showing that `type_precat` has exponentials, I'm trying to prove a goal of the form `Y → X → X × Y` which would be solved elegantly with `flip dirprodpair`.

siddharthist:
I've updated this to account for where I added another version of `flip` in #927

UniMath/MoreFoundations/PartA.v
siddharthist: Fixed! Thanks.

PR #925, opened 2018-02-27
Bicategories "from scratch", displayed bicategories

benediktahrens:
- maps to and from the existing bicategories
- some examples of displayed bicategories
- the bicategory of CwFs

with @maggesi


UniMath/CategoryTheory/PlainBicat/BicatFromBicategory.v
```diff
@@ -0,0 +1,190 @@
...
+    apply whisker_right_right.
+  - (* 11 vcomp_whisker *)
+    apply twomor_naturality.
+  - (* 12a lunitor_linvunitor *)
+    apply (iso_inv_after_iso (left_unitor f)).
+  - (* 12b linvunitor_lunitor *)
+    apply (iso_after_iso_inv (left_unitor f)).
+  - (* 13a runitor_rinvunitor *)
+    apply (iso_inv_after_iso (right_unitor f)).
+  - (* 13b rinvunitor_runitor *)
```
DanGrayson: These numbered comments look mysterious.

benediktahrens: Oh, thanks for pointing that out - it refers to a list of axioms on the ncatlab. We have added an explanation in https://github.com/UniMath/UniMath/pull/925/commits/47f7a6f6f0639c44d3fe4559cec18aac93d55c09 .

benediktahrens:
We should mention that this is WIP, and needs more work and cleanup. However, it would be good to have it in UniMath now to avoid divergence.

siddharthist:
This is not a very important point, but I think it's nice to have a cleaner commit history. One can do some minor cleanup with a `rebase` which doesn't have to be painful and take forever. For example, 579617aecc98c1756f923e4f8f3b5d0395626c3f could easily be squashed into the previous commit.

siddharthist:
Is there any reason in particular to not have this replace `bicategories` as it stands now? I think it's generally good to present only a single interface to a concept to avoid duplication of efforts.

benediktahrens:
On 01/03/18 20:10, Langston Barrett wrote:
> Is there any reason in particular to not have this replace 
> |bicategories| as it stands now? I think it's generally good to present 
> only a single interface to a concept to avoid duplication of efforts.

I think of the following two reasons:

- Our PR does not implement all the material covered by the existing 
code on bicategories, in particular, not bi-univalence.
- The definition in terms of enrichment in 1-categories as given by 
Mitchell is concise, and easy to read, while the one we implement is 
easier to work with (at least we expect this to be the case), but more 
difficult to read. In particular, it is not obvious that the list of 
axioms of a bicategory are all there should be, without comparing it to 
another definition.

The second point warrants keeping the current "bicategories" indefinitely.

benediktahrens:
On 28/02/18 19:13, Langston Barrett wrote:
> This is not a very important point, but I think it's nice to have a 
> cleaner commit history. One can do some minor cleanup with a |rebase| 
> which doesn't have to be painful and take forever. For example, 579617a 
> <https://github.com/UniMath/UniMath/commit/579617aecc98c1756f923e4f8f3b5d0395626c3f> 
> could easily be squashed into the previous commit.

I agree with everything you are saying, but for my peace of mind, I 
decide to not worry about the git history.

UniMath/CategoryTheory/PlainBicat/BicatOfCats.v
```diff
@@ -0,0 +1,141 @@
...
+  use tpair.
+  - intro C. cbn in C. exact (functor_identity C).
+  - cbn. intros a b c f g.
+    exact (functor_composite f g).
+Defined.
+
+Definition cat_prebicat_1_id_comp_cells : prebicat_1_id_comp_cells.
+Proof.
+  exists cat_precategory_data.
+  intros a b f g. cbn in *. exact (nat_trans f g).
```
siddharthist: In general, I think it's nice not to introduce specific names for things that won't get used (e.g. this could be `intros ? ? f g.`). This also makes it easier to move the code around.

siddharthist:
@benediktahrens I'm afraid I don't know anything about bicategories...

DanGrayson:
I've been meaning to take a look for days now ...

peterlefanulumsdaine:
I’ve been meaning to look as well!  Based on a rough look over the outline of the new material, it looks generally great; I haven’t looked at the code in detail.

My one question so far is similar to @siddharthist ’s earlier, about the relationship with the existing `bicategories`.  I agree with @benediktahrens about the benefits of keeping both in some form.  This version is more elementary, while the version given by Mitchell and me in `bicategories/prebicategory.v` is packaged in a way that’s closer to most presentations in the classical literature (though not all — some are like this PR’s version).  So it’s both theoretically reassuring to know that they’re equivalent, and practically useful to have both available as interfaces for constructing/using them.

However, I agree with @siddharthist ’s concerns about avoiding duplication of effort in future work.  Might it make sense to consolidate them to a single subpackage, clearly indicating one version as the “definitive” one, and the other as a module within it (or a few modules) giving an alternative interface?  At the moment, naming makes the older version look more definitive, but the new one is much more developed and presumably is intended to be definitive going forward.  So one could e.g. name the new version just `Bicategories` rather than `PlainBicats`, so that users clearly see it as the definitive thing to import, and then move the files of the old `bicategories` to live in the same subpackage under a name something like `WkCatEnrichment`?

If the present PR authors want to keep focusing on the development of new material, then I can happily do this.

benediktahrens:
Your plan sounds very good!

We are not currently working on new material, but will come back to it 
soon. The new material might fit better in UniMath/TypeTheory, though - 
to be discussed in a separate thread.

If you have a clear vision of how you would like to organize the 
material on bicategories, then I would be glad if you could implement 
it. Otherwise I will attempt a reorganisation in the vein of what you 
suggest myself.

UniMath/CategoryTheory/PlainBicat/Bicat.v
```diff
@@ -0,0 +1,1046 @@
...
+ *)
+
+
+(** Chaotic bicat *)
+
+Section chaotic_bicat.
+
+Variable C : precategory.
+
+Definition chaotic_prebicat_data : prebicat_data.
```
DanGrayson: A little less verbose would be:
```
Definition chaotic_prebicat_data : prebicat_data.
Proof.
  use tpair.
  - use tpair.
    + exact C.
    + cbn. intros _ _ _ _. exact unit.
  - cbn; repeat (use tpair); cbn ; intros ; exact tt.
Defined.
```

UniMath/CategoryTheory/PlainBicat/DispBicat.v
```diff
@@ -0,0 +1,742 @@
+(* ******************************************************************************* *)
+(** * Bicategories
+    Benedikt Ahrens, Marco Maggesi
+    February 2018
```
DanGrayson: I suppose the things in this file are called "display bicategories".  A comment stating that and saying what they are and giving a reference or two might be good.

PR #927, opened 2018-02-27
type_precategory has binary coproducts

siddharthist:
Thanks @DanGrayson and @benediktahrens for the pointers to the right lemmas!

This PR has a TODO, I wanted to ask the maintainers if the lemma I added was a duplicate (this seems like it should be somewhere, but I didn't find anything when fuzzy-searching for "paths", "weq", and "dirprod"). If not, where should it go?


benediktahrens:
I am not aware of the lemma `pathsdirprodweq` existing already.

I think that it might be useful to prove it slightly differently: after the insertion of the first equivalence `total2_paths_equiv`, one can use a function of the form
```
x = x' -> (x = y) \simeq (x' = y)
```
which would often come in handy in other contexts, too.

siddharthist:
@benediktahrens Any feedback on where those lemmas should live?

benediktahrens:
On 27/02/18 23:32, Langston Barrett wrote:
> @benediktahrens <https://github.com/benediktahrens> Any feedback on 
> where those lemmas should live?


In as basic a spot as possible, meaning MoreFoundations/PartA.

siddharthist:
Hmm, I'm not sure I understand how to prove Benedikt's lemma (where are the quantifiers?). I've moved the auxiliary lemma into `MoreFoundations` and removed `transport_const`, which is duplicated from `Foundations.PartA`.

siddharthist:
I've tacked on binary products, since the proof is near-identical.

siddharthist:
Okay, sorry to add even more but I've also put in exponential objects, and in doing so, have split the results about `type_precat` into their own file, and made a few other small organizational changes. cf the commit message

DanGrayson:
Here's a shorter proof:
```
Definition pathsdirprodweq {X Y : UU} {x1 x2 : X} {y1 y2 : Y} :
  (( x1 ,, y1 ) = ( x2 ,, y2 )) ≃ (x1 = x2) × (y1 = y2).
Proof.
  intermediate_weq (dirprodpair x1 y1 ╝ dirprodpair x2 y2).
  - apply total2_paths_equiv.
  - apply weqfibtototal; intro p.
    rewrite transportf_const.
    apply idweq.
Defined.
```

benediktahrens:
On 28/02/18 12:18, Daniel R. Grayson wrote:
> Here's a shorter proof:
> 
> |Definition pathsdirprodweq {X Y : UU} {x1 x2 : X} {y1 y2 : Y} : (( x1 ,, 
> y1 ) = ( x2 ,, y2 )) ≃ (x1 = x2) × (y1 = y2). Proof. intermediate_weq 
> (dirprodpair x1 y1 ╝ dirprodpair x2 y2). - apply total2_paths_equiv. - 
> apply weqfibtototal; intro p. rewrite transportf_const. apply idweq. 
> Defined. |

That is more or less the proof I am suggesting, except that instead of 
the rewrite (which is ugly in a transparent term), I am suggesting to 
use the construction
```
forall x x' y, x = x' -> (x = y) simeq (x' = y)
```
which is given by precomposing with the given identity or its inverse in 
either direction.

siddharthist:
Okay, I will try again to implement that at some point today!

siddharthist:
I've made another change. Since you've already reviewed this @DanGrayson, I'm happy to split it into a separate PR after this one is merged.

The proofs that `weqfunfromcoprodtoprod` and `weqfuntoprodtoprod` show that the product and coproduct types have their categorical universal properties are actually more general than previously stated. I've rephrased them in entirely category-theoretic language, so that one can instead use similar statements as alternative definitions of coproducts and products. 

A similar approach should help me prove `type_precat` has all limits in #928, I'll try that next.

DanGrayson:
lgtm

PR #928, opened 2018-02-28
Limits in `type_precat` a la "Homotopy limits in type theory"

siddharthist:
As noted in the comment, these are in the style of "Homotopy limits in type theory" by Jeremy Avigad, Chris Kapulkin, and Peter LeFanu Lumsdaine ([arXiv:1304.0680v3](https://arxiv.org/abs/1304.0680v3)).

I prove the following universal property:
```coq
  (** The above weak equivalence specialized to the case of [standard_limit]s *)
  Definition standard_limit_up_weq {X} : (X → standard_limit d) ≃ cone d X :=
    weqpair (into_cone_to_cone (type_cone d)) (limit_universal X).
```
I haven't yet managed to translate it back into `isLimCone`.

cc @peterlefanulumsdaine


siddharthist:
I believe that to translate that universal property back into category-theoretic language will require a characterization of cone paths similar to that [in the original work](https://github.com/peterlefanulumsdaine/hott-limits/blob/master/Limits1.v#L155). This is somewhat outside of the scope of my goals, so I likely won't work on it soon.

benediktahrens:
@siddharthist : Travis says
```
COQC UniMath/Induction/M/Limits.v
File "./UniMath/Induction/M/Limits.v", line 24, characters 37-48:
Error: The reference type_precat was not found in the current environment.
```

siddharthist:
@benediktahrens Ah, I moved `type_precat` to a different file in #927. Fixed now!

PR #935, opened 2018-03-09
Groupoids

siddharthist:
Elementary definitions and lemmas about groupoids
 - Pregroupoids: categories where every arrow is an iso
 - Groupoids (pregroupoids with homsets)
 - Univalent groupoids
 - The maximal subgroupoid of a type that consists of all objects and only the isomorphisms
 - The pregroupoid of paths of a type
 - The pregroupoid of paths is a univalent groupoid when the type is of hlevel 3


siddharthist:
Ah, Travis failed... I'll try to fix this soon!

siddharthist:
I've found a few more places in the library that use similar constructions, I'll try to unify them and re-open this PR.

PR #939, opened 2018-03-11
Makefile: Partially revert 7bba938ebc

nyraghu:
* Makefile (doc): Ensure that the in-place sed command works with GNU
  sed.  Restores the command to its previous invocation.


nyraghu:
[sed in-place flag that works both on Mac (BSD) and Linux](https://stackoverflow.com/a/22084103)

DanGrayson:
Thanks, again.  I guess Benedikt knew that when he added the backup option.  Benedikt, why back up the files at all?

benediktahrens:
I copied the line from another project, without understanding what exactly it is doing.

nyraghu:
> why back up the files at all?

Because `sed -i'.bk'` is a portable way of in-place replacement that works with both BSD sed and GNU sed, as explained in the SO posting that I'd linked to.  If desired, one can add another command to the target, to remove the backups, but it doesn't seem necessary to me.

I should have perhaps mentioned that with the command as it is in master, GNU sed in Linux produces an error as follows:

``` bash
$ make doc
mkdir -p enhanced-html
cp util/enhanced-doc/proofs-toggle.js enhanced-html/proofs-toggle.js
COQDOC
sed -i '.bk' -f util/enhanced-doc/proofs-toggle.sed enhanced-html/*html
sed: can't read .bk: No such file or directory
Makefile:209: recipe for target 'doc' failed
make: *** [doc] Error 2
```
The change in the PR eliminates this error.

DanGrayson:
Oh, thank you.

PR #941, opened 2018-03-12
add installation of "num" library to instructions

DanGrayson:
... for those with opam.  We've had a report that it is now needed, and
I've noticed the same thing on another branch.


DanGrayson:
I've cancelled the travis job, because this change affects only INSTALL.md.

PR #942, opened 2018-03-12
Displayed adjunctions

tamaravonglehn:
- a displayed adjunction gives natural weak equivalences of homsets
(FFc -> d) <-> (c -> GGd)
- the right adjoint functor of a displayed adjunction preserves cartesian morphisms


benediktahrens:
There are two things to fix before this can be merged. The first one is that we need to get rid of 
```
Require Import TypeTheory.Displayed_Cats.ComprehensionC.
```
For this, we have to move the stuff from that file that is needed here to a file in UniMath/UnIMath. Specifically, insert
```
Definition is_cartesian_disp_functor
  {C C' : category} {F : functor C C'}
  {D : disp_cat C} {D' : disp_cat C'} (FF : disp_functor F D D') : UU
:= ∏  {c c' : C} {f : c' --> c}
      {d : D c} {d' : D c'} (ff : d' -->[f] d),
  is_cartesian ff -> is_cartesian (#FF ff).

Lemma isaprop_is_cartesian
    {C : category} {D : disp_cat C}
    {c c' : C} {f : c' --> c}
    {d : D c} {d' : D c'} (ff : d' -->[f] d)
  : isaprop (is_cartesian ff).
Proof.
  repeat (apply impred_isaprop; intro).
  apply isapropiscontr.
Qed.
```
somewhere into UniMath/CategoryTheory/DisplayedCats/Fibrations, after the definition of `is_cartesian`.
Then delete the `Require Import...` line above.

benediktahrens:
The second thing in need of fixing is the merge conflict in UniMath/CategoryTheory/.package/files.
We can fix that for you, or you do it yourself, as you prefer.

Nice proof!

PR #949, opened 2018-03-24
category theory: hint databases for auto and autorewrite

siddharthist:
I wanted to bring up another discussion about the use of automation and tactics in UniMath (cf #828). Very often, I see scripts like `rewrite <- assoc; rewrite id_left; rewrite id_right; reflexivity.`. It seems like when we're using `rewrite` in particular, we don't care about the final proof term (it has a lot of `internal_**` stuff). This seems like an ideal place to save human effort with automation! 

What do you all think about the use of `auto` and `autorewrite` in situations where `rewrite` is already accepted?

If helpful, I can attach some proofs in this file that are simplified by `auto` and `autorewrite`.

### DIFF
--- a/UniMath/CategoryTheory/Categories.v
+++ b/UniMath/CategoryTheory/Categories.v
@@ -213,6 +213,15 @@ Arguments id_left [C a b] f.
 Arguments id_right [C a b] f.
 Arguments assoc [C a b c d] f g h.
 
+(** For use with `auto with cat` and `autorewrite with cat`. *)
+Hint Resolve id_left : cat.
+Hint Resolve id_right : cat.
+Hint Resolve assoc : cat.
+Hint Rewrite id_left : cat.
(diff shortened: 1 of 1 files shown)


benediktahrens:
I agree with you that we should avoid using the `rewrite` tactic.

The "effort" one would save with this kind of automation mainly seems to be "typing effort", not "thinking effort". I am not sure that it is worth it.

siddharthist:
@benediktahrens That's fair. To specify a little more, I really mean "remembering the names of the lemmas"-effort. It's not particularly fun to mode-switch between, say, lemmas for associativity in monoids and those in categories. 

I should also say that other Coq practitioners (notably many people at MIT) seem to think that this sort of small-scale automation is greatly helpful, and part of my reason for suggesting its implementation is a sense of curiosity at their claims. Might it be worth experimenting with?

Issue #608, opened 2017-02-24
Prove converse of a result on adjunctions


mortberg:
See comment on https://github.com/UniMath/UniMath/pull/606#issuecomment-282258761. This could be a nice exercise for someone who wants to learn UniMath.

siddharthist:
I wonder if an even stronger result holds: that there is a weak equivalence between such "natural weak equivalences of hom-types" and adjunctions. I.e. this logical equivalence might be a a full weak equivalence.

m-lindgren:
I think this can be closed. Fixed by #950 and #952.

benediktahrens:
@m-lindgren : thanks for finding this - closing.

PR #950, opened 2018-03-26
add adjunction from natural weq of homsets

tamaravonglehn:
Proof of converse to adjunction_hom_weq #608.

### DIFF
--- a/UniMath/CategoryTheory/Adjunctions.v
+++ b/UniMath/CategoryTheory/Adjunctions.v
@@ -628,3 +628,79 @@ Section HomSetIso_from_Adjunction.
   Qed.
 
 End HomSetIso_from_Adjunction.
+
+
+(** * Adjunction defined from a natural isomorphism on homsets (F A --> B) ≃ (A --> G B) *)
+
+Section Adjunction_from_HomSetIso.
+
+  Context {C D : precategory} {F : functor C D} {G : functor D C}
+          {hom_weq :  ∏ {A : C} {B : D}, F A --> B ≃ A --> G B}
+          {natural_precomp : ∏ (A : C) (B : D) (f : F A --> B) (X : C) (h : X --> A),
+                             hom_weq (#F h · f) = h · hom_weq f}
+          {natural_postcomp : ∏ (A : C) (B : D) (f : F A --> B) (Y : D) (k : B --> Y),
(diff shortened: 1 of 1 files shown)


UniMath/CategoryTheory/Adjunctions.v
```diff
@@ -628,3 +628,79 @@ Section HomSetIso_from_Adjunction.
...
+      apply id_right.
+  Defined.
+
+  Let η := unit_from_hom.
+  Let ε := counit_from_hom.
+
+  Definition adj_from_homsetiso : are_adjoints F G.
+  Proof.
+    apply (mk_are_adjoints F G η ε).
+    apply tpair.
```
siddharthist: Is there a more specific function than `tpair` that could be used here?

UniMath/CategoryTheory/Adjunctions.v
```diff
@@ -628,3 +628,79 @@ Section HomSetIso_from_Adjunction.
...
+    - intros B B' k. cbn.
+      rewrite <- inv_natural_postcomp.
+      rewrite <- inv_natural_precomp.
+      apply maponpaths.
+      rewrite id_left.
+      apply id_right.
+  Defined.
+
+  Let η := unit_from_hom.
+  Let ε := counit_from_hom.
```
siddharthist: Since these are only used in one further line, they probably don't need special names.

UniMath/CategoryTheory/Adjunctions.v
```diff
@@ -628,3 +628,79 @@ Section HomSetIso_from_Adjunction.
...
+Section Adjunction_from_HomSetIso.
+
+  Context {C D : precategory} {F : functor C D} {G : functor D C}
+          {hom_weq :  ∏ {A : C} {B : D}, F A --> B ≃ A --> G B}
+          {natural_precomp : ∏ (A : C) (B : D) (f : F A --> B) (X : C) (h : X --> A),
+                             hom_weq (#F h · f) = h · hom_weq f}
+          {natural_postcomp : ∏ (A : C) (B : D) (f : F A --> B) (Y : D) (k : B --> Y),
+                              hom_weq (f · k) = hom_weq f · #G k}.
+
+  Definition hom_inv {A : C} {B : D}: A --> G B → F A --> B
```
siddharthist: This can be a `Local Definition`.

PR #952, opened 2018-03-28
add weak equivalence between adjunctions and natural weqs of homsets

tamaravonglehn:
Suggested in comment to #608. (It only seems to work for categories not precategories though).

### DIFF
--- a/UniMath/CategoryTheory/Adjunctions.v
+++ b/UniMath/CategoryTheory/Adjunctions.v
@@ -634,15 +634,22 @@ End HomSetIso_from_Adjunction.
 
 Section Adjunction_from_HomSetIso.
 
-  Context {C D : precategory} {F : functor C D} {G : functor D C}
-          {hom_weq :  ∏ {A : C} {B : D}, F A --> B ≃ A --> G B}
-          {natural_precomp : ∏ (A : C) (B : D) (f : F A --> B) (X : C) (h : X --> A),
-                             hom_weq (#F h · f) = h · hom_weq f}
-          {natural_postcomp : ∏ (A : C) (B : D) (f : F A --> B) (Y : D) (k : B --> Y),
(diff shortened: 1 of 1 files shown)


UniMath/CategoryTheory/Adjunctions.v
```diff
@@ -700,4 +707,80 @@ Section Adjunction_from_HomSetIso.
...
+
+  Definition nathomweq_from_adj : (are_adjoints F G) → (natural_hom_weq F G)
+    := λ H, (adjunction_hom_weq H,, (φ_adj_natural_precomp H,, φ_adj_natural_postcomp H)).
+
+  Lemma adj_after_nathomweq (H : are_adjoints F G)
+    : adj_from_nathomweq (nathomweq_from_adj H) = H.
+  Proof.
+    apply subtypeEquality'.
+    - apply dirprod_paths; cbn.
+      + apply subtypeEquality'.
```
benediktahrens: Use `apply nat_trans_eq` here instead. Same in the other component.

UniMath/CategoryTheory/Adjunctions.v
```diff
@@ -700,4 +707,80 @@ Section Adjunction_from_HomSetIso.
...
+    apply subtypeEquality'.
+    - cbn.
+      unfold adjunction_hom_weq.
+      do 2 (apply funextsec; intro).
+      apply subtypeEquality'.
+      + cbn.
+        unfold φ_adj, adj_from_nathomweq. cbn.
+        apply funextsec.
+        intro f.
+        rewrite <- (pr2 (pr2 H)).
```
benediktahrens: Could you add access functions for `natural_hom_weq` and use those? Then the code would be easier to read. (The lack of access functions is of course an oversight of the person who first implemented `natural_hom_weq`. We will never know who that was.)

PR #953, opened 2018-04-04
Vectors, matrices, and magnitude

siddharthist:
Currently, `Sequence`s in `X` are sigma-types ` ∑ n, stn n -> X`. Here, I give a name to the type `stn n -> X`, namely `Vector X n`. A `Matrix` is a vector of vectors. 

I show that most algebraic properties of binary operations hold when said operations are extended pointwise to vectors and entrywise to matrices. I then define matrix multiplication, weighting and coweighting, and the magnitude of a matrix with entries in a semiring (as in [The magnitude of metric spaces](https://arxiv.org/abs/1012.5857)).

### DIFF
--- a/UniMath/Algebra/Matrix.v
+++ b/UniMath/Algebra/Matrix.v
@@ -311,10 +311,83 @@ Section MatrixMult.
     - exact (rigunel1). (* The additive identity *)
   Defined.
 
-  (*
-  Lemma matrix_mult_assoc {m n : nat} (mat1 : Matrix R m n)
-                          {p : nat} (mat2 : Matrix R n p)
-                          {q : nat} (mat3 : Matrix R p q) :
-    ((mat1 ** mat2) ** mat3) = (mat1 ** (mat2 ** mat3)).
-  *)
 End MatrixMult.
+
+Local Notation Σ := (iterop_fun rigunel1 op1).
+Local Notation "R1 ^ R2" := ((pointwise _ op2) R1 R2).
+Local Notation "A ** B" := (matrix_mult A B) (at level 80).
+
+
+(** The following is based on "The magnitude of metric spaces" by Tom Leinster
+    (arXiv:1012.5857v3). *)
+Section Weighting.
+
+  Context {R : rig}.
+
+  (** Definition 1.1.1 in arXiv:1012.5857v3 *)
+  Definition weighting {m n : nat} (mat : Matrix R m n) : UU :=
+    ∑ vec : Vector R (S n), (mat ** (col_vec vec)) = col_vec (const_vec (1%rig)).
+
+  Definition coweighting {m n : nat} (mat : Matrix R m n) : UU :=
+    ∑ vec : Vector R (S m), ((row_vec vec) ** mat) = row_vec (const_vec (1%rig)).
+
+  Lemma matrix_mult_vectors {n : nat} (vec1 vec2 : Vector R (S n)) :
+    ((row_vec vec1) ** (col_vec vec2)) = weq_matrix_1_1 (Σ (vec1 ^ vec2)).
+  Proof.
+    apply funextfun; intro i; apply funextfun; intro j; reflexivity.
+  Defined.
+
+  (** Multiplying a column vector by the identity row vector is the same as
+      taking the sum of its entries. *)
+  Local Lemma sum_entries1 {n : nat} (vec : Vector R (S n)) :
+    weq_matrix_1_1 (Σ vec) = ((row_vec (const_vec (1%rig))) ** (col_vec vec)).
+  Proof.
+    refine (_ @ !matrix_mult_vectors _ _).
+    do 2 apply maponpaths.
+    apply pathsinv0.
+    refine (pointwise_lunit 1%rig _ vec).
+    apply riglunax2.
+  Defined.
+
+  Local Lemma sum_entries2 {n : nat} (vec : Vector R (S n)) :
+      weq_matrix_1_1 (Σ vec) = (row_vec vec ** col_vec (const_vec 1%rig)).
+  Proof.
+    refine (_ @ !matrix_mult_vectors _ _).
+    do 2 apply maponpaths.
+    apply pathsinv0.
+    refine (pointwise_runit 1%rig _ vec).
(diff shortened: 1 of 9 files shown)


UniMath/Combinatorics/FiniteSequences.v
```diff
@@ -1,13 +1,217 @@
...
+    <<
+        <--- n --->
+      | [ * * * * ]
+      m [ * * * * ]
+      | [ * * * * ]
+    >>
+    Since [Vector]s are encoded as functions ⟦n⟧ → X, a matrix is a function (of
+    two arguments). Thus, the (i, j)-entry of a matrix Mat is simply Mat i j.
+
+    We don't allow for empty matrices.
```
DanGrayson: You should allow for empty matrices, m by n, with m and n any natural numbers, because
matrices are used to encode homomorphisms between free modules, and the rank of a free module
can be any natural number.

UniMath/Combinatorics/FiniteSequences.v
```diff
@@ -1,13 +1,217 @@
...
+  flip mat.
+
+Definition row {X : UU} {m n : nat} (mat : Matrix X m n) : ⟦ S m ⟧ → Vector X (S n) := mat.
+
+Definition col {X : UU} {m n : nat} (mat : Matrix X m n) : ⟦ S n ⟧ → Vector X (S m) := transpose mat.
+
+Definition row_vec {X : UU} {n : nat} (vec : Vector X (S n)) : Matrix X 0 n :=
+  λ i j, vec j.
+
+Definition col_vec {X : UU} {n : nat} (vec : Vector X (S n)) : Matrix X n 0 :=
```
DanGrayson: What's the point of col_vec and row_vec?

UniMath/Algebra/Matrix.v
```diff
@@ -0,0 +1,393 @@
...
+  Defined.
+
+  Definition pointwise_abmonoid (X : abmonoid) (n : nat) : abmonoid.
+  Proof.
+    use abmonoidpair.
+    - apply pointwise_setwithbinop; [exact X|assumption].
+    - apply pointwise_abmonoidop; exact (pr2 X).
+  Defined.
+
+  Definition pointwise_rig (X : rig) (n : nat) : rig.
```
DanGrayson: Vectors are normally not made into elements of a ring or rig, so this function is unmotivated.  Perhaps eliminate it.

UniMath/Algebra/Matrix.v
```diff
@@ -0,0 +1,393 @@
...
+    use mk_isabmonoidop.
+    - apply entrywise_monoidop; exact (pr1isabmonoidop _ _ abmonoidax).
+    - apply entrywise_comm; exact (pr2 abmonoidax).
+  Defined.
+
+End OneOpMat.
+
+(** *** Standard conditions on a pair of binary operations *)
+
+Section TwoOpsMat.
```
DanGrayson: Matrices are not normally given a pair of binary operations entrywise, so this section is unmotivated and could be deleted.  (Matrix multiplication is the important thing, not entrywise multiplication.)

UniMath/Combinatorics/FiniteSequences.v
siddharthist: They are just for the sake of clarity when reading code. Its helpful to be able to denote "we're considering this a row vector now". For an example of this usage, see Matrix.v

siddharthist:
Thanks for the review @DanGrayson! I believe I've addressed all your comments so far (except perhaps for the inclusion of `row_vec` and `col_vec`).

UniMath/Algebra/Matrix.v
benediktahrens: What is unmotivated or not, or what is "normally" done, is not up to a central authority to decide.

I think the removed code should be restored.

benediktahrens:
lgtm

UniMath/Algebra/Matrix.v
DanGrayson: Central authority?  This is a discussion.

DanGrayson: Why restore the removed code, unless one can expect to find it useful for something?

benediktahrens:
On 07/05/18 18:28, Daniel R. Grayson wrote:
> 
> 
> In UniMath/Algebra/Matrix.v 
> <https://github.com/UniMath/UniMath/pull/953#discussion_r186490326>:
> 
>> +
> +  Definition pointwise_monoid (X : monoid) (n : nat) : monoid.
> +  Proof.
> +    use monoidpair.
> +    - apply pointwise_setwithbinop; [exact X|assumption].
> +    - apply pointwise_monoidop; exact (pr2 X).
> +  Defined.
> +
> +  Definition pointwise_abmonoid (X : abmonoid) (n : nat) : abmonoid.
> +  Proof.
> +    use abmonoidpair.
> +    - apply pointwise_setwithbinop; [exact X|assumption].
> +    - apply pointwise_abmonoidop; exact (pr2 X).
> +  Defined.
> +
> +  Definition pointwise_rig (X : rig) (n : nat) : rig.
> 
> Why restore the removed code, unless one can expect to find it useful 
> for something?

What might look useless today might become useful tomorrow, without 
anyone expecting that today.

Also, whose expectations would be checked against?


I wouldn't suggest writing code that I don't expect to be useful, but 
the code under discussion has already been written, so it might as well 
remain.

siddharthist:
I don't have a strong opinion on the matter. I think you both have good points in that
 1. It is never clear what might be useful in the future and the effort is already spent, but
 2. Unnecessary bloat is undesirable, it takes time and energy to maintain every line of code. What if Coq 9.0 makes some syntax used these files obsolete, and it must be changed throughout?

If we leave this code in the commit history, it's at least feasible to restore it later. This might be a good compromise.

benediktahrens:
Error on Travis:
```
File "./UniMath/NumberSystems/NaturalNumbersAlgebra.v", line 29, characters 26-27:
Error:
Ltac call to "split with (bindings)" failed.
In environment
a : ⟦ ?n ⟧
The term "0" has type "nat" while it is expected to have type "⟦ ?n ⟧".
```

PR #955, opened 2018-04-08
Lemmas about equivalences of categories, fully faithful functors

siddharthist:
Results:
 * Composites of faithful functors are faithful
 * Composites of full functors are full
 * Composites of essentially sufjective functors are essentially surjective
 * Composites of full and faithful functors are full and faithful
 * Composites of fully faithful functors are fully faithful
 * Composites of the underlying left functors in two equivalences are full and faithful, so if the domain is univalent, they constitute an equivalence

Plus: 
 * Some small cleanup in proof scripts
 * Reorganization to put the definition of functor composition before the results on full, faithful, fully faithful functors
 * Improved documentation in each of the files used


siddharthist:
In the latest commit:
 * Define split essentially surjective functors
 * Show that for univalent domain, a fully faithful essentially surjective functor is actually split essentially surjective
* Define the "inverse on objects" of an essentially surjective functor, use this to simplify a proof

siddharthist:
In the latest commit: 
 * Define conservative functors
 * Show fully faithful functors are conservative
 * Define the forgetful functor HSET -> Type, show it's conservative

I thought I might be able to weaken the hypotheses of `isaprop_sigma_iso`, but didn't end up being able to. Still, I figured this is worth defining!

siddharthist:
Conflicts fixed!

UniMath/CategoryTheory/functor_categories.v
```diff
@@ -331,6 +315,98 @@ Proof.
   apply functor_id.
 Qed.
 
+(** ** Conservative functors *)
+
+(** The generic property of "reflecting" a property of a morphism. *)
+
+Definition reflects_morphism {C D : precategory} (F : functor C D)
```
benediktahrens: I suggest renaming into `reflects_morphism_property`, despite that being uncomfortably long. Reason: the current name sounds like morphisms are to be reflected, not their properties.

UniMath/CategoryTheory/equivalences_lemmas.v
```diff
@@ -106,4 +102,110 @@ Proof.
...
+
 End from_equiv_to_fully_faithful.
+
+(** ** Composition of equivalences *)
+
+(** There is probably a more general way to do this. We assume
+    that C is univalent, and use the fact that a fully faithful, essentially
+    surjective functor out of a univalent category is an equivalence. *)
+
+Lemma compose_equivalences_univalent {C : univalent_category} (D E : precategory)
```
benediktahrens: This construction should not require C to be univalent, as you point out in the comment.

PR #956, opened 2018-04-12
define primality for natural numbers

siddharthist:


UniMath/MoreFoundations/NaturalNumbers.v
```diff
@@ -0,0 +1,243 @@
...
+
+Require Import UniMath.Foundations.PartA.
+Require Import UniMath.Foundations.NaturalNumbers.
+Require Import UniMath.MoreFoundations.PartA.
+
+(** Strong induction (or complete induction) ([stronginduction]) *)
+
+Lemma stronginduction {P : nat -> UU} (p : P 0)
+      (q : ∏ n : nat, n ≠ 0 → (∏ m : nat, m < n → P m) → P n) :
+  ∏ n : nat, P n.
```
DanGrayson: Is there any good reason to have this lemma, when the next one, `stronginduction'`, is phrased so much more naturally?

UniMath/MoreFoundations/NaturalNumbers.v
```diff
@@ -0,0 +1,243 @@
...
+           ++ intros k s.
+              rewrite right in s.
+              apply (IHn k).
+              assumption.
+Defined.
+
+(** The following unifies the case of n = 0 with the rest. *)
+Lemma stronginduction'
+{P : nat -> UU} (q : ∏ n : nat, (∏ m : nat, m < n → P m) → P n) :
+  ∏ n : nat, P n.
```
DanGrayson: Since `q` is not used in the statement, why not phrase it this way?:
```
Lemma stronginduction' {P : nat -> UU} :
  ( ∏ n : nat, (∏ m : nat, m < n → P m) → P n) -> ∏ n : nat, P n.
```

UniMath/MoreFoundations/NaturalNumbers.v
```diff
@@ -0,0 +1,243 @@
...
+      refine (pr2 p @ _).
+      exact (!pr2 p').
+Defined.
+
+Notation "n | m" := (nat_divides n m) (at level 50) : nat.
+
+Local Open Scope nat.
+
+(** Note that due to the definition of [natdivmod], 0 divides everything. *)
+Lemma nat_divides_0 (n : nat) : 0 | n.
```
DanGrayson: This should be fixed: 0 should not divide everything.

UniMath/MoreFoundations/NaturalNumbers.v
```diff
@@ -0,0 +1,243 @@
...
+(** First definition: the remainder when dividing m by n is 0. *)
+Definition nat_divides (n m : nat) : hProp.
+Proof.
+  use hProppair.
+  - exact (natrem m n = 0).
+  - apply isasetnat.
+Defined.
+
+(** Alternative definition: there exists some p such that n * p = m. *)
+Definition nat_divides' (n m : nat) : hProp.
```
DanGrayson: What's the point of having two definitions?

UniMath/MoreFoundations/NaturalNumbers.v
```diff
@@ -0,0 +1,243 @@
...
+  intros g g'.
+  apply subtypeEquality'.
+  - apply isantisymmnatleh.
+    + apply nat_gcd_ge_divisors; [apply nat_gcd_div_n|apply nat_gcd_div_m].
+    + apply nat_gcd_ge_divisors; [apply nat_gcd_div_n|apply nat_gcd_div_m].
+  - apply propproperty.
+Defined.
+
+(** "The nat_gcd of n and m is g", i.e. every nat_gcd of n and m is equal to g. *)
+Definition nat_gcd_is (n m g : nat) : hProp.
```
DanGrayson: What's the point of this if you already have `is_nat_gcd`?

UniMath/MoreFoundations/NaturalNumbers.v
```diff
@@ -0,0 +1,243 @@
...
+  apply weqimplimpl.
+  - apply nat_gcd_sym_lemma.
+  - apply nat_gcd_sym_lemma.
+  - apply propproperty.
+  - apply propproperty.
+Defined.
+
+(** [nat_gcd] with 0 and 1. *)
+
+Lemma nat_gcd_0 : ∏ (n : nat), nat_gcd_is n 0 1.
```
DanGrayson: The gcd of n and 0 should be n, not 1.

UniMath/MoreFoundations/NaturalNumbers.v
```diff
@@ -0,0 +1,243 @@
...
+Lemma nat_gcd_1 : ∏ (n : nat), nat_gcd_is n 1 1.
+Abort.
+
+(** *** Relative primality ([nat_rel_prime]) *)
+
+(** Two numbers are relatively prime if their [nat_gcd] is 1. *)
+Definition nat_rel_prime (n : nat) (m : nat) : hProp.
+Proof.
+  use hProppair.
+  - exact (∏ g : nat_gcd n m, nat_gcd_num g = 1).
```
DanGrayson: Why isn't this line `exact (is_nat_gcd n m 1)`?

UniMath/MoreFoundations/NaturalNumbers.v
```diff
@@ -0,0 +1,243 @@
...
+
+(** Relative primality is symmetric *)
+
+(** *** Primality ([nat_prime]) *)
+
+(** A number is prime if it is relatively prime with every number less than it. *)
+Definition nat_prime (n : nat) : hProp.
+Proof.
+  use hProppair.
+  - exact (1 < n × (∏ m, m < n -> nat_rel_prime n m)).
```
DanGrayson: This is not the standard definition.  The standard and preferred definition is that 1 < n and m | n -> m = 1 or m = n.  For "or" here, you can use either the sum or the disjunction, since m=1 and m=n don't both occur.  Using the sum would be simpler.

siddharthist:
This was done a bit on a whim, I'll definitely incorporate this feedback should I ever update this pull request.

PR #958, opened 2018-04-12
Define rig ideals

siddharthist:
* Ideals in rigs
    - Ideals from kernels
    - Right and left ideals are equivalent for commutative rings
 * Cosets in monoids and groups (eventually, to get to quotients by normal subgroups and then quotients by ideals)
    - Definitions
    - Proof that being in the same coset is an equivalence relation on a group
 * Move results about natural numbers to`NumberSystems.NaturalNumbersAlgebra`

### DIFF
--- a/UniMath/Algebra/Domains_and_Fields.v
+++ b/UniMath/Algebra/Domains_and_Fields.v
@@ -29,7 +29,7 @@ Unset Kernel Term Sharing.
 
 (** Imports *)
 
-Require Export UniMath.Algebra.Rigs_and_Rings.
+Require Export UniMath.Algebra.RigsAndRings.
 
 (** To upstream files *)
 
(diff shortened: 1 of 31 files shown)


UniMath/Algebra/Domains_and_Fields.v
```diff
@@ -269,15 +269,7 @@ Local Open Scope ring_scope.
...
-
-Lemma isapropisnonzeroring (X : ring) : isaprop (isnonzeroring X).
-Proof.
-  intros X. use isapropneg.
-Defined.
-Opaque isapropisnonzeroring.
-
-Lemma isnonzerolinvel (X : ring) (is : isnonzeroring X) (x : X) (x' : multlinvpair X x) :
+Lemma isnonzerolinvel (X : ring) (is : isnonzerorig X) (x : X) (x' : multlinvpair X x) :
   neg ((pr1 x') = 0).
```
DanGrayson: How about writing `pr1 x' != 0` here?

siddharthist: Good idea, will do.

siddharthist:
@DanGrayson I've changed that instance and a few others in that file.

PR #960, opened 2018-04-15
Fix build when BUILD_COQ=no and undefined COQBIN

SkySkimmer:
With COQBIN set, `$(COQBIN)coq_makefile` is an actual file but with
empty COQBIN it's intended to be resolved through PATH.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -170,7 +170,13 @@ describe:; git describe --dirty --long --always --abbrev=40 --all
 	echo '# End:' ;\
 	) >$@
 # the '' above prevents emacs from mistaking the lines above as providing local variables when visiting this file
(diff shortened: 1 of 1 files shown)


DanGrayson:
Good catch.  But it could be a 1-line patch, with this new line:
```
build/CoqMakefile.make .coq_makefile_output.conf: $(shell command -v $(COQBIN)coq_makefile) .coq_makefile_input
```

SkySkimmer:
Changed.

Gaëtan Gilbert

DanGrayson:
Thanks!

SkySkimmer:
Reverted as that didn't work when BUILD_COQ=yes: that sets COQBIN but coq isn't built yet so $(COQBIN)coq_makefile doesn't exist and command fails on it.

PR #962, opened 2018-04-16
Characterization of paths for categories

siddharthist:
While in general, `is_precategory` is not a proposition, it is for categories. We can use this to show that every path between categories arises from one between their data.


siddharthist:
I'm making good progress on a (more significant) lemma for `precategory_ob_mor`, namely
```coq
  Lemma precategory_ob_mor {A B : precategory_ob_mor} :
    A = B ≃ ∑ w1 : pr1 A ≃ pr1 B, ∏ a1 a2 : A, A ⟦ a1, a2 ⟧ ≃ B ⟦ w1 a1, w1 a2 ⟧.
```
It may be good to hold off on merging this until that one is done, then we can put them in their own file.

PR #963, opened 2018-04-17
Univalence for categories (partial)

siddharthist:
This gives progress towards and an outline of univalence for categories. I realized that it will take significantly more effort than I can put in right now, but I think that what's here is useful on its own as well.


DanGrayson:
I think you'll have difficulty later on, because the maps underlying the equivalences don't seem to be the expected ones.  For example, try this:
```
Definition precategory_ob_mor_paths_map {A B : precategory_ob_mor} :
  A = B -> ∑ w : pr1 A ≃ pr1 B, ∏ a1 a2 : A, A ⟦ a1, a2 ⟧ ≃ B ⟦ w a1, w a2 ⟧.
Proof.
  intro e.
  induction e.
  exists (idweq _).
  simpl.
  intros.
  exact (idweq _).
Defined.

Lemma precategory_ob_mor_paths_weq_compute {A B : precategory_ob_mor} :
  pr1weq (@precategory_ob_mor_paths_weq A B) = @precategory_ob_mor_paths_map A B.
Proof.
  Fail reflexivity.
```

UniMath/MoreFoundations/Univalence.v
```diff
@@ -15,3 +15,24 @@ Lemma toforallpaths_funextsec {T : UU} {P : T -> UU} {f g : ∏ t : T, P t} :
 Proof.
   intro h; exact (homotweqinvweq (weqtoforallpaths _ _ _) h).
 Defined.
+
+(** Paths in a fibration over the universe are equivalent to weak equivalences
+    such that, when transporting the second projections over the equality
+    given by univalence, they're equal.
+
+    Compare to [total2_paths_equiv]. *)
+Lemma total2_paths_UU_equiv {B : UU -> UU} (s s' : ∑ x, B x) :
```
DanGrayson: This should be a Definition, not a Lemma, since it is proof relevant.

UniMath/MoreFoundations/Univalence.v
```diff
@@ -15,3 +15,24 @@ Lemma toforallpaths_funextsec {T : UU} {P : T -> UU} {f g : ∏ t : T, P t} :
...
+    s = s' ≃
+    (∑ p : pr1 s ≃ pr1 s', transportf (λ x, B x) (weqtopaths p) (pr2 s) = pr2 s').
```
DanGrayson: It might be better to phrase this as a statement that some obvious map is an equivalence.

siddharthist:
Closing until I have time to update it.

PR #965, opened 2018-04-18
Compatibility with Coq PR #7257 fixing the result of a "random" unification heuristic.

herbelin:
I'm very sorry, but a bug-fix PR, number coq/coq#7257, is causing an incompatibility in UniMath. Indeed, we discovered that some unification heuristic was sensitive to the alphabetic order and a PR has been proposed to use a more stable heuristic. In UniMath's file `Algebra/Modules/Core.v`, there is indeed an example of a unification problem which is sensitive to the name of some variable. The problem works because a variable is called `y` but it would not have worked if the variable had been called `v` because it involves another variable `x`, and `v`<`x`<`y` alphabetically :(((

We fix the file by giving a more refined pattern to help unification.


siddharthist:
@herbelin Thank you for contributing a fix! It's great that the Coq developers think so much about their users :smile:

PR #970, opened 2018-04-20
replace the deprecated elements (as of Coq 8.8.0)

rmatthes:
It concerns Import in sections and Focus.

This in particular makes proofs cleaner where goals were spawned within the focused proof development.

### DIFF
--- a/UniMath/CategoryTheory/DisplayedCats/Equivalences.v
+++ b/UniMath/CategoryTheory/DisplayedCats/Equivalences.v
@@ -59,7 +59,7 @@ Proof.
     { apply eq_iso. cbn. simpl. unfold precomp_with.
       etrans. apply maponpaths_2. apply id_right.
       etrans. eapply pathsinv0. apply functor_comp.
-      etrans. Focus 2. apply functor_id.
+      etrans. 2: apply functor_id.
       apply maponpaths. apply iso_after_iso_inv.
    }
     set (XRT := transportf (λ r, iso_disp r (FF x dd) yy )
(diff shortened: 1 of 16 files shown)


UniMath/CategoryTheory/DisplayedCats/Equivalences.v
```diff
@@ -591,46 +591,48 @@ Proof.
...
-          apply transport_f_f.
-        etrans. apply maponpaths_2, transport_f_f.
-        apply mor_disp_transportf_postwhisker.
-      etrans. apply transport_f_f.
-      etrans. apply transport_f_f.
-      etrans. apply transport_f_f.
-      etrans. apply transport_f_f.
-      etrans. apply transport_f_f.
-      (* A trick to hide the huge equality term: *)
-      apply maponpaths_2. shelve.
```
benediktahrens: Some formatting has been lost in the conversion here. Was that formatting helpful?

rmatthes: I had the feeling that ProofGeneral thought this now to be the most appropriate indentation. I'll check that again.

benediktahrens:
rebased version available as #973 .

UniMath/CategoryTheory/DisplayedCats/Equivalences.v
benediktahrens: The problem here is that this part of the proof does not use bullets, which would usually guide ProofGeneral's indentation. You hence cannot rely on PG doing the right thing here.

rmatthes: this is now hopefully completely solved in PR #974 - I apologise for this careless re-indentation (of course, I ought to have seen that PG did not do what was intended)

Issue #940, opened 2018-03-12
Reorganize the two developments on bicategories


benediktahrens:
following @peterlefanulumsdaine 's suggestion in https://github.com/UniMath/UniMath/pull/925#issuecomment-372103581

siddharthist:
Was this solved by #971? What remains to do?

PR #971, opened 2018-04-23
Reorganization of Bicategories (Issue #940)

maggesi:


benediktahrens:
Merging, since it is the suggestion already discussed and linked to from #940

PR #973, opened 2018-04-23
replace the deprecated (as of Coq 8.8.0) uses of Import in sections a…

benediktahrens:
…nd Focus

supersedes #970 

@rmatthes : this is a rebased version of your PR.

PR #974, opened 2018-04-23
replaces the deprecated elements of Coq 8.8.0

rmatthes:
supersedes #973 (taking it as the basis) in order to solve a problem of #970


DanGrayson:
Should the other two be closed?

PR #975, opened 2018-04-27
Groupoids and discrete categories

siddharthist:
This unifies a bunch of stuff about groupoids and discrete categories that's scattered all over UniMath.


UniMath/CategoryTheory/Categories.v
```diff
@@ -1566,3 +1567,59 @@ Lemma transport_source_is_iso {C : precategory} {x y z : ob C} (f : x --> y) (H
...
+    <<
+      X₀ → X₁ ← X₂ → ⋯ → Xₙ₋₁ ← Xₙ
+    >>
+
+    Note that [dni] is the (order-preserving) inclusion [stn n → stn (S n)] that
+    excludes its first argument from the range. Thus, [dni firstelement]
+    adds 1 to the first projection of its argument, whereas [dni lastelement]
+    doesn't.
+ *)
+Definition is_connected (C : precategory) : UU :=
```
benediktahrens: This is structure, not property. I hence suggest to rename the definition into "connection" or something similar.

UniMath/CategoryTheory/Categories.v
```diff
@@ -1566,3 +1567,59 @@ Lemma transport_source_is_iso {C : precategory} {x y z : ob C} (f : x --> y) (H
...
+  ∏ a b : ob C,
+    ∑ (n : nat) (seq : stn (S n) → (ob C)),
+      (a = seq firstelement) ×
+      (b = seq lastelement) ×
+      ∏ i : stn n,
+        (C⟦seq (dni lastelement i), seq (dni firstelement i)⟧) ⨿
+        (C⟦seq (dni firstelement i), seq (dni lastelement i)⟧).
+
+(** Note: this is non-standard terminology. *)
+Definition is_path_connected (C : precategory) : UU := ∏ a b : ob C, C ⟦a, b⟧.
```
benediktahrens: This is structure, not property, and the name should reflect that. Better(?): path_connection

UniMath/CategoryTheory/Groupoids.v
```diff
@@ -0,0 +1,244 @@
...
+Definition univalent_groupoid_is_pregroupoid :
+  ∏ ugpd : univalent_groupoid, is_pregroupoid (pr1 ugpd) := pr2.
+Definition univalent_groupoid_to_groupoid :
+  univalent_groupoid -> groupoid :=
+  λ ugpd, mk_groupoid ugpd (univalent_groupoid_is_pregroupoid ugpd).
+Coercion univalent_groupoid_to_groupoid :
+  univalent_groupoid >-> groupoid.
+
+(** An alternative characterization of univalence for groupoids *)
+Definition is_univalent_pregroupoid (pgpd : pregroupoid) :=
```
benediktahrens: The comment suggests an equivalence of definitions, but I have only seen one implication. Is the other somewhere?
If not, could you add a comment saying that one implication is missing?

UniMath/CategoryTheory/Categories.v
DanGrayson: Alternatively, consider truncating it to a proposition.  Similarly for `is_path_connected`.

siddharthist: I've removed it for now. While playing around with a few other proofs in the area, I reworked the definition a few times. So I'm not ultimately convinced that this is ready for the prime time yet. Perhaps soon!

UniMath/CategoryTheory/Categories.v
DanGrayson: Another thing that makes this choice of name not so good is that the condition coded here has nothing to do with being connected by a path (a concept which already exists), so another name should be chosen.

This function probably has no anticipated use.

UniMath/CategoryTheory/Groupoids.v
siddharthist: Done! I kept this in because it was present in `StandardCategories`, but wasn't able to prove the reverse implication as easily.

UniMath/CategoryTheory/Groupoids.v
```diff
@@ -0,0 +1,246 @@
...
+
+  Author: Langston Barrett (@siddharthist), March 2018 *)
+
+(** ** Contents
+
+  - Definitions
+    - Pregroupoid
+    - Groupoid
+    - Univalent groupoid
+  - An alternative characterization of univalence for groupouds
```
benediktahrens: groupouds -> groupoids

siddharthist:
Fixed!

PR #976, opened 2018-04-27
Initial implementation of relative modules (modules over relative mon…

maggesi:
…ads).

### DIFF
--- a/UniMath/CategoryTheory/Monads/RelativeModules.v
+++ b/UniMath/CategoryTheory/Monads/RelativeModules.v
@@ -1,0 +1,358 @@
+(* =================================================================================== *)
+(** * Modules over relative monads                                                     *)
+(* =================================================================================== *)
+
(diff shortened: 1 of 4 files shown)


UniMath/CategoryTheory/Monads/RelativeModules.v
```diff
@@ -0,0 +1,358 @@
+(* =================================================================================== *)
+(** * Modules over relative monads                                                     *)
+(* =================================================================================== *)
+
+(* ----------------------------------------------------------------------------------- *)
+(* Contents:                                                                           *)
```
siddharthist: This should probably be a block cmment starting with `(**` so that it shows up in the HTML documentation.

maggesi: Thanks for reporting.  Fixed in 4b363e93b6bfe43784451b9e39790d582633510e.

PR #982, opened 2018-05-17
add test to travis script ...

### LINKED ISSUE #930: testing with modern coq
See issue #921 - Coq is tested against UniMath, so
our travis testing should ensure that UniMath builds also with the forthcoming version
of Coq, as well as with the past one.  For example, from coq 8.7 to 8.8 there is a change 
in the way the coq_makefile program works, and thus a change to our Makefile could break
the building.

DanGrayson:
that builds a part of UniMath with the master branch of Coq.

Resolves #930


benediktahrens:
Travis does not seem to run on this PR?

.travis.yml
```diff
@@ -62,6 +63,14 @@ script:
...
        fi &&
        . /root/.opam/opam-init/init.sh &&
+       git submodule update --init sub/coq &&
+       if [ $COQ_BRANCH != "" ]
+       then ( cd sub/coq && 
+       	      git checkout $COQ_BRANCH )
+       fi &&
+       ( cd sub/coq && 
+         git branch &&
+	 git describe --dirty --long --always --abbrev=40 --tags ) &&
```
favonia: The indentation is off at this line.

favonia: Well, tabs and spaces have different semantics. This is equivalent to claiming randomly replacing $0$ by $\emptyset$ is fine because they will probably be the same set in ZFC.

DanGrayson: TAB and SPACE have equivalent semantics in the vast majority of languages.  The only ones I can think of where that's not true are TeX and make.  Anyway, I've removed the tabs, so we'll see if that's a problem here.

DanGrayson:
With the TABs removed, the build is now running.  Thanks, Favonia!

DanGrayson:
It's a well known bug in travis: https://github.com/travis-ci/travis-ci/issues/4947

DanGrayson:
That's good, it detected a real problem:
```
cd sub/coq && ./configure -coqide "no" -with-doc no -annotate -local
-annotate has been removed. Please use -annot or -bin-annot instead.
Configuration script failed!
Makefile:201: recipe for target 'sub/coq/config/coq_config.ml' failed
make: *** [sub/coq/config/coq_config.ml] Error 1
```

The coq people haven't complained, so the coq travis script must operate by first building coq and then building UniMath with the instruction not to build coq -- the error above is in the way UniMath builds coq.

DanGrayson:
It seems okay to me now.

DanGrayson:
This one is ready now.

PR #990, opened 2018-07-13
Kleisli category of a monad

bcldoherty:
Contents: 
 - Definition of the Kleisli category of a monad, using the traditional definition of a monad as an endofunctor with a unit and multiplication.
 - The canonical adjunction between a category C and the Kleisli category of a monad on C.
 - A couple of lemmas involving the bind operation for a monad, similar to those in the Monads file.
 
One difficulty that came up in writing this code was that, because the composition of morphisms in the Kleisli category is defined in terms of the composition in the original category, the notation appearing in goals is sometimes confusing, as the same symbol · is used for both - so one might see an expression like f · g · h in some goal, where the two · symbols refer to compositon in different categories, and it's not immediately obvious which is which. Furthermore, using "unfold compose" in these situations unfolds composition in both the original category and the Kleisli category. It is often possible to work around this by using rewrite with the lemma Kleisli_compose_def which I have defined here, which simply restates the definition of composition in the Kleisli category, although attempting to do this often leads to an error message if the goal term has been simplified using simpl. For an example of these kinds of issues, see the definition of Kleisli_homset_iso in the code I've submitted here.

Does anyone know if there is any way I could define a special notation for composition in a Kleisli category in order to make the goal terms easier to read, and/or a more natural way to unfold composition in one category without unfolding all compositions appearing in the goal?


siddharthist:
@bcldoherty It might be helpful to read about notations (and especially scoping): https://coq.inria.fr/refman/user-extensions/syntax-extensions.html

bcldoherty:
Following a suggestion by @maggesi on the UniMath mailing list, I've removed the awkward workaround Kleisli_compose_def and rewritten some of the proofs, using "unfold" at specific positions to unfold composition only in the Kleisli category, e.g. "unfold compose at 2" when the second occurrence of composition in the goal is in the Kleisli category.

@siddharthist, thanks for the reference on notations and scoping; I've read through it and experimented a bit, but I still can't seem to define a separate notation for composition in a Kleisli category. It may not be a big deal if it can't be done, since it's possible to determine whether a given composition is in the Kleisli category or the original category by comparing the domains and codomains of the morphisms (e.g., if a morphism from a to T b is being composed with a morphism from b to T c, that must be in the Kleisli category of T).

UniMath/CategoryTheory/Monads/KleisliCategory.v
```diff
@@ -127,8 +118,8 @@ Proof.
     unfold Left_Kleisli_functor_data; simpl.
     apply id_left.
   - intros a b c f g.
-    rewrite Kleisli_compose_def.
     unfold Left_Kleisli_functor_data; simpl.
+    unfold compose at 3; simpl.
```
maggesi: In most cases, you may want to use "cbn" instead of "simpl".
There are little practical differences between the two from the user viewpoint.
However, the (new) tactic "cbn" is supposed to be better engineered and more predictable than the (old) "simpl".
Also, it seems that occasionally simpl generate proof terms that require more time to be typechecked by the kernel wrt cbn.

bcldoherty: Thanks for the tip. I've now replaced all instances of simpl in the file with cbn, which has made the code a bit simpler in places (cbn seems to unfold composition in the Kleisli category automatically, removing the need to do this manually). I'll upload the new version of the file shortly.

PR #993, opened 2018-07-30
Work on Distributive Laws, Liftings and Adjoint Folds

Marmann:
2018 Internship in Toulouse under the supervision of Ralph Matthes.
Work on the article "Unifying structured recursion schemes" by HINZE and WU.


UniMath/CategoryTheory/DistributiveLaws.v
```diff
@@ -0,0 +1,667 @@
...
+Variables C C' D D' : precategory.
+Variable F : functor C D.
+Variable F' : functor C' D'.
+Variable H : functor C' C.
+Variable K : functor D' D.
+*)
+
+   Definition DistrLaw {C C' D D' : precategory} (F : functor C D) (F' :
+functor C' D') (H : functor C' C) (K : functor D' D) : UU := nat_trans
+(functor_composite H F) (functor_composite F' K).
```
DanGrayson: In what sense is this a "distributive" law?

UniMath/CategoryTheory/DistributiveLaws.v
```diff
@@ -0,0 +1,667 @@
...
+
+Definition DistrLaw_data_type {C C' D D' : precategory} (F : functor C
+D) (F' : functor C' D') (H : functor C' C) (K : functor D' D) : UU :=  ∏
+x : ob C', (functor_composite H F) x --> (functor_composite F' K) x.
```
DanGrayson: This definition seems superfluous, in light of the one just above, which includes something equivalent.

UniMath/CategoryTheory/DistributiveLaws.v
```diff
@@ -0,0 +1,667 @@
...
+
+
+Definition is_distr_law {C C' D D' : precategory} (F : functor C D) (F'
+: functor C' D') (H : functor C' C) (K : functor D' D) (t :
+DistrLaw_data_type F F' H K) := is_nat_trans _ _ t.
```
DanGrayson: This one also seems superfluous.

UniMath/CategoryTheory/DistributiveLaws.v
```diff
@@ -0,0 +1,667 @@
...
+
+Print DistrLaw.
+
+Section OperationsDistrLaws.
+
+   Definition comp_distr_laws {C C' C'' D D' D'' : precategory} {F :
+functor C D} {F' : functor C' D'}  {F'' : functor C'' D''} {H : functor
+C' C} {H' : functor C'' C'} {K : functor D' D} {K' : functor D'' D'}
+(lambda : DistrLaw F F' H K) (lambda' : DistrLaw F' F'' H' K') :
+DistrLaw F F'' (functor_composite H' H ) (functor_composite K' K).
```
DanGrayson: Isn't there a unicode symbol for `functor_composite`?

UniMath/CategoryTheory/DistributiveLaws.v
```diff
@@ -0,0 +1,667 @@
...
+     use (nat_trans_comp _ _ _ _ (α_functor _ _ _)).
+     apply (nat_trans_comp _ _ _ (pre_whisker H' lambda)).
+     apply (nat_trans_comp _ _ _ (α_functor_inv _ _ _)).
+     exact (post_whisker lambda' K).
+     Defined.
+
+
+
+   Definition id_distr_law  {C D : precategory} (F : functor C D) :
+DistrLaw F F (functor_identity C) (functor_identity D).
```
DanGrayson: This line should be indented more than the one before.

UniMath/CategoryTheory/DistributiveLaws.v
```diff
@@ -0,0 +1,667 @@
...
+(lambda : DistrLaw F F' H K) (lambda' : DistrLaw F' F'' H' K') :
+DistrLaw F F'' (functor_composite H' H ) (functor_composite K' K).
+   Proof.
+     red.
+     apply (nat_trans_comp _ _ _ (α_functor _ _ _)).
+     use (nat_trans_comp _ _ _ _ (α_functor _ _ _)).
+     apply (nat_trans_comp _ _ _ (pre_whisker H' lambda)).
+     apply (nat_trans_comp _ _ _ (α_functor_inv _ _ _)).
+     exact (post_whisker lambda' K).
+     Defined.
```
DanGrayson: This line should be indented the same as the "Proof" line above.

UniMath/CategoryTheory/DistributiveLaws.v
```diff
@@ -0,0 +1,667 @@
...
+    apply (nat_trans_eq hs).
+    intro c.
+    simpl.
+    repeat rewrite id_left.
+    apply id_right.
+  Qed.
+
+  Locate "·".
+
+    Definition id_distr_law_right {C' D' : precategory} (F' : functor C' D') :
```
DanGrayson: The indentation here also needs fixing.  Perhaps there are other places, too.

UniMath/CategoryTheory/DistributiveLaws.v
```diff
@@ -0,0 +1,667 @@
...
+
+
+Definition is_lifting {C D: precategory} (hsC: has_homsets C) (hsD: has_homsets D) {F: functor C C} {G: functor D D} (H: functor C D) (HH: functor (FunctorAlg F hsC) (FunctorAlg G hsD)): UU.
+Proof.
+  set (UF := forget_algebras hsC F).
+  set (UG := forget_algebras hsD G).
+  exact (UF ∙ H = HH ∙ UG).
+Defined.
+
+Print FunctorAlg.
```
DanGrayson: Don't include any print statements, it makes for noisy compilation.

UniMath/CategoryTheory/DistributiveLaws.v
```diff
@@ -0,0 +1,667 @@
...
+  - exact (φ_adj h x).
+  - red.
+    apply pathsinv0.
+    apply Hyp.
+Defined.
+
+Lemma φ_adj_h_x_equals_initial_arrow (x : C ⟦ L μDD, B ⟧) : (# DD (φ_adj h x) · alg_map DD
+                                       ((lifting_from_distr_law hsC hsD τ)
+                                          (AlgConstr' B b)) = inDD · φ_adj h x) ->
+
```
DanGrayson: This blank line shouldn't be here.

UniMath/CategoryTheory/DistributiveLaws.v
```diff
@@ -0,0 +1,667 @@
...
+  simpl in aux'.
+  change (φ_adj h x) with (mor_from_algebra_mor _ _ _ (φ_adj_h_x x Hyp)).
+  rewrite aux.
+  apply idpath.
+Qed.
+
+
+Theorem TheoremOfHinzeAndWu : iscontr (∑ x : L μDD --> B, #L inDD · x = nat_trans_data σ μDD · # CC x · b).
+Proof.
+red.
```
DanGrayson: Indent bodies of proofs.

UniMath/CategoryTheory/DistributiveLaws.v
```diff
@@ -0,0 +1,667 @@
+(** Author : Hichem Saghrouni
+Internship at IRIT, 2018
+Under the supervision of Ralph Matthes *)
```
DanGrayson: If this file is about the paper of Hinze and Wu, give a reference and state the purpose of the file.

UniMath/CategoryTheory/DistributiveLaws.v
rmatthes: In fact, Hinze and Wu observed that they do not need any extra requirement. I agree that this should be mentioned in the comments.

UniMath/CategoryTheory/DistributiveLaws.v
rmatthes: This definition helped in the development later. Of course, it is just an instance of the non-existent nat_trans_data that should be analogous to functor_data. The existing nat_trans_data is only analogous to functor_data_from_functor. I would rather discuss this issue separately form the present PR.

UniMath/CategoryTheory/DistributiveLaws.v
rmatthes: The identifier functor_composite currently appears more than 400 times in the library. But, yes, one could simple replace it by the symbol typed with first backslash then period.

rmatthes:
The PR should be fit for pulling - none of the previous criticism should be applicable to the current state.

DanGrayson:
Okay, it seems the term "distributive law" is unrelated to the traditional use of term.  If it's possible to come up with a better name for the file "DistributiveLaws.v", that would be good.  What about "RecursionSchemes.v"?

In "which are plainly natural transformations" the word "plainly" doesn't sound right.  I suggest removing it.

In "distributive laws bound by certain identities" it's hard to guess what "bound by" means.  Perhaps you mean "distributive laws satisfying certain identities".

Re "use these notions to prove the Theorem of Hinze and Wu"  -- strictly speaking, they proved it, and you are formalizing their proof, I guess.

benediktahrens:
On 31/07/18 16:06, Daniel R. Grayson wrote:
> Okay, it seems the term "distributive law" is unrelated to the 
> traditional use of term. If it's possible to come up with a better name 
> for the file "DistributiveLaws.v", that would be good. What about 
> "RecursionSchemes.v"?
> 

The term "distributive law" is standard in category theory, see, e.g.,

https://golem.ph.utexas.edu/category/2017/02/distributive_laws.html
or
https://arxiv.org/abs/1503.02447

The definition here seems to be more general than the ones in those 
references.

rmatthes:
Distributive laws w.r.t. monads might get a name that points to monads. Such a development would naturally be contained within the Monads subdirectory.

DanGrayson:
Re: "The term "distributive law" is standard in category theory"

That may be, but here it's being used differently.  There it was a natural transformation ST->TS, where S and T were monads, so at least the idea of distributing one operation over another was preserved.  Here it's lost almost all connection with the original idea.

DanGrayson:
Re: "Would "use these notions to reprove the Theorem of Hinze and Wu formally" be okay?"

That would be okay if the proof formalized is a new proof, in which case, it would be good to advertise the respect in which the new proof is novel, so you get credit.

rmatthes:
The formalized proof is the formalization of the published proof - with all the subtleties that UniMath requires. Mathematically, there is nothing new. Of course, already the statement is not that of Hinze and Wu: who would write in JFP that a certain Sigma-type is contractible? My question is now: what makes "reprove ... formally" a claim for originality?

rmatthes:
Re: "The term "distributive law" is standard in category theory"

Hinze and Wu introduce functor squares and call the diagonal there distributive law - "for brevity". While some lemmas are stated with four arbitrary functors, the situation gets more specific for the theorem. There, the first two functor arguments of σ are the same, and likewise for the last two functor arguments of τ. Thus, there is still the idea of getting a functor from one side of the composition to the other side - the L resp. R mediate between the Control functor and the Data functor.

I see rather an achievement when someone observes that some results do not need the strict format in which distributive laws have been studied, while being led by the intuition gained from them.

DanGrayson:
Re: "My question is now: what makes "reprove ... formally" a claim for originality?"

To my mind, the word "reprove" means "prove again", i.e., to prove again in a different way.

PR #994, opened 2018-08-07
one lemma of Hinze & Wu

rmatthes:
The task looked simple but took more effort than expected since it seems to be rather difficult to argue about equality of functor data. The UniMath library has only very few examples where the object transformations are not already definitionally equal. I provide a lemma of more general interest on that question. 

Some other small changes stem from identifying `double_transport` in some constructions.

### DIFF
--- a/UniMath/CategoryTheory/Categories.v
+++ b/UniMath/CategoryTheory/Categories.v
@@ -1220,12 +1220,6 @@ Definition double_transport {C : precategory_ob_mor} {a a' b b' : ob C}
    (p : a = a') (q : b = b') (f : a --> b) : a' --> b' :=
   transportf (λ c, a' --> c) q (transportf (λ c, c --> b) p f).
 
-(*
-Definition double_transport_perm {C : precategory_ob_mor} {a a' b b' : ob C}
-   (p : a = a') (q : b = b') (f : a --> b) : a' --> b' :=
-   transportf (λ c, c --> b') p (transportf (λ c, a --> c) q f).
- *)
-
 Lemma idtoiso_postcompose (C : precategory) (a b b' : ob C)
   (p : b = b') (f : a --> b) :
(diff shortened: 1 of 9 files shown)


UniMath/CategoryTheory/Categories.v
```diff
@@ -1216,10 +1216,16 @@ Qed.
...
-Definition double_transport {C : precategory} {a a' b b' : ob C}
+Definition double_transport {C : precategory_ob_mor} {a a' b b' : ob C}
    (p : a = a') (q : b = b') (f : a --> b) : a' --> b' :=
   transportf (λ c, a' --> c) q (transportf (λ c, c --> b) p f).
 
+(*
+Definition double_transport_perm {C : precategory_ob_mor} {a a' b b' : ob C}
+   (p : a = a') (q : b = b') (f : a --> b) : a' --> b' :=
+   transportf (λ c, c --> b') p (transportf (λ c, a --> c) q f).
+ *)
```
benediktahrens: Do you consider it better to keep this commented items or to delete it?

UniMath/CategoryTheory/Categories.v
```diff
@@ -1265,6 +1271,19 @@ Proof.
...
+  (p : a = a') (q : b = b')  (f : a --> b) :
+  double_transport_perm p q f = inv_from_iso (idtoiso p) · f · idtoiso q.
+Proof.
+  destruct p.
+  destruct q.
+  intermediate_path (identity _ · f).
+  - apply pathsinv0; apply id_left.
+  - apply pathsinv0; apply id_right.
+Defined.
+*)
```
benediktahrens: Do you consider it better to keep this commented items or to delete it?

rmatthes:
I did a notational change in `functor_data_eq` and in fact removed `double_transport_perm`. This PR should be considered as terminated.

Issue #543, opened 2017-01-14
revisit some proofs in CategoryTheory


benediktahrens:
Some proofs in CT were written by people who clearly did not know their stuff (me). Consequently, these proofs are long, roundabout, cumbersome. Somebody should write better proofs. Specifically, I am having in mind:

- Precomposition with an essentially surjective functor is essentially surjective (https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/precomp_ess_surj.v). Here it should be possible to get rid of the hypothesis of full faithfulness.
- Precomposition with a fully faithful functor is fully faithful (https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/precomp_fully_faithful.v).
- A functor category is univalent if the target category is (https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/functor_categories.v). The equivalence should probably be constructed as a composition of several equivalences, instead of via the grad theorem.

The current proofs are good in the sense that they survived upgrades of Coq without any significant changes. It is my hope that this property would be maintained.

arnoudvanderleer:
@benediktahrens What is the status of this issue?
Are you sure that precomposition is essentially surjective if the functor is not fully faithful?

PR #996, opened 2018-08-13
Remove hypothesis of faithfulness; simplify

tamaravonglehn:
A first attempt at #543 .

### DIFF
--- a/UniMath/CategoryTheory/precomp_fully_faithful.v
+++ b/UniMath/CategoryTheory/precomp_fully_faithful.v
@@ -12,7 +12,7 @@ january 2013
 
 Contents :
 
-Precomposition with a fully faithful and
+Precomposition with a full and
            essentially surjective functor yields
            a full and faithful, i.e. a fully faithful,
            functor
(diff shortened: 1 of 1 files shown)


UniMath/CategoryTheory/precomp_fully_faithful.v
```diff
@@ -96,280 +86,150 @@ Section full.
...
 Lemma is_nat_trans_pdelta :
      is_nat_trans F G pdelta.
 Proof.
   intros b b' f.
+  apply (p b (hProppair _ (hsC _ _ _ _) )).
+  intro t; induction t as [a h]; simpl.
+  apply (p b' (hProppair _ (hsC _ _ _ _) )).
+  intro t; induction t as [a' h']; simpl.
+  unfold pdelta.
+  rewrite (pr2 ((pr1 (iscontr_aux_space b))) a h).
```
benediktahrens: Here it would be good to isolate the equation before rewriting with it, i.e., projecting it out or something.

PR #997, opened 2018-08-20
toward univalence for the category of (relative) monads

rmatthes:
univalence established for the category of (relative) monads when the target category is univalent

before that some simple commits with a bit more beautiful presentation of the univalence proof for the categories of algebras for a functor and the functor categories

This is now a true pull request.


benediktahrens:
I feel that univalence for the category of (relative) monads would be best done using displayed categories.

UniMath/CategoryTheory/slicecat.v
```diff
@@ -356,20 +356,21 @@ Defined.
...
+  rewrite transport_of_functor_map_is_pointwise; simpl in *.
+  unfold double_transport. unfold slicecat_mor, slice_precat_ob_mor, slicecat_mor.
+  simpl.
+  rewrite transportf_total2.
+  apply subtypePairEquality; [intro; apply hsC | ].
+  rewrite transportf_total2; simpl.
+  unfold slicecat_functor_identity_ob.
+  rewrite toforallpaths_funextsec; simpl.
+  case (id_right f).
+  now case (id_right g).
```
benediktahrens: While we are at it, we could remove "now" and replace "case" by something else.

rmatthes:
@benediktahrens : thanks for your strategic comment; however, I would have to understand how this would ease the task; I was too advanced with the "pedestrian approach", which is now completed

UniMath/CategoryTheory/Monads/RelativeMonads.v
```diff
@@ -394,6 +441,628 @@ Defined.
...
+
+Lemma is_univalent_RelMonad {C : precategory_data} {D : precategory}
+      (H: is_univalent D) (J : functor C D)(R R': RelMonad J)
+  : is_univalent (precategory_RelMonad (pr2 H) J).
+Proof.
+  split.
+  - apply isweq_idtoiso_RelMonad.
+  - clear R R'. intros R R'.
+    apply (has_homsets_RelMonad (pr2 H) J).
+Defined.
```
benediktahrens: Out of interest: statement and proof do not seem to use "pr1 H". But I would expect the theorem to be wrong when D is not univalent. What is happening?

rmatthes: I'll look into this. By the way, is there a lemma in the library that univalence is inherited through weak equivalence?

benediktahrens:
On 22/08/18 11:48, Ralph Matthes wrote:
> *@rmatthes* commented on this pull request.
> 
> 
> In UniMath/CategoryTheory/Monads/RelativeMonads.v 
> <https://github.com/UniMath/UniMath/pull/997#discussion_r211894066>:
> 
>> +Proof.
> +  apply (isweqhomot (relmonadmor_idtoiso H J R R')).
> +  - intro p. induction p.
> +    apply idpath.
> +  - apply (pr2 _ ).
> +Qed.
> +
> +Lemma is_univalent_RelMonad {C : precategory_data} {D : precategory}
> +      (H: is_univalent D) (J : functor C D)(R R': RelMonad J)
> +  : is_univalent (precategory_RelMonad (pr2 H) J).
> +Proof.
> +  split.
> +  - apply isweq_idtoiso_RelMonad.
> +  - clear R R'. intros R R'.
> +    apply (has_homsets_RelMonad (pr2 H) J).
> +Defined.
> 
> I'll look into this. By the way, is there a lemma in the library that 
> univalence is inherited through weak equivalence?

No, because that is not true. Univalence is not even preserved under 
adjoint equivalence. For instance, take the one-object category with 
only an identity arrow, and the two-object category with exactly one 
morphism between each pair of objects (that is, two isormorphic but 
distinct objects).
The former is univalent, the latter is not. But they are equivalent.

UniMath/CategoryTheory/Monads/RelativeMonads.v
rmatthes: isweq_idtoiso_RelMonad uses the hypothesis of univalence when it is applied - no mystery here

benediktahrens:
On 22/08/18 11:55, Ralph Matthes wrote:
> 
> isweq_idtoiso_RelMonad uses the hypothesis of univalence when it is 
> applied - no mystery here

Thanks a lot!

rmatthes:
I'll add further material. Do not pull, please.

rmatthes:
@benediktahrens : how can a category with one object be in the ≃ (weq) relation with a category with two objects?

rmatthes:
A more reasonable question: Is there a lemma in the library that univalence is inherited through catiso?

rmatthes:
Please pull. I'll not pursue this topic for some days, but I think the PR is sufficiently complete now.

benediktahrens:
On 22/08/18 16:27, Ralph Matthes wrote:
> A more reasonable question: Is there a lemma in the library that 
> univalence is inherited through catiso?

Yes, it is. The proof is as follows: by univalence for types, (catiso A 
B) \simeq (A = B).

rmatthes:
@benediktahrens : thanks for this express reply (and the merge), but the whole file catiso.v depends on univalence - could there be an elementary proof?

benediktahrens:
On 22/08/18 17:15, Ralph Matthes wrote:
> @benediktahrens <https://github.com/benediktahrens> : thanks for this 
> express reply (and the merge), but the whole file catiso.v depends on 
> univalence - could there be an elementary proof?

Yes, it can be shown without univalence. But remember that many relevant 
examples of univalent categories require the univalence axiom to be 
univalent, starting with the category of sets. So avoiding UA for this 
proof might not be very important.

Issue #932, opened 2018-03-02
document the type theory used in UniMath precisely


DanGrayson:
See the discussion at https://groups.google.com/forum/#!topic/univalent-mathematics/c21KD03Jh8g .
See also our style guide in https://github.com/UniMath/UniMath/blob/master/UniMath/README.md .

DanGrayson:
Maybe something like this (Thierry pointed out some of the distinctions to me):

UniMath-1 is MLTT79 except:
- the variables in λ-expressions are decorated with their type
- we omit W-types
- just the finite types of cardinality 0, 1, and 2 are used
- we omit reflection from identities to judgmental equalities
- we add the resizing rules

(the univalence axiom is not part of the type theory, but we do accept it)

UniMath-2 is UniMath-1 except:
- we add η for pairs

DanGrayson:
resolved by #999

PR #999, opened 2018-08-22
add a description of the UniMath language

DanGrayson:
resolves see #932

-- no need to run the tests

### DIFF
--- a/UniMath/README.md
+++ b/UniMath/README.md
@@ -37,6 +37,46 @@ accidentally commit and push, add its name to the head of the list in
 "../build/Makefile-configuration", which is created from
 "../build/Makefile-configuration-template".
 
+## The UniMath formal language
+
+The formal language used in the UniMath project is based on Martin-Löf type
+theory, as present in MLTT79 below.  We are currently on version 2.
+
+UniMath-1 is MLTT79 except:
+
+- the bound variable in a λ-expression is annotated with its type
+- we omit W-types
+- just the finite types of cardinality 0, 1, and 2 are used, although there would be no problem with introducing further ones
+- we omit reflection from identities to judgmental equalities
+- we add the resizing rules from the [slides](https://www.math.ias.edu/vladimir/sites/math.ias.edu.vladimir/files/2011_Bergen.pdf) of Voevodsky's 2011 talk in Bergen
+
+UniMath-2 is UniMath-1 except:
+
+- we add η for pairs
+
+The axioms accepted are: the univalence axiom, the law of excluded middle, the
+axiom of choice, and a few new variants of the axiom of choice, validated by
+the semantic model.
+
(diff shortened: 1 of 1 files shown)


DanGrayson:
I cancelled all the checks.

andrejbauer:
As for referring to Martin-Löf type theory, I would recommend

> Per Martin-Löf: *An intuitionistic theory of types*, published in
> *Twenty Five Years of Constructive Type Theory*, edited by Giovanni Sambin and Jan M. Smith, Oxford Logic Guides 36, Clarendon press, 1998.

This seems to be a mostly the same 
[PDF version](http://citeseerx.ist.psu.edu/viewdoc/download;jsessionid=24245C7C75F71FD0307EAF98B49BAA29?doi=10.1.1.131.926&rep=rep1&type=pdf) of the paper.

This version of type theory is intensional, builds terms together with their types, etc. The earlier versions (1979, 1984) are extensional and considered obsolete by those who have informed opinions on these matters.

UniMath/README.md
```diff
@@ -37,6 +37,46 @@ accidentally commit and push, add its name to the head of the list in
...
 "../build/Makefile-configuration-template".
 
+## The UniMath formal language
+
+The formal language used in the UniMath project is based on Martin-Löf type
+theory, as present in MLTT79 below.  We are currently on version 2.
+
+UniMath-1 is MLTT79 except:
+
+- the bound variable in a λ-expression is annotated with its type
```
andrejbauer: With the newer 1998 version of MLTT the bound variables in λ-expressions are already tagged, and there is no reflection for identity types.

andrejbauer: For having bound variables explicitly typed, see section 2.3.3 in Martin-Löf's paper: "If `x` is a variable of type `A` and `b[x]` is a term of type `B[x]` then `(λ x) b[x]` is a term of type `∏(x ∈ A) B[x]". I read this as having `x` explicitly typed. In contrast, earlier versions of MLTT (say in 1984 paper) say that `λ x . x` is a term which has many types.

DanGrayson: You're referring to the 1998 paper.   I read "x is a variable of type A" as a statement of something appended to the context in a prerequisite of an inference rule, and I read `(λ x) b[x]` as a description of the term -- with no `A` in it.

DanGrayson: One advantage of referring to the older paper is that it gives credit to Per for developing the ideas by that early date.  (significantly earlier)

What if I were to refer to the later paper, not as part of our definition of the UniMath language, but as an alternative exposition?

siddharthist:
@DanGrayson When you don't want to run tests for a commit, you can put `[skip ci]` in the commit message and it will do so automatically

DanGrayson:
Thanks!

PR #1004, opened 2018-08-24
Construction of a monad from an adjunction

bcldoherty:
This includes the construction of a monad from an adjunction, a proof that applying this construction to the Kleisli adjunction of a monad returns the original adjunction, and some minor cleanup to a couple of proofs in KleisliCategories.v (mainly replacing simpl with cbn).

### DIFF
--- a/UniMath/CategoryTheory/Monads/KleisliCategory.v
+++ b/UniMath/CategoryTheory/Monads/KleisliCategory.v
@@ -71,15 +71,15 @@ Proof.
   split.
   - split.
     + intros a b f.
-      unfold identity; unfold compose; simpl.
+      unfold identity; unfold compose; cbn.
       apply η_bind.
     + intros a b f.
-      unfold identity; unfold compose; simpl.
+      unfold identity; unfold compose; cbn.
       rewrite <- id_right.
       apply cancel_precomposition.
       apply bind_η.
   - intros a b c d f g h.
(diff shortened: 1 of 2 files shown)


UniMath/CategoryTheory/Monads/KleisliCategory.v
```diff
@@ -177,23 +176,52 @@ Defined.
...
+  : is_left_adjoint (Left_Kleisli_functor T) 
+    := are_adjoints_to_is_left_adjoint (Left_Kleisli_functor T)
+    (Right_Kleisli_functor T) (Kleisli_functors_are_adjoints T).
+
+Definition Right_Kleisli_is_right_adjoint {C : precategory} (T : Monad C) 
+  : is_right_adjoint (Right_Kleisli_functor T) 
+    := are_adjoints_to_is_right_adjoint (Left_Kleisli_functor T)
+    (Right_Kleisli_functor T) (Kleisli_functors_are_adjoints T).
+
+Theorem Kleisli_adjunction_monad_eq {C : precategory} (hs : has_homsets C) (T : Monad C)  : Monad_from_adjunction (Kleisli_functors_are_adjoints T) = T.
```
benediktahrens: It is great that you can get an identity here. An isomorphism would maybe be sufficient for most purposes, though.

PR #1009, opened 2018-10-08
Monoidal precategories, functors and precategories of monoids

varkor:
https://github.com/UniMath/UniMath/pull/722 has been open for over a year and doesn't look any closer to completion, but monoidal (pre)categories are very useful, so I've provided a new definition of monoidal precategories here. It's less ambitious than https://github.com/UniMath/UniMath/pull/722 and thus will hopefully be easier to merge more quickly.

Included:
- Definition of a natural isomorphisms.
- Definition of (strict) monoidal precategories.
- Definition of lax/strong monoidal functors.
- Definition of precategories of monoids for monoidal precategories.

(I haven't made a contribution of this size before, so I expect I'll have made some stylistic faux pas!)

PR #1010, opened 2018-10-09
Generalised actions and strengths

varkor:
This includes definitions of actions over monoidal categories and strengths between actions, generalising the traditional notion of tensorial strength, based upon the definitions in _Second-order and dependently-sorted abstract syntax_ by Marcelo Fiore.

I'm not sure how general/specific the definitions in UniMath are meant to be, so let me know if this is too specific. I have results that make use of these, but they may take a little while to prepare/clean up.

(I've defined various functors here in single definitions like in https://github.com/UniMath/UniMath/pull/1009#discussion_r223658051; I could refactor them as a follow-up if that's acceptable.)

### DIFF
--- a/UniMath/CategoryTheory/Categories.v
+++ b/UniMath/CategoryTheory/Categories.v
@@ -660,11 +660,6 @@ Proof.
     apply (pr1 H). apply idpath.
 Defined.
 
-Definition is_iso_id {C : precategory} {c : C} : is_iso (identity c).
-Proof.
-  exact (is_iso_qinv (identity c) (identity c) (is_inverse_in_precat_identity c)).
-Defined.
-
 Definition iso_comp_left_weq {C:precategory} {a b:C} (h:iso a b) (c:C) :
  (c --> a) ≃ (c --> b) := weqpair _ (iso_comp_left_isweq h c).
 
--- a/UniMath/CategoryTheory/Categories.v
+++ b/UniMath/CategoryTheory/Categories.v
@@ -349,7 +349,7 @@ Definition isopair {C: precategory_data}{a b : C} (f : a --> b) (fiso: is_iso f)
 Definition inv_from_iso {C:precategory_data}{a b : C} (f : iso a b) : b --> a :=
    invmap (weqpair (precomp_with f) (pr2 f a)) (identity _ ).
 
-Definition iso_inv_after_iso {C : precategory} {a b : C} (f: iso a b) :
+Definition iso_inv_after_iso {C : precategory_data} {a b : C} (f: iso a b) :
    f · inv_from_iso f = identity _ .
 Proof.
   set (T:=homotweqinvweq (weqpair (precomp_with f) (pr2 f a ))).
--- a/UniMath/CategoryTheory/Categories.v
+++ b/UniMath/CategoryTheory/Categories.v
@@ -349,7 +349,7 @@ Definition isopair {C: precategory_data}{a b : C} (f : a --> b) (fiso: is_iso f)
 Definition inv_from_iso {C:precategory_data}{a b : C} (f : iso a b) : b --> a :=
    invmap (weqpair (precomp_with f) (pr2 f a)) (identity _ ).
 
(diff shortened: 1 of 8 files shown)


UniMath/CategoryTheory/Categories.v
```diff
@@ -349,15 +349,15 @@ Definition isopair {C: precategory_data}{a b : C} (f : a --> b) (fiso: is_iso f)
 Definition inv_from_iso {C:precategory_data}{a b : C} (f : iso a b) : b --> a :=
    invmap (weqpair (precomp_with f) (pr2 f a)) (identity _ ).
 
-Definition iso_inv_after_iso {C : precategory_data}{a b : C} (f: iso a b) :
+Definition iso_inv_after_iso {C : precategory} {a b : C} (f: iso a b) :
```
rmatthes: Is there a reason why this definition is now restricted in its domain?

varkor: Not sure what happened there; I've reverted the change.

UniMath/CategoryTheory/Categories.v
```diff
@@ -660,6 +660,11 @@ Proof.
     apply (pr1 H). apply idpath.
 Defined.
 
+Definition is_iso_id {C : precategory} {c : C} : is_iso (identity c).
```
benediktahrens: There is `identity_is_iso` already, of the same type. Is there a reason why you want your version?

UniMath/CategoryTheory/Monoidal/Actions.v
```diff
@@ -0,0 +1,275 @@
...
+    reflexivity.
+  - unfold functor_compax.
+    simpl.
+    intros.
+    repeat rewrite (binprod_proj_comp f); repeat rewrite binprod_proj_comp.
+    do 2 (rewrite binprod_comp; rewrite (functor_comp odot)).
+    reflexivity.
+Defined.
+
+Definition odot_x_otimes_y_functor : (A ⊠ V) ⊠ V ⟶ A.
```
benediktahrens: In the definition of this functor, it would be good to opacify the "proof" part. 
This can be achieved either by defining first `functor_data` and `is_functor` separately, and end the second one by `Qed.` instead of `Defined.`, or by using the tactical `abstract (...)`.

UniMath/CategoryTheory/Monoidal/Actions.v
```diff
@@ -0,0 +1,275 @@
...
+Let tensor_A := monoidal_precat_tensor Mon_A.
+Notation "X ⊗_A Y" := (tensor_A (X , Y)) (at level 31).
+Notation "f #⊗_A g" := (#tensor_A (f #, g)) (at level 31).
+Let α_A := monoidal_precat_associator Mon_A.
+Let λ_A := monoidal_precat_left_unitor Mon_A.
+Let ρ_A := monoidal_precat_right_unitor Mon_A.
+
+Context (U : strong_monoidal_functor Mon_V Mon_A).
+
+Definition otimes_U_functor : A ⊠ V ⟶ A.
```
benediktahrens: In the definition of this functor, it would be good to opacify the "proof" part. 
This can be achieved either by defining first `functor_data` and `is_functor` separately, and end the second one by `Qed.` instead of `Defined.`, or by using the tactical `abstract (...)`.

UniMath/CategoryTheory/Monoidal/Actions.v
```diff
@@ -0,0 +1,275 @@
...
+    simpl.
+    intros.
+    rewrite (functor_comp U).
+    rewrite binprod_comp.
+    rewrite (functor_comp tensor_A).
+    reflexivity.
+Defined.
+
+Definition U_action_struct : action_struct.
+Proof.
```
benediktahrens: It would be good to prove some subgoals of this proof in separate lemmas.

varkor:
I've split all the definitions up into their `*_data` and `is_*` components as suggested!

PR #1013, opened 2018-10-12
add a file on double negation

DanGrayson:
a few facts, well known, potentially useful


UniMath/MoreFoundations/DoubleNegation.v
```diff
@@ -0,0 +1,37 @@
+Require Import UniMath.Foundations.All.
+
+Lemma wma_dneg {X} P : ¬¬ P -> (P -> ¬ X) -> ¬ X.
+(* when proving a negation, we may undo a double negation *)
```
siddharthist: It might be nice to have these comments start before the `Lemma` and be in coqdoc-compatible comments.

DanGrayson: Thanks, done.

DanGrayson:
This is a new file, and the failure is another file.  ???
```
COQC UniMath/Algebra/Lattice.v
File "./UniMath/Algebra/Lattice.v", line 596, characters 21-22:
Error:
In environment
X : hSet
is : latticedec X
x, y : X
H : ¬ latticedec_gt_so x y
H0 : neg (Lle is x y)
Unable to unify "hProptoType (¬ latticedec_gt_so x y)" with 
"∅".
```

UniMath/MoreFoundations/DoubleNegation.v
```diff
@@ -0,0 +1,37 @@
+Require Import UniMath.Foundations.All.
+
+(** When proving a negation, we may undo a double negation. *)
+Lemma wma_dneg {X} P : ¬¬ P -> (P -> ¬ X) -> ¬ X.
```
benediktahrens: What is the type of `X` here? 
In general, I would prefer if types of arguments were always given explicitly.

DanGrayson: Okay, done.

PR #1014, opened 2018-10-14
Natural number objects

mortberg:
Define NNO's
Prove that nat is one
Relate this to the already existing initial algebra definition in FunctorAlgebras


benediktahrens:
lgtm

PR #1015, opened 2018-10-15
Make combined all.v file

DanGrayson:
create a combined "All.v" file, which loads all files from all packages.
    
This involved settling upon a uniform parsing precedence for the notation "A ⇒ B".


DanGrayson:
Anders, I like that idea.  What about making it this way?:
```

Reserved Notation "A ⇒ B" (at level 95, left associativity).
(* to input: type "\Rightarrow" or "\r=" or "\r" or "\Longrightarrow" or "\=>" with Agda input method *)

Reserved Notation "X ⇐ c"   (at level 94, right associativity).
(* to input: type "\Leftarrow" or "\Longleftarrow" or "\l=" or "\l" with Agda input method *)
```

benediktahrens:
Isn't the function type constructor right associative?

It will confuse people a lot if implication has a different 
associativity from function type.

mortberg:
Oops, I of course meant "**right** associative"... Sorry for the confusion!

PR #1020, opened 2018-10-17
introduce a submodule called "AddNotation"

### LINKED ISSUE #1019: display of monoid operations
It would be nice if there were some way to control how the operation of a monoid is displayed.
Currently, when it occurs in a term, it is always displayed as "x * y", regardless of which of
the two scopes below is open.

```
Notation "x + y" := (op x y) : addmonoid_scope.
Notation "0" := (unel _) : addmonoid_scope.

Delimit Scope addmonoid_scope with addmonoid.

Notation "x * y" := (op x y) : multmonoid_scope.
Notation "1" := (unel _) : multmonoid_scope.

Delimit Scope multmonoid_scope with multmonoid.
```

DanGrayson:
... to allow the user to select whether a monoid operation should be
displayed additively (or multiplicatively, the default.)

Resolves #1019 .

On branch fix-monoid-notation
Changes to be committed:
	modified:   UniMath/Algebra/Archimedean.v
	modified:   UniMath/Algebra/Lattice.v
	modified:   UniMath/Algebra/Modules/Core.v
	modified:   UniMath/Algebra/Monoids_and_Groups.v
	modified:   UniMath/Algebra/Tests.v
	modified:   UniMath/CONTENTS.md
	modified:   UniMath/Ktheory/AbelianMonoid.v


DanGrayson:
Yes, but every Coq file creates a Module of the same name, so we can't avoid using them.  If we really wanted to enforce that rule, I could make a separate file 2 lines long called "AddNotation", but that wouldn't be an improvement.

Maybe the naive use of modules to implement importation units should be allowed, but the advanced usage should be disallowed (e.g., Module Type, etc.).

mortberg:
This sounds reasonable to me. Maybe this PR could also update the style guide?

DanGrayson:
I cancelled the travis build, because all I changed was a README.

benediktahrens:
@DanGrayson : see https://coderwall.com/p/f7fusq/how-to-skip-making-a-build-in-travis-ci for how to skip travis build by a special msg in commit msg

DanGrayson:
> @DanGrayson : see https://coderwall.com/p/f7fusq/how-to-skip-making-a-build-in-travis-ci for how to skip travis build by a special msg in commit msg

Thanks!

PR #1024, opened 2018-10-21
Properties of products

siddharthist:
- Uniqueness up to isomorphism of general (pre)categorical products
- Associativity of binary products in categories


siddharthist:
I wanted to prove these facts to show that categories with finite products have a monoidal structure, but I don't think it will be easy to show that this isomorphism between `(a x b) x c` and `a x (b x c)` is natural. 

Does anyone know of a "cleaner" isomorphism, one that would be easier to prove is natural?

benediktahrens:
You have the three projections from `(a x b) x c` to `a`, `b`, and `c`, 
respectively.
Use `BinProductArrow` twice to construct an arrow into `a x (b x c)`, 
first making an arrow into `b x c`, then into `a x (b x c)`.
This construction can be shown to be natural, hopefully using the lemmas 
that are already there.

varkor:
I'm happy to see this, as I had been about to define cartesian monoidal categories for a proof, so this is very helpful!

benediktahrens:
@siddharthist : I don't think that the comparison with the three-indexed coproduct is terribly useful. But now that it is done, it would be a bad idea to throw it away. I am not sure where to put it - maybe best to have it in the file on binary products, in a section.

benediktahrens:
On 22/10/2018 00:47, Langston Barrett wrote:
> @benediktahrens <https://github.com/benediktahrens> You're right, it was 
> much easier to prove that one is natural! Is there somewhere I should 
> put the alternative proof (the one based on the three-way product)? Or 
> should I just replace it with the one you suggest?

I don't think that the comparison with the three-indexed coproduct is 
terribly useful. But now that it is done, it would be a bad idea to 
throw it away. I am not sure where to put it - maybe best to have it in 
the file on binary products, in a section.

siddharthist:
@varkor I've completed everything but the pentagon proof in the definition of cartesian monoidal structure. You can check out my progress here: https://github.com/UniMath/UniMath/compare/master...siddharthist:cartesian-2?expand=1

Feel free to adapt it or finish it and submit a pull request, I likely won't be able to complete it until next weekend.

benediktahrens:
On 22/10/2018 00:47, Langston Barrett wrote:
> @benediktahrens <https://github.com/benediktahrens> You're right, it was 
> much easier to prove that one is natural! Is there somewhere I should 
> put the alternative proof (the one based on the three-way product)? Or 
> should I just replace it with the one you suggest?

I am a bit confused, because I don't see naturality proved in the code, 
though you seem to say that you did prove it. Am I missing something?

siddharthist:
@benediktahrens It's proved in the branch I linked to (https://github.com/UniMath/UniMath/compare/master...siddharthist:cartesian-2). Sorry for the lack of clarity there, I'll be updating this pull request when I finish the pentagon identity proof.

siddharthist:
Closing in favor of #1031

PR #1026, opened 2018-10-22
Definition of the diagonal functor

varkor:
It's very simple, but it's useful to include for thoroughness.


varkor:
Ah, I see! I've modified the pull request to include a comment mentioning the existing diagonal functor instead, to make text searching more convenient.

PR #1030, opened 2018-10-26
Fix category duality

DanGrayson:
... so C^op^op = C judgmentally

Also (C × D)^op = (C^op × D^op) and
product_precategory (λ i, (C i)^op) = (product_precategory C)^op.

This involves changing the end of some proofs of "is_precategory" from
Qed to Defined, for transparency.

See PR #852, which this replaces.

This resolves #602.

PR #1031, opened 2018-10-28
Monoidal structure on cartesian categories

langston-barrett:
With certain properties of products, some of which were necessary for the proofs.


DanGrayson:
lgtm.

Wait for #1030 and it will be an easy corollary that categories with coproducts have a monoidal structure.

UniMath/CategoryTheory/limits/binproducts.v
```diff
@@ -754,4 +759,285 @@ Proof.
...
+  Hint Resolve compose : binprod.
+
+  Local Notation "c '⊗' d" := (BinProductObject C (BC c d)) : cat.
+  Local Notation "f '××' g" := (BinProductOfArrows _ _ _ f g) (at level 80) : cat.
+  Local Open Scope cat.
+
+  (** The behavior of eauto here is to [apply] the above hints until successful *)
+  Definition binprod_assoc_r {x y z} : C⟦x ⊗ (y ⊗ z), (x ⊗ y) ⊗ z⟧.
+  Proof.
+    eauto with binprod.
```
benediktahrens: Can you please replace this automated tactic by an explicit constructon?

UniMath/CategoryTheory/limits/binproducts.v
```diff
@@ -754,4 +759,285 @@ Proof.
...
+  Defined.
+
+  Definition binprod_assoc_l {x y z} : C⟦(x ⊗ y) ⊗ z, x ⊗ (y ⊗ z)⟧.
+  Proof.
+    eauto with binprod.
```
benediktahrens: Can you please replace this automated tactic by an explicit constructon?

siddharthist:
While I certainly understand the concerns about automation in Unimath, there are certain proofs I'd rather not write out in full. Perhaps I'll put these in a separate repository. Feel free to cherry-pick the less problematic commits from the PR (e.g. uniqueness of products).

DanGrayson:
I don't know a definition of "automation" in proof scripts, Benedikt.

benediktahrens:
On 29/10/2018 06:58, Daniel R. Grayson wrote:
> I don't know a definition of "automation" in proof scripts, Benedikt.

Surely, using a tactic called `auto` or `eauto` or `autorewrite` counts 
as such?

benediktahrens:
It seems particularly worrying that the automation is being used to define a structure (`Defined.`), not to do a proof (`Qed.`).

DanGrayson:
I've never used any of those, nor do I use eapply, but let me take a look at this pull request and see whether there is something problematic about such proofs.

We currently have no appeals to `eauto`, so portability would be enhanced by not admitting its use.  On the other hand, we have some additions to hint databases that Vladimir installed.  Did he use them?

UniMath/CategoryTheory/limits/binproducts.v
DanGrayson: The construction found by `eauto` here turns out to be this one:
```
  Proof.
    apply BinProductArrow.
    - apply BinProductArrow.
      + apply BinProductPr1.
      + exact (BinProductPr2 _ _ · BinProductPr1 _ _).
    - exact (BinProductPr2 _ _ · BinProductPr2 _ _).
  Defined.
```

siddharthist:
I'm fine rewriting the eauto proofs, that would be easy. Note that `autorewrite` is only used in the proofs of axioms, it could be made opaque with `abstract` or similar.

DanGrayson:
> > We currently have no appeals to eauto, so portability would be enhanced by not admitting its use.
> 
> Is there anywhere the reasoning behind the portability arguments is set forth? I don't see how UniMath could realistically transition to another proof assistant without effectively reproving everything, as the tactic language (even without automation) seems heavily specific to Coq — but perhaps such methods have been discussed in the past?

Well, yes, it's probably a vain hope, and no, there is no "argument" set forth, especially since now we have 960274 lines of proofs.

PR #1040, opened 2018-11-10
Initial (pre/set/univalent)category, notes on equalizers

siddharthist:
- Strengthen the lemma about the type of functors from the empty category being contractible
 - Give the initial object in subcategories of `Precat`
 - Small writeup of the difficulties involved in equalizers


UniMath/CategoryTheory/categories/Cats.v
```diff
@@ -264,4 +300,18 @@ Proof.
...
 Defined.
 
+(** *** Notes on equalizers *)
+
+(** From Emily Rhiel's Category Theory In Context:
+    "The equalizer of a pair of functors [(F G : C --> D)] is the subcategory
+    of [C] consisting of those objets [c : ob C] so that [F c = G c]
+    and those morphisms [f] such that [# F f = # G f]."
+
+    Unfortunately, it is nontrivial to construct equalizers for categories
```
DanGrayson: Re: "it is nontrivial to construct equalizers for categories"

Do we want to do that?

I wonder what the purpose of this entire comment is.

siddharthist: The purpose of the comment is just to outline some of the difficulties involved in constructing equalizers.

I initially thought that it might be easy to do, but this is what I ran into. Anyway I'm happy to remove it if you dont find it helpful.

UniMath/CategoryTheory/categories/Cats.v
```diff
@@ -264,4 +300,18 @@ Proof.
   + apply weqproperty.
 Defined.
 
+(** *** Notes on equalizers *)
+
+(** From Emily Rhiel's Category Theory In Context:
```
benediktahrens: s/Rhiel/Riehl

siddharthist: Good catch!

siddharthist: Since this was a small typo, I went ahead and revised the commit and force-pushed, instead of making a new commit.

UniMath/CategoryTheory/categories/Cats.v
DanGrayson: She's working with strict categories, which are the objects of a category.  But our categories, which have a type of objects that may not be a set, aren't the objects of a category.

benediktahrens: I think this comment is not very useful in the univalent setting. It is probably best to remove it.

siddharthist: Will do.

siddharthist:
Oh, thanks for fixing this @benediktahrens!

PR #1042, opened 2018-11-11
Alternative universal property for equalizers

siddharthist:
I was hoping to prove these without the homset condition so that I could use the results in `Induction/M/Limits.v` to deduce that the precategory of types has all limits (in the categorical sense). 

In any case, the alternative definition could be useful in the meantime. It's in the same spirit as the alternative universal property for binary products, which has a precedent in Foundations.


siddharthist:
Right, that question would only be relevant if (as with the analogous result for products) I could avoid the homset condition. I'll go ahead and write out the weak equivalence here before merging.

siddharthist:
I've updated this.

benediktahrens:
@siddharthist : great, thanks. In general, I would prefer if PRs were amended instead of overridden, that is, if force-push were avoided. It makes checking the amendments easier/possible.

PR #1043, opened 2018-11-11
Reflection of limits, limits in full subcategories

siddharthist:


siddharthist:
I forgot to push the commit that gives the lemma referenced here. I'll do that when I'm back on wifi!

siddharthist:
I've made several additions and changes to this merge request, some more substantive and useful content.

DanGrayson:
lgtm

PR #1049, opened 2018-11-16
Exact categories

DanGrayson:
Here is a beginning of a formalization of Quillen's notion of "exact category".  These
categories have slightly weaker properties than abelian categories do, but they still support
proofs of sufficiently many of the standard diagram chasing lemmas that much can be done with
them.  In particular, Quillen defines higher algebraic K-theory for any exact category.

In this first installment I prove several diagram-chasing lemmas, formalize the modern definition of exact category, prove that it's equivalent to Quillen's original definition (one of whose properties is known as his "obscure axiom"), show that an additive category becomes an exact category by regarding the split sequences as the exact ones, and show that a function X -> ob M from a type X to the objects of an exact category M, whose image is closed under extensions up to isomorphism, induces an exact category structure on X.

Forthcoming: show that an abelian category is exact; construct, for a ring R, the exact categories of finitely generated left R-modules and of finitely generated left projective R-modules; construct, for an exact category M, the exact category of bounded chain complexes over M.

This is part of a larger project to formalize the results of my paper at https://arxiv.org/abs/1310.8644 .


DanGrayson:
The only error now is that the Foundations package has changed.  The only change there is to the reserved notations in Init.v.

PR #1051, opened 2018-11-16
Builder for bicategories

nmvdw:
I added a build function for bicategory. The file `one_types.v` compiles, but emacs cannot open it.

### DIFF
--- a/UniMath/CategoryTheory/Bicategories/Examples/one_types.v
+++ b/UniMath/CategoryTheory/Bicategories/Examples/one_types.v
@@ -1,0 +1,80 @@
+Require Import UniMath.Foundations.All.
+Require Import UniMath.CategoryTheory.Bicategories.Bicat.
+
+Definition one_type
+  := ∑ (A : Type), isofhlevel 3 A.
+
+Definition one_type_to_type : one_type -> UU := pr1.
+Coercion one_type_to_type : one_type >-> UU.
+
+Definition one_type_bicat
+  : bicat.
+Proof.
+  simple refine (build_bicategory
+                   _ _ _ _ _ _ _
+                   _ _ _ _ _ _ _
+                   _ _ _ _ _ _ _
+                   _ _ _ _ _ _ _
+                   _ _ _ _ _ _ _
+                   _ _
+                ).
+  - exact one_type.
+  - exact (fun X Y => X -> Y).
+  - exact (fun _ _ f g => f = g).
+  - exact (fun _ x => x).
+  - exact (fun _ _ _ f g x => g(f x)).
+  - reflexivity.
+  - cbn ; intros X Y f g h p q.
+    exact (p @ q).
+  - cbn ; intros X Y Z f g h p.
+    exact (maponpaths (λ φ x, φ (f x)) p).
+  - cbn ; intros X Y Z f g h p.
+    exact (maponpaths (λ φ x, h(φ x)) p).
+  - reflexivity.
+  - reflexivity.
+  - reflexivity.
+  - reflexivity.
+  - reflexivity.
+  - reflexivity.
+  - abstract (reflexivity).
+  - abstract (intros X Y f g p ;
+              apply pathscomp0rid).
+  - abstract (intros X Y f g h k p q r ;
+              apply path_assoc).
+  - abstract reflexivity.
+  - abstract reflexivity.
+  - abstract (intros X Y Z f g h i p q ;
+              induction p, q ; cbn ;
+              reflexivity).
+  - abstract (intros X Y Z f g h i p q ;
+              induction p, q ; cbn ;
+              reflexivity).
+  - abstract (intros X Y f g p ;
+              induction p ; cbn ;
+              reflexivity).
+  - abstract (intros X Y f g p ;
+              induction p ; cbn ;
+              reflexivity).
+  - abstract (intros W X Y Z f g h i p ;
+              induction p ; cbn ;
+              reflexivity).
+  - abstract (intros W X Y Z f g h i p ;
+              induction p ; cbn ;
+              reflexivity).
+  - abstract (intros W X Y Z f g h i p ;
+              induction p ; cbn ;
+              reflexivity).
+  - abstract (cbn ; intros X Y Z f g h i p q ;
+              induction p, q ; cbn ;
+              reflexivity).
+  - abstract reflexivity.
+  - abstract reflexivity.
+  - abstract reflexivity.
+  - abstract reflexivity.
+  - abstract reflexivity.
+  - abstract reflexivity.
+  - abstract reflexivity.
+  - abstract reflexivity.
+  - abstract (cbn ; intros X Y f g ;
+              exact (impredfun 3 X Y (pr2 Y) f g)).
+Defined.
--- a/UniMath/CategoryTheory/.package/files
+++ b/UniMath/CategoryTheory/.package/files
@@ -145,6 +145,7 @@ DisplayedCats/Examples.v
 DisplayedCats/Adjunctions.v
 
 Bicategories/Bicat.v
+Bicategories/Examples/one_types.v
 Bicategories/StandardBicategories.v
 Bicategories/OpCellBicat.v
 Bicategories/OpMorBicat.v
--- a/UniMath/CategoryTheory/Bicategories/Bicat.v
+++ b/UniMath/CategoryTheory/Bicategories/Bicat.v
@@ -479,7 +479,7 @@ Proof.
            *** exact lwhisk.
            *** exact rwhisk.
     + cbn.
-      repeat (use tpair ; try abstract assumption).
+      repeat (use tpair) ; abstract assumption.
   - cbn.
     unfold isaset_cells.
     apply cell_is_set.
--- a/UniMath/CategoryTheory/Bicategories/Bicat.v
+++ b/UniMath/CategoryTheory/Bicategories/Bicat.v
@@ -354,6 +354,137 @@ Definition isaset_cells (C : prebicat) : UU
 Definition bicat : UU
   := ∑ C : prebicat, isaset_cells C.
 
+Definition build_bicategory
+           (ob : UU)
+           (mor : ob -> ob -> UU)
+           (cell : ∏ {X Y : ob}, mor X Y -> mor X Y -> UU)
+           (id₁ : ∏ (X : ob), mor X X)
+           (comp : ∏ {X Y Z : ob}, mor X Y -> mor Y Z -> mor X Z)
+           (id₂ : ∏ {X Y : ob} (f : mor X Y), cell f f)
+           (vcomp : ∏ {X Y : ob} {f g h : mor X Y}, cell f g -> cell g h -> cell f h)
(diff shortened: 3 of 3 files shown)


UniMath/CategoryTheory/Bicategories/Examples/one_types.v
```diff
@@ -0,0 +1,80 @@
+Require Import UniMath.Foundations.All.
```
siddharthist: It would be nice to have a comment at the top of the file explaining what's in it.

siddharthist:
@nmvdw Thanks, it's always good to have more explicit constructors. Are there some repeated uses of `tpair` that this could replace?

benediktahrens:
@DanGrayson : both Niels and I have difficulties opening the file `one_types.v` of this PR in Emacs.
Could you try?

DanGrayson:
Proof General goes into an almost-infinite loop trying to colorize the code because of that long
sequence of underscores.  Change that line to `use build_bicategory.` and all will be with emacs.

I suspect it's an inefficient regular expression that's involved in the colorization.

DanGrayson:
Actually, `use build_bicategory` will not work either, as it goes only up to 9 arguments.

benediktahrens:
On 17/11/2018 11:06, Daniel R. Grayson wrote:
> Actually, |use build_bicategory| will not work either, as it goes only 
> up to 9 arguments.

Is it likely that, if we extended the definition of the `use` tactic to 
up to 37 arguments, we would experience the same Emacs problem in that file?

nmvdw:
>It would be nice to have a comment at the top of the file explaining what's in it.
Indeed. Will be added later.

>Thanks, it's always good to have more explicit constructors. Are there some repeated uses of tpair that this could replace?

It can simplify `bicat_of_cats` in `BicatOfCats.v`.

DanGrayson:
> On 17/11/2018 11:06, Daniel R. Grayson wrote: Actually, |use build_bicategory| will not work either, as it goes only up to 9 arguments.
> Is it likely that, if we extended the definition of the `use` tactic to up to 37 arguments, we would experience the same Emacs problem in that file?

Yes, probably.  You could do something ridiculous like this:

```
Ltac simple_rapply_2 p :=
  simple refine p ||
  simple refine (p _) ||
  simple refine (p _ _) ||
  simple refine (p _ _ _) ||
  simple refine (p _ _ _ _) ||
  simple refine (p _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _ _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _ _ _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _ _ _ _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _ _ _ _ _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _ _ _ _ _ _ _ _ _ _).
Ltac simple_rapply_1 p :=
  simple refine p ||
  simple refine (p _) ||
  simple refine (p _ _) ||
  simple refine (p _ _ _) ||
  simple refine (p _ _ _ _) ||
  simple refine (p _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _ _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _ _ _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _ _ _ _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _ _ _ _ _ _ _ _ _) ||
  simple_rapply_2 p.
Ltac simple_rapply p :=
  simple refine p ||
  simple refine (p _) ||
  simple refine (p _ _) ||
  simple refine (p _ _ _) ||
  simple refine (p _ _ _ _) ||
  simple refine (p _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _ _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _ _ _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _ _ _ _ _ _ _ _) ||
  simple refine (p _ _ _ _ _ _ _ _ _ _ _ _ _ _) ||
  simple_rapply_1 p.

```

Functions with 37 arguments are ungainly.  I think it would be better if we didn't have any.

benediktahrens:
On 17/11/2018 12:37, Daniel R. Grayson wrote:
>     On 17/11/2018 11:06, Daniel R. Grayson wrote: Actually, |use
>     build_bicategory| will not work either, as it goes only up to 9
>     arguments.
>     Is it likely that, if we extended the definition of the |use| tactic
>     to up to 37 arguments, we would experience the same Emacs problem in
>     that file?
> 
> Yes, probably. You could do something ridiculous like this:
> 
> |Ltac simple_rapply_2 p := simple refine p || simple refine (p _) || 
> simple refine (p _ _) || simple refine (p _ _ _) || simple refine (p _ _ 
> _ _) || simple refine (p _ _ _ _ _) || simple refine (p _ _ _ _ _ _) || 
> simple refine (p _ _ _ _ _ _ _) || simple refine (p _ _ _ _ _ _ _ _) || 
> simple refine (p _ _ _ _ _ _ _ _ _) || simple refine (p _ _ _ _ _ _ _ _ 
> _ _) || simple refine (p _ _ _ _ _ _ _ _ _ _ _) || simple refine (p _ _ 
> _ _ _ _ _ _ _ _ _ _) || simple refine (p _ _ _ _ _ _ _ _ _ _ _ _ _) || 
> simple refine (p _ _ _ _ _ _ _ _ _ _ _ _ _ _) || simple refine (p _ _ _ 
> _ _ _ _ _ _ _ _ _ _ _ _). Ltac simple_rapply_1 p := simple refine p || 
> simple refine (p _) || simple refine (p _ _) || simple refine (p _ _ _) 
> || simple refine (p _ _ _ _) || simple refine (p _ _ _ _ _) || simple 
> refine (p _ _ _ _ _ _) || simple refine (p _ _ _ _ _ _ _) || simple 
> refine (p _ _ _ _ _ _ _ _) || simple refine (p _ _ _ _ _ _ _ _ _) || 
> simple refine (p _ _ _ _ _ _ _ _ _ _) || simple refine (p _ _ _ _ _ _ _ 
> _ _ _ _) || simple refine (p _ _ _ _ _ _ _ _ _ _ _ _) || simple refine 
> (p _ _ _ _ _ _ _ _ _ _ _ _ _) || simple refine (p _ _ _ _ _ _ _ _ _ _ _ 
> _ _ _) || simple_rapply_2 p. Ltac simple_rapply p := simple refine p || 
> simple refine (p _) || simple refine (p _ _) || simple refine (p _ _ _) 
> || simple refine (p _ _ _ _) || simple refine (p _ _ _ _ _) || simple 
> refine (p _ _ _ _ _ _) || simple refine (p _ _ _ _ _ _ _) || simple 
> refine (p _ _ _ _ _ _ _ _) || simple refine (p _ _ _ _ _ _ _ _ _) || 
> simple refine (p _ _ _ _ _ _ _ _ _ _) || simple refine (p _ _ _ _ _ _ _ 
> _ _ _ _) || simple refine (p _ _ _ _ _ _ _ _ _ _ _ _) || simple refine 
> (p _ _ _ _ _ _ _ _ _ _ _ _ _) || simple refine (p _ _ _ _ _ _ _ _ _ _ _ 
> _ _ _) || simple_rapply_1 p. |
> 
> Functions with 37 arguments are ungainly. I think it would be better if 
> we didn't have any.

What does "ungainly" mean? What do you suggest as an upper limit for the 
number of arguments to a function?

Such a function is the only way I can see to abstract away from how a 
bicategory is implemented (association of the components of the tuple). 
Do you have any other ideas?

nmvdw:
>Such a function is the only way I can see to abstract away from how a
bicategory is implemented (association of the components of the tuple).
Do you have any other ideas?

As one function, yes. An alternative way would be:
1. A function which combines `prebicat_data` with `prebicat_laws`.
2. Tactic which construct `prebicat_data` and `prebicat_laws` by repeatedly applying `use tpair`.
 
(I fear that this might give a slower build function though)

benediktahrens:
On 17/11/2018 12:59, Niels van der Weide wrote:
>     Such a function is the only way I can see to abstract away from how a
>     bicategory is implemented (association of the components of the tuple).
>     Do you have any other ideas?
> 
> As one function, yes. An alternative way would be:
> 
>  1. A function which combines |prebicat_data| with |prebicat_laws|.
>  2. Tactic which construct |prebicat_data| and |prebicat_laws| by
>     repeatedly applying |use tpair|.
> 
> (I fear that this might give a slower build function though)

Do you want to try your suggestion 1.?

DanGrayson:
>> Functions with 37 arguments are ungainly. I think it would be better if we didn't have any.
> What does "ungainly" mean? What do you suggest as an upper limit for the number of arguments to a function? Such a function is the only way I can see to abstract away from how a bicategory is implemented (association of the components of the tuple). Do you have any other ideas?

Certainly 37 is excessive.  Why not define a hierarchy of builder functions?  Then each one would take just a few arguments.  I.e., build not just bicategories, but also prebicategories, prebicategory_data, prebicategory_laws, and so on.  

Suppose you were teaching the concept to a human.  Would you present it as a list of 37 items?  The ncatlab presents it with 8 bullet points.

benediktahrens:
Good, but
- I don't think the tactic `build_bicat_laws` is very helpful.
- Should we delete the commented-out function with 37 arguments?

nmvdw:
>I don't think the tactic build_bicat_laws is very helpful.

`build_bicat_laws` is a bit more informative than `repeat (use tpair)`. I think the former looks a bit nicer, but it is not very helpful indeed.

>Should we delete the commented-out function with 37 arguments?

If you agree this version is better, then yes.

benediktahrens:
On 17/11/2018 14:56, Niels van der Weide wrote:
>     I don't think the tactic build_bicat_laws is very helpful.
> 
> |build_bicat_laws| is a bit more informative than |repeat (use tpair)|. 
> I think the former looks a bit nicer, but it is not very helpful indeed.

It does look nicer, but it is obscuring what is happening. Please delete it.

>     Should we delete the commented-out function with 37 arguments?
> 
> If you agree this version is better, then yes.

Yes, please delete.

nmvdw:
Okay. I deleted both of them now.

PR #1065, opened 2018-11-19
Stuff about bicategories ported from D&N library

benediktahrens:


siddharthist:
@benediktahrens It might be good to add some top-level comments in each of these files noting the author and original provenance.

DanGrayson:
lgtm

benediktahrens:
On 19/11/2018 22:54, Langston Barrett wrote:
> @benediktahrens <https://github.com/benediktahrens> It might be good to 
> add some top-level comments in each of these files noting the author and 
> original provenance.

Good point, I'll file an issue for that.

PR #1066, opened 2018-11-20
Classifying monos and epis in HSET

siddharthist:


UniMath/CategoryTheory/categories/HSET/MonoEpiIso.v
```diff
@@ -13,22 +15,112 @@ Require Import UniMath.MoreFoundations.Tactics.
...
+Require Import UniMath.CategoryTheory.Monics.
 Require Import UniMath.CategoryTheory.Epis.
 
 Require Import UniMath.CategoryTheory.categories.HSET.Core.
 
 Local Open Scope cat.
 
+(** ** Points as global elements *)
+
+Definition global_element_HSET {A : hSet} (a : A) : HSET⟦unitset, A⟧ := fun _ => a.
```
DanGrayson: Why define this function when `(λ_,a)` is so easy to type?

UniMath/CategoryTheory/categories/HSET/MonoEpiIso.v
```diff
@@ -13,22 +15,112 @@ Require Import UniMath.MoreFoundations.Tactics.
...
+(** TODO: I think there is a name in UniMath for the constant function at [a],
+    what is it? *)
+
+Definition global_element_HSET_paths_weq {A : hSet} (x y : A) :
+  (x = y) ≃ (global_element_HSET x = global_element_HSET y).
+Proof.
```
DanGrayson: Derive this as a corollary of : 
```
weqfunfromunit: ∏ X : UU, (unit → X) ≃ X
```
and
```
weqonpathsincl: ∏ (X Y : UU) (f : X → Y), isincl f → ∏ x x' : X, x = x' ≃ f x = f x'
```

UniMath/CategoryTheory/categories/HSET/MonoEpiIso.v
```diff
@@ -13,22 +15,112 @@ Require Import UniMath.MoreFoundations.Tactics.
...
+  apply global_element_HSET_paths_weq.
+Qed.
+
+(** ** Monomorphisms are exactly injective functions [MonosAreInjective_HSET] *)
+
+Lemma MonosAreInjective_HSET {A B : HSET} (f: HSET ⟦ A, B ⟧) :
+      isMonic f ≃ isInjective f.
+Proof.
+  apply weqimplimpl.
+  - intro isM.
```
DanGrayson: This can be simplified to
```
  - intro isM.
    apply incl_injectivity; intros b.
    apply invproofirrelevance; intros a1 a2.
    apply subtypeEquality; [intro; apply setproperty|].
    apply global_element_HSET_paths_weq.
    apply isM.
    apply funextsec; intro t.
    exact (pr2 a1 @ ! pr2 a2).
```

UniMath/CategoryTheory/categories/HSET/MonoEpiIso.v
```diff
@@ -13,22 +15,112 @@ Require Import UniMath.MoreFoundations.Tactics.
...
+      apply global_element_HSET_paths_weq.
+      exact (pr2 a1).
+    + refine (_ @ !global_element_HSET_comp _ _).
+      apply global_element_HSET_paths_weq.
+      exact (!pr2 a2).
+  - intro isI.
+    unfold isMonic.
+    intros ? ? ? eq.
+    apply funextfun; intro.
+    apply (invweq (weqpair _ (isI _ _))).
```
DanGrayson: There's a function for this `weqpair` application:
```
    apply (Injectivity _ isI).
```
(and "apply" is smart enough that `invweq` is not needed, although that's perhaps a bit confusing).

UniMath/CategoryTheory/categories/HSET/MonoEpiIso.v
siddharthist: I think it's more searchable, and makes the intent/perspective clearer. Haskell programmers, for instance, generally prefer to use [`const`](http://hackage.haskell.org/package/base-4.12.0.0/docs/Prelude.html#v:const) rather than `\_ -> a`. 

In fact, I thought there was something like `const` in UniMath already, but couldn't find it.

UniMath/CategoryTheory/categories/HSET/MonoEpiIso.v
siddharthist: Ah, it makes sense that those would already be available. Thanks for the pointers!

UniMath/CategoryTheory/categories/HSET/MonoEpiIso.v
siddharthist: I'm happy to make it `Local` if you prefer?

siddharthist:
Okay, I think I've incorporated all the feedback. The proofs are a lot shorter now, thanks @DanGrayson!

UniMath/CategoryTheory/categories/HSET/MonoEpiIso.v
DanGrayson: Ah.  Maybe you could just call it "element", since in this context, every element is global.

siddharthist:
Okay, I believe this is ready now.

PR #1068, opened 2018-11-20
Move definition of subsets from RealNumbers to MoreFoundations

siddharthist:
Adds a TOC and removes a duplicate result as well.


benediktahrens:
```
COQC UniMath/RealNumbers/NonnegativeRationals.v
File "./UniMath/RealNumbers/NonnegativeRationals.v", line 19, characters 23-29:
Error: The reference subset was not found in the current environment.
```

DanGrayson:
lgtm

PR #1073, opened 2018-11-20
Uniqueness of adjoints

nmvdw:
I transferred most of the proof that internal adjoints are unique. The only part which I didn't do yet, is showing that it's a mere proposition.

### DIFF
--- a/UniMath/CategoryTheory/Bicategories/adjoint_unique.v
+++ b/UniMath/CategoryTheory/Bicategories/adjoint_unique.v
@@ -1,0 +1,593 @@
+(*
+Right adjoints, units, counits are unique.
+
+Authors: Dan Frumin, Niels van der Weide
+
+Ported from: https://github.com/nmvdw/groupoids
+*)
+
+Require Import UniMath.Foundations.All.
+Require Import UniMath.MoreFoundations.All.
+Require Import UniMath.CategoryTheory.Categories.
+Require Import UniMath.CategoryTheory.Bicategories.Bicat. Import Bicat.Notations.
+Require Import UniMath.CategoryTheory.Bicategories.Unitors.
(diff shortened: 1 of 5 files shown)


UniMath/CategoryTheory/Bicategories/adjoint_unique.v
```diff
@@ -0,0 +1,593 @@
...
+        refine (vcomp_move_R_pM _ _ _ _) ; simpl.
+        rewrite <- transport_vcomp₁_l.
+        rewrite eisretr ; cbn.
+        symmetry.
+        apply transport_counit.
+  Defined.
+  *)
+End UniquenessAdjoint.
+
+(*
```
DanGrayson: What's the purpose of these large commented-out sections of code?

nmvdw: We are transferring code from our library (in HoTT) to UniMath. The commented-out section is still to-do. I can remove it if you want.

DanGrayson: That's okay, thanks.

UniMath/CategoryTheory/Bicategories/transport_laws.v
```diff
@@ -0,0 +1,119 @@
+(*
```
siddharthist: It's best if comments start with `(**` so they show up in the output of `coqdoc`

nmvdw: Sure! I fixed it now.

UniMath/CategoryTheory/Bicategories/adjoint_unique.v
nmvdw: Now it is finished.

PR #1078, opened 2018-11-21
fix compilation error in CategoryTheory/All

benediktahrens:
Maybe we should stop tracking those $P/All files?


co-dan:
> I guess it doesn't matter whether we track the All.v files, since the Makefile will generate them, anyway, and no one has to see them in the source repository.

It is also a bit inconsistent in that some of the All.v files are not tracked, whilst others are being tracked.
What is point of having them in the source tree if they have to be generated anyway?

PR #1083, opened 2018-11-23
put notations in a scope

### LINKED ISSUE #1082: notations in a scope
The notations in the file `UniMath/SubstitutionSystems/Notation.v` should be in a scope, so if someone loads all of UniMath, the notations don't become active.

DanGrayson:
Put some notations into a scope and make the style guide require that.

Resolves #1082.


DanGrayson:
oops, I have to fix something ...

PR #1090, opened 2018-11-24
hfiber as a pullback of arrows from unit

siddharthist:
For context, I'm hoping this well-known result will enable me to prove that `HSET` has a subobject classifier.

For a reference, see the "Definition" section of https://arxiv.org/pdf/1012.5647.pdf


UniMath/MoreFoundations/WeakEquivalences.v
```diff
@@ -28,6 +28,17 @@ Proof.
     reflexivity.
 Defined.
 
+Definition transitive_paths_weq_reverse {X : UU} {x y z : X} :
+  x = y -> (z = x ≃ z = y).
+Proof.
```
DanGrayson: I don't think it makes much sense to have such a lemma with x=y as a hypothesis, since the version with x and y the same can suffice for all applications, in which case this is just an instance of `idweq`.

siddharthist: I find it useful for backwards reasoning, e.g. when I haven't already proved that x = y.

See the proof I use it in for an example of where it's useful.

DanGrayson: Maybe you do find it useful, but you didn't use it in this pull request.  Why not wait until you need it, and then, in the middle of your proof, phrase the lemma like this?:
```
Local Lemma A {X : UU} {x y z : X} : x = y -> z = x ≃ z = y.
Proof.
  intros p. induction p. apply idweq.
Qed.
```
Or alternatively you could write
```
  match goal with |- _ = ?x ≃ _ = ?y => enough (x=y) as p end.
  induction p; apply idweq.
```
You could also write this:
```
  use (transportf (λ t, _ = _ ≃ _ = t) _ (idweq _)).
```
Otherwise we'll get a proliferation of trivial lemmas in `MoreFoundations`.

siddharthist: Oh, my bad, I used it in an earlier version of the proof. Thanks for pointing that out.

I will remove it.

DanGrayson: Here's an even quicker way to go:
```
  apply eqweqmap.
  apply maponpaths.
```

UniMath/CategoryTheory/categories/HSET/Limits.v
```diff
@@ -246,6 +248,90 @@ Proof.
...
+      |            | y
+      V            V
+      X --- f -->  Y
+    >>
+    makes [Z] (isomorphic to) the inverse image of a point [y : Y] under [f].
+ *)
+
+(** A translation of [weqfunfromunit] into the language of category theory,
+    to make the statement of the next lemmas more concise. *)
+Lemma weqfunfromunit_HSET (X : hSet) : HSET⟦TerminalHSET, X⟧ ≃ X.
```
benediktahrens: You don't seem to use this thing elsewhere?

UniMath/CategoryTheory/categories/HSET/Limits.v
```diff
@@ -246,6 +248,90 @@ Proof.
...
+        -- apply funextfun; intro; reflexivity.
+      * intros t.
+        apply subtypeEquality.
+        -- intro; apply has_homsets_HSET.
+        -- cbn.
+           apply funextfun; intro; cbn.
+           apply subtypeEquality.
+           ++ intro; apply setproperty.
+           ++ apply (toforallpaths _ _ _ (pr2 t)).
+Qed.
```
benediktahrens: `isPullback` contains some information important for computation, so you probably do not want to render the whole thing opaque. It would probably be good, on the other hand, to make some subproofs of this construction opaque.

UniMath/CategoryTheory/categories/HSET/Limits.v
siddharthist: It is used in the statement of `hfiber_is_pullback`

UniMath/CategoryTheory/categories/HSET/Limits.v
siddharthist: I don't see a simple way to make only some of the subproofs opaque (none of them make much sense as lemmas on their own). I've made the whole proof transparent for now.

siddharthist:
This has been fixed and is ready for another round of review/merge

PR #1099, opened 2018-11-26
Functor categories as displayed categories

co-dan:
- Define a category of functors `[C,D]` in several iterations, starting from a base category of functions on objects.
- Show that the resulting display category is univalent if `D` is univalent.

### DIFF
--- a/UniMath/CategoryTheory/DisplayedCats/Constructions.v
+++ b/UniMath/CategoryTheory/DisplayedCats/Constructions.v
@@ -7,15 +7,15 @@ Partial contents:
 - Displayed category given by a structure on objects and a proposition
    on morphisms of the base category
 - Direct products of displayed categories (and their projections)
-  - [dirprod_cat D1 D2]
-  - [dirprodpr1_functor], [dirprodpr2_functor]
+  - [dirprod_disp_cat D1 D2]
+  - [dirprodpr1_disp_functor], [dirprodpr2_disp_functor]
 - Sigmas of displayed categories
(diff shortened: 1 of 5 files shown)


UniMath/CategoryTheory/DisplayedCats/Constructions.v
```diff
@@ -7,15 +7,15 @@ Partial contents:
...
+  - [dirprod_disp_cat D1 D2]
+  - [dirprodpr1_disp_functor], [dirprodpr2_disp_functor]
 - Sigmas of displayed categories
 - Displayed functor cat
 - Fiber cats
 *)
 
 Require Import UniMath.Foundations.Sets.
-Require Import UniMath.MoreFoundations.PartA.
+Require Import UniMath.MoreFoundations.All.
```
siddharthist: I think it's best to avoid importing the `All` files. This makes the compilation of this file depend (unnecessarily) on the compilation of everything that might be added to `MoreFoundations` in the future. 

This is just a matter of preference, though.

UniMath/CategoryTheory/DisplayedCats/Constructions.v
```diff
@@ -26,6 +26,26 @@ Local Open Scope mor_disp_scope.
 
 Section Auxiliary.
 
+(* TODO: Move this to another file *)
```
siddharthist: These could probably go in `MoreFoundations.WeakEquivalences`.

PR #1102, opened 2018-11-26
fix 2 admitted lemmas

benediktahrens:


DanGrayson:
Oops, we shouldn't have accepted the admitted lemmas before.

Issue #1044, opened 2018-11-12
ambiguous coercions


DanGrayson:
I'm beginning to suspect that having multiple coercion paths between two classes is a bad idea, both for making proofs easily and for execution time.  Does anyone else think so?

Here is what's bugging me today:
```
         (precategory_data_from_precategory
-            (category_to_precategory
-               (categoryWithAbgrops_category
                  (PreAdditive_categoryWithAbgrops M))))

         (precategory_data_from_precategory
-            (precategoryWithBinOps_precategory
-               (categoryWithAbgrops_precategoryWithBinOps
                  (PreAdditive_categoryWithAbgrops M))))
```
The notion of "precategory with binary operations on each Hom(A,B)" should be removed so we can avoid this.

DanGrayson:
If we got rid of `precategoryWithBinOps`, we could probably also get rid of this silly rewriting rule:

```
Lemma rewrite_op {A:PreAdditive} (x y:A) :
  @to_binop (categoryWithAbgrops_precategoryWithBinOps (PreAdditive_categoryWithAbgrops A)) x y
  =
  @op (pr1monoid (grtomonoid (abgrtogr (@to_abgr (PreAdditive_categoryWithAbgrops A) x y)))).
Proof.
  reflexivity.
Defined.
```

benediktahrens:
On 12/11/2018 14:04, Daniel R. Grayson wrote:
> If we got rid of |precategoryWithBinOps|, we could probably also get rid 
> of this silly rewriting rule:
> 
> |Lemma rewrite_op {A:PreAdditive} (x y:A) : @to_binop 
> (categoryWithAbgrops_precategoryWithBinOps 
> (PreAdditive_categoryWithAbgrops A)) x y = @op (pr1monoid (grtomonoid 
> (abgrtogr (@to_abgr (PreAdditive_categoryWithAbgrops A) x y)))). Proof. 
> reflexivity. Defined. |

Sounds good!

siddharthist:
I agree that it is generally a bad idea to have multiple coercion paths. Perhaps there is a way that we can get Coq to throw an error (or at least a warning) when this occurs?

DanGrayson:
Loading the global All.v file does produce plenty of warning messages:
```
[Loading ML file ltac_plugin.cmxs ...
Ambiguous paths:
[hProp_to_hSet; pr1hSet] : hProp >-> Preamble.UU
[...]

PR #1112, opened 2018-11-28
remove ambiguous coercions from Foundations,

DanGrayson:
... MoreFoundations, and Combinatorics

Begins to address #1044 

In the process I moved my stuff on negative propositions into its own file in
`MoreFoundations`, leaving `Foundations` tidier.


siddharthist:
@DanGrayson Did you forget to include the file `NegativePropositions`? 

Other than that omission, this LGTM!

DanGrayson:
Oops, I've added it.

DanGrayson:
I'm reluctant to add tables of contents to any files, because nothing checks that they are correct.

siddharthist:
As someone who has fixed a few, I understand that worry. Sections (with `(** **`-style comments), however, are next to the code they divide (so are easy to maintain) and still provide good organization. If used with `outline-mode`, they can even provide folding. So including those might be nice, even in lieu of a TOC.

DanGrayson:
Yes, I have two of those.

Issue #980, opened 2018-05-07
Upstream code on foundations and categories from UniMath/TypeTheory


benediktahrens:
The file https://github.com/UniMath/TypeTheory/blob/master/TypeTheory/Auxiliary/Auxiliary.v contains stuff on
- Foundations (equivalences of types, transport
- Adjunctions and equivalences of categories
which should be upstreamed asap, to avoid (further) duplication of efforts (see #955 ).

benediktahrens:
@siddharthist : would you be interested in upstreaming the category-theoretic part? I suggest merging your PR #955 first, and then you could replace, e.g., your composition-of-adj-eqs-is-adj-eq with the more general result from UniMath/TypeTheory.

siddharthist:
Yes, I would!

And no worries on the previously-done results -- that was an exercise from a textbook that I wanted to complete, rather than a lemma I wanted to establish merely to move on to something using it.

arnoudvanderleer:
@benediktahrens Has this been resolved by these pull requests?

benediktahrens:
@arnoudvanderleer : I think this issue can be closed. If necessary, a new one on the same topic can be opened.

PR #1118, opened 2018-12-02
Port composites and inverses of equivalences from TypeTheory

siddharthist:
Also, split material into a few separate files (for the usual reasons: speed, clarity, organization).

Partially addresses #980


benediktahrens:
Thanks a lot for taking care of this, and for not organizing the material better!

PR #1125, opened 2018-12-04
upstream part of file on comprehension categories from TypeTheory

benediktahrens:
@peterlefanulumsdaine


rmatthes:
It is worse than that: TypeTheory does not compile now. This is bad if someone is interested in C-systems or B-systems, for example.

benediktahrens:
On 07/12/2018 13:42, Ralph Matthes wrote:
> It is worse than that: TypeTheory does not compile now. This is bad if 
> someone is interested in C-systems or B-systems, for example.

This is an issue to be reported against UniMath/TypeTheory.

Note that there is already one:
https://github.com/UniMath/TypeTheory/issues/144

rmatthes:
Agreed. And the problem reported there is the one I have. But awareness of those problems should be ensured also on the side of UniMath/UniMath on which UniMath/TypeTheory depends.

siddharthist:
The way the software development community more broadly deals with the problem of downstream breakage is versioning/releases. UniMath could consider doing something similar. I'd imagine every release would be API-breaking, but it would guarantee that other packages could refer to a consistent version.

Another option would be for TypeTheory to include UniMath as a git submodule.

benediktahrens:
On 07/12/2018 14:02, Ralph Matthes wrote:
> Agreed. And the problem reported there is the one I have. But awareness 
> of those problems should be ensured also on the side of UniMath/UniMath 
> on which UniMath/TypeTheory depends.

Fixing the compilation error takes 5 minutes, only twice as long as 
awareness-raising does. I am not sure it is a good use of my time, or of 
anybody else's, to raise awareness.

PRs to fix the problem are welcome, and will be handled prioritarily. 
Otherwise it'll be fixed when I take the time to fix it.

PR #1130, opened 2018-12-08
Restructured the file UnivalentCat and BicatOfCats

nmvdw:
Our focus is univalent bicategories, so we mainly are interested in the bicategory of univalent categories. For this reason, in BicatOfCats, we define this bicategory rather than the bicategory of ordinary categories.
Furthermore, the file UnivalentCat was rather large and contained a lot of intermediate results, which are actually about category theory. Those are now moved to other places in the category theory library.


UniMath/CategoryTheory/CategoryEquality.v
```diff
@@ -0,0 +1,262 @@
...
+Require Import UniMath.Foundations.All.
+Require Import UniMath.MoreFoundations.All.
+Require Import UniMath.CategoryTheory.Categories.
+Require Import UniMath.CategoryTheory.functor_categories.
+Require Import UniMath.CategoryTheory.whiskering.
+Require Import UniMath.CategoryTheory.catiso.
+
+Open Scope cat.
+
+(* MOVE SOMEWHERE ELSE? *)
```
benediktahrens: This lemma should probably go into MoreFoundations.PartA or similar.

PR #1139, opened 2018-12-19
one dir of correspondence between distributive laws and monad liftings

tamaravonglehn:


UniMath/CategoryTheory/Monads/MonadAlgebras.v
```diff
@@ -231,3 +237,179 @@ Proof.
...
+
+
+Section Liftings.
+
+Context {C : category} (S T : Monad C).
+
+(** A lifting of (T, η, μ) is a monad (T', η', μ') on (MonadAlg S) which commutes with the
+forgetful functor:
+
+                  T'
```
siddharthist: This diagram should be in between `<<` and `>>` so it shows up properly in `coqdoc`

PR #1141, opened 2018-12-20
Subobject classifiers

siddharthist:
Plus, an example in `HSET`


UniMath/CategoryTheory/SubobjectClassifier.v
```diff
@@ -0,0 +1,95 @@
...
+Require Import UniMath.CategoryTheory.limits.pullbacks.
+Require Import UniMath.CategoryTheory.Monics.
+
+Local Open Scope cat.
+
+(** ** Definition *)
+
+Definition subobject_classifier {C : precategory} (T : Terminal C) : UU :=
+  ∑ (O : ob C) (true : C⟦T, O⟧), ∏ (X Y : ob C) (m : Monic _ X Y),
+    iscontr (∑ (chi : C⟦Y, O⟧)
```
mortberg: Why not use `∃!` instead of `iscontr (∑ ...)`?

UniMath/CategoryTheory/categories/HSET/Structures.v
```diff
@@ -383,3 +386,197 @@ Proof.
   refine (hset_equiv_is_iso a b (weqpair f _)).
   apply (type_iso_is_equiv _ _ (isopair _ is_iso_forget_f)).
 Defined.
+
+(** ** Subobject classifier *)
+
+Definition hProp_set : hSet := (_,, isasethProp).
```
mortberg: This should probably be upstreamed to `Foundations` or `MoreFoundations`?

UniMath/CategoryTheory/categories/HSET/Structures.v
```diff
@@ -383,3 +386,197 @@ Proof.
...
+
+Lemma isaprop_hfiber_monic {A B : hSet} (f : HSET⟦A, B⟧) (isM : isMonic f) :
+  isPredicate (hfiber f).
+Proof.
+  intro; apply incl_injectivity, MonosAreInjective_HSET; assumption.
+Qed.
+
+(* Require Import UniMath.Foundations.All. *)
+(* Require Import UniMath.MoreFoundations.All. *)
```
mortberg: Why are these 2 comments here?

UniMath/CategoryTheory/categories/HSET/Structures.v
```diff
@@ -383,3 +386,197 @@ Proof.
...
+      -- intro.
+          apply isapropdirprod.
+          ++ apply (setproperty (hset_fun_space _ _)).
+          ++ apply (setproperty (hset_fun_space _ unitset)).
+      -- cbn.
+          apply funextsec; intro; cbn.
+          (** Precompose with [m] and use the commutative square *)
+          apply (invweq (weqpair _ (MonosAreInjective_HSET m (MonicisMonic _ m) _ _))).
+          eapply pathscomp0.
+          ++ apply (toforallpaths _ _ _ (pr1 (pr2 t))).
```
mortberg: Was `t` introduced by the `intro` after `apply funextsec` above? If so please name it there. It's not very good practice to rely on the automatic naming of Coq as it makes proof scripts brittle.

UniMath/CategoryTheory/categories/HSET/Structures.v
```diff
@@ -383,3 +386,197 @@ Proof.
...
+          ++ apply pathsinv0.
+             apply hfiberpr2.
+Defined.
+
+(** For any subset [s : Y -> hProp], the carrier [∑ y : Y, s y] is a pullback
+    of [s] with the constantly-true arrow. *)
+Lemma HSET_carrier_is_pullback {Y : HSET} (chi : HSET ⟦ Y, hProp_set ⟧) :
```
mortberg: I don't understand the naming convention here... Shouldn't the above lemma be the "is_pullback" (or better "isPullback") one and this the Pullback one? Also I don't understand why it is "HSET_carrier_..."?

UniMath/CategoryTheory/categories/HSET/Structures.v
```diff
@@ -383,3 +386,197 @@ Proof.
...
+        -- apply toforallpaths in H; cbn.
+           specialize (H p); cbn in H.
+           abstract (rewrite H; exact tt).
+      * split; [reflexivity|].
+        apply proofirrelevance, hlevelntosn, iscontrfuntounit.
+    + intro.
+      apply subtypeEquality'; [|apply isapropdirprod; apply setproperty].
+      apply funextfun; intro.
+      apply subtypeEquality'; [|apply propproperty].
+      refine (_ @ toforallpaths _ _ _ (pr1 (pr2 t)) x).
```
mortberg: Once again, where did `t` come from?

UniMath/CategoryTheory/categories/HSET/Structures.v
mortberg: Also, what is the type of `t`? It seems to me like either some coercion or some projection function is missing as you are using `pr1` and `pr2`.

UniMath/CategoryTheory/categories/HSET/Structures.v
```diff
@@ -383,3 +386,197 @@ Proof.
...
+Proof.
+  intros.
+  use hfiberpair.
+  - exact (hfiberpr1 _ _ z).
+  - reflexivity.
+Qed.
+
+Definition subobject_classifier_HSET : subobject_classifier TerminalHSET.
+Proof.
+  unfold subobject_classifier.
```
mortberg: Why do you have to unfold it? The `exists` should still work without unfolding?

UniMath/CategoryTheory/categories/HSET/Structures.v
```diff
@@ -383,3 +386,197 @@ Proof.
...
+      }
+
+      apply funextfun; intro y.
+      apply weqtopathshProp, weqimplimpl.
+      * intro isO.
+
+        (** We know that [carrier (pr1 O')] is a pullback of [pr1 O'] and [const_htrue].
+            By hypothesis, X is as well. Thus, we have a canonical isomorphism
+            [carrier (pr1 O') -> X], which commutes with the pullback projections.
+            In particular, the following triangle commutes:
```
mortberg: Which triangle? :)

UniMath/CategoryTheory/categories/HSET/Structures.v
```diff
@@ -383,3 +386,197 @@ Proof.
...
+            >>
+         *)
+
+        pose (PBO' := mk_Pullback (pr1 O') (@const_htrue unitHSET) X m (TerminalArrow TerminalHSET X) (pr1 (pr2 O')) (pr2 (pr2 O'))).
+        pose (PBC := HSET_carrier_is_pullback (pr1 O')).
+        pose (pbiso := pullbackiso PBC PBO').
+
+        use hfiberpair.
+        -- exact (morphism_from_z_iso _ _ _ (pr1 pbiso) (y,, isO)).
+        -- replace (pr1 m (morphism_from_z_iso _ _ _ (pr1 pbiso) (y,, isO))) with (((pr1 pbiso) · pr1 m) (y,, isO)); [|reflexivity].
```
mortberg: Why not use `change ... with ...`? It should be the same as `replace ... with ...; [|reflexivity]`?

UniMath/CategoryTheory/categories/HSET/Structures.v
```diff
@@ -383,3 +386,197 @@ Proof.
...
+            X ------- ! ---> unit
+            |                 |
+            |                 |
+            V                 V
+            Y -- pr1 O' --> hProp
+          >>
+          then [pr1 O' = hfiber m].
+       *)
+
+      assert (eq : m · pr1 O' ~ m · (fun z => (hfiber (pr1 m) z,, isaprop_hfiber_monic (pr1 m) (pr2 m) z : pr1hSet hProp_set))).
```
mortberg: This line is very long, as are a few others in this proof.

siddharthist: Do you have a suggestion on how to shorten it? Since it's an `assert`, the thing I'm trying to assert has to be fully specified... I suppose I could introduce an extraneous `pose` for the RHS or something like that.

UniMath/CategoryTheory/categories/HSET/Structures.v
siddharthist: I always forget about `change`... Thanks!

UniMath/CategoryTheory/categories/HSET/Structures.v
siddharthist: The type of `t` comes from the universal property of the pullback, it is another "candidate pullback": `t : ∑ hk : Z → X, (λ x : Z, m (hk x)) = f × (λ _ : Z, tt) = g`. There are no accessors for such things.

siddharthist:
I've addressed all the review comments (modulo one of Anders', which I responded to above). Thank you both for the detailed review!

PR #1142, opened 2018-12-21
Restriction of adjunctions to equivalences on their fixed points

### LINKED ISSUE #1120: Rename EquivalenceExamples, reorganize Adjunctions
The file EquivalenceExamples doesn't actually use any material on equivalences. It seems to be examples of adjunctions.

I propose that we make a new subdirectory `CategoryTheory/Adjunctions/` to house this file and the content in `Adjunctions.v` (with the latter possibly split up among a few files).

siddharthist:
- Reorganize material on adjunctions, fixes #1120 
- Loosen some hypotheses on results about isos in full subcategories


UniMath/CategoryTheory/Adjunctions/Restriction.v
```diff
@@ -0,0 +1,247 @@
...
+
+Require Import UniMath.CategoryTheory.Adjunctions.Core.
+
+Local Open Scope cat.
+
+(** ** Lemmas about inverses *)
+
+(** The right inverse of an invertible morphism must be equal to the known (two-sided) inverse. *)
+(** TODO: Did I switch up right and left here vis a vis the conventional use? *)
+Lemma right_inverse_of_iso_is_inverse {C : precategory} {c c' : C}
```
benediktahrens: These first lemmas should probably go into other files.

siddharthist: I will wait to do this until #1144 is merged, since that will change the file to which these will be moved.

siddharthist:
This has been fixed.

PR #1143, opened 2018-12-28
Instructions for installing UniMath using Nix

maggesi:
Installing UniMath from scratch using Nix is pretty easy.
It is enough to copy paste the instructions below:
just 8 lines that work independently on any major Unix-system.
(No differences between Mac and Linux; no differences either between the various Linux distributions: dpkg/rpm etc.)
I think it worth to inform the users of this alternative method and perhaps consider it for adoption as the default one. 
```bash
curl https://nixos.org/nix/install | sh
git clone https://github.com/UniMath/UniMath.git
nix-shell -p ocaml ocamlPackages.findlib ocamlPackages.camlp4 ocamlPackages.camlp5 ocamlPackages.num gnumake git
git clone https://github.com/UniMath/UniMath.git
cd UniMath
make
exit
```

NB: This is a preliminary draft.  E.g., these instructions can be further refined for installing and configuring Emacs and Proof-General in one shot.  (For now, we simply suggest to use the default method.)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,6 +1,9 @@
 Installation of UniMath
 =======================
 
+NB: This file describes the default method for installing UniMath.  An
+alternative method using the [Nix Package Manager](https://nixos.org/nix/) is available in the file [INSTALL\_NIX.md](https://github.com/UniMath/UniMath/blob/master/INSTALL_NIX.md).
+
 Prepare for installation by installing the OCAML compiler and a more modern
 version of `bash` on your system.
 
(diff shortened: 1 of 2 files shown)


DanGrayson:
Probably the first line should be `sudo mkdir /nix ; sudo chown $USER: /nix`.

maggesi:
The installation script takes care to create the Nix store with the appropriate attributes, so I believe it is better to delegate this kind of details that may change in the future.
However, it is true that the script requires sudo access to root.
I added a notice to warn the user about this in the last commit.

PR #1146, opened 2019-01-01
Type family over A equivalent to map into A

benediktahrens:
- Construction by @marcbezem
- This PR updates and supersedes PR #1086


UniMath/MoreFoundations/PartD.v
```diff
@@ -0,0 +1,68 @@
...
+Defined.
+
+Definition display {A: Type} :
+  (∑ B, B -> A) -> (A -> Type).
+Proof.
+  intro Bf. exact (hfiber (pr2 Bf)).
+Defined.
+
+Definition totalfst {A: Type} :
+  (A -> Type) -> (∑ B, B -> A).
```
DanGrayson: If we think of "displaying" as the operation of putting the fibers of a map side by side in a total space, then it's this map that one would name `display`.  The other map could be named something like `fibersof`.

DanGrayson:
Modulo one remark, this looks fine.

PR #1150, opened 2019-01-14
update INSTALL

DanGrayson:
Using Homebrew under Mac OS X, we no longer need to build camlp5 specially, because
my pull request was accepted, see https://github.com/Homebrew/homebrew-core/pull/35916 ;
and we no longer need an extra tap for getting ocaml-findlib.

Under Arch Linux we now need to install ocaml-num.

[skip ci]


DanGrayson:
I killed travis.

PR #1152, opened 2019-01-20
Refactor material on slice categories

siddharthist:


mortberg:
I don't have any particularly strong opinions for or against splitting the file on slice categories like this. So if @benediktahrens and @DanGrayson approves I'm fine with merging this PR.

About where to put results about subobject classifiers in slice categories: I would put them in `SubobjectClassifiers` (at least after the splitting of the material on slice categories). If I'm working on a development that need this I would have imported the file on subobject classifiers anyway so `Search` would help me. However if you are planning to split SubobjectClassifiers into multiple files in a separate directory as well then there is definitely a question on where to put the construction of subobject classifiers in slice categories. That is maybe another argument for why we shouldn't split related results and constructions into too fine grained files.

DanGrayson:
With emacs TAGS search and Coq Search, it's almost irrelevant to me what file results are in, or what the sizes of the files are, but I can imagine that it might make a difference to others.  I'll defer to Benedikt, since it's category theory.

siddharthist:
Closing due to conflicts with further developments on slice categories. Will reopen if there's more interest.

PR #1155, opened 2019-02-01
Change some Qed into Defined for some proofs of equivalences

amblafont:
The underlying equivalences really need to be known and unfolded
for some use cases.


DanGrayson:
lgtm

PR #1156, opened 2019-02-04
Redefine the bicategory of 1-types.

co-dan:
Define two-cells between f and g to be homotopies f~g instead of equalities f=g.


siddharthist:
Why not have both definitions in different files and a proof of their equivalence using function extensionality?

nmvdw:
That would indeed be interesting and better. 

However, the equivalence proof would take a while. For that, we need the notion of ‘biequivalence between bicategories’ and that takes a while to set up.

PR #1161, opened 2019-02-05
Use univalence 2 when possible

maggesi:


benediktahrens:
@maggesi @co-dan 
Merge conflict

PR #1163, opened 2019-02-11
Clean up Ktheory and move upstream

DanGrayson:
Move a lot of stuff in the Ktheory package upstream.

Also move some lemmas from CategoryTheory upstream and make some arguments implicit.


DanGrayson:
Let me check compilation time first, before pulling it...

benediktahrens:
Euh, some of the files you are suggesting to move contain "Inductive" types. I have never liked those files to be in Ktheory, but I would even less like those files to go into CategoryTheory.

DanGrayson:
What if I comment them out and everything that depends on them, so I can work on replacing them with something kosher later on?


> Euh, some of the files you are suggesting to move contain "Inductive" types. I have never liked those files to be in Ktheory, but I would even less like those files to go into CategoryTheory.

siddharthist:
A lot of this material should go further upstream, for instance in `RepresentableFunctors/AbelianMonoid.v` there are results about naturals, standard finite sets, and free monoids, all of which belong in separate files further upstream than `CategoryTheory`. It might be easier to move this material further, in smaller steps.

DanGrayson:
> A lot of this material should go further upstream, for instance in `RepresentableFunctors/AbelianMonoid.v` there are results about naturals, standard finite sets, and free monoids, all of which belong in separate files further upstream than `CategoryTheory`. It might be easier to move this material further, in smaller steps.

... except this is a construction of a free monoid different from the one upstream.

DanGrayson:
I'll move upstream what can be moved, though...

DanGrayson:
Okay, I've removed the files with inductive constructions and upstreamable lemmas, so now it could be considered again.

DanGrayson:
The travis failure is due to some blank lines being deleted in Foundations/Sets.v.  I must have done that absent-mindedly.  I'll leave it, since it seems beneficial, anyway.

DanGrayson:
The compilation time is fine.

UniMath/SyntheticHomotopyTheory/Halfline.v
```diff
@@ -0,0 +1,90 @@
...
+           (s:target_paths target_points) (n:ℕ) :
+  maponpaths (halfline_map s) (squash_path n (S n)) = s n.
+Proof.
+  intros. refine (_ @ map_path_check s n _).
+  apply pathsinv0. apply maponpathscomp.
+Defined.
+
+(*
+Local Variables:
+compile-command: "make -C ../.. UniMath/SyntheticHomotopyTheory/Halfline.vo"
```
benediktahrens: Can we remove these?

UniMath/SyntheticHomotopyTheory/Circle.v
```diff
@@ -0,0 +1,339 @@
...
+  { set (y := f (basepoint circle)). set (l := maponpaths f circle_loop).
+    set (P := λ T : underlyingType circle, circle_map _ T = f T).
+    apply transport_fun_path. rewrite pathscomp0rid.
+    change (idpath y @ maponpaths f circle_loop) with (maponpaths f circle_loop).
+    exact (! circle_map_check_paths l). }
+Defined.
+
+(*
+Local Variables:
+compile-command: "make -C ../.. UniMath/SyntheticHomotopyTheory/Circle.vo"
```
benediktahrens: Remove?

UniMath/Combinatorics/MetricTree.v
```diff
@@ -0,0 +1,107 @@
...
+          { exact a. } }
+        { assert (a := natleh0n m).
+          assert (b := natlehlthtrans _ _ _ a h).
+          assert (c := natlthtolehsn _ _ b).
+          exact (nat_dist_minus 1 n c). } } } }
+Defined.
+
+(*
+Local Variables:
+compile-command: "make -C ../.. TAGS UniMath/Ktheory/MetricTree.vo"
```
benediktahrens: Remove?

UniMath/MoreFoundations/QuotientSet.v
```diff
@@ -0,0 +1,60 @@
+(* -*- coding: utf-8 -*- *)
```
benediktahrens: We do not usually have that in our files. Do you think it would be good to add this everywhere?

benediktahrens:
I am seeing two `Record` types defined in the moved code. It would be great to convert them to iterated Sigmas. If we decide to not fix this now, then an issue should be opened.

UniMath/SyntheticHomotopyTheory/Halfline.v
DanGrayson: I think they're useful for anyone.

UniMath/SyntheticHomotopyTheory/Circle.v
DanGrayson: ditto

UniMath/Combinatorics/MetricTree.v
DanGrayson: ditto

UniMath/MoreFoundations/QuotientSet.v
DanGrayson: If we start to observe problems associated with its lack, maybe then.  In my .emacs file I have this:
```
(set-language-environment "UTF-8")
```
, so any file I edit is in that coding system, by default.

DanGrayson:
> I am seeing two `Record` types defined in the moved code. It would be great to convert them to iterated Sigmas. If we decide to not fix this now, then an issue should be opened.

I'd prefer to modify the coding style rules to allow it, because the Coq developers have said that any deviation of the behavior of records from iterated sums is a bug in Coq.

DanGrayson:
Okay, you've convinced me, the latest commit removes that use of "Record".

UniMath/CategoryTheory/RepresentableFunctors/DirectSum.v
```diff
@@ -12,26 +12,29 @@ Require Import UniMath.CategoryTheory.Core.NaturalTransformations.
...
       : Hom C (universalObject B) (universalObject D).
-Proof. intros. apply RawMatrix.from_matrix. apply identity_matrix.
-       assumption. assumption. Defined.
+Proof.
+  intros. apply RawMatrix.from_matrix. apply identity_matrix.
+  - assumption.
+  - assumption.
+Defined.
 
 Record DirectSum {C:category} (h:ZeroMaps C) I (dec : isdeceq I) (c : I -> ob C) :=
```
benediktahrens: Here is another record.

DanGrayson: Oops, removing...

benediktahrens:
There is one last record in UniMath/Algebra/GroupAction.v, 
Record ActionStructure (G:gr) (X:hSet) :=...
There is also ActionStructure' (prime) which is an iterated Sigma-type. So it should be very easy to replace.

DanGrayson:
Speaking of records, what about this one of yours?
```
  Record disp_cat (C : precategory) : UU :=
    { ob_disp : C -> UU
    ; mor_disp {x y : C} : (x --> y) -> ob_disp x -> ob_disp y -> UU
    ; id_disp {x : C} (xx : ob_disp x) : mor_disp (identity x) xx xx
    ; comp_disp {x y z : C} {f : x --> y} {g : y --> z}
                   {xx : ob_disp x} {yy : ob_disp y} {zz : ob_disp z}
        : mor_disp f xx yy -> mor_disp g yy zz -> mor_disp (f ;; g) xx zz
    ; id_left_disp {x y} {f : x --> y} {xx} {yy} (ff : mor_disp f xx yy)
        : comp_disp (id_disp xx) ff
          = transportb (λ g, mor_disp g xx yy) (id_left _) ff
    ; id_right_disp {x y} {f : x --> y} {xx} {yy} (ff : mor_disp f xx yy)
        : comp_disp ff (id_disp yy)
          = transportb (λ g, mor_disp g xx yy) (id_right _) ff
    ; assoc_disp {x y z w} {f : x --> y} {g : y --> z} {h : z --> w}
        {xx} {yy} {zz} {ww}
        (ff : mor_disp f xx yy) (gg : mor_disp g yy zz) (hh : mor_disp h zz ww)
        : comp_disp ff (comp_disp gg hh)
          = transportb (λ k, mor_disp k _ _) (assoc _ _ _)
            (comp_disp (comp_disp ff gg) hh)
    ; homsets_disp {x y} {f : x --> y} {xx} {yy} : isaset (mor_disp f xx yy)
    }.
```

benediktahrens:
On 12/02/2019 17:49, Daniel R. Grayson wrote:
> Speaking of records, what about this one of yours?

It is there for documentation purposes, and is used nowhere.
To prevent accidental usage, it is in a module that is not exported into 
other files.

Afterwards, the same structure is defined as an iterated Sigma-type, 
whose accessor functions have the same names.

DanGrayson:
So it's not really needed, right?

DanGrayson:
Let's convert it to a comment, so no one is tempted to use it.

benediktahrens:
On 12/02/2019 17:55, Daniel R. Grayson wrote:
> So it's not really needed, right?

Depends on what you mean by "needed". It should be considered to be part 
of the documentation, not of the code. In that sense, it is not needed, 
and everything will still compile if we remove it.

benediktahrens:
On 12/02/2019 17:57, Daniel R. Grayson wrote:
> Let's convert it to a comment, so no one is tempted to use it.
> 

Ok!

siddharthist:
Did you mean to change the coq submodule in https://github.com/UniMath/UniMath/pull/1163/commits/18ac16ad1ef0292f45241a43c2f8fd773c976d85?

PR #1164, opened 2019-02-12
stop suppressing all but one type of warning message

DanGrayson:


benediktahrens:
@DanGrayson : will this PR not clash with #1163 ? Which one would you like to have merged first? (I.e., which one is easier to fix when merge conflicts arise?)

DanGrayson:
I'd like the other one first.

This one has to be fixed, anyway, because scopes can't be declared until 8.10.

DanGrayson:
Oops, that empty commit was a mistake here.

The only travis failure was the check for changes to Foundations -- they're intentional.

PR #1184, opened 2019-03-06
Update USAGE.md by a screenshot of Emacs variables

benediktahrens:
Describe what to do when Emacs wants to set variables


DanGrayson:
lgtm

PR #1194, opened 2019-04-02
Declare hint databases in the prelude

ppedrot:
Adapt to coq/coq#8984.

### DIFF
--- a/UniMath/Foundations/Init.v
+++ b/UniMath/Foundations/Init.v
@@ -188,3 +188,7 @@ Ltac exact_op x := (* from Jason Gross: same as "exact", but with unification th
   let T := type of x in
   let G := match goal with |- ?G => constr:(G) end in
   exact (((λ g:G, g) : T -> G) x).
+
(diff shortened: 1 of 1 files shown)


benediktahrens:
@ppedrot : Thanks a lot for the PR. Is it important for us to understand what these commands do?
I assume that this is something that is part of the Coq Prelude that we do not load (-noinit)?

ejgallego:
> I assume that this is something that is part of the Coq Prelude that we do not load (-noinit)?

They were loaded by ML code, they just declare some empty databases that rewrite needs to work; they shouldn't affect the logic.

ppedrot:
Indeed. Before, these hint bases used to be created by the OCaml files.

benediktahrens:
On 02/04/2019 17:55, Pierre-Marie Pédrot wrote:
> Indeed. Before, these hint bases used to be created by the OCaml files.

Thanks a lot to you and the Coq developers in general for taking care of 
this, and for keeping UniMath compatible with Coq in general!

PR #1196, opened 2019-04-03
Minor cleaning

vbgl:
While doing some changes to Coq, I hit this proof and realized part of the script was not obviously useful.


siddharthist:
lgtm

PR #1202, opened 2019-04-05
Monoidal precategories as prebicategories

mroman42:
Monoidal categories as an example of bicategories. This could be useful to recycle existing results in the bicategories (sub)library, e.g.: both unitors coincide in the monoidal unit.


mroman42:
Ops, it is not clear to me what is failing right now. Any hint would be appreciated.

DanGrayson:
```
--- checking every proof file is listed in one of the packages ---
error: *** file UniMath/CategoryTheory/Monoidal/BicategoryFromMonoidal.v not listed in appropriate file UniMath/*/.package/files
Makefile:391: recipe for target 'check-listing-of-proof-files' failed
make: *** [check-listing-of-proof-files] Error 1
```

mroman42:
Thank you very much, I missed the error message when reading the log.

benediktahrens:
lgtm

The one question I am asking myself: should this go in Monoidal or in Bicategories?

mroman42:
I was trying to use general results from Bicategories in Monoidal (while I would not expect to use results from Monoidal in Bicategories), so my reasoning was that Monoidal was going to depend on Bicategories anyway and I preferred to avoid Bicategories also depending on Monoidal

UniMath/CategoryTheory/Monoidal/BicategoryFromMonoidal.v
```diff
@@ -0,0 +1,195 @@
...
+Defined.
+
+Definition two_cells_from_monoidal : prebicat_2cell_struct (one_cells_data_from_monoidal)
+  := λ _ _ a b, pM ⟦ a , b⟧.
+
+Definition prebicat_data_from_monoidal : prebicat_data.
+Proof.
+  use mk_prebicat_data.
+  - apply (one_cells_data_from_monoidal ,, two_cells_from_monoidal).
+  - split. { intros a b f. apply (identity f). }
```
siddharthist: I often find it tiresome to think up names that I don't end up using. You can always put `?` instead of an identifier if you don't use it later.

siddharthist:
@mroman42 Looks like there's a merge conflict.

PR #1204, opened 2019-04-07
Injective pairing is inconsistent with univalence

JLimperg:
Injective pairing is the proposition

  (a ,, b) = (a ,, b') -> b = b'.

This is inconsistent with univalence, which Benedikt and I agreed is surprising enough to warrant a proof.


martinescardo:
Shouldn't it be the case that injective pairing is equivalent to the K axiom (all types are sets), and this is the reason for the inconsistency?

benediktahrens:
On 07/04/2019 22:09, Martin Escardo wrote:
> Shouldn't it be the case that injective pairing is equivalent to the K 
> axiom (all types are sets), 

Yes, I think so. For a more systematic analysis of the various 
incompatibilities, this would be a good result to have.

> and this is the reason for the inconsistency?

Not sure about the "the" in "the reason", but the construction can 
certainly be factorized in this way. The counterexample here uses the 
fact that bool has a non-identity automorphism.

JLimperg:
We can indeed play a similar game with pairs of the form `(x ,, x = y)` to get UIP. I've added a commit to that effect.

UniMath/MoreFoundations/NoInjectivePairing.v
```diff
@@ -0,0 +1,121 @@
...
+
+(** [idfun] is a left identity of [funcomp]. *)
+Lemma funcomp_id_l {A B} (f : A → B) :
+  idfun B ∘ f = f.
+Proof.
+  apply funextsec; intro. apply idpath.
+Defined.
+
+(** Weak equivalences are injective. *)
+Lemma isinj_weq {A B} (eq : A ≃ B) : isInjective (pr1weq eq).
```
siddharthist: This should be upstreamed to where isInjective is defined

UniMath/MoreFoundations/NoInjectivePairing.v
```diff
@@ -0,0 +1,121 @@
...
+  b = b'.
+
+(** Proof sketch by Folkmar Frederik Ramcke. Formalisation by Jannis Limperg with help from Joj
+   Helfer. *)
+
+
+(** * Preliminaries *)
+
+(** [idfun] is a left identity of [funcomp]. *)
+Lemma funcomp_id_l {A B} (f : A → B) :
```
siddharthist: This should be upstreamed to where idfun or funcomp is defined, and used in the construction of the precategory of types.

UniMath/MoreFoundations/NoInjectivePairing.v
```diff
@@ -0,0 +1,121 @@
...
+    + induction y; apply idpath.
+  - cbn. intro t.
+    induction t as [b eq].
+    induction b; induction eq; apply idpath.
+Defined.
+
+(** Thus, we can construct two equal pairs whose second components are different: *)
+Definition negb_weq_pair
+  : @paths (∑ A : UU, A → bool) (bool ,, idfun bool) (bool ,, negb).
+Proof.
```
DanGrayson: One can give a proof that avoids existential variables this way:
```
Definition negb_weq_pair
  : @paths (∑ A : UU, A → bool) (bool ,, idfun bool) (bool ,, negb).
Proof.
  refine (total2_paths_f (B := λ A, A → bool) (s := bool ,, idfun bool) (s' := bool ,, negb)
                         (weqtopaths negb_weq) _ ).
  refine (transportf_fun (idfun UU) (weqtopaths negb_weq) (idfun (idfun UU bool)) @ _).
  change (transportb (idfun UU) (weqtopaths negb_weq) = negb).
  refine (maponpaths _ (pathsinv0_weqtopaths _ ) @ _).
  apply weqpath_transport.
Defined.
```

UniMath/MoreFoundations/NoInjectivePairing.v
JLimperg: I've removed the lemma for now since it's not needed for the proof Dan suggested. Of course, I can re-add it to `MoreFoundations.PartB` or wherever.

UniMath/MoreFoundations/NoInjectivePairing.v
JLimperg: The two sides aren't quite judgmentally equal; we need to apply `funext` first. I can inline the lemma in the proof, of course, or avoid it by applying `funext` there.

UniMath/MoreFoundations/NoInjectivePairing.v
```diff
@@ -0,0 +1,121 @@
...
+Lemma isinj_weq {A B} (eq : A ≃ B) : isInjective (pr1weq eq).
+Proof.
+  apply isweqonpathsincl.
+  apply isofhlevelfweq.
+Defined.
+
+(** Inverting the path obtained from an equivalence is the same as inverting the equivalence. *)
+Lemma pathsinv0_weqtopaths {A B : UU} (eq : A ≃ B) :
+  ! (weqtopaths eq) = weqtopaths (invweq eq).
+Proof.
```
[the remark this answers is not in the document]
JLimperg: That's indeed a much nicer proof! However, I was under the impression that `rewrite` is considered evil because it may introduce `match`es. If so, I can rewrite the proof in terms of `!`, `@`, `maponpaths` and friends.

UniMath/MoreFoundations/NoInjectivePairing.v
JLimperg: Is there a reason to avoid evars? My proof contains a bit less 'term noise', but I don't mind much either way.

JLimperg:
@DanGrayson, thank you for your helpful comments! I've addressed some of them in new commits and responded inline to the others.

UniMath/MoreFoundations/NoInjectivePairing.v
```diff
@@ -0,0 +1,85 @@
+Require Import UniMath.Foundations.All.
+Require Import UniMath.MoreFoundations.Univalence.
+
+(** We prove that the following proposition ("injective pairing") is inconsistent with univalence: *)
```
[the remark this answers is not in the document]
DanGrayson: How did this get resolved?

UniMath/MoreFoundations/NoInjectivePairing.v
DanGrayson: The two sides are judgmentally equal:
```
Lemma funcomp_id_l {A B} (f : A → B) :
  idfun B ∘ f = f.
Proof.
  reflexivity.
Defined.
```

UniMath/MoreFoundations/NoInjectivePairing.v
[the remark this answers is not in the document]
DanGrayson: You're right that rewrite uses a lemma that uses match:
```
internal_paths_rew = 
λ (A : UU) (a : A) (P : A → Type) (f : P a) (y : A) (p : a = y),
match p in (_ = y0) return (P y0) with
| idpath _ => f
end
     : ∏ (A : UU) (a : A) (P : A → Type), P a → ∏ y : A, a = y → P y
```
And you're right that it's good to avoid `rewrite`.  My main complaints about it are that it can be slow and fragile.

However, `rewrite`  is being used by many authors, as we haven't put a ban on it into our style guide.

UniMath/MoreFoundations/NoInjectivePairing.v
DanGrayson: Well, we don't have a ban on it, but I think proofs using them look cluttered.  Others may have different opinions.

UniMath/MoreFoundations/NoInjectivePairing.v
JLimperg: Ah, eta to the rescue. I've removed the lemma; thanks for pointing this out.

UniMath/MoreFoundations/NoInjectivePairing.v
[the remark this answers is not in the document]
JLimperg: I changed "proposition" to "lemma" in the comment. Is there a good word for "statement which looks like it should be true but actually isn't"?

UniMath/MoreFoundations/NoInjectivePairing.v
JLimperg: Fair enough. I've adopted your proof (minus the `change ...`).

UniMath/MoreFoundations/NoInjectivePairing.v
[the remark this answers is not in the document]
JLimperg: Okay, I've rewritten your proof to avoid `rewrite`. It now involves a little more plumbing, but if that makes it more maintainable, that's probably a good tradeoff.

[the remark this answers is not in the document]
benediktahrens: @JLimperg : I am confused: the proof currently still uses `rewrite`. Have you pushed your new proof yet?

benediktahrens:
On 21/04/2019 17:49, Jannis Limperg wrote:
> *@JLimperg* commented on this pull request.
> 
> 
> In UniMath/MoreFoundations/NoInjectivePairing.v 
> <https://github.com/UniMath/UniMath/pull/1204#discussion_r277174621>:
> 
>> @@ -0,0 +1,85 @@
> +Require Import UniMath.Foundations.All.
> +Require Import UniMath.MoreFoundations.Univalence.
> +
> +(** We prove that the following proposition ("injective pairing") is inconsistent with univalence: *)
> 
> I changed "proposition" to "lemma" in the comment. Is there a good word 
> for "statement which looks like it should be true but actually isn't"?

"statement" is neutral - it can be provable or not. It would be an 
option here.

UniMath/MoreFoundations/NoInjectivePairing.v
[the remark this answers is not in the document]
JLimperg: No you're right, I forgot to push the new proof.

UniMath/MoreFoundations/NoInjectivePairing.v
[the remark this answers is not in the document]
DanGrayson: Being a proposition means having any two elements equal - that's what I was talking about.  And the word "lemma" is no better, since the type asserted to be inhabited by a lemma should be proposition.  Probably you want say something like: there is no way to construction a function of the following type.

PR #1208, opened 2019-04-13
working on the package on monoidal categories

rmatthes:
This is not yet a request for being pulled but rather an announcement of ongoing work.

The code conforms better with the style guide now, but the main change is the use of the library constructions of functors and natural transformations, so that the "hand coding" of those can be avoided. In particular, this makes lengthy proofs of functoriality or naturality unnecessary. This latter part is far from completion, but it can be done incrementally. 
A good example is `U_action_ρ_is_nat_trans` that could be deleted from `Actions.v` - at the price of an upstream lemma, but that one could be useful in the future.

### DIFF
--- a/UniMath/CategoryTheory/Monoidal/Actions.v
+++ b/UniMath/CategoryTheory/Monoidal/Actions.v
@@ -34,96 +34,21 @@ Section Actions_Definition.
 
 Section Actions_Natural_Transformations.
 
-Context (A : precategory) (odot : functor (precategory_binproduct A V) A).
+Context {A : precategory} (odot : functor (precategory_binproduct A V) A).
 
 Notation "X ⊙ Y" := (odot (X , Y)) (at level 31).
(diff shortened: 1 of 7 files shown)


benediktahrens:
This is great, thanks for starting this cleaning effort!

rmatthes:
I have completed the first phase of what I intended to do with monoidal categories: make the code cleaner and shorter (by taking more advantage of the library - that had to be expanded a little bit).
In conclusion: 12 definitions of functors and 2 definitions of natural transformations treated, overall efforts on the code (bullets, use of red and cbn, line breaks).
Please check that `strength` in file `Strengths.v` is now more appropriate. Please also check if the upstream material in file `PrecategoryBinProduct.v` did not introduce duplication of code or simply follows from existing material by way of instantiation.

Now, I would consider this as a proper pull request.

rmatthes:
The merged material has strictly nothing to do with my PR.

PR #1211, opened 2019-04-19
Define the syntax of propositional logic

siddharthist:


UniMath/CategoryTheory/Inductives/PropositionalLogic.v
```diff
@@ -0,0 +1,191 @@
...
+  - assumption. (* vs *)
+  - exact not.
+  - apply (invweq (weqfunfromdirprod _ _ _)); exact and.
+  - apply (invweq (weqfunfromdirprod _ _ _)); exact or.
+  - apply (invweq (weqfunfromdirprod _ _ _)); exact impl.
+Defined.
+
+(** The fold, or catamorphism: given the same structure of operations on any
+    other set, we can construct an interpretation of PL in that set. *)
+(** TODO: why are the pr1s necessary? *)
```
benediktahrens: If you say "X : hSet" instead of "X : HSET", then the "pr1" are not necessary. This is because there is no beta reduction happening before the coercion is inserted.

siddharthist:
I split the material on booleans into a file in `MoreFoundations`, and removed the unnecessary `pr1`s

UniMath/MoreFoundations/Bool.v
```diff
@@ -0,0 +1,37 @@
+(** * Booleans *)
+
+Require Import UniMath.Foundations.Init.
+Require Import UniMath.Foundations.Sets.
+
+Definition bool_set : hSet := hSetpair _ isasetbool.
```
DanGrayson: This is already present as
```
Definition boolset : hSet := hSetpair bool isasetbool.
```

UniMath/MoreFoundations/Bool.v
```diff
@@ -0,0 +1,37 @@
...
+  intros b1 b2; induction b1; [exact true|exact b2].
+Defined.
+
+Definition implb : bool -> bool -> bool.
+Proof.
+  intros b1 b2; induction b1; [exact b2|exact true].
+Defined.
+
+(* Double check they have the right truth tables: *)
+Local Definition andbtt1 : andb true true = true := (idpath _).
```
DanGrayson: These 12 names are not needed, as here you are just doing tests.  It would be better to formulate them using "Goal" and putting them in a new file called "Tests.v".

siddharthist: @DanGrayson Thanks for the tip! Will do.

siddharthist:
I addressed Dan's comments, but I tried to add some computation lemmas and the induction principle and ran into performance problems. Any tips @mortberg? It looks like you may have run into similar issues with the list functor.

benediktahrens:
I'll merge this now, but this should not discourage @mortberg from commenting...

mortberg:
> I addressed Dan's comments, but I tried to add some computation lemmas and the induction principle and ran into performance problems. Any tips @mortberg? It looks like you may have run into similar issues with the list functor.

I think I made something Opaque and that solved the problem. I really don't remember the details, but if you step through my file and uncomment Opaque/Qed then you might see which definitions makes thing slow.

mortberg:
@siddharthist : I even left some comments in my file:

https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/Inductives/Lists.v#L124

I think "to typecheck" means "to typecheck reasonably fast" here:

https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/Inductives/Lists.v#L146

So try analogous things for your file and hopefully that will make things faster.

siddharthist:
@mortberg Yeah I saw those and attempted a few things. I'll keep trying. Thanks!

PR #1212, opened 2019-04-21
Proposal for a new package: Books

langston-barrett:
See the README for motivation and details. I have just created a skeleton here, but I'm more than happy to fill in a bunch more results if everyone thinks this is a good idea. 

This has a precedent in [the HoTT/HoTT library](https://github.com/HoTT/HoTT): https://github.com/HoTT/HoTT/blob/master/contrib/HoTTBook.v


UniMath/Books/Topoi.v
```diff
@@ -0,0 +1,27 @@
+(** * Topoi: The Categorical Analysis of Logic - Robert Goldblatt *)
```
maggesi:
```suggestion
(** * Topoi: The Categorial Analysis of Logic - Robert Goldblatt *)
```

siddharthist: @maggesi Thanks!

co-dan:
I think the chapter list from Category Theory in Context has ended up in the Topoi.v file as well.

siddharthist:
Thats true. Again, this PR is just a skeleton , a suggestion. I'll redo it with a lot more content if folks are interested in the idea.

UniMath/Books/README.md
```diff
@@ -0,0 +1,24 @@
+# Books
+
+UniMath is a vast library. This package is a navigational and organizational aid,
+its aim is to establish correspondences between the results in UniMath and those
+presented in some familiar textbooks.
+
+The names serve as a guide only, they are not intended to be referenced in other
```
DanGrayson: If the names are not intended to be referenced, then why have them?

DanGrayson: Maybe they should all be local.  (Some are, some are not.)

siddharthist: Yes, I think they should all be local. They are intended to be indicative of where the result is from in the text (see the examples). An alternative would be to have them in comments, but then we don't get automatic checking for duplicate entries as we do with local definitions.

siddharthist: More generally, the point of having these listed even though they aren't to be used directly in UniMath is to have them as a navigational aid: a third option besides `Search` and grepping through files for finding the names of results you need.

DanGrayson: Perhaps there is a way to put the reference to "Definition 1.1.9" in a comment instead of in an identifier.  To do that, we'd have to find another way to ensure that @CategoryTheory.Core.Isos.iso exists...  I don't know.

One problem is that nothing ensures the definition of @CategoryTheory.Core.Isos.iso doesn't changed later on, so that it's no longer identical to the thing in the book.

siddharthist: @DanGrayson That's true, I hadn't thought of that. However, since most of these books are based in set theory, the comparison is already somewhat of a metaphor. I'd be quite surprised if the definition changed to something conceptually unrelated.

Issue #687, opened 2017-04-18
Uniformity


vladimirias:
We need uniformity in naming. It is crucial.

1. Now we have four different ways of forming a name from two (or more) words. Let's say monoidfun can be named as:

monoidfun
monoid_fun
MonoidFun
Monoid_Fun

we *must* for the integrity of the library choose one of these four and apply it to *all* names in the library.

Those who want to preserve there own preferred way of doing it if it differs from the UniMath one will have to develop his own own UniMath-based library in his own GitHub account.

I suggest that we avoid capitalization as it makes it difficult to remember a composite name as one word and choose one of the first two.

DanGrayson:
This is a good place for me to advertise my past proposal: #176.

An advantage of capitalization is that it makes it possible to separate an identifier into words, especially where the words are unrecognizable to English speakers, such as in "coconustot".

I think we should also be uniform in *never* including arbitrary parameter names in identifier names, such as the "t" in coconustot, since there is no way to guess whether it is coconustot or coconustox or coconustoy, etc.

By the way, why is there "coconustot" but not "conustot"?

siddharthist:
@DanGrayson I like your proposal, and think it should be implemented whenever possible. I also think that UniMath should have a Travis script that checks for conformity. Would you be willing to write up these guidelines in `STYLE.md` so future contributors can refer to it? Or is there some level of consensus that should be reached first?

peterlefanulumsdaine:
I agree it would be good to have a guideline for this.  General conventions on identifier naming (as discussed in Dan’s #176) would be great, but are very hard to design well — but “how to concatenate words” is a specific comparatively simple one that’s easy to decide on.

Following @DanGrayson above, I would argue strongly against the form `monoidfun` — i.e. concatenation with *no* marker of word boundaries.  These are really just unnecessarily difficult to read.

Between the rest, my preference isn’t so strong.  But overall, I vote for the form `monoid_fun` — i.e. always use underscores for concatenation of words, without usually capitalising them.  This seems visually clear, plus it leaves capitalisation free to be either avoided entirely, or used for some other distinction.  (E.g. a convention like “plain words are never capitalised; acronyms and single-letter names may be capitalised iff they would be in traditional usage”.)

[...]

PR #1215, opened 2019-04-22
uniform use of underscores

DanGrayson:
see https://github.com/UniMath/UniMath/issues/687#issuecomment-485281491

### DIFF
--- a/UniMath/README.md
+++ b/UniMath/README.md
@@ -84,6 +84,8 @@ semantics is more likely to be rigorously verifiable and portable to new proof
 checking systems, and we follow a style of coding designed to render proofs
 less fragile and to make the files have a more uniform and pleasing appearance.
 
+* Constructing identifiers:
+    * form the identifier by concatenating English words or existing identifiers, separating them by underscores
 * Do not use `Admitted` or introduce new axioms.
 * Do not use `apply` with a term that needs no additional arguments filled in,
   because using `exact` would be clearer.


siddharthist:
What do we think about changing some existing identifiers to match? I'm working on a parser for (some subset of) the Coq grammar that should allow us to do some batch renaming if we're interested.

DanGrayson:
Emacs'  tag facility makes replacement easy:
```
M-R runs the command tags-query-replace (found in global-map), which is an
interactive autoloaded compiled Lisp function in ‘etags.el’.

It is bound to M-R, <menu-bar> <edit> <replace> <tags-repl>.

(tags-query-replace FROM TO &optional DELIMITED FILE-LIST-FORM)

Do ‘query-replace-regexp’ of FROM with TO on all files listed in tags table.
Third arg DELIMITED (prefix arg) means replace only word-delimited matches.
If you exit (M-x keyboard-quit, RET or q), you can resume the query replace
with the command M-,.
Fourth arg FILE-LIST-FORM non-nil means initialize the replacement loop.

If FILE-LIST-FORM is non-nil, it is a form to evaluate to
produce the list of files to search.
```

Anything in `Foundations` should be left to me.

siddharthist:
@DanGrayson Let's make an issue after this PR is merged to consider changing names in Foundations to be consistent with this schema. 

It's amazing what Emacs can do!

DanGrayson:
That might be premature, since there are other things to consider when choosing a name, such as capitalization.

Issue #1096, opened 2018-11-26
Unify names of builder functions


benediktahrens:
We have different ways to name builder functions, e.g., there are `makecategory` and `mk_precategory`. 
We should have one way, e.g., always use prefix `mk_`.

DanGrayson:
Let's not implement a change to the code without first agreeing on what it should be.  It should be part of a comprehensive proposal about how identifiers should be named, as discussed in #687 and in #176.

siddharthist:
Don't forget the the `*pair` names, like `weqpair` and `iscontrpair`.

I actually think that this can be done separately from a more comprehensive overhaul. If the builder/constructor names are consistent with one another, they will be easy to make consistent with whatever overall scheme we adopt (i.e. with a script).

To that end, I suggest a naming scheme for builders in particular following the `no_caps_with_underscores`, and beginning with `make_`. But I would prefer to see any uniformity adopted rather than bikeshed about which naming convention is best.

benediktahrens:
+1 for `make_`
+1 for fixing this issue as soon as possible

siddharthist:
The proposal here is also in line with what we've decided on #1215.

arnoudvanderleer:
@benediktahrens Has this been fully resolved with Marco's PR?

arnoudvanderleer:
For now I'll assume this has been resolved. If not, you may reopen it

PR #1216, opened 2019-04-23
Uniform use of prefix 'make_' in builders names.

maggesi:
Address #1096.

Caveat: This is an "aggressive" commit that changes most of the files in the repository.
(The reason is that I obtained it with "semiautomatic" manipulation using bash and emacs.)

NB: Once this commit is merged, it might be wise to revise the identifiers (not too many) that match the regex
`make[^_] `
or
`[a-zA-Z_]make`
etc.


DanGrayson:
lgtm

The only travis error is the check for changing the Foundations package.  The change there is acceptable.

siddharthist:
Since this touches basically everything in UniMath, it will probably conflict with every other PR. So, we might want to merge it before anything else to prevent @maggesi from doing too much more work :smile:

benediktahrens:
@maggesi : would you be willing to also work your magic on UniMath/TypeTheory and UniMath/largecatmodules?

PR #1217, opened 2019-04-24
general work towards a more abstract description of (semantic) signatures

rmatthes:
All of this is currently only preparatory work. The commits of April 18 are of hygienic nature. The commit of April 19 provides auxiliary results for all monoidal categories. The commit of April 22 stresses that (semantic) signatures need not refer to endofunctors (neither be higher-order endofunctors nor work on endofunctors). The first commit of April 24 does harmless work: the endofunctors for a given category give rise to a monoidal category.

I am asking for help on the second commit of April 24 that should be completely independent of all the previous ones: I do the straightforward proof that by fixing an object of a bicategory, the endomorphisms constitute a monoidal category. The proofs work well, but the final "Defined" takes forever. I have already avoided the use of "rewrite" in the proof, but without effect.

The PR is anyway not yet a sufficient contribution on signatures - as it will be meant to be.


benediktahrens:
On 24/04/2019 21:40, Ralph Matthes wrote:
> All of this is currently only preparatory work. The commits of April 18 
> are of hygienic nature. The commit of April 19 provides auxiliary 
> results for all monoidal categories. The commit of April 22 stresses 
> that (semantic) signatures need not refer to endofunctors (neither be 
> higher-order endofunctors nor work on endofunctors). The first commit of 
> April 24 does harmless work: the endofunctors for a given category give 
> rise to a monoidal category.
> 
> I am asking for help on the second commit of April 24 that should be 
> completely independent of all the previous ones: I do the 
> straightforward proof that by fixing an object of a bicategory, the 
> endomorphisms constitute a monoidal category. The proofs work well, but 
> the final "Defined" takes forever. I have already avoided the use of 
> "rewrite" in the proof, but without effect.

I think it is a problem that `mk_is_precategory` does not `Qed.` its 
arguments.

(The reason for this presumably is that some people think it is 
desirable to have C^op^op convertibly equal to C.)

One could imagine two variants of `mk_is_precategory`, one that is 
transparent, and one that is opaque (i.e., closed with `Qed.`).

The opaque one should be the default, and the other could be called 
`mk_is_precategory_transparent`.

I then recommend the use of the opaque version in most situations.


Furthermore, I suggest making the components of `is_functor` in 
`tensor_from_prebicat_and_ob` opaque, by using (making?) a suitable 
constructor.


You could also try deleting all `cbn` and `red` in 
`monoidal_precat_from_prebicat_and_ob` and check if that helps.


Finally, the proof of `monoidal_precat_from_prebicat_and_ob` is far too 
long. Some subgoals should be proved separately, closed with Qed, and 
then be referred to in this proof.

rmatthes:
Thanks a lot for these hints. By the way, compliation does terminate - here is the data found in the file `.BicategoryFromMonoidal.aux` on my computer:

``0 0 VernacProof "tac:no using:no"
1649 1657 proof_build_time "0.007"
1649 1657 proof_check_time "0.001"
0 0 VernacProof "tac:no using:no"
2487 2495 proof_build_time "0.060"
2487 2495 proof_check_time "0.064"
0 0 VernacProof "tac:no using:no"
7769 7777 proof_build_time "1.711"
7769 7777 proof_check_time "10.239"
0 0 VernacProof "tac:no using:no"
8840 8848 proof_build_time "0.038"
8840 8848 proof_check_time "0.015"
0 0 VernacProof "tac:no using:no"
9428 9436 proof_build_time "0.019"
9428 9436 proof_check_time "0.016"
0 0 VernacProof "tac:no using:no"
12198 12206 proof_build_time "0.561"
12198 12206 proof_check_time "1936.336"
0 0 vo_compile_time "1951.401"``

rmatthes:
already down at less than 6 min compilation time w/o working on `mk_is_precategory`
will continue

rmatthes:
below 3min compilation time for the problematic file, but still very unsatisfying

mortberg:
Just to repeat what Benedikt said: make as much as you can opaque! 

A good rule of thumb is that you should keep structure transparent and properties opaque when formalizing category theory. This is quite easy for many of the categorical structures in UniMath as they are often organized as a Sigma type where the first component is the operations and the second are their properties. So instead of building the structure interactively in one go just factor it into two lemmas. The first of these define the operations which you end with Defined and the second proves the properties of the operations and you end it with Qed. This way the terms are kept small and Coq doesn't have to unfold gigantic proof terms to compare things during conversion and unification.

UniMath/CategoryTheory/Monoidal/EndofunctorsMonoidal.v
```diff
@@ -84,22 +87,111 @@ Definition α_functor_inv (X : functor C D)(Y : functor D E)(Z : functor E F) :
...
+
+Section Endofunctors_as_monoidal_category.
+
+  Context {C : category}.
+  Variable hs : has_homsets C.
+
+(** The category of endofunctors on [C] *)
+Local Notation "'EndC'":= ([C, C, hs]) .
+
+Definition monoidal_precat_of_endofunctors: monoidal_precat.
```
mortberg: I bet large parts of this could be made opaque. Try running `Eval cbn in monoidal_precat_of_endofunctors` before and after you do this. I conjecture that the difference in size will be huge.

UniMath/CategoryTheory/Monoidal/MonoidalCategories.v
```diff
@@ -150,3 +158,86 @@ Definition strict_monoidal_precat : UU :=
...
+Section swapped_tensor.
+
+  Context (M : monoidal_precat).
+
+  Let C := monoidal_precat_precat M.
+  Let tensor := monoidal_precat_tensor M.
+
+Definition swapping_of_tensor: C ⊠ C ⟶ C := functor_composite binswap_pair_functor tensor.
+
+Definition swapping_of_monoidal_precat: monoidal_precat.
```
mortberg: This one could also be refactored and opacified as appropriately.

UniMath/CategoryTheory/Monoidal/BicategoryFromMonoidal.v
```diff
@@ -193,3 +194,177 @@ Definition prebicat_from_monoidal : prebicat :=
...
+*)
+
+Local Definition build_associator: associator tensor.
+Proof.
+  use mk_nat_iso.
+  - exact nat_trans_associator.
+  - intro c; apply is_iso_rassociator.
+Defined.
+
+Definition monoidal_precat_from_prebicat_and_ob: monoidal_precat.
```
mortberg: I would refactor this one and make parts of it opaque as well

mortberg:
I conjecture that if you refactor the 3 constructions of `monoidal_precat` and make as much as possible of them opaque then this will be a lot faster.

rmatthes:
I see that I really have to opacify everything by hand, see also my comment on #1218. Thanks for your help.

rmatthes:
I destroyed my branch with interactive rebasing and a forced push. Therefore, I am closing this PR and reopen with the one I had in mind.

PR #1221, opened 2019-04-28
Displayed precategories, the displayed precategory of categories

langston-barrett:
@benediktahrens I commented out two proofs that I wasn't able to fix, any ideas on what to do?

### DIFF
--- a/UniMath/CategoryTheory/DisplayedCats/Adjunctions.v
+++ b/UniMath/CategoryTheory/DisplayedCats/Adjunctions.v
@@ -168,7 +168,7 @@ Section DispHomSetIso_from_Adjunction.
     rewrite triangle1.
     unfold transportb.
     rewrite mor_disp_transportf_postwhisker.
-    rewrite id_left_disp.
+    rewrite (@id_left_disp C' D').
     unfold transportb.
     rewrite 2 transport_f_f.
     intermediate_path (transportf _ (idpath _) beta).
@@ -193,7 +193,7 @@ Section DispHomSetIso_from_Adjunction.
     rewrite triangle2.
     unfold transportb.
     rewrite mor_disp_transportf_prewhisker.
-    rewrite id_right_disp.
+    rewrite (@id_right_disp C D).
     unfold transportb.
     rewrite 2 transport_f_f.
     intermediate_path (transportf _ (idpath _ ) alpha).
@@ -264,6 +264,7 @@ End DispHomSetIso_from_Adjunction.
 
 (** * The right adjoint functor of a displayed adjunction preserves cartesian morphisms. *)
 
+(*
 Lemma right_over_adj_preserves_cartesianness : is_cartesian_disp_functor GG.
 Proof.
   unfold is_cartesian_disp_functor.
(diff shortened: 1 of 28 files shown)


UniMath/CategoryTheory/DisplayedCats/Adjunctions.v
```diff
@@ -287,11 +288,11 @@ Proof.
...
     + intro p.
-      set (equiv1 := homset_conj'_after_conj_inv _ h).
-      set (equiv2 := homset_conj'_after_conj_inv _ (gg;; # GG ff)).
+      set (equiv1 := homset_conj'_after_conj_inv d h).
+      set (equiv2 := homset_conj'_after_conj_inv d (gg;; # GG ff)).
       unfold homset_conj' in equiv1, equiv2.
       rewrite <- equiv1.
-      rewrite <- equiv2.
+      refine (!equiv2 @ _).
       rewrite eq2.
```
benediktahrens: Here, the `rewrite eq2.` fails. Instead,
```
      etrans. apply maponpaths. apply maponpaths. apply maponpaths. apply eq2.
```
works.

The way I tackle such issues is as follows: I make a second clone of the repository, and open this file in another instance of emacs. Stepping through the original version, I can see where the `rewrite` would act, and replace it with a more fine-grained, manual approach as above.

This should be a warning to everyone who thinks that using `rewrite` is a good idea.

siddharthist: @benediktahrens Thanks for the tips on debugging, I should make a second clone as you describe. The restriction on one Coq process per Emacs is a little frustrating...

DanGrayson: Rewrite is indeed fragile, but sometimes it seems to be the best approach.

benediktahrens:
On 29/04/2019 18:50, Daniel R. Grayson wrote:
> 
> 
> In UniMath/CategoryTheory/DisplayedCats/Adjunctions.v 
> <https://github.com/UniMath/UniMath/pull/1221#discussion_r279469375>:
> 
>>        unfold homset_conj' in equiv1, equiv2.
>         rewrite <- equiv1.
> -      rewrite <- equiv2.
> +      refine (!equiv2 @ _).
>         rewrite eq2.
> 
> Rewrite is indeed fragile, but sometimes it seems to be the best approach.

When?

DanGrayson:
Here, for example:

https://github.com/UniMath/UniMath/blob/d5e570ba972bca92ec6b9023e5166034b0d79654/UniMath/CategoryTheory/ExactCategories/ExactCategories.v#L239

peterlefanulumsdaine:
It looks to me like this PR has staled too much to be easily salvageable in its current form, unfortunately.

Certainly its changes overlap so much with subsequent development that a direct merge is very difficult — it would almost certainly be simpler to “port” it instead, rewriting these changes by hand starting from the current state of the library.  But it would also need some mathematical rethinking — it interacts with the change in bundling of `has_homsets` and the emphasis on categories rather than precategories.

So it would be nice to have this, but also quite a bit of work.  If anyone feels like working on it now, that would be great! Otherwise, should we leave this PR open to increase the chance someone finds it in future and is inspired to do this, or close it to reduce clutter in the outstanding PR’s?

benediktahrens:
I have a slight preference for leaving the PR open, but decorating it with a tag making explaining its status.

PR #1225, opened 2019-05-20
Make sure bound scopes are imported correctly.

maximedenes:
Scopes used for arguments depend on the bound scopes (in the sense of
`Bind Scope`) that have been imported (as in `Import`). A Coq bug was
making some non-imported bound scopes visibles, and one file of UniMath
was relying on this bug. We fix it here.

This is in preparation for https://github.com/coq/coq/pull/10181, but
should be backward compatible (remains to be tested).


benediktahrens:
Thanks a lot for preparing this PR, and for keeping UniMath compatible with Coq!

maximedenes:
@benediktahrens thanks, but was this tested already?

PR #1227, opened 2019-05-20
Characterization of path type of coproduct type via encode-decode

benediktahrens:
Not sure this should be in SyntheticHomotopyTheory; it could go into MoreFoundations as well.


DanGrayson:
I think it would be better in `MoreFoundations/PartA.v`, as it is basic.  And you could increase the number of final lemmas from 2 to 4.

DanGrayson:
There is a lot of overlap with this code from `Foundations/PartA.v`:
```
Definition equality_cases {P Q : UU} (x x' : P ⨿ Q) : UU.
Proof.
                           (* "codes" *)
  intros. induction x as [p|q].
  - induction x' as [p'|q'].
    + exact (p = p').
    + exact empty.
  - induction x' as [p'|q'].
    + exact empty.
    + exact (q = q').
Defined.

Definition equality_by_case {P Q : UU} {x x' : P ⨿ Q} :
  x = x'-> equality_cases x x'.
Proof.
  intros e. induction x as [p|q].
  - induction x' as [p'|q'].
    + simpl.
      exact (maponpaths (@coprod_rect P Q (λ _, P) (λ p, p) (λ _, p)) e).
    + simpl.
      exact (transportf (@coprod_rect P Q (λ _, UU) (λ _, unit) (λ _, empty))
                        e tt).
  - induction x' as [p'|q'].
    + simpl.
      exact (transportb (@coprod_rect P Q (λ _, UU) (λ _, unit) (λ _, empty))
                        e tt).
    + simpl.
      exact (maponpaths (@coprod_rect P Q (λ _,Q) (λ _, q) (λ q, q)) e).
Defined.

Definition inv_equality_by_case {P Q : UU} {x x' : P ⨿ Q} :
  equality_cases x x' -> x = x'.
Proof.
  intros e.
  induction x as [p|q].
  - induction x' as [p'|q'].
    + exact (maponpaths (@ii1 P Q) e).
    + induction e.
  - induction x' as [p'|q'].
    + induction e.
    + exact (maponpaths (@ii2 P Q) e).
Defined.

(* the same proof proves 4 lemmas: *)

Lemma ii1_injectivity {P Q : UU} (p p' : P) :
  ii1 (B := Q) p = ii1 (B := Q) p' -> p = p'.
Proof.
  exact equality_by_case.
Defined.

Lemma ii2_injectivity {P Q : UU} (q q' : Q) :
  ii2 (A := P) q = ii2 (A := P) q' -> q = q'.
Proof.
  exact equality_by_case.
Defined.

Lemma negpathsii1ii2 {X Y : UU} (x : X) (y : Y) : ii1 x != ii2 y.
Proof.
  exact equality_by_case.
Defined.

Lemma negpathsii2ii1 {X Y : UU} (x : X) (y : Y) : ii2 y != ii1 x.
Proof.
  exact equality_by_case.
Defined.
```

Could you take advantage of the existing code fruitfully?

benediktahrens:
> There is a lot of overlap with this code from |Foundations/PartA.v|:
>
> Could you take advantage of the existing code fruitfully?

Regarding the last four maps (two of which should not be called `Lemma`, by the way), the results above are more general.

Regarding the "equality by cases" constructions: I don't think I can use them. Specifically, from what I understand, it is an important feature of the encode-decode method to not rely on induction on the type under consideration, except for in the definition of some type family over that type.
On the other hand, all the "equality by cases" results do use such an induction.
I thus don't believe these results can/should be used in the encode-decode method. 

The opposite is true, of course: the existing code could fruitfully take advantage of the code above.

DanGrayson:
Hmm?  But your `decode` and `decode_encode` do use induction.

DanGrayson:
I just noticed my code can be simplified:
```
diff --git a/UniMath/Foundations/PartA.v b/UniMath/Foundations/PartA.v
index 6be38ee8..3225e009 100644
--- a/UniMath/Foundations/PartA.v
+++ b/UniMath/Foundations/PartA.v
@@ -2692,19 +2692,9 @@ Defined.
 Definition equality_by_case {P Q : UU} {x x' : P ⨿ Q} :
   x = x'-> equality_cases x x'.
 Proof.
-  intros e. induction x as [p|q].
-  - induction x' as [p'|q'].
-    + simpl.
-      exact (maponpaths (@coprod_rect P Q (λ _, P) (λ p, p) (λ _, p)) e).
-    + simpl.
-      exact (transportf (@coprod_rect P Q (λ _, UU) (λ _, unit) (λ _, empty))
-                        e tt).
-  - induction x' as [p'|q'].
-    + simpl.
-      exact (transportb (@coprod_rect P Q (λ _, UU) (λ _, unit) (λ _, empty))
-                        e tt).
-    + simpl.
-      exact (maponpaths (@coprod_rect P Q (λ _,Q) (λ _, q) (λ q, q)) e).
+  intros e. induction e. induction x as [p|q].
+  - apply idpath.
+  - apply idpath.
 Defined.
 
 Definition inv_equality_by_case {P Q : UU} {x x' : P ⨿ Q} :
@@ -2713,11 +2703,11 @@ Proof.
   intros e.
   induction x as [p|q].
   - induction x' as [p'|q'].
-    + exact (maponpaths (@ii1 P Q) e).
+    + induction e. reflexivity.
     + induction e.
   - induction x' as [p'|q'].
     + induction e.
-    + exact (maponpaths (@ii2 P Q) e).
+    + induction e. reflexivity.
 Defined.
 
 (* the same proof proves 4 lemmas: *)
```

benediktahrens:
On 30/05/2019 21:30, Daniel R. Grayson wrote:
> There is a lot of overlap with this code from |Foundations/PartA.v|:
> 
> Could you take advantage of the existing code fruitfully?

Regarding the last four maps (two of which should not be called `Lemma`, 
by the way), the results above are more general.

Regarding the "equality by cases" constructions: I don't think I can use 
them. Specifically, from what I understand, it is an important feature 
of the encode-decode method to not rely on induction on the type under 
consideration, except for in the definition of some type family over 
that type.
On the other hand, all the "equality by cases" results do use such an 
induction.
I thus don't believe these results can/should be used in the 
encode-decode method.

benediktahrens:
On 30/05/2019 22:13, Daniel R. Grayson wrote:
> I just noticed my code can be simplified:

It might be good to file a PR with these simplifications.

benediktahrens:
On 30/05/2019 21:30, Daniel R. Grayson wrote:
> 
> Could you take advantage of the existing code fruitfully?

I don't see how, but if you have a suggestion, I would be happy to see it.

DanGrayson:
> On 30/05/2019 21:30, Daniel R. Grayson wrote: Could you take advantage of the existing code fruitfully?
> I don't see how, but if you have a suggestion, I would be happy to see it.

What about this proof?:

```
Lemma coprod_paths {A B} (t u : A ⨿ B) : (t = u) ≃ equality_cases t u.
Proof.
  use weq_iso.
  - apply equality_by_case.
  - apply inv_equality_by_case.
  - intros []. induction t; now induction u.
  - induction t; induction u; now intros [].
Defined.
```

benediktahrens:
On 02/06/2019 15:00, Daniel R. Grayson wrote:
>     On 30/05/2019 21:30, Daniel R. Grayson wrote: Could you take
>     advantage of the existing code fruitfully?
>     I don't see how, but if you have a suggestion, I would be happy to
>     see it.
> 
> What about this proof?:
> 
> |Lemma coprod_paths {A B} (t u : A ⨿ B) : (t = u) ≃ equality_cases t u. 
> Proof. use weq_iso. - apply equality_by_case. - apply 
> inv_equality_by_case. - intros []. induction t; now induction u. - 
> induction t; induction u; now intros []. Defined. |
> 
> —

My goal here was to demonstrate the encode-decode method, not to be 
economical. This is why I suggested putting it in 
SyntheticHomotopyTheory in the first place.

DanGrayson:
Isn't my proof also using the encode-decode method?

Why should encode-decode proofs go into SyntheticHomotopyTheory?  I would think that the statement of the theorem would determine the package it goes into, not the proof.

benediktahrens:
On 02/06/2019 16:13, Daniel R. Grayson wrote:
> Isn't my proof also using the encode-decode method?
> 
> Why should encode-decode proofs go into SyntheticHomotopyTheory? 

Because the same recipe can be used to characterize the loop space of 
the circle and other non-sets.

DanGrayson:
The method of proof is used many times already in the Foundations.  For an example, consider the proof that nat is a set.  It goes by positing a family of types to serve as the family of path spaces.  So we can't use the method of proof as a way to separate foundational material from synthetic homotopy theory.

peterlefanulumsdaine:
This would be nice to merge in some form, since (as far as I can find) we still don’t have the full characterisation of path spaces of coproducts anywhere.  It feels like it got bogged down because of a tension between two different motivations for the material:

1. provide a characterisation of paths in coproduct types in the library
2. pedagogical: illustrate a simple (but still non-trivial) case of the encode-decode method, as a warmup for more advanced cases

(1) seems the more important use-case, and motivates @DanGrayson’s comments above — putting this in MoreFoundations, and simplifying proofs.

But (2) seems worthwhile too, and motivates keeping proofs self-contained as @benediktahrens was suggesting, and putting this in SyntheticHomotopyTheory.

I propose:

- (definitely): add this material in MoreFoudations, with proofs simplified as @DanGrayson suggests, and lemma names changed to something more global-friendly (eg `coprod_paths_code` not just `code`)
- (possibly, if others support this): add another version of this material in SyntheticHomotopyTheory, with proofs self-contained and written to illustrate the encode-decode method, and this pedagogical intention clearly documented, perhaps with results also marked as `Local` to discourage importing them downstream

benediktahrens:
I had produced these constructions for a talk on UF. (In the end, I did not even get to that part in my talk...) I don't have a strong opinion (any more) on what should happen with this material.

benediktahrens:
lgtm - thanks, @peterlefanulumsdaine !

peterlefanulumsdaine:
I think this is now complete and ready to merge (I should have marked Friday’s version as draft, sorry).

m-lindgren (approved):

PR #1230, opened 2019-05-22
Created Normal Subgroups section.

olynch:
This is my first time writing any serious formalization in Coq, if you could give me code feedback that would be great! Also, I hope that this is useful.

I think I want to next write some stuff up about the normal subgroup generated by a set, in order to do presentations by generators and relations in a more ergonomic way, thoughts on that are also welcome.

### DIFF
--- a/UniMath/Algebra/Groups.v
+++ b/UniMath/Algebra/Groups.v
@@ -10,6 +10,7 @@
   - Subobjects
   - Quotient objects
   - Cosets
+  - Normal Subgroups
   - Direct products
  - Abelian groups
   - Basic definitions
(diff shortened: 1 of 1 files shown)


UniMath/Algebra/Groups.v
```diff
@@ -569,6 +570,105 @@ Section GrCosets.
   Defined.
 End GrCosets.
 
+
+(** *** Normal Subgroups *)
+
+Section NormalSubGroups.
+  Local Open Scope multmonoid.
+
+  Definition isnormalsubgr {X : gr} (N : subgr X) : UU := ∏ g : X, ∏ n1 : N, ∑ n2 : N, (g * (pr1 n1)) * (grinv X g) = pr1 n2.
```
siddharthist: Is this a proposition (`isaprop`)? If so, it would be good to provide that proof and package this as an `hProp` (try grepping for `use make_hProp` for examples).

UniMath/Algebra/Groups.v
```diff
@@ -569,6 +570,105 @@ Section GrCosets.
...
+  Definition lcoset_equal_rcoset {X : gr} (N : subgr X) : UU := lcoset_in_rcoset N × rcoset_in_lcoset N.
+
+  Lemma lcoset_equal_rcoset_impl_normal {X : gr} (N : subgr X) : lcoset_equal_rcoset N -> isnormalsubgr N.
+  Proof.
+    intros pf.
+    unfold isnormalsubgr.
+    intros g n1.
+    set (pfcond := (pr1 pf) g n1).
+    set (n2 := pr1 pfcond).
+    split with n2.
```
siddharthist: We generally use bullets when a tactic introduces more than 1 subgoal:
```coq
split.
- solve_first_goal.
- split.
  * solve_second_goal_pt1.
  * solve_second_goal_pt2.
```
and so on.

siddharthist:
Thanks for your contribution! This definitely seems like useful material to have in UniMath.

UniMath/Algebra/Groups.v
DanGrayson: What's the point of introducing n2 in this definition?  Why not just check directly whether `g*n1*g^-1` is an element of N?  Just "apply" N to the expression.

DanGrayson: Using `make_hProp` would be slow way to make an hProp.  Better would be to use the operators that make new hProps from old, especially `∀`.

UniMath/Algebra/Groups.v
```diff
@@ -589,42 +602,43 @@ Section NormalSubGroups.
...
-    set (pfcond := (pr1 pf) g n1).
-    set (n2 := pr1 pfcond).
-    split with n2.
-    apply (grrcan X g).
-    rewrite (assocax _ _ _ _).
-    rewrite (grlinvax X _).
-    rewrite (runax X).
-    exact (pr2 pfcond).
+    induction ((pr1 pf) g n1) as [n2 eq_n2].
+    assert (g * pr1 n1 * (grinv X g) = pr1 n2).
```
siddharthist: It's best to give such hypotheses explicit names, like so:
```
    assert (my_hypothesis_name : g * pr1 n1 * (grinv X g) = pr1 n2).
```
This makes it easier to read your code without stepping through it, and doesn't rely on Coq not changing the auto-generated names.

UniMath/Algebra/Groups.v
```diff
@@ -589,42 +602,43 @@ Section NormalSubGroups.
...
       intros g n1.
-      set (pfcond := pf g n1).
-      set (n2 := pr1 pfcond).
-      split with n2.
-      apply (grrcan X (grinv X g)).
-      refine (_ @ !assocax _ _ _ _ ).
-      rewrite (grrinvax X _).
-      rewrite (runax X).
-      exact (pr2 pfcond).
+      use (tpair _).
```
siddharthist: Why not
```
use tpair.
```

UniMath/Algebra/Groups.v
```diff
@@ -641,32 +655,35 @@ Section NormalSubGroups.
...
-      set (pf2 := pr2 (normal_impl_lcoset_equal_rcoset (pr1 N) (pr2 N)) c n1).
-      set (n2 := pr1 pf2).
-      split with n2.
-      refine (assocax _ _ _ _ @ _).
-      refine (maponpaths (λ x, _ * x) (!pr2 pf2) @ _).
-      refine (!assocax _ _ _ _ @ _).
-      apply (maponpaths (λ x, x * c)).
-      exact (pr2 pf1).
+      use (tpair _).
+      + exact (pr1 (pr2 (normal_impl_lcoset_equal_rcoset (pr1 N) (pr2 N)) c (pr1 pf1))).
```
siddharthist: Usually, we have named "accessor functions" that we prefer to use rather than `pr1` and `pr2`. This makes the code clearer. (You should look near the definition of `subgr` to find some)

UniMath/Algebra/Groups.v
siddharthist: Yes, thanks for the pointer Dan!

olynch:
I made isnormalsubgr into a `hProp`, and fixed up the style of my proofs.

Couple questions:
It would be nice to have cosets as objects that I could work with, but I can't seem to find any code for generating the image of a subset under a function. I could write this if it doesn't already exist (maybe in MoreFoundations.Subtypes), but I thought that it was more likely that I was missing something because this seems very basic. I was thinking about using univalence to get a nice statement like `rcoset a H = lcoset a H`, or `lcoset a (rcoset (grinv _ a) H) = H`, do you think this is a good idea?

UniMath/Algebra/Groups.v
olynch: Damn you are fast :)

olynch: Yeah, and that's kind of a horrendous line, I agree. Would breaking it up into set statements also be a good idea?

siddharthist: It might be nice, but that's a bit more subjective. Up to you!

siddharthist:
@olynch Have you tried:
```
Require Import UniMath.Algebra.All.
Require Import UniMath.MoreFoundations.All.
Search hsubtype.
```
You can also do (for example)
```
Search hsubtype tpair.
```
to find lemmas mentioning `hsubtype` _and_ `tpair`. 

I think the lemmas you mention would be nice, but the application of univalence is the easy part :smile:. The hard (and interesting) part is to produce a weak equivalence (`≃`). Because of univalence, we know this is equivalent to the equality, but in general it's better to provide the equivalence (because this is usually the one with meaningful content, and the one we give a computationally meaningful proof of).

olynch:
My impression of the definition of weak equivalence for subtypes is that it should be pretty easily implied by my lcoset_equals_rcoset type, as long as lcoset and rcoset are suitably defined. I still can't find anything about image subtypes though (I searched for`hsubtype incl` and `hsubtype isincl`). Should I have at it in MoreFoundations.Subtypes?

siddharthist:
@DanGrayson or @benediktahrens might know if it's somewhere else... it can be hard to find things in such a huge library :)

I think that'd be a good place for it if you want to give it a go.

UniMath/Algebra/Groups.v
```diff
@@ -509,12 +510,25 @@ Section GrCosets.
 
   (** The property of being in the same coset defines an equivalence relation. *)
 
+  Definition in_same_left_coset_prop : X -> X -> hProp.
```
DanGrayson: This lemma is redundant, since the result is packaged up as part of `in_same_left_coset_eqrel`.

UniMath/Algebra/Groups.v
```diff
@@ -509,12 +510,25 @@ Section GrCosets.
...
+  Proof.
+    intros x1 x2.
+    use make_hProp.
+    + exact (in_same_left_coset Y x1 x2).
+    + apply isaprop_in_same_left_coset.
+  Defined.
+
+  Definition in_same_right_coset_prop : X -> X -> hProp.
```
DanGrayson: Maybe make this redundant, too, by creating `in_same_right_coset_eqrel`, analogous to and adjacent to `in_same_left_coset_eqrel`.

DanGrayson:
The appropriate notion of equality for subtypes is probably this one, from Subtypes.v: 
```
Definition subtype_equal {X:UU} (S T : hsubtype X) : hProp := ∀ x, S x ⇔ T x.

Notation " S ≡ T " := (subtype_equal S T) (at level 70) : subtype.
```

DanGrayson:
What's your definition of "weak equivalence for subtypes"?

olynch:
Yeah, that was the definition that I was thinking of.

UniMath/Algebra/Groups.v
peterlefanulumsdaine: This is fixed by 8626d3b.

peterlefanulumsdaine:
This PR provides nice material — as far as I can see it just got bogged down during review.  Looking at the review comments, it seems to me that the most important of the requested changes have been fixed.  @olynch If you’re still interested in working further on this to implement some of the further suggestions before it’s merged, say so here — otherwise I suggest we merge this soon (either as-is, or with a little further polishing, which I can do easily and quickly) so it’s not sitting around unmerged and staling any longer!

peterlefanulumsdaine:
OK, I’ve made a few of minor style/organisation tweaks, in particular replacing some `pr1`/`pr2` with access functions, which I think was the main remaining review comment.  @DanGrayson are you happy with this to merge now?

DanGrayson (approved):

PR #1233, opened 2019-06-05
Computational M-types

rmatthes:
a fruit of the UniMath School 2019: a streamlined version of the construction by Felix Rech of M-types (coinductive trees) that enjoy the equational rule of coiteration definitionally: `Lemma corec_computation` is proved merely by `idpath`. As input we take any final coalgebra for the respective polynomial functor (hence, with the provable coiteration rule), the output is the refined final coalgebra with the law holding definitionally (of course, both coalgebras are equal - provably). The proof tries to use as much as possible already existing results in the library. Only one lemma of a general nature is added to `MoreFoundations.PartA`.

The commit history contains also the development made by Scoccola and Rech at UniMath School 2017 and other stuff that is deleted later. For reviewing this PR, looking at the diff will be much easier than following the commit history.


UniMath/Induction/M/ComputationalM.v
```diff
@@ -0,0 +1,300 @@
...
+    use total2_paths_f; try apply carriers_eq.
+    apply funextfun. intro m0.
+    rewrite eq1. rewrite eq2. rewrite eq3.
+    cbn. unfold polynomial_functor_obj.
+    rewrite transportf_total2_const.
+    use total2_paths_f; try apply idpath.
+    cbn. apply funextsec. intros b. rewrite <- helper_A.
+    unfold carriers_eq. rewrite weqpath_transport.
+    cbn. rewrite eq_corecM0. apply idpath.
+  Qed.
```
benediktahrens: Would it be better to use `Defined.` here?

benediktahrens:
The code looks very good. I would only like to have confirmation that the many `Qed.` are indeed the right choice. @rmatthes @dominik-kirst , can you confirm?

rmatthes:
@benediktahrens : I cannot confirm that `Qed` was the right choice. Compilation time increased by very little with `Defined` (almost) everywhere. One would have to observe the impact in case studies using M-types.

benediktahrens:
@rmatthes : thanks a lot for your feedback. Why is the body of `corec_computation` opaque, even though it is `idpath`? Typically, working with an `idpath` is as pleasant as it gets, and thus hiding it in an opaque constant seems a shame. Or are you suggesting never using `corec_computation`?

rmatthes:
@benediktahrens : I thought that such equations are rarely used for rewriting since one would normally solve the goal at hand by `idpath`, again. But it is true that we do not need to discourage their use so brutally.

PR #1236, opened 2019-06-15
Extend a bit the theory of graphs.

maggesi:
- Introduces graphs (pregraphs whose nodes and edges form sets).
- Build the category of graphs.


siddharthist:
Note that the file [`colimits.v`](https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/limits/graphs/colimits.v#L36) also defines pregraphs. My preference would be to have the material on graphs upstreamed to `Combinatorics.Graphs`, which, oddly enough, currently has no explicit definition of a graph. Then this file and `colimits.v` could import it from `Combinatorics`.

maggesi:
> My preference would be to have the material on graphs upstreamed to Combinatorics.Graphs

@siddharthist: seems a reasonable proposal, indeed.  I will prepare a commit in that direction.

maggesi:
I did a minimal reorganisation.  I didn't touch colimits.v (because pregraphs are called graphs there, and because so far no relevant material from the theory of graph is used there, so I see no strong point in refactoring the code at the moment).

maggesi:
The definition of pregraph coincides with the one of precategory_ob_mor.
Indeed, some theorems can be shared between the twos.

maggesi:
I committed sub/coq by mistake.
Can someone help with this?
Which is the fastest way to get out of this situation?

benediktahrens:
On 17/06/2019 15:16, Marco Maggesi wrote:
> I committed sub/coq by mistake.
> Can someone help with this?
> Which is the fastest way to get out of this situation?

Looking at

https://github.com/UniMath/UniMath/pull/1236/files

I see three tiny dots on the right for each file.

Clicking on the dots, I see "Delete file".

Maybe doing this for "sub/coq" does the right thing?

DanGrayson:
Deleting it might delete it, which doesn't sound right.  Do this: on your computer, on your graph branch, look at the result from
```
git diff master...graph sub/coq
```
You'll see two commit numbers, the old number and the new number.  Use `(cd sub/coq ; git checkout XXXX)`, where XXXX is the old number, and then use `git commit -m 'fix sub/coq'` to check in the reversion.  After that, the first command above, when repeated, should give a null result.  Then use `git push` to push it to github.

maggesi:
@DanGrayson : Thanks, it worked.
Maybe we should save this explanation somewhere.

PR #1237, opened 2019-06-18
Fiber bicategory

nmvdw:
I finished the construction of the fiber bicategory. 

In the beginning of this file, I added some lemmas. I am unsure on where to put those.


nmvdw:
`COQC UniMath/CategoryTheory/Bicategories/DisplayedBicats/Fibration.v
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.`

I think the file `Fibration.v` took too long to build. I think I can speed up type checking by around 1 or 2 minutes. I will see whether that helps.

UniMath/CategoryTheory/Bicategories/DisplayedBicats/Fibration.v
```diff
@@ -18,8 +18,68 @@ Require Import UniMath.CategoryTheory.Bicategories.Bicategories.BicategoryLaws.
 Local Open Scope cat.
 Local Open Scope mor_disp_scope.
 
-Section LocalIsoFibration.
+Definition disp_lwhisker_vcomp_alt
```
benediktahrens: Maybe this lemma and the following could go into DispBicat.v ? 
In this way, the lemmas definitely show up in searches when working with displayed bicategories.

nmvdw:
I broke up `Fibration.v` into two files (with an export file), because in one file, type checking would take too long.

PR #1252, opened 2019-07-15
Avoid relying on `Export` bugs

maximedenes:
See https://github.com/coq/coq/issues/10474 and
https://github.com/coq/coq/issues/10480

This is in preparation for https://github.com/coq/coq/pull/10476 which fixes these two bugs, and is backward compatible (tested on Coq's master).

### DIFF
--- a/UniMath/Algebra/ConstructiveStructures.v
+++ b/UniMath/Algebra/ConstructiveStructures.v
@@ -3,9 +3,9 @@
 
 Unset Kernel Term Sharing.
 
-Require Export UniMath.Algebra.Apartness.
 Require Export UniMath.Algebra.DivisionRig.
 Require Export UniMath.Algebra.Domains_and_Fields.
+Require Export UniMath.Algebra.Apartness.
 
 Require Import UniMath.MoreFoundations.Tactics.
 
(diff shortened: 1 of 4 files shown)


benediktahrens:
@maximedenes : Thanks a lot for preparing this PR!

Could you briefly explain, e.g., why the fix in  UniMath/CategoryTheory/ExactCategories/ExactCategories.v is necessary, moving the 
```
Require ExportUniMath.MoreFoundations.Propositions.
```
to a later place? Does it have to do with some constant in that file (module) being shadowed by another one of the same name in another module?

DanGrayson:
Thanks, looks good to me.

maximedenes:
> Does it have to do with some constant in that file (module) being shadowed by another one of the same name in another module?

It is because there is an unfortunate clash of notations: `∃! x, P` is defined as `iscontr [...]` in `Fondations.PartA` and `iscontr_hProp [...]` in `MoreFoundations.Propositions`.

benediktahrens:
@maximedenes : Thanks for your explanation, and thanks again for your help in maintaining UniMath!

@DanGrayson : I have opened https://github.com/UniMath/UniMath/issues/1253 regarding the clash of notation.

PR #1257, opened 2019-07-22
Uniqueness of binitial objects and theory of bifinal objects.

maggesi:
- Prove uniqueness of binitial objects in univalent bicategories.
- Also provide the dual theory of bifinal objects from the theory of
  initial ones mutatis mutandis.


UniMath/CategoryTheory/Bicategories/Colimits/Final.v
```diff
@@ -0,0 +1,501 @@
...
+Require Import UniMath.CategoryTheory.Bicategories.Bicategories.Adjunctions.
+Require Import UniMath.CategoryTheory.Bicategories.Bicategories.AdjointUnique.
+Require Import UniMath.CategoryTheory.Bicategories.Bicategories.EquivToAdjequiv.
+Require Import UniMath.CategoryTheory.Bicategories.Bicategories.Examples.OneTypes.
+Require Import UniMath.CategoryTheory.Bicategories.Bicategories.Examples.BicatOfCats.
+Require Import UniMath.CategoryTheory.categories.StandardCategories.
+Require Import UniMath.CategoryTheory.Bicategories.Bicategories.Univalence.
+Require Import UniMath.CategoryTheory.Core.Univalence.
+Require Import UniMath.CategoryTheory.catiso.
+(* Require Import UniMath.CategoryTheory.Bicategories.Bicategories.Colimits.Inital. *)
```
benediktahrens:
```suggestion
```

maggesi:
I now realise that `isweq_isinj` is duplicated:
it is both in Bicategories/Colimits/Final.v and in Bicategories/Colimits/Initial.v.

UniMath/CategoryTheory/Bicategories/Colimits/Final.v
```diff
@@ -0,0 +1,500 @@
...
+  Definition bifinal_eq
+             {X : C}
+             (HX : is_bifinal X)
+             {Y : C}
+    : bifinal_eq_property X Y.
+  Proof.
+    intros f g α β.
+    pose (L := functor_to_unit (univ_hom C_is_univalent_2_1 Y X)).
+    pose (R := (left_adjoint_right_adjoint (HX Y) : functor _ _)).
+    assert (#R(#L α) = #R(#L β)) as p.
```
benediktahrens: Why is this reflexivity term important?

maggesi:
> I now realise that `isweq_isinj` is duplicated:
> it is both in Bicategories/Colimits/Final.v and in Bicategories/Colimits/Initial.v.

@benediktahrens : would be correct to move `isweq_isinj` at the end of MoreFoundations/PartA.v ?

benediktahrens:
> > I now realise that `isweq_isinj` is duplicated:
> > it is both in Bicategories/Colimits/Final.v and in Bicategories/Colimits/Initial.v.
> 
> @benediktahrens : would be correct to move `isweq_isinj` at the end of MoreFoundations/PartA.v ?

@maggesi : Yes, I think so.

DanGrayson:
> > I now realise that `isweq_isinj` is duplicated:
> > it is both in Bicategories/Colimits/Final.v and in Bicategories/Colimits/Initial.v.
> 
> @benediktahrens : would be correct to move `isweq_isinj` at the end of MoreFoundations/PartA.v ?

Probably not.  The proof is so simple that every use of the lemma could probably be replaced by a single use of `invmaponpathsincl`.  Try it and see.

Here is the simple proof:
```
Definition isweq_isinj {A B : UU} {f : A → B} : isweq f -> ∏ x y, f x = f y → x = y.
Proof.
  exact (invmaponpathsincl f).
Defined.
```

maggesi:
Use of `isweq_isinj` replaced with `invmaponpathsincl`.
(However, if I can say, a lemma containing the substring "inj" in its name would be much more easier to search and remember.)

UniMath/CategoryTheory/Bicategories/Colimits/Final.v
```diff
@@ -0,0 +1,485 @@
...
+  apply is_bifinal'_to_is_bifinal.
+  exact (HX₁ ,, λ Y, (HX₂ Y ,, HX₃ Y)).
+Defined.
+
+Lemma isofhlevel_unit n : isofhlevel n unit.
+  apply isofhlevelcontr.
+  exact iscontrunit.
+Qed.
+
+Definition isofhleve_unit n : isofhlevel n unit
```
benediktahrens: Typo in the name of this definition. Also, the same is proved just above.

benediktahrens:
What is the purpose of having the two definitions, `bifinal` and `bifinal'`? In what situation is one more convenient than the other?

nmvdw:
> What is the purpose of having the two definitions, `bifinal` and `bifinal'`? In what situation is one more convenient than the other?

They represent two ways of stating the universal property. `bifinal` says the category of morphisms is contractible while `bifinal'` gives a more explicit formulation, which is more convenient to use. On the other hand, it's straight forward to prove that `isBifinal` is a proposition.

UniMath/CategoryTheory/Bicategories/Colimits/Final.v
maggesi: > Typo in the name of this definition. Also, the same is proved just above.

Thanks. Should be fixed now.

PR #1267, opened 2019-07-24
alternative to isweqfibtototal

benediktahrens:
I am suggesting a much simpler (to me, at least) alternative to `isweqfibtototal`. 

It is my hope that this alternative has better computational behaviour, but I cannot back that up right now.

I suggest not removing `isweqfibtototal`.


DanGrayson:
Well, it certainly looks simpler, so let's try it.

Issue #1261, opened 2019-07-22
Travis out of memory when compiling CategoryTheory/Bicategories/DisplayedBicats/Fibration/Fibration2


benediktahrens:
Travis currently fails on master with
```
COQC UniMath/CategoryTheory/Bicategories/DisplayedBicats/Fibration/Fibration2.v
Fatal error: out of memory.
Command exited with non-zero status 2
``` 
On my machine, the file compiles fine.

DanGrayson:
Can you point to the test result on master?

DanGrayson:
Probably the only way to debug this at home is under Linux, since Mac OS X has no way of limiting memory to processes.

DanGrayson:
Pull request #1258 was merged, even though travis didn't pass it.

DanGrayson:
I have an Ubuntu machine here, so if you don't I could try to debug it.

benediktahrens:
On 23/07/2019 00:52, Daniel R. Grayson wrote:
> Pull request #1258 <https://github.com/UniMath/UniMath/pull/1258> was 
> merged, even though travis didn't pass it.

The failure is due to a limitation of Travis, not due to the code of the PR.

DanGrayson:
That may be, but we depend solely on travis for checking pull requests, so pull requests have to pass the test before being accepted.

benediktahrens:
I presume the culprit is `discrete_fiber_data_laws_lassociator_lassociator`, with high memory use at `Qed.` time. 
The easy fix would be to cut the proof in smaller pieces. This would be easy, since this is a proof of an equality, were it not for the `cbn.` at the beginning of the proof, which yields an unreadably large goal.

Any other ideas?

DanGrayson:
I don't have any ideas at all.  How do you know memory use is high at Qed time for that particular result?

benediktahrens:
On 23/07/2019 12:35, Daniel R. Grayson wrote:
> I don't have any ideas at all. How do you know memory use is high at Qed 
> time for that particular result?

Looking at `htop` when stepping through the file shows a peak in memory 
usage when doing the `Qed.`, and again when closing the section at the 
very end of the file.

Note that Travis has just succeeded compiling current master:
https://travis-ci.org/UniMath/UniMath/builds/562530181?utm_source=github_status&utm_medium=notification

This shows that it is not deterministic whether Travis fails or succeeds.

DanGrayson:
[...]

PR #1271, opened 2019-08-01
More work on displayed bicategories and biequivalence

maggesi:
- Fiber category of a displayed bicategory.
- Biequivalence between Monads and Ktriples.
- Biequivalence between 1-types and pointed groupoids.


UniMath/CategoryTheory/Bicategories/DisplayedBicats/DispBicat.v
```diff
@@ -1336,6 +1336,62 @@ Proof.
...
 
+(* ----------------------------------------------------------------------------------- *)
+(** ** Useful properties                                                               *)
+(* ----------------------------------------------------------------------------------- *)
+
+Definition disp_2cells_isaprop
+           {B : bicat} (D : disp_bicat B)
+  := ∏ (a b : B) (f g : a --> b) (x : f ==> g)
+       (aa : D a) (bb : D b) (ff : aa -->[f] bb) (gg : aa -->[g] bb),
+     isaprop (disp_2cells x ff gg).
```
benediktahrens: Is there a notation for `disp_2cells x ff gg` ?

UniMath/CategoryTheory/Bicategories/DisplayedBicats/DisplayToFiber.v
```diff
@@ -0,0 +1,392 @@
+(* ******************************************************************************* *)
+(** * Fiber category of a displayed bicategory
```
benediktahrens: This comment is misleading, if the displayed bicategory is not a general one, but one with special properties, such as locally groupoidal.

benediktahrens: Also, the file name should maybe reflect the conditions on the displayed bicategory.

UniMath/CategoryTheory/Bicategories/DisplayedBicats/DisplayToFiber.v
```diff
@@ -0,0 +1,392 @@
...
+Require Import UniMath.CategoryTheory.Bicategories.DisplayedBicats.DispTransformation.
+Require Import UniMath.CategoryTheory.Bicategories.DisplayedBicats.DispModification.
+Require Import UniMath.CategoryTheory.Bicategories.DisplayedBicats.DispBiequivalence.
+Require Import UniMath.CategoryTheory.Bicategories.Bicategories.BicategoryLaws.
+Require Import UniMath.CategoryTheory.Bicategories.Bicategories.Invertible_2cells.
+
+Local Open Scope cat.
+Local Open Scope mor_disp_scope.
+
+Definition transportb_transpose
```
benediktahrens: I do think that this result is somewhere in UniMath already. 
Otherwise, this construction should be moved to an appropriate spot, e.g., MoreFoundations/PartA.

UniMath/CategoryTheory/Bicategories/DisplayedBicats/DisplayToFiber.v
```diff
@@ -0,0 +1,392 @@
...
+           {a₁ a₂ : A}
+           {p : a₁ = a₂}
+           {y : Y a₁} {z : Y a₂}
+  : transportf Y p y = z → y = transportb Y p z.
+Proof.
+  induction p.
+  apply idfun.
+Defined.
+
+Definition transportb_disp_psfunctor
```
benediktahrens: This construction probably belongs into the file where disp_psfunctors are defined.

UniMath/CategoryTheory/Bicategories/DisplayedBicats/Examples/KleisliTriple.v
```diff
@@ -0,0 +1,446 @@
...
+  : bicat
+  := total_bicat kleisli_triple_disp_bicat.
+
+Definition disp_2cells_isaprop_kleisli
+  : disp_2cells_isaprop kleisli_triple_disp_bicat.
+Proof.
+  exact @disp_2cellsisaprop.
+Qed.
+
+(* Improve Qed is slow *)
```
benediktahrens: Is this comment still pertinent?

UniMath/CategoryTheory/Bicategories/DisplayedBicats/Fibration/Fibration1.v
```diff
@@ -1,2833 +0,0 @@
-(* ******************************************************************************* *)
```
benediktahrens: I am not sure I like seeing this file go away. What's wrong with it? 
Same for Fibration2, of course.

UniMath/CategoryTheory/Bicategories/DisplayedBicats/Fibration.v
```diff
@@ -1,6 +1,387 @@
 (* ******************************************************************************* *)
-(** Export file for fibrations
+(** * Fiber category of a displayed bicategory
```
benediktahrens: Misleading comment: here, one does not construct the fiber category of an arbitrary displayed bicategory.

UniMath/CategoryTheory/Bicategories/DisplayedBicats/Examples/ProdCommutativity.v
```diff
@@ -1,205 +0,0 @@
-(* ******************************************************************************* *)
```
benediktahrens: Why does this file get deleted?

benediktahrens:
Good stuff. I only have a few minor comments/questions.

UniMath/CategoryTheory/Bicategories/DisplayedBicats/DispBicat.v
nmvdw: Yes, but not at that point. They are defined in the module `Notations`, which are defined after it.

UniMath/CategoryTheory/Bicategories/DisplayedBicats/Fibration/Fibration1.v
nmvdw: I think the construction of the general fiber bicategory does not have much merit. It requires a lot of work, but it gives something which is too complicated to use.

If you want to go from displayed pseudofunctors to fiber pseudofunctors, then this will require a large amount of tedious equational reasoning since you need to prove the coherencies. It cannot be used to formulate local univalence of the total bicategory via local univalence of the base and the fibers, since local univalence talks about multiple fibers. For global univalence, again a large amount of tedious equational reasoning is required to do the necessary steps.

For this reason, I think the fiber category in the more specific case is more interesting. Also, note that all our examples satisfy these conditions.

UniMath/CategoryTheory/Bicategories/DisplayedBicats/Examples/ProdCommutativity.v
nmvdw: In the current version, only a displayed pseudofunctor is constructed. This is no enough to show commutativity. Showing that it gives rise to a displayed biequivalence, was too difficult due to the coherencies. We had a different version where we showed it gives rise to an isomorphism, but that one suffered from a very slow Qed.

UniMath/CategoryTheory/Bicategories/DisplayedBicats/DisplayToFiber.v
maggesi: The head is an instance of transportf_transpose, however it is not easy to use in practice (eg. cannot be used for rewriting).  So I think it is useful to add this. 

There is another variant which is
transportf_transpose_alt
in DispBicat.

In addition, they have different choice of implicit arguments.

I think we should write the four possible variants (transposef/transposeb on left/right) in a consistent way and put it in MoreFoundations/PartA.

UniMath/CategoryTheory/Bicategories/DisplayedBicats/Examples/KleisliTriple.v
maggesi: I would say so: it takes more than 20s on my machine.
But I'm afraid we also have worse than this elsewhere.

nmvdw: This comment was there to indicate that we need to put some effort in making the `Qed` faster.

UniMath/CategoryTheory/Bicategories/DisplayedBicats/Fibration/Fibration1.v
maggesi: As you know, Fibration2 is the file of issue #1261.

@DanGrayson, @nmvdw: do you agree to put it back?

- Reason to remove Fibration1+Fibration2: takes a lot of time and memory to compile, it is too general, we do not have examples for it.
- Reason to keep them: it is a nontrivial work that can be useful one day(?)

nmvdw: >Reason to keep them: it is a nontrivial work that can be useful one day(?)

We can also put it in another repository and retrieve it if it is needed. It is not necessary to keep it in UniMath.

UniMath/CategoryTheory/Bicategories/DisplayedBicats/Examples/ProdCommutativity.v
maggesi: It was intended as a simple elucidating example of biequivalence.
But, it turned out that it was not so simple.
Not being necessary, we didn't finish it (he had various `Admitted`).
So I think it is better to remove it for now.

UniMath/CategoryTheory/Bicategories/DisplayedBicats/DisplayToFiber.v
nmvdw: > Also, the file name should maybe reflect the conditions on the displayed bicategory.

Any suggestions in this regard? There are quite some conditions.

UniMath/CategoryTheory/Bicategories/DisplayedBicats/Examples/ProdCommutativity.v
DanGrayson: It doesn't use "Admitted" now, nor does any file in UniMath.

UniMath/CategoryTheory/Bicategories/DisplayedBicats/DisplayToFiber.v
benediktahrens: Yes, that sounds good!

UniMath/CategoryTheory/Bicategories/DisplayedBicats/DisplayToFiber.v
```diff
@@ -0,0 +1,281 @@
+(* ******************************************************************************* *)
+(** * Displayed biequivalences give rise to equivalences on the fiber.
+      This requires that the involved displayed bicategoies are propositions
```
benediktahrens:
```suggestion
      This requires that the involved displayed bicategoies have propositions
```

UniMath/CategoryTheory/Bicategories/DisplayedBicats/Fibration.v
```diff
@@ -1,6 +1,388 @@
 (* ******************************************************************************* *)
-(** Export file for fibrations
+(** * Fiber category of a displayed bicategory whose displayed 2-cells form a
+      proposition. In addition, the displayed needs to be locally univalent and
+      there needs to be a local isofibration on the 2-cells.
```
benediktahrens: Do you mean this:
```suggestion
      to be equipped with a local iso-cleaving.
```
?

UniMath/CategoryTheory/Bicategories/DisplayedBicats/Fibration.v
```diff
@@ -1,6 +1,388 @@
 (* ******************************************************************************* *)
-(** Export file for fibrations
+(** * Fiber category of a displayed bicategory whose displayed 2-cells form a
+      proposition. In addition, the displayed needs to be locally univalent and
```
benediktahrens:
```suggestion
      proposition. In addition, we ask the displayed bicategory to be locally univalent and
```

UniMath/CategoryTheory/Bicategories/DisplayedBicats/Examples/ProdCommutativity.v
maggesi: > It doesn't use "Admitted" now, nor does any file in UniMath.

True.  I was confused with our own version of the same file that was not merged yet.
I think we can just delete this file for now and put it when it is ready.

UniMath/CategoryTheory/Bicategories/DisplayedBicats/DisplayToFiber.v
maggesi: > Yes, that sounds good!

Done in 4aeb21d1a9d3cdce336312448229fc6b519df6ed

UniMath/CategoryTheory/Bicategories/DisplayedBicats/DisplayToFiber.v
benediktahrens: No idea. Let's leave it as is for the time being.

UniMath/CategoryTheory/Bicategories/DisplayedBicats/Fibration/Fibration1.v
benediktahrens: I would not know of any better place to put this code.

I would like to keep this code in UniMath. There is no compelling reason to remove it.
If a `Qed.` at the end of a construction takes too long, we replace it with `Abort.`, and add a comment about this. In this way, at least the tactics are still checked.

Subsequent results that use this construction can use an `admit.` for the corresponding subgoal, and an `Abort.` at the end.

DanGrayson: There is a compelling reason to do something about `Fibration2` -- travis sometimes can't handle it, because of the overwhelming memory usage.  I would like to wait a little while for #1261 to result in a simplification of the proof that reduces the memory usage.

DanGrayson: Replacing Qed with Abort seems dangerous, as over time, Qed may stop working completely with the proof, and then we would have lost the proof.

nmvdw: > I would not know of any better place to put this code.
> 
> I would like to keep this code in UniMath. There is no compelling reason to remove it.
> If a `Qed.` at the end of a construction takes too long, we replace it with `Abort.`, and add a comment about this. In this way, at least the tactics are still checked.
> 
> Subsequent results that use this construction can use an `admit.` for the corresponding subgoal, and an `Abort.` at the end.

We can keep it in UniMath via https://github.com/UniMath/UniMath/pull/1272.

benediktahrens:
On 03/08/2019 14:57, Niels van der Weide wrote:
> *@nmvdw* commented on this pull request.
> 
> 
> In 
> UniMath/CategoryTheory/Bicategories/DisplayedBicats/Fibration/Fibration1.v 
> <https://github.com/UniMath/UniMath/pull/1271#discussion_r310360152>:
> 
>> @@ -1,2833 +0,0 @@
> -(* ******************************************************************************* *)
> 
>     I would not know of any better place to put this code.
> 
>     I would like to keep this code in UniMath. There is no compelling
>     reason to remove it.
>     If a |Qed.| at the end of a construction takes too long, we replace
>     it with |Abort.|, and add a comment about this. In this way, at
>     least the tactics are still checked.
> 
>     Subsequent results that use this construction can use an |admit.|
>     for the corresponding subgoal, and an |Abort.| at the end.
> 
> We can keep it in UniMath via #1272 
> <https://github.com/UniMath/UniMath/pull/1272>.

That's great. #1272 is merged, now this PR needs to have the removal of 
Fibration{1,2} undone.

maggesi:
> That's great. #1272 is merged, now this PR needs to have the removal of Fibration{1,2} undone.

Done: b1bd2504d60c00d6837a1106bc4fdf6f254acab8

PR #1272, opened 2019-08-03
Attempt to improve Fibration2

nmvdw:
I think this version of Fibration2.v consumes less memory. I split the proof of `discrete_fiber_data_laws_lassociator_lassociator` in several smaller parts so that the ultimate proof term becomes smaller.

### DIFF
--- a/UniMath/CategoryTheory/Bicategories/DisplayedBicats/Fibration/Fibration2.v
+++ b/UniMath/CategoryTheory/Bicategories/DisplayedBicats/Fibration/Fibration2.v
@@ -26,6 +26,13 @@ Section LocalIsoFibration.
 
   Local Arguments transportf {_} {_} {_} {_} {_} _.
   Local Arguments transportb {_} {_} {_} {_} {_} _.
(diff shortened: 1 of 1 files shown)


DanGrayson:
This refers to #1261

nmvdw:
Yes. The current version in master needs 3920000K to compile. I suspect this one requires less.

If the method works, I can apply it to other definitions in Fibration1 and Fibration2 to reduce the memory consumption further.

DanGrayson:
This one runs in 2000000K under Ubuntu 19.04, and the goals all fit on the screen after each "cbn", so this is a vast improvement, thank you.

After this we should enforce a memory limit ourselves, for the sake of reproducibility, and for catching proofs that waste memory.

nmvdw:
> After this we should enforce a memory limit ourselves, for the sake of reproducibility, and for catching proofs that waste memory.

Depending on the limit, it might be good to make a list of files which consume the most memory, so people can try to make those more efficient (if necessary).

PR #1284, opened 2019-09-13
Global completion

nmvdw:
Two main changes:
- I added the global completion of locally univalent bicats
- I made "left biadjoints preserve biinitial objects" compatible with non-univalent bicategories


UniMath/Bicategories/Core/YonedaLemma.v
```diff
@@ -846,3 +848,39 @@ Section YonedaLocalEquivalence.
...
+      (psfunctor_bicat_is_univalent_2_1
+         (op1_bicat B) _
+         univalent_cat_is_univalent_2_1)
+      (y B_is_univalent_2_1).
+Proof.
+  intros x y.
+  apply yoneda_mor_is_equivalence.
+Defined.
+
+Definition global_completion
```
benediktahrens: Maybe call it `rezk_completion_2_0` or something in that vein?
```suggestion
Definition rezk_completion_2_0
```

PR #1285, opened 2019-09-16
Strict bicategories and 2-categories

nmvdw:


UniMath/Bicategories/Core/Strictness.v
```diff
@@ -0,0 +1,428 @@
...
+    + exact (@prunitor _ S).
+    + exact (@plassociator _ S).
+    + exact (@idtoiso_plunitor _ S).
+    + exact (@idtoiso_prunitor _ S).
+    + exact (@idtoiso_plassociator _ S).
+    + abstract (simpl ; intros ; apply HB).
+    + abstract (simpl ; intros ; apply HB).
+Defined.
+
+Definition two_cat
```
benediktahrens: I think that this is not a good definition. 

A 2-category should be a 1-category with extra structure. In particular, there should be a coercion from 2-categories to 1-categories allowing one to use the usual 1-categorical constructions, in particular, the accessor functions for 1-categories.

nmvdw: What about this version?

PR #1295, opened 2019-10-16
fix the opam installation instructions

DanGrayson:


DanGrayson:
I canceled the travis build ...

PR #1297, opened 2019-10-24
Finished biequiv between monads and K-triples

nmvdw:
Only the unit and counit were there. Now the whole biequivalence.

### DIFF
--- a/UniMath/Bicategories/DisplayedBicats/Examples/MonadKtripleBiequiv.v
+++ b/UniMath/Bicategories/DisplayedBicats/Examples/MonadKtripleBiequiv.v
@@ -530,45 +530,50 @@ Proof.
   apply (unit_bind k).
 Qed.
 
-Definition Monad_biequiv_Ktriple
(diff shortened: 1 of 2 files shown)


benediktahrens:
I am puzzled by the Travis result:
```
--- checking for changes to the Foundations package ---
test -z "`git diff --stat master -- UniMath/Foundations`"
make: *** [Makefile:433: check-for-change-to-Foundations] Error 1
```
There does not seem to be a change in this PR to any file in Foundations.
@DanGrayson : am I missing something?

DanGrayson:
Oh, I see what's wrong.  The test in the Makefile depends on the master branch of the cloned repository being up to date -- in the docker image I made the repository is present, but, of course, it will get more and more out of date.  Probably what it should do instead is to fetch origin/master and compare against that remote branch.  Will fix.

DanGrayson:
If the other tests passed, go ahead and pull and I'll submit the fix separately.

PR #1298, opened 2019-10-24
fix Makefile sanity checks

DanGrayson:
... so it fetches from the "origin" and compares to "origin/master", in case the
master branch of the current repository is not up to date (as is now the case
when running under travis with the new docker image).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -426,16 +426,16 @@ check-listing-of-proof-files:
 
 # Here we check for changes to UniMath/Foundations, which normally does not change.
 # One step of the travis job will fail, if a change is made, see .travis.yml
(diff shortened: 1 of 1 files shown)


benediktahrens:
This test fails: 
```
--- checking for changes to the Foundations package ---
git fetch origin
test -z "`git diff --stat origin/master -- UniMath/Foundations`"
--- checking for submodule changes ---
git fetch origin
test -z "`git diff master sub`"
make: *** [Makefile:439: check-for-submodule-changes] Error 1
```
This is a different test that fails, so maybe we should merge this PR and then see about the other test?

DanGrayson:
I think that ought to fix it ...

PR #1304, opened 2019-11-21
Switch Qed to Defined in [Exponentials_functor_HSET]

mortberg:
I got stuck on an evil Qed today. I have no clue why it was there as I've apparently been very careful to make all the irrelevant parts abstract in the proof...


DanGrayson:
lgtm

PR #1307, opened 2019-12-12
Tiny representable objects in presheaf categories

elisabethbonnevier:
If a category C is small and has binary products then yoneda of any object is tiny, that is, exponentiation by it has a right adjoint in the presheaf category on C.

The PR also contains the fact that yoneda commutes with binary products (up to iso) and a useful lemma for constructing isos of functors between presheaf categories.


benediktahrens:
@elisabethbonnevier : the code looks very good, thanks a lot for the contribution.

There are a few small things that could be improved:

1. Travis shows the following error:
```
--- checking the ordering prescribed by the files UniMath/*/.packages/files ---
error: *** UniMath/CategoryTheory/yoneda_binproducts.vo should not require UniMath/CategoryTheory/Presheaf.vo
make: *** [Makefile:318: .check-prescribed-ordering.okay] Error 1
```
This can be fixed by putting the line containing "yoneda_binproducts.v" *after* the line containing "Presheaf" in the file UniMath/CategoryTheory/.package/files.

2. The file name "tiny.v" is not very descriptive. Maybe something like "ExponentiationLeftAdjoint.v" would be a bit more informative? 

3. We seem to capitalize file names now. This is not consistent throughout the library, many old files do not adhere to this convention. But new files should be capitalized.

Would you mind amending your PR to incorporate these comments?

PR #1310, opened 2020-01-25
Prove interval in cartesian cubical sets has decidable equality

elisabethbonnevier:
Proof that the interval in cartesian cubical sets has decidable equality for every object in the cartesian cube category.

### DIFF
--- a/UniMath/CategoryTheory/categories/CartesianCubicalSets.v
+++ b/UniMath/CategoryTheory/categories/CartesianCubicalSets.v
@@ -138,7 +138,7 @@ Local Definition I : cartesian_cubical_sets :=
 
 (** The interval in cartesian cubical sets has two distinct elements *)
 Lemma interval_cartesian_cubical_sets_two_elements :
-  ∏ n : cartesian_cube_category, ∑ f g : I ⟨ n ⟩, f != g.
+  ∏ n : cartesian_cube_category, ∑ f g : ((I : functor _ _) n : hSet), f != g.
 Proof.
   intro n.
   exists (λ _ : stn 1, @inr (stn n) (stn 2) (make_stn 2 0 (idpath _))).
@@ -153,7 +153,7 @@ Defined.
 
 (** The interval in cartesian cubical sets has decidable equality *)
 Lemma interval_cartesian_cubical_sets_dec_eq :
-  ∏ (n : cartesian_cube_category), isdeceq I ⟨ n ⟩.
+  ∏ (n : cartesian_cube_category), isdeceq ((I : functor _ _) n : hSet).
 Proof.
   intro n.
   use isdeceqweqb.
--- a/UniMath/CategoryTheory/categories/CartesianCubicalSets.v
+++ b/UniMath/CategoryTheory/categories/CartesianCubicalSets.v
@@ -1,6 +1,11 @@
 (** ****************************************************************************
 
-Theory about the cartesian cube category and cartesian cubical sets.
+We define the cartesian cubical sets and show that the interval satisfies
+axioms B1, B2 and B3 in:
+
+A Survey of Constructive Presheaf Models of Univalence (2018),
+Thierry Coquand
+https://dl.acm.org/doi/abs/10.1145/3242953.3242962
 
 Contents:
 
@@ -8,11 +13,11 @@ Contents:
 - Binary products in the cartesian cube category ([cartesian_cube_category_binproducts])
 - The empty set is a terminal object in the cartesian cube category
   ([empty_is_terminal_cartesian_cube_category])
-- The interval in cartesian cubical sets has two distinct elements
+- The interval in cartesian cubical sets has two distinct elements (axiom B1)
   ([interval_cartesian_cubical_sets_two_elements])
-- The interval in cartesian cubical sets has decidable equality
+- The interval in cartesian cubical sets has decidable equality (axiom B2)
   ([interval_cartesian_cubical_sets_dec_eq])
-- The interval in cartesian cubical sets is tiny
+- The interval in cartesian cubical sets is tiny (axiom B3)
   ([interval_cartesian_cubical_sets_is_tiny])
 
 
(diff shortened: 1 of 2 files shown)


UniMath/CategoryTheory/categories/CartesianCubicalSets.v
```diff
@@ -144,6 +146,20 @@ Proof.
   apply (negpaths0sx 0 e).
 Defined.
 
+(** The interval in cartesian cubical sets has decidable equality *)
+Lemma interval_cartesian_cubical_sets_dec_eq :
+  ∏ (n : cartesian_cube_category), isdeceq (pr1 (pr1 I n)).
```
benediktahrens: What is "(pr1 (pr1 I n))"? Would it be useful to introduce a name for that thing? A beautiful name would make the statement of this result easier to read.

mortberg: Yeah, or figuring out how to make the coercions trigger properly so that we can just write `I n`... 

`I` is an element of `PreShv(C)`, so I think the innermost `pr1` gets the object part of the functor. I then think the outermost pr1 gets the underlying type of the set `pr1 I n`.

benediktahrens: Ok, thanks, I think I understand. It may well be difficult (read: impossible) to trigger those coercions without manual intervention.
Maybe the best one could come up with is something like `isdeceq ((I : functor _ _) n : hSet)`? (I have not checked that in Coq, so take it with a grain of salt.)

mortberg: I guess that's a bit more readable. One could maybe even drop the innermost type declaration, but I'm also not sure and haven't checked in Coq. 

The long term solution is probably to add notations to the file on presheaves for this.

mortberg:
Btw, this might be a good point to add a pointer to https://dl.acm.org/doi/abs/10.1145/3242953.3242962 and say which of the axioms we have checked in the file

UniMath/CategoryTheory/categories/CartesianCubicalSets.v
elisabethbonnevier: I is a presheaf on the cartesian cube category so I need the first projection to get the functor data so I can apply it to the object n. Then I need the first projection again to get the set that I wish to show has decidable equality. But I agree that it is not very nice to read...

How about
Definition I_on_objects : cartesian_cube_category -> UU :=
\lambda n : cartesian_cube_category, pr1 (pr1 I n)?

It might be worth introducing this because I use pr1 (pr1 I n) in interval_cartesian_cubical_sets_two_elements as well.

mortberg: What about adding a general notation to the file on presheaves instead, something that lets us write I[n] or some other symbols around n?

mortberg: We had a discussion about this and the "I [n]" notation will clash with some other notations in UniMath, so we decided to go for "I ⟨ n ⟩" and @elisabethbonnevier will add it to the PreSheaf file in this PR. This could make some of my old presheaf related code nicer as well.

benediktahrens: I am interested to see if it works in both parsing and printing.

DanGrayson: What happened to the idea of using a coercion?  Did it prove to be not possible?

mortberg: > I am interested to see if it works in both parsing and printing.

Yeah, we got it to work also for printing by adding a call to `functor_on_objects` in addition to the `pr1`'s in the notation. See the latest commit.

mortberg: > What happened to the idea of using a coercion? Did it prove to be not possible?

At least I couldn't figure out how to do it. If someone can then it would be nicer than the notation, but at least the notation is a lot better than what it used to be.

benediktahrens: > > What happened to the idea of using a coercion? Did it prove to be not possible?
> 
> At least I couldn't figure out how to do it. If someone can then it would be nicer than the notation, but at least the notation is a lot better than what it used to be.

Have you tried the thing I suggested?

UniMath/CategoryTheory/categories/CartesianCubicalSets.v
```diff
@@ -148,7 +153,7 @@ Defined.
 
 (** The interval in cartesian cubical sets has decidable equality *)
 Lemma interval_cartesian_cubical_sets_dec_eq :
-  ∏ (n : cartesian_cube_category), isdeceq (pr1 (pr1 I n)).
+  ∏ (n : cartesian_cube_category), isdeceq I ⟨ n ⟩.
```
mortberg: I'm surprised that no parantheses are needed here. Maybe the notation binds a bit too hard. We could try with 40 or something.

benediktahrens: I would prefer the manual triggering of the coercions over introducing another definition and/or notation. Please try once more.

I might also give it a try, but need some time to do so.

UniMath/CategoryTheory/categories/CartesianCubicalSets.v
elisabethbonnevier: Yes, it works with ((I : functor _ _) n : hSet).

benediktahrens: @elisabethbonnevier : then I suggest using this. It is a bit annoying, but less so then introducing more constants.

elisabethbonnevier: Alright :)

PR #1312, opened 2020-02-14
Characterize adding structure/properties via hlevels

nmvdw:
On nLab, there is a definition of when a forgetful functor forgets structure and when it forgets properties:
https://ncatlab.org/nlab/show/stuff%2C+structure%2C+property#definitions
We study these definitions using displayed categories. Say that a displayed category adds properties if the forgetful functor is full&faithful and that a displayed category adds structure if the forgetful functor is faithful. More specifically, we show:
1. A displayed category adds properties if and only if the displayed morphisms are contractible (```adds_properties_weq_locally_contractible``` in ```Projection.v```)
2. A displayed category adds structure if and only if the displayed morphisms are a proposition (```adds_structure_weq_locally_propositional``` in ```Projection.v```)

In ```UnitalBinop```, two ways of defining sets with a unital binary operation are studied. We can construct the category of sets with a unital binary operation from the category of sets with a binary operation in two ways:
1. We can add the unit element as property
2. We can add the unit element as structure (this one will be preferred in practice).

The point is to illustrate the difference between adding properties versus adding structure from the perspective given in ```Projection.v```.


UniMath/CategoryTheory/DisplayedCats/Examples/UnitalBinop.v
```diff
@@ -0,0 +1,410 @@
...
+
+Local Open Scope cat.
+
+(**
+Let us start by defining sets with a binary operation on them.
+To do so, we use algebras on the diagonal functor.
+Let us repeat the diagonal.
+After that, we show we have a univalent category of sets with a binary operation.
+ *)
+Definition diag
```
langston-barrett: Could this also be defined as the composition of the pre-existing `bindelta_functor` and `binproduct_functor`?

nmvdw: Good point, thank you. It is fixed now.

UniMath/CategoryTheory/DisplayedCats/Examples/UnitalBinop.v
```diff
@@ -0,0 +1,409 @@
+(**
+Author: Niels van der Weide
+
+Suppose, we have a set `X` with a binary operation `f : X → X → X`.
+Since init elments for `f` are unique, the type of unit elements of `f` is a proposition.
```
benediktahrens:
```suggestion
Since unit elments for `f` are unique, the type of unit elements of `f` is a proposition.
```

PR #1314, opened 2020-03-14
Remove obsolete [camlp4] from list of installation dependencies 

peterlefanulumsdaine:
If I understand right, `camlp4` has not been required by Coq since v8.7 (see https://github.com/coq/coq/pull/461), so is no longer needed as a `UniMath` installation dependency.

Moreover, `camlp4` is now somewhat deprecated, and is no longer available from Homebrew, so its presence makes the Mac OS build instructions broken.

I’ve confirmed with a build just now that `camlp4` is not required at least on Mac OS X.  I haven’t tested on other systems — in particular, I haven’t tested the Nix install instructions — but I can’t see any way that it could still be required.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -19,7 +19,7 @@ A more flexible, but complex, installation method is given in [INSTALL\_OPAM.md]
 1. Install "Homebrew", available from http://brew.sh/.
 2. Using Homebrew, install ocaml with the following command:
 ```bash
-$ brew install objective-caml ocaml-num camlp4 camlp5 bash ocaml-findlib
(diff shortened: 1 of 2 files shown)


DanGrayson:
Thanks, I'll test it under various flavors of linux.

DanGrayson:
It works under Arch Linux.  Travis tests it under Ubuntu, and my notes for the docker image in .travis.yml state that only camlp5 was installed on the docker image where the tests run.  I don't feel the need to test it under Nixos, so will merge it now.

Thanks!

maggesi:
It work with Nix too.

PR #1317, opened 2020-04-22
Need to require module Ltac rather than Notations to import tactics with -nois

herbelin:
Hi, there is a change in how tactics are imported with option `-nois`. They are now imported by requiring module `Coq.Init.Ltac` rather than `Coq.Init.Notations`.

This PR adapts UniMath to this change, as noticed on Coq CI. It should be merged synchronously with the Coq PR.

### DIFF
--- a/UniMath/Foundations/Init.v
+++ b/UniMath/Foundations/Init.v
@@ -7,7 +7,7 @@ Require Export Coq.Init.Logic.
 Require Export Coq.Init.Notations.
 (* get the standard Coq reserved notations *)
 
-Require Export Coq.Init.Ltac.
+From Coq Require Ltac.
 (* get the tactics *)
 
 (** Notations *)
--- a/UniMath/Foundations/Init.v
+++ b/UniMath/Foundations/Init.v
@@ -7,7 +7,7 @@ Require Export Coq.Init.Logic.
 Require Export Coq.Init.Notations.
 (* get the standard Coq reserved notations *)
 
-From Coq Require Ltac.
+From Coq Require Export Ltac.
 (* get the tactics *)
 
 (** Notations *)
--- a/UniMath/Foundations/Init.v
+++ b/UniMath/Foundations/Init.v
@@ -7,6 +7,9 @@ Require Export Coq.Init.Logic.
 Require Export Coq.Init.Notations.
 (* get the standard Coq reserved notations *)
 
+Require Export Coq.Init.Ltac.
+(* get the tactics *)
+
 (** Notations *)
 
 Notation "'∏'  x .. y , P" := (forall x, .. (forall y, P) ..)
--- a/Makefile
+++ b/Makefile
@@ -218,7 +218,7 @@ distclean::          ; - $(MAKE) -C sub/lablgtk arch-clean
 # building coq:
 export PATH:=$(shell pwd)/sub/coq/bin:$(PATH)
 CONFIGURE_OPTIONS := -coqide "$(COQIDE_OPTION)" -with-doc no -local -no-custom
-BUILD_TARGETS := coqbinaries tools states
+BUILD_TARGETS := coqbinaries tools states ltac
 ifeq ($(DEBUG_COQ),yes)
 CONFIGURE_OPTIONS += -annot
 BUILD_TARGETS += byte


benediktahrens:
Dear @herbelin , thanks for preparing this PR. I will keep an eye on this and the related Coq PR.

DanGrayson:
Thanks, Hugo!

UniMath/Foundations/Init.v
```diff
@@ -7,6 +7,9 @@ Require Export Coq.Init.Logic.
 Require Export Coq.Init.Notations.
 (* get the standard Coq reserved notations *)
 
+Require Export Coq.Init.Ltac.
```
ejgallego, line 10: I'm worried this will create backwards compatibility problems; I'd suggest instead
`From Coq Require Ltac` cc: @herbelin

ejgallego:
Hi folks, this can be merged, however my proposed change may work better for you if you still want to use both ` master` and 8.11

UniMath/Foundations/Init.v
benediktahrens, line 10: Following @ejgallego 's suggestion:
```suggestion
From Coq Require Ltac.
```

ejgallego: Sorry I had an obvious typo here; I forgot the `Export`.

DanGrayson: Backward compatibility would be good, so people can use UniMath with older distributed versions of Coq.  What line precisely do you recommend as the alternative?

ejgallego: `From Coq Require Export Ltac.`

DanGrayson: We already have had these lines for a long time:
```
Require Export Coq.Init.Logic.
Require Export Coq.Init.Notations.
```
Should we change them, too?  To
```
From Coq Require Export Logic.
From Coq Require Export Notations.
```
?

ejgallego: No, you need to add a new one for `Ltac` , as it is not exported by `Notations` anymore.

DanGrayson:
@benediktahrens -- so apparently we need
```
From Coq Require Export Ltac.
```
instead.

UniMath/Foundations/Init.v
```diff
@@ -7,6 +7,9 @@ Require Export Coq.Init.Logic.
 Require Export Coq.Init.Notations.
 (* get the standard Coq reserved notations *)
 
+From Coq Require Ltac.
```
benediktahrens, line 10:
```suggestion
From Coq Require Export Ltac.
```

ppedrot:
What's blocking this? The Coq CI is broken in the meantime...

DanGrayson:
Currently travis tells us that the build fails with this message:
```
*** Warning: in file UniMath/Foundations/Init.v, library Coq.Ltac is required and has not been found in the loadpath!
COQC UniMath/Foundations/Init.v
File "./UniMath/Foundations/Init.v", line 10, characters 24-28:
Error: Unable to locate library Ltac with prefix Coq.
```

Wasn't the new line (`From Coq Require Export Ltac.`) supposed to be backward compatible with the version of Coq our build is still using?  (It's "commit b7c4ca96e555a68553aaa51b42f3d42e919ceb0a (HEAD, tag: V8.11.0)".)

DanGrayson:
Indeed, the file `theories/Init/Ltac.v` doesn't exist in 8.11.0.

DanGrayson:
What if we put the contents of Ltac.v into our Init.v, namely these two lines, instead?:
```
Declare ML Module "ltac_plugin".
Export Set Default Proof Mode "Classic".
```

herbelin:
> Wasn't the new line (From Coq Require Export Ltac.) supposed to be backward compatible with the version of Coq our build is still using? 

It is supposed to be compatible and it works with my brew-installed version of 8.11.

> Indeed, the file theories/Init/Ltac.v doesn't exist in 8.11.0.

The 8.11 installation does not have `theories/Init/Ltac.v` but it should have `plugins/ltac/Ltac.v`.

> What if we put the contents of Ltac.v into our Init.v, namely these two lines, instead?:
> 
> ```
> Declare ML Module "ltac_plugin".
> Export Set Default Proof Mode "Classic".
> ```

That's less elegant, but it is a possible alternative.

DanGrayson:
Currently, when building Coq ourselves, we use just `coqbinaries tools states` as make targets.  Perhaps we have to build the plugins, too, by adding `plugins` to the list.  (There is no file `Ltac.vo` after we build.)

ppedrot:
The `ltac` target should be enough.

DanGrayson:
It worked at home, so it ought to work under travis...

ejgallego:
Indeed I was proposing you used the `plugins/ltac/Ltac.v` file for 8.11 instead. That will pick `Init.Ltac` in 8.12.

PR #1329, opened 2020-09-22
Opaque PathOver (needed for coq/coq#13037)

gares:
This little change makes coq/coq#13037 show no regression on Circle2.v

### DIFF
--- a/UniMath/SyntheticHomotopyTheory/Circle2.v
+++ b/UniMath/SyntheticHomotopyTheory/Circle2.v
@@ -209,6 +209,7 @@ Proof.
   apply (maponpaths (λ f, pr1weq f v)). apply cp_irrelevance. apply torsor_hlevel.
 Defined.
 
+Opaque PathOver.
 Section A.
 
   Context (A : circle -> Type) (a : A pt) (p : PathOver a a loop).


DanGrayson:
Could you clarify what " regression on Circle2.v " means?

DanGrayson:
Why is this a draft?

gares:
It's a draft because I'm still not sure coq/coq#13037 will be merged. If we decide to do so, then I'll flip the draft bit (I should have explained this, sorry).

With coq/coq#13037 the computation of default implicit arguments changes. This computation puts the type in head normal form, in order to expose `forall` quantifiers, and it was performed using a call by name procedure. Some users had performance problems, so I tried to use a call by need procedure. In the benchmak I run only one development did show a regression, UniMath, and only one file did see a regression (in memory usage, mostly, and time), Circle2.v. See https://coq.gitlab.io/-/coq/-/jobs/748058922/artifacts/_bench/html/coq-unimath/UniMath/SyntheticHomotopyTheory/Circle2.v.html (yellow is new, red is old, here you only see time but the extra memory consumption comes from the same lines, I did test it locally).

After a bit of fiddling, I found out that the root problem is the reductions allowed by the body of `PathOver`. I have no clue if it is desirable to make it Opaque, at least in this file, but the proof works, so I've opened the PR to get feedback. If it's OK, then you can just merge it. If it's not ideal, but tolerable, let's first decide what to do with the PR on the Coq side. If it's not OK, please tell me now.

DanGrayson:
Thanks for all the work.  I think we could just merge it without pain, and in doing so I would add a link to this pull request.  But about those timings: the first time I looked at your timings file I managed to see one display of timings, but now I don't see any.  Should I be seeing them just by hovering over bi-colored lines of code?

gares:
Yes, hovering should pop up the line number and the two timings.

DanGrayson:
Oh, now I see them.  I think I just wasn't patient enough -- it takes time for them to pop up.

Do the timings recover fully, back to their original times, with that addition of that one line?

gares:
I don't know. Locally it came back to instantaneous and memory usage under 1G (it was up to 3G). If you `coqc -time` you see live what is the used to generated the html.
 
Unfortunately it's not trivial to run our bench infrastructure on a branch. But I can do it if you want to be 100% sure.

DanGrayson:
Thanks!

PR #1334, opened 2020-11-20
Adapt to fixing dropped implicit arguments in Context

herbelin:
Hi, implicit arguments in `Context` were discarded. Coq PR [#11390](https://github.com/coq/coq/pull/11390) preserves them. The present PR adapts UniMath to the new semantics.

I went for a backwards-incompatible change preserving the intent of providing an implicit argument to an hypothesis named `H` in file `slicecat.v`. As such it'll have to be merged synchronously with the Coq PR.

Note that, if you prefer, a backward-compatible could be done by instead replacing the `{ }` around `{c c' : C}` with parentheses in the type of `H`.

Best,

### DIFF
--- a/UniMath/CategoryTheory/slicecat.v
+++ b/UniMath/CategoryTheory/slicecat.v
@@ -840,10 +840,10 @@ Defined.
 *)
 Section dependent_product.
 
-Context (H : ∏ {c c' : C} (g : C⟦c,c'⟧), is_left_adjoint (base_change_functor g)).
+Context (H : ∏ (c c' : C) (g : C⟦c,c'⟧), is_left_adjoint (base_change_functor g)).
 
 Let dependent_product_functor {c c' : C} (g : C⟦c,c'⟧) :
-  functor (C / c) (C / c') := right_adjoint (H g).
+  functor (C / c) (C / c') := right_adjoint (H c c' g).
 
 Let BPC c : BinProducts (C / c) := @BinProducts_slice_precat C hsC PC c.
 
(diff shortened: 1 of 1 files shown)


DanGrayson:
Thanks, Hugo!

Backwards compatibility is probably better for us, as then we don't have to synchronize versions of coq.  In fact, I expect the travis build to fail for that reason, as the coq submodule of our repository hasn't been updated in your pull request.

I'd be happy to take care of that.  Benedikt, do you agree? ( @benediktahrens )

benediktahrens:
> Thanks, Hugo!
> 
> Backwards compatibility is probably better for us, as then we don't have to synchronize versions of coq. In fact, I expect the travis build to fail for that reason, as the coq submodule of our repository hasn't been updated in your pull request.
> 
> I'd be happy to take care of that. Benedikt, do you agree? ( @benediktahrens )

Yes, sounds good! Thanks a lot @DanGrayson and @herbelin !

DanGrayson:
Hugo, now that github actions provide so much cpu time to us, it would probably be good if we follow coq more closely, so we can detect these errors ourselves, by running an overnight test against the latest coq.  If you think that's a good idea, which branch should we follow?

herbelin:
Hi Dan, regarding this particular error, there would have been no way to detect it as of now besides running the submitted Coq PR against UniMath, what our CI is doing. So, I'm not sure what more could be done. If you'd like to receive a notification for any failure of a PR against UniMath, that's certainly possible though (we already do it for fiat-crypto to Jason, if I remember correctly).

DanGrayson:
Okay, then I guess it's best to leave it to you, as a failure in a PR might be due to the PR rather than to our code.

PR #1335, opened 2020-11-29
With Coq PR #13513, "intros []" uses "destruct" rather than "case" and becomes a bit "smarter"

herbelin:
As a consequence, some `clear` on a residual dependent argument of an inductive type becomes useless.

We leave a `try` so as to be backwards-compatible (you may prefer a clean non-compatible script though).

Cf coq/coq#13513.

### DIFF
--- a/UniMath/.dir-locals.el
+++ b/UniMath/.dir-locals.el
@@ -7,7 +7,7 @@
 	     (make-local-variable 'coq-prog-args)
 	     (setq coq-prog-args
 		   ;; these options should match what's used in ../Makefile
-		   `("-emacs" "-noinit" "-indices-matter" "-type-in-type" "-w" "-notation-overridden" "-Q" ,(concat unimath-topdir "UniMath") "UniMath" )
+		   `("-quiet" "-emacs" "-noinit" "-indices-matter" "-type-in-type" "-w" "-notation-overridden" "-Q" ,(concat unimath-topdir "UniMath") "UniMath" )
 		   )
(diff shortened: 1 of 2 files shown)


DanGrayson:
Vladimir was in favor of using the induction principle only, hence using neither "destruct" nor "[]".  In fact, we have these guidelines in UniMath/README.md:
```
* Do not use `destruct`, `match`, `case`, square brackets with `intros`, or
  nested square brackets with `induction`.  (The goal is to prevent generation of
  proof terms using `match`.)
```

So what if we use this instead?:
```diff
diff --git a/UniMath/SubstitutionSystems/GenMendlerIteration_alt.v b/UniMath/SubstitutionSystems/GenMendlerIteration_alt.v
index 41b3c38e..9d9373f6 100644
--- a/UniMath/SubstitutionSystems/GenMendlerIteration_alt.v
+++ b/UniMath/SubstitutionSystems/GenMendlerIteration_alt.v
@@ -120,7 +120,7 @@ Proof.
 use make_cocone.
 - intro n.
   apply (pr1 (Pow n) _ z).
-- abstract (intros n m []; clear m; apply Pow_cocone_subproof).
+- abstract (intros n m k; induction k; apply Pow_cocone_subproof).
 Defined.
 
 Local Definition CC_LchnF : ColimCocone LchnF.
```

I'd be happy to take care of that, and the result would be clearer to the user.

herbelin:
> So what if we use this instead?: ...

That's fine to me.

Issue #1316, opened 2020-04-15
algebraic geometry


DanGrayson:
It would be good to start formalizing classical algebraic geometry, starting with the Zariski topology on Spec R and its sheaf of rings.  Kevin Buzzard mentioned this in a recent talk online.

PR #1341, opened 2021-01-25
Zariski topology on Spec R

timcherganov:
#1316


DanGrayson:
Could you merge our master branch into your branch?  Also, the travis error results from not adding the new package to the test script.  I suggest adding it this way:
```diff
diff --git a/.travis.yml b/.travis.yml
index a16328fb..b9efea75 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ env:
     - PACKAGES=Bicategories
     - PACKAGES="MoreFoundations Ktheory SyntheticHomotopyTheory"
     - PACKAGES=HomologicalAlgebra
-    - PACKAGES=Topology
+    - PACKAGES="Topology AlgebraicGeometry"
     - PACKAGES=RealNumbers
     - PACKAGES=SubstitutionSystems
     - PACKAGES=Tactics BUILD_COQIDE=yes
```

timcherganov:
Done

UniMath/AlgebraicGeometry/Spec.v
```diff
@@ -0,0 +1,141 @@
...
+
+  Definition make_spec_set (p : ideal R) (isp : is_prime p) : spec_set :=
+    p ,, isp.
+
+  Definition spec_set_ideal (S : spec_set) : ideal R := pr1 S.
+  Coercion spec_set_ideal : spec_set >-> ideal.
+
+  Definition spec_set_is_prime (S : spec_set) : is_prime S := pr2 S.
+
+  Definition D (a : hsubtype R) : hsubtype spec_set :=
```
benediktahrens, line 22: Could you use a longer name for this, and reserve one-letter identifiers for variables?

benediktahrens (approved):
Looks good, thanks a lot for the PR!

UniMath/AlgebraicGeometry/Spec.v
timcherganov: Don't like this name either, but don't know better. The only other name that comes to my mind is 'prime_ideals_not_containing', which is even worse. Maybe you have a suggestion?

DanGrayson: What about `principalOpen` or `principalOpenSet`?  

Certainly not `D`.

benediktahrens: > Don't like this name either, but don't know better. The only other name that comes to my mind is 'prime_ideals_not_containing', which is even worse. Maybe you have a suggestion?

I don't find this suggestion worse; I find it much better than `D`. This identifier explains so well what it does that one does not need to read its definition.

timcherganov: > What about `principalOpen` or `principalOpenSet`?

This name is better suited for the following definition
`Definition principal_open (f : R) : hsubtype spec_set := λ p, ¬ p f`

timcherganov: > This identifier explains so well what it does that one does not need to read its definition.

But the definition is shorter than its identifier. Maybe I should define the Zariski topology this way?
```
Definition zariski_topology : (spec_set -> hProp) -> hProp :=
  λ U, ∃ A, U ≡ (λ p, A ⊈ p).
```

DanGrayson: That looks good -- it's easy enough to read.

UniMath/Algebra/RigsAndRings/Ideals.v
```diff
@@ -113,3 +114,18 @@ Proof.
     refine (monoidfunmul (rigmultfun f) _ _ @ _); cbn.
     abstract (rewrite ss; refine (rigmult0x _ (pr1 f r) @ _); reflexivity).
 Defined.
+
+(** ** Prime ideal *)
+
+Section prime.
+  Context {R : rig}.
```
DanGrayson, line 121: The definition of prime ideal below should be restricted to commutative rings, not rigs.  (The definition for noncommutative rings is different.)

UniMath/AlgebraicGeometry/Spec.v
```diff
@@ -0,0 +1,141 @@
...
+    λ p, a ⊈ p.
+
+  Definition zariski_topology : (spec_set -> hProp) -> hProp :=
+    λ U, ∃ A, U ≡ D A.
+
+  Lemma union_not_contained_in {X : UU} (U : (X -> hProp) -> hProp) (S : X -> hProp) :
```
DanGrayson, line 28: This could go in Topology/Prelim.v, for use by others.

UniMath/AlgebraicGeometry/Spec.v
```diff
@@ -0,0 +1,141 @@
...
+          apply (transportb (λ y, p y) Hab).
+          apply (ideal_isr p b a), Hpa.
+      + exists b. use make_dirprod.
+        * exact Hb.
+        * use (negf _ Hpx). intro Hpb.
+          apply (transportb (λ y, p y) Hab).
+          apply (ideal_isl p a b), Hpb.
+  Defined.
+
+  Definition spec_top : TopologicalSet :=
```
DanGrayson, line 133: Hmm, this term `TopologicalSet` (not due to you) sounds strange.  Why isn't it called `TopologicalSpace`?  @benediktahrens - do you know?

DanGrayson:
Thank you for the contribution!

UniMath/AlgebraicGeometry/Spec.v
benediktahrens: > Hmm, this term `TopologicalSet` (not due to you) sounds strange. Why isn't it called `TopologicalSpace`? @benediktahrens - do you know?

I seem to remember a discussion about that between Catherine and Vladimir at the IAS, mentioning that "space" sounds like it were about an arbitrary type, not about a type of hlevel 2. But I am not sure that that was it.

DanGrayson: You're probably right -- Vladimir was probably looking forward to topological spaces where the underlying type had a higher truncation.  Would you mind if I just change it?  I think it should agree with current mathematical terminology, so current mathematicians can read the formalization.

benediktahrens: > Would you mind if I just change it? 

Not at all!

DanGrayson: On second thought, if I make that change, something will be wrong, because the definition of TopologicalSet doesn't seem to require that the underlying type is a set:
```coq
Definition isTopologicalSet (X : UU) :=
  ∑ O : (X → hProp) → hProp, isSetOfOpen O.
Definition TopologicalSet := ∑ X : UU, isTopologicalSet X.
```

benediktahrens: I would assume that the truncation assumption is at least required to obtain a (univalent?) category of topological spaces/sets?

DanGrayson: Yes, otherwise Hom(X,Y) isn't even a set.  At least if you define it to be X->Y.

benediktahrens: Let me summarize to check that I understand correctly:
- the current def of topological set does not have a set-condition (which is likely to be a mistake, since that is required to obtain a category, in particular, a univalent category).
- you would like to rename it to topological space, to align it with classical mathematics.

It seems to me that if we add a set-condition, then calling those things "topological sets" (or topological hsets) is useful.

Maybe someone would like to study if/to what extent the set-condition is necessary for a Stone-type duality.

DanGrayson: I wouldn't call it a mistake -- maybe they had something in mind.  For example, perhaps they could define Hom(X,Y) to be the set-truncation of X->Y.

But where does the name "TopologicalSet" come from, if the type is not a set?  Is the name appropriate?

Adding the condition that the underlying type is a set, in my mind, is desirable mainly because that aligns it with classical mathematics.

martinescardo:
I find it surprising that you speak of topological sets rather than topological spaces in UniMath. At first it sounds you are considering a new notion. But then you find that it is just the notion of topological space from traditional mathematics (assuming that the underlying type is indeed a set, as the name suggests) with a different name.

UniMath/AlgebraicGeometry/Spec.v
benediktahrens: Of course there is a *set* of subsets - maybe that's why Cath and Vladimir chose to call it "topological *set*"?

martinescardo: I find it surprising that you speak of topological sets rather than topological spaces in UniMath. At first it sounds you are considering a new notion. But then you find that it is just the notion of topological space from traditional mathematics (assuming that the underlying type is indeed a set, as the name suggests) with a different name.

DanGrayson: It is surprising.  Maybe I should add the assumption that the underlying type is a set and change the name, too.

martinescardo: Another remark is that I don't think allowing the underlying type to be more general than a set is useful, because if `X` is any type and `Y` is its set truncation, then the canonical map `((X->hProp)->hProp) -> ((Y->hProp)->hProp)` is an equivalence. That is, a topology on a type is the same thing as a topology on its set truncation. And, as already remarked, the topology itself is a set, regardless of whether `X` is a set or not.

DanGrayson: Well, maybe their idea was that when X and Y are types with topologies, the subtype of X->Y given by the "continuous" functions is an interesting one to consider, for some reason.

PR #1343, opened 2021-02-27
Changed some arguments in CategoryTheory.DisplayedCats.Core.v from implicit to explicit.

Tobias-Schmude:
All arguments of assoc_disp were implicit. Now they match those of assoc_disp_var.


Tobias-Schmude:
The arguments of id_left/right_disp(_var), homset_disp and pathscomp0_disp were implicit as well. Changed those that are necessary to infer the others to explicit, and updated all definitions that depended on these (in Constructions.v).

benediktahrens:
This is a goo change - it is odd to have all arguments to be implicit and maximally inserted.

PR #1344, opened 2021-04-05
Move notation for `eqset` to scope `logic` instead of `set`?

peterlefanulumsdaine:
Currently, the notation of `x = y` for `eqset x y` (i.e. the `hProp` of paths in a `hSet`) lives in the scope `set`.

This seems a bit odd, and often inconvenient.  The scope `set` is otherwise for constructions producing sets, e.g. `total2_hSet`, whereas `eqset` produces a proposition, and is used most when working in the prop-valued logic.  So it seems like this notation for `eqset` should live in scope `logic`, with notations like `p ∧ q` and `p ∨ q`.

In this PR, I’ve tried out that change (and the same for `!=` as `neqset`), and updated various files to make use of the notation (since it no longer requires scope-juggling).  As hoped, it gives (I think) much better readability in many places, especially in `Algebra` — do others agree?

(Besides what shows up in this PR, I was prompted to make this change by getting annoyed by the scope-juggling needed in https://github.com/peterlefanulumsdaine/UniMath/tree/bourbaki-witt , where I’ve been working on formalising fixed-point theorems. I think this scope change will be very useful there.)

One caveat: I know we generally want to be *very* conservative about making changes to files in `Foundations`.  However, I think changing the scope of a couple of notations doesn’t affect Vladimir’s mathematical or expository intent, so seems to me worthwhile in this instance.

### DIFF
--- a/UniMath/Algebra/BinaryOperations.v
+++ b/UniMath/Algebra/BinaryOperations.v
@@ -34,6 +34,8 @@
 
 Require Export UniMath.Foundations.Sets.
 
+Local Open Scope logic.
+
 (** To upstream files *)
 
 
@@ -326,7 +328,7 @@ Proof.
         rewrite e. apply idpath.
     -  apply (is x').
   }
-  assert (is' : ∏ x : X, hexists (λ x0 : X, eqset (opp x0 x) un0)).
+  assert (is' : ∏ x : X, hexists (λ x0 : X, (opp x0 x) = un0)).
   {
     intro x. apply (λ f : _ , hinhuniv f (is x)). intro s1.
     destruct s1 as [ x' eq ]. apply hinhpr. split with x'. simpl.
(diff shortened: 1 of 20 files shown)


benediktahrens:
I agree that 
1. this is an improvement
2. does not alter the nature of Foundations.

LGTM

peterlefanulumsdaine:
Noting for visibility: the failing check in the CI is just [check for change to Foundations](https://travis-ci.org/github/UniMath/UniMath/jobs/765940971#L1471) — [everything else succeeds](https://travis-ci.org/github/UniMath/UniMath/builds/765940970).

DanGrayson:
Thanks, Peter, that looks good!

peterlefanulumsdaine:
A quick note, by the way, on the changes this may allow and/or necessitate downstream:

- Most code doesn’t need changing
- If you have opened the scope `logic` but are using type-valued equality `paths` at the same time, you will need to either add type annotations `(x = y)%type`, or else (if you use a lot of `paths`) open scope `type` after opening `logic`.
- The one case where this change causes nasty scope-annotation that wasn’t necessary before: when you want `=` generally interpreted as `paths`, but `∃!` interpreted as `iscontr_hProp`, with `paths` inside the `∃!`.  This now requires a double scope annotation, e.g. `(∃! AA : extension Γ0, (ext_extension Γ0 AA = Γ)%type)%logic`  (from https://github.com/UniMath/TypeTheory/commit/230e419572e5d2e34410d8da5cca4aa451448968 ).
- On the plus side, grepping for occurrences of `eqset`, `neqset`, `%set`, and `Open Scope set` (among other things) may find places where the new scoping allows you to improve readability by just using `=` in `logic`.

PR #1345, opened 2021-04-07
Improve reduction behaviour of funcomp and idfun

peterlefanulumsdaine:
A frequent cause of minor annoyance was the need to manually unfold `funcomp` and `idfun` in a lot of places.  This commit aims to reduce that, by tweaking their reduction behaviour so that they automatically unfold more easily (but not too easily):
```
(** Make [simpl], [cbn], etc. unfold [idfun X x] but not [ idfun x ]: *)
Arguments idfun _ _ / .

(** Make [simpl], [cbn], etc. unfold [ (f ∘ g) x ] but not [ f ∘ g ]: *)
Arguments funcomp {_ _ _} _ _ _ /.
```

This allows most instances of `unfold funcomp` and `unfold idfun` to be either removed entirely or replaced by `simpl`/`cbn` — much easier when writing proofs.  Before this change, the library had 132 instances of `unfold funcomp` and 95 of `unfold idfun`; afterwards, there are 42 and 5 respectively (and about a third of those are in `Foundations`, where I didn’t try to change/remove them).

When removing/changing uses of `unfold`, I was careful to avoid hurting both human-readability of the interactive proof states, and overall compilation speed.

This reduction tweak is an upstreaming from https://github.com/UniMath/TypeTheory/tree/master/TypeTheory .  As the maintainers may have noticed, I’m working on upstreaming a lot of things from there at the moment.  To try to keep the reviewing burden as easy as possible, I’m doing most of the miscellaneous upstreaming in a single branch, but for larger and potentially more impactful changes like this one, I’m splitting them out into separate branches+PR’s.  I hope that sounds good to the maintainers — please let me know if you’d like me to work differently, either consolidating or splitting up the PR’s more.

### DIFF
--- a/UniMath/Algebra/BinaryOperations.v
+++ b/UniMath/Algebra/BinaryOperations.v
@@ -33,7 +33,6 @@
 (** Imports *)
 
 Require Export UniMath.Foundations.Sets.
-Require Import UniMath.MoreFoundations.PartA.
 
 Local Open Scope logic.
 
--- a/UniMath/Algebra/BinaryOperations.v
+++ b/UniMath/Algebra/BinaryOperations.v
@@ -33,6 +33,7 @@
 (** Imports *)
 
 Require Export UniMath.Foundations.Sets.
+Require Import UniMath.MoreFoundations.PartA.
 
 Local Open Scope logic.
 
@@ -1385,7 +1386,7 @@ Lemma isbinopfuncomp {X Y Z : setwithbinop} (f : binopfun X Y) (g : binopfun Y Z
   isbinopfun (funcomp (pr1 f) (pr1 g)).
 Proof.
   set (axf := pr2 f). set (axg := pr2 g).
-  intros a b. unfold funcomp.
+  intros a b. simpl.
   rewrite (axf a b). rewrite (axg (pr1 f a) (pr1 f b)).
   apply idpath.
 Defined.
@@ -1527,7 +1528,7 @@ Proof.
   apply (isincltwooutof3a (λ x0 : X, op x x0) f (pr2 (pr1 f))).
   assert (h : homot (funcomp f (λ y0 : Y, op (f x) y0)) (funcomp (λ x0 : X, op x x0) f)).
   {
-    intro x0. unfold funcomp. apply (pathsinv0 ((pr2 f) x x0)).
+    intro x0; simpl. apply (pathsinv0 ((pr2 f) x x0)).
   }
   apply (isinclhomot _ _ h).
   apply (isinclcomp f (make_incl _ is)).
@@ -1540,7 +1541,7 @@ Proof.
   apply (isincltwooutof3a (λ x0 : X, op x0 x) f (pr2 (pr1 f))).
   assert (h : homot (funcomp f (λ y0 : Y, op y0 (f x))) (funcomp (λ x0 : X, op x0 x) f)).
   {
-    intro x0. unfold funcomp. apply (pathsinv0 ((pr2 f) x0 x)).
+    intro x0; simpl. apply (pathsinv0 ((pr2 f) x0 x)).
   }
   apply (isinclhomot _ _ h). apply (isinclcomp f (make_incl _ is)).
 Defined.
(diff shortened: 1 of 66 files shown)


benediktahrens:
> This reduction tweak is an upstreaming from https://github.com/UniMath/TypeTheory/tree/master/TypeTheory . As the maintainers may have noticed, I’m working on upstreaming a lot of things from there at the moment. To try to keep the reviewing burden as easy as possible, I’m doing most of the miscellaneous upstreaming in a single branch, but for larger and potentially more impactful changes like this one, I’m splitting them out into separate branches+PR’s. I hope that sounds good to the maintainers — please let me know if you’d like me to work differently, either consolidating or splitting up the PR’s more.

This PR looks great, thanks a lot!

Yes, it is best to have one PR per change, if the change is substantial. Then uncontroversial changes can be pulled immediately, and controversial ones can be discussed and modified without impacting other changes.

benediktahrens:
> This allows most instances of `unfold funcomp` and `unfold idfun` to be either removed entirely or replaced by `simpl`/`cbn` — much easier when writing proofs. Before this change, the library had 132 instances of `unfold funcomp` and 95 of `unfold idfun`; afterwards, there are 42 and 5 respectively (and about a third of those are in `Foundations`, where I didn’t try to change/remove them).

I am not sure that the restriction on changing `Foundations` applies to such small changes. I would be in favour of moving the `Arguments idfun...` et al into `Foundations`. The arguments for this are multifold:
1. The resulting terms remain the same modulo convertibility.
2. It is confusing to have different behaviour of the same constants in different places of the library. (More strongly, we could discourage changing the argument and reduction behaviour of a constant anywhere other than right after its definition.)

What do you think, @DanGrayson ?

@peterlefanulumsdaine : I don't mean to ask you to do even more work - I would be happy to do this change myself, if we can agree to accept it.

peterlefanulumsdaine:
> I am not sure that the restriction on changing Foundations applies to such small changes. I would be in favour of moving the Arguments idfun... et al into Foundations. 

Yes, I agree with moving the annotation into `Foundations` if you think that’s OK, and I can do that very easily and quickly.

> Yes, it is best to have one PR per change, if the change is substantial. Then uncontroversial changes can be pulled immediately, and controversial ones can be discussed and modified without impacting other changes.

Great — that was exactly my thinking, glad it sounds good to you!

DanGrayson:
This is an improvement, Peter, and I think it's a Foundations-friendly one, so please feel free to move the commands to just after the corresponding functions are defined and to remove any uses of `unfold` you wish to.

It's a good idea to push techniques useful to others upstream from your repository -- we'll benefit from that.

I was unaware of the syntax.  The manual says this: "the function will be unfolded only if it’s applied to at least the arguments appearing before the /".   Would it be better phrased as "the function will be unfolded by simpl if it’s applied to the arguments appearing before the /"?  Or what?  That can't be exactly right, because you have, in some cases, simply removed the unfold command without replacing it by "simpl".  How does that work?

I know in my own work, I regard "simpl" as unpredictable and perhaps time consuming, whereas if I can unfold one thing, that seems precise and quick, so I might still prefer "unfold" to "simpl".  I'm glad that you have been paying attention also to speed.

About the frequency and size of pull requests, I like the way you have been doing it.  It is good to have pull requests devoted to a single topic.  It makes them easier to review.  Additionally, if a pull request consists of several unrelated parts, it is hard to reject just one part, especially because "reverting" a commit screws up the history.

I looked at https://github.com/UniMath/TypeTheory -- you may wish to add the names of the authors to its file `README.md`.

peterlefanulumsdaine:
@benediktahrens OK, I’ve moved the annotations up to `Foundations` as you suggest.  I have no other pending changes, so I’m good for this to be merged if the maintainers are both happy!

And I agree, it seems generally good practice that argument annotations for a function should go at its site of definition — or at least, if made elsewhere, they should be kept `Local`.

peterlefanulumsdaine:
Just to confirm again, btw — the CI red light is just due to the “no changes in `Foundations`” check failing.

DanGrayson:
Thanks!

PR #1346, opened 2021-04-10
Provide better [maponpaths] lemmas for multi-argument functions, and unify duplicates elsewhere

peterlefanulumsdaine:
The library (before this PR) contains quite a lot of duplicate lemmas providing [maponpaths]-stype principles for multiple-argument functions.

Some of these are special cases that are worth specifically naming (e.g. `base_paths`, `cancel_precomposition`), but most seem like just redundant code duplication, hand-rolled each time because suitable general lemmas weren’t available or discoverable at the time of writing.

In this PR, I have:

1. provided a family of uniformly-named [maponpaths]-style lemmas for multi-argument functions, in `MoreFoundations.PartA`, upstreamed from https://github.com/UniMath/TypeTheory/

2. refactored away special cases/duplicates elsewhere, when they seemed like just unnecessary duplication

3. when special cases elsewhere seemed clearly intended/justified, added comments documenting their added value

A couple of points remain slightly unsatisfactory after this PR.  In particular, `Foundations` itself contains a pair of duplicate functions `map_on_two_paths`, `two_arg_paths`, and this family adds a third, `maponpaths_12` (to fit the family’s general naming convention).  I’ve agonised over what to do about this, and it seems difficult — the desiderata “avoid duplication”, “have consistently-named family of lemmas”, and “don’t change `Foundations`” are all in conflict here.  So to avoid blockage/bikeshedding, I suggest we defer it to a separate issue after this PR is merged, if the maintainers agree that this PR is an improvement on the current situation.

### DIFF
--- a/UniMath/CategoryTheory/CategoriesWithBinOps.v
+++ b/UniMath/CategoryTheory/CategoriesWithBinOps.v
@@ -28,12 +28,6 @@ Section def_precategory_with_binops.
   (** Gives the binop of the homs from x to y. *)
   Definition to_binop {BC : precategoryWithBinOps} (x y : BC) : binop (BC⟦x, y⟧) := (pr2 BC) x y.
 
-  Lemma to_binop_eq {BC : precategoryWithBinOps} {x y : BC} {f1 f2 g1 g2 : x --> y}
-        (e1 : f1 = f2) (e2 : g1 = g2) : to_binop _ _ f1 g1 = to_binop _ _ f2 g2.
-  Proof.
-    induction e1, e2. apply idpath.
-  Qed.
-
 End def_precategory_with_binops.
 
 Definition oppositePrecategoryWithBinOps (M : precategoryWithBinOps) : precategoryWithBinOps
--- a/UniMath/CategoryTheory/Chains/OmegaCocontFunctors.v
+++ b/UniMath/CategoryTheory/Chains/OmegaCocontFunctors.v
@@ -55,6 +55,7 @@ Require Import UniMath.Foundations.PartD.
 Require Import UniMath.Foundations.Propositions.
 Require Import UniMath.Foundations.Sets.
 Require Import UniMath.Foundations.NaturalNumbers.
+Require Import UniMath.MoreFoundations.PartA.
 Require Import UniMath.MoreFoundations.Tactics.
 
 Require Import UniMath.CategoryTheory.Core.Categories.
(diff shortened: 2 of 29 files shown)


benediktahrens:
LGTM

Regarding `map_on_two_paths`, `two_arg_paths`, and `maponpaths_12`: the first two probably do not come from Vladimir. Also, Vladimir would likely appreciate a consistent naming scheme for the variants of `maponpaths`. Also, he detests duplication. Thus, I would argue for moving `maponpaths_12` into `Foundations` and eliminating the other two variants. (I know that you don't want to discuss this as part of this PR, but maybe @DanGrayson agrees and then there would be no lengthy discussion at all. Otherwise we suspend this discussion.)

peterlefanulumsdaine:
I’d be very happy to move `maponpaths_12` up to `Foundations` and consolidate it with `two_arg_paths` and `map_on_two_paths` as you suggest (providing @DanGrayson agrees, of course), but I would still slightly prefer to do it in a separate PR.  The consolidation will take a bit of work and thought, and possibly discussion, so it’d be nice to keep it independent of other things that build on this PR.

DanGrayson:
Thanks, Peter.   

`map_on_two_paths` comes from Vladimir, commit #30f3ff57d

`two_arg_paths` comes from me, commit #49213394d, later (oops, they're the same!)

It would be good to have just one of them, as your pull request accomplishes.

It would be good to preserve Vladimir's name as a synonym, at least, since he liked the names he chose.

In this PR or another one, as you wish.  Just let us know.

peterlefanulumsdaine:
If it’s OK with you both, I’d prefer to have this one merged now, and then open an issue about the further unification (which can be assigned to me).  I have a couple of other refactoring branches on the go which depend on this PR, and I’d like to be able to continue working on them for the moment rather than getting them blocked on dealing with the further unification (though I should be able to come back to it soon).

benediktahrens (approved):

benediktahrens:
I am interpreting @DanGrayson 's comment as an approval for this PR, merging.

benediktahrens:
@peterlefanulumsdaine : I reintroduced some of the `*_eq` lemmas in another library using UniMath/UniMath (see https://github.com/UniMath/largecatmodules/commit/947215575c03938d2b70200eb7cba38a2b454ff0), since I could not get the `maponpaths_*` variants to work there. If you have any advice, I would be very happy to know.

PR #1348, opened 2021-05-20
Fixed point theorems on posets

peterlefanulumsdaine:
This PR develops some fixed-point theorems on posets, along with the background constructions they depend on — in particular, completeness properties and function posets.

The main results are on the interplay of chain-completeness, directed-completeness, and monotone and progressive endo-maps: these are the classical Bourbaki–Witt and Knaster–Tarski fixed-point theorems, plus Pataraia’s (constructive) fixed-point theorem.


DanGrayson:
Thanks, it looks good to me!  But 2 remarks: 

You mention moving various results upstream later.  Why not move them upstream now?

Does this stuff count as "Algebra"?  It seems more basic.  I think of algebra as involving binary operations such as plus and times.  Maybe the whole thing could go upstream.

peterlefanulumsdaine:
Re the upstreaming: yes, I probably should do much of it now.  There are a few things that I think may be a bit more involved or warrant discussion, which was why I put off the upstreaming.  But honestly that was partly just procrastination — most of it isn’t so involved, so I should just do it.  Will try to get to it soon, and update this PR!

Re `Algebra`: I agree, I wouldn’t quite count this as “algebra”.  I put this in `Algebra` to follow the existing `Algebra.Dcpo` — that’s the closest existing material, and also this file uses (a few things from) that.  The other main development of posets is in `Combinatorics.WellOrderedSets` — though I wouldn’t really think of it as `Combinatorics` either.  Overall I’d support reorganising/consolidating/moving these files, and I’d be happy to do that, but I think it definitely wants a bit of discussion and should be left to a subsequent PR.

DanGrayson:
Okay, we can discuss it later.  Or maybe I'll just do it.

Maybe Dcpo.v should be moved, too.  By the way, "Dcpo" is a horrible name for a file.

Thanks!

Benedikt, merging this is okay with me.

UniMath/Algebra/FixedPointTheorems.v
```diff
@@ -0,0 +1,1414 @@
...
+  Defined.
+
+End LowerBounds.
+
+Section UpperBounds.
+(* NOTE: this should be kept in sync with [Section LowerBounds] above. *)
+
+  (* TODO: it would be natural to just refactor [isupperbound] in [Algebra.Dcpo]
+  as an hProp from the start, using [∀] and so on.  However, that turns out
+  to slightly interfere with the use of [apply] on lemmas in that file, espcially
```
benediktahrens, line 331:
```suggestion
  to slightly interfere with the use of [apply] on lemmas in that file, especially
```

benediktahrens (approved):

Issue #1324, opened 2020-06-12
github actions instead of travis


DanGrayson:
We've just started using "github actions" for Macaulay2, see our first result here: https://github.com/Macaulay2/M2/actions/runs/132744336 -- much more CPU time is provided than with travis, so we may want to switch and stop using a docker image, which adds a complication.

DanGrayson:
"GitHub runs jobs in parallel, by default, up to a limit of 20 for a free account. (Note that there is further limit of 5 for macOS jobs.) Each job can run for up to 6 hours, and the workflow itself is limited to 72. Details [here](https://help.github.com/en/actions/getting-started-with-github-actions/about-github-actions#usage-limits)."

benediktahrens:
That sounds good! If I understand correctly, the action is also configured via a .yml file that is part of the repository, and the file format is similar to that for travis?

If UniMath/UniMath transitions to a new CI, it would make sense to transition also UniMath/TypeTheory and UniMath/largecatmodules.

DanGrayson:
Yes, here's our action file now: https://github.com/Macaulay2/M2/blob/master/.github/workflows/test_build.yml

benediktahrens:
When making the switch, we could also rename the `master` branch: https://www.theregister.com/2020/06/15/github_replaces_master_with_main/
I would suggest `trunk` instead of `main`.

DanGrayson:
It could cause confusion when people do "git pull", and it no longer works.

peterlefanulumsdaine:
Just for reference, here’s the issue thread from [the HoTT library](https://github.com/HoTT/HoTT) for its migration from Travis to Github Actions: https://github.com/HoTT/HoTT/issues/1247

benediktahrens:
FYI: 

1. There is a documentation page at https://docs.github.com/en/actions/learn-github-actions/migrating-from-travis-ci-to-github-actions explaining some of the differences between travis and github actions.
2. We will need to switch from travis-ci.org to travis-ci.com if we want to continue using travis, see https://blog.travis-ci.com/2021-05-07-orgshutdown.

@DanGrayson : what do you think, is it time to switch to Github actions? It might be sufficient to slightly adapt the travis configuration file and move it to a subdirectory. There is also a quickstart button on https://github.com/UniMath/UniMath/actions/new.

benediktahrens:
[...]

PR #1352, opened 2021-05-29
First draft for Github action

benediktahrens:
This is a very naive first draft for a github action. It is based on the installation instructions for Debian/Ubuntu, and I have not tested them - I would not know how to test them.

relates to #1324

### DIFF
--- a/.github/workflows/build-unimath.yml
+++ b/.github/workflows/build-unimath.yml
@@ -29,20 +29,15 @@ jobs:
       # Install everything needed for building UniMath (and more)
       - name: Install build dependencies
         run: sudo apt-get install build-essential git ocaml ocaml-nox ocaml-native-compilers camlp5 libgtk2.0 libgtksourceview2.0 liblablgtk-extras-ocaml-dev ocaml-findlib libnum-ocaml-dev emacs
-  
-      # Change into UniMath directory and run sanity checks
-      - name: Change into UniMath directory and run sanity checks
-        run: |
-          cd $GITHUB_WORKSPACE
-          time make TIMECMD=time sanity-checks
-         
+
       # Change into UniMath directory and build UniMath
       - name: Build UniMath
         run: |
           cd $GITHUB_WORKSPACE
           time make build-coq BUILD_COQIDE=$BUILD_COQIDE && time make TIMECMD=time
-          
-          
-          
-          
-  
+
+      # Change into UniMath directory and run sanity checks
+      - name: Change into UniMath directory and run sanity checks
+        run: |
+          cd $GITHUB_WORKSPACE
+          time make TIMECMD=time sanity-checks
--- a/.github/workflows/build-unimath.yml
+++ b/.github/workflows/build-unimath.yml
@@ -28,7 +28,7 @@ jobs:
       
       # Install everything needed for building UniMath (and more)
       - name: Install build dependencies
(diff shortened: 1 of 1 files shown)


.github/workflows/build-unimath.yml
```diff
@@ -0,0 +1,57 @@
...
+
+      # Runs a set of commands using the runners shell
+      - name: Run a multi-line script
+        run: |
+          echo Add other actions to build,
+          echo test, and deploy your project.
+      
+      # Install everything needed for building UniMath (and more)
+      - name: Install build dependencies
+        run: sudo apt-get install build-essential git ocaml ocaml-nox ocaml-native-compilers camlp5 libgtk2.0 libgtksourceview2.0 liblablgtk-extras-ocaml-dev ocaml-findlib emacs
```
benediktahrens, line 40:
```suggestion
        run: sudo apt-get install build-essential git ocaml ocaml-nox ocaml-native-compilers camlp5 libgtk2.0 libgtksourceview2.0 liblablgtk-extras-ocaml-dev ocaml-findlib libnum-ocaml-dev emacs
```

.github/workflows/build-unimath.yml
```diff
@@ -0,0 +1,57 @@
...
+
+      # Runs a single command using the runners shell
+      - name: Run a one-line script
+        run: echo Hello, world!
+
+      # Runs a set of commands using the runners shell
+      - name: Run a multi-line script
+        run: |
+          echo Add other actions to build,
+          echo test, and deploy your project.
```
benediktahrens, line 36:
```suggestion
```

benediktahrens:
@DanGrayson : 

Maybe an improvement would be to split the current job "build" into two, one for doing the sanity checks, and one for the actual building? 

Running the sanity checks builds Coq, for some reason, and I am not sure if the Coq built for the sanity checks would be persistent across jobs and be available for the "build" job. Worth checking in case we split the jobs.

benediktahrens:
Here is more explanation of jobs depending on other jobs: https://docs.github.com/en/actions/learn-github-actions/managing-complex-workflows

We do not want the "build" job to depend on the "sanity check" job: sanity-checking includes checking for changes in Foundations, which we do want to allow in some cases.

Different jobs that are not dependent on each other run in parallel, so using the Coq built during job "sanity check" is not possible when running job "build".

We could make the sanity check depend on "build", but it seems a good idea to run both checks on each PR/commit in any case. So maybe the current setup is the best?

DanGrayson:
Having 1 job should be enough.   Why don't we just put the sanity check last?  Then the proofs will get checked, too, even if the sanity test doesn't succeed?  The proof checking step will build coq.

benediktahrens:
> Having 1 job should be enough. Why don't we just put the sanity check last? Then the proofs will get checked, too, even if the sanity test doesn't succeed? The proof checking step will build coq.

Yes, good idea - done in the latest commit.

benediktahrens:
@DanGrayson : is this ready for being merged? Once this is done, I'll deactivate travis on this repository.

DanGrayson:
Something went wrong here:

https://github.com/UniMath/UniMath/pull/1352/checks?check_run_id=2738595713#step:5:12

DanGrayson:
I'll fix that later...

DanGrayson:
I think I've disabled travis now, too.

PR #1357, opened 2021-06-06
make with 2 jobs on github action

benediktahrens:
The runners have 2 cores, so we could speed this up a bit: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

However, Coq sometimes does not like being compiled with more than 1 job, so I am leaving it as is.

### DIFF
--- a/.github/workflows/build-unimath.yml
+++ b/.github/workflows/build-unimath.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Build UniMath
         run: |
           cd $GITHUB_WORKSPACE
-          time make build-coq BUILD_COQIDE=$BUILD_COQIDE && time make TIMECMD=time
+          time make build-coq BUILD_COQIDE=$BUILD_COQIDE && time make -j2 TIMECMD=time
 
       # Change into UniMath directory and run sanity checks
       - name: Change into UniMath directory and run sanity checks


benediktahrens:
There is still a Travis residue in this PR's status message. The Github Action succeeds, the Travis action seems to be hanging. 

I have suspended Travis' access to the repositories in the UniMath organization - how about uninstalling it completely?

DanGrayson:
> There is still a Travis residue in this PR's status message. The Github Action succeeds, the Travis action seems to be hanging.
> 
> I have suspended Travis' access to the repositories in the UniMath organization - how about uninstalling it completely?

I tried that, but it warned that it would affect all 16 repositories, so it seems to be a bad idea.

DanGrayson:
> > There is still a Travis residue in this PR's status message. The Github Action succeeds, the Travis action seems to be hanging.
> > I have suspended Travis' access to the repositories in the UniMath organization - how about uninstalling it completely?
> 
> I tried that, but it warned that it would affect all 16 repositories, so it seems to be a bad idea.

We should check whether the suspension affects all 16 repositories.

benediktahrens:
> We should check whether the suspension affects all 16 repositories.

It does affect all 16 repositories, but from what I see, only UniMath/Schools still relies on it - and I am planning to switch that one over to github actions.

Travis has been horribly unreliable anyway.

benediktahrens:
Obsolete, closing.

PR #1358, opened 2021-06-06
Github Action: build TypeTheory and largecatmodules

benediktahrens:


benediktahrens:
I am confused by this error in the building of UniMath when building the other libraries on top:
```
COQC UniMath/Algebra/Universal/Signatures.v
COQC UniMath/Algebra/Universal/Algebras.v
COQC UniMath/Algebra/Universal/Examples/Nat.v
File "./UniMath/Algebra/Universal/Examples/Nat.v", line 13, characters 15-46:
Error: Unable to locate library UniMath.Algebra.Universal.Terms.

make[2]: *** [build/CoqMakefile.make:716: UniMath/Algebra/Universal/Examples/Nat.vo] Error 1
make[1]: *** [build/CoqMakefile.make:339: all] Error 2
make[1]: Leaving directory '/home/runner/work/UniMath/UniMath'
make: *** [Makefile:90: all] Error 2
```
@DanGrayson : any idea what is happening here?

DanGrayson:
What if we build those repositories in actions attached to those repositories, instead?

benediktahrens:
> What if we build those repositories in actions attached to those repositories, instead?

Is it possible to trigger github actions in different repositories, and have the results reported back to the repository that has triggered the build? That would be good, but it sounds much more complicated to set up.

DanGrayson:
> > What if we build those repositories in actions attached to those repositories, instead?
> 
> Is it possible to trigger github actions in different repositories, and have the results reported back to the repository that has triggered the build? That would be good, but it sounds much more complicated to set up.

I don't know a way.  Such builds could, however, happen daily or weekly.  Oh, well, let's try it this way.

PR #1363, opened 2021-06-11
Displayed inserter and 2-inserter

benediktahrens:


benediktahrens:
UniMath build works, problem in satellite libs unrelated (since satellites do not use Bicategories package). Merging.

PR #1373, opened 2021-07-19
The beginning of the definition of the structure sheaf

timcherganov:
This PR defines the map on open sets of Spec R that will be used in the definition of the structure sheaf.
#1316

### DIFF
--- a/UniMath/Algebra/RigsAndRings/Ideals.v
+++ b/UniMath/Algebra/RigsAndRings/Ideals.v
@@ -11,7 +11,9 @@ Author: Langston Barrett (@siddharthist)
   - Two-sided ideals ([ideal])
   - The above notions coincide for commutative rigs
 - Kernel ideal
+- Unit ideal
 - Prime ideal
+- Localization at a prime ideal
  *)
 
 Require Import UniMath.Algebra.RigsAndRings.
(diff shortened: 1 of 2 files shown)


UniMath/Algebra/RigsAndRings/Ideals.v
```diff
@@ -115,17 +117,73 @@ Proof.
...
 Defined.
 
+(** ** Unit ideal *)
+
+Lemma ideal_rigunel2 {R : rig} (I : ideal R) : I 1 -> forall x, I x.
+Proof.
+  intros H x.
+  apply (transportf (λ x, I x) (rigrunax2 _ x)).
+  exact (ideal_isl _ _ _ H).
+Defined.
```
benediktahrens, line 127: Would it be preferable to `Qed` this lemma?

UniMath/Algebra/RigsAndRings/Ideals.v
```diff
@@ -115,17 +117,73 @@ Proof.
...
+Section prime_facts.
+  Context {R : commring} (p : prime_ideal R).
+
+  Definition prime_ideal_ax1_contraposition :
+    ∀ a b : R, ¬ p a ⇒ ¬ p b ⇒ ¬ p (a * b).
+  Proof.
+    intros a b Ha Hb.
+    apply (negf (prime_ideal_ax1 p a b)), toneghdisj.
+    exact (make_dirprod Ha Hb).
+  Defined.
```
benediktahrens, line 158: Would it be preferable to `Qed` this lemma?

UniMath/Algebra/RigsAndRings/Ideals.v
```diff
@@ -115,17 +117,73 @@ Proof.
...
+
+  Lemma prime_ideal_rigunel2 : ¬ p 1.
+  Proof.
+    intro H.
+    apply (hexistsnegtonegforall _ (prime_ideal_ax2 p)).
+    exact (ideal_rigunel2 _ H).
+  Defined.
```
benediktahrens, line 165: Would it be preferable to `Qed` this lemma?

UniMath/Algebra/RigsAndRings/Ideals.v
timcherganov: I just followed the coding style:

> - Do not end a proof with Qed., except with Goal, for that may prevent later computations.
> - Start all proofs with Proof. on a separate line and end it with Defined. on a separate line, as this makes it possible for us to generate HTML with expansible/collapsible proofs.

UniMath/Algebra/RigsAndRings/Ideals.v
```diff
@@ -115,17 +117,73 @@ Proof.
...
+Defined.
+
 (** ** Prime ideal *)
 
 Section prime.
   Context {R : commring}.
 
-  Definition is_prime (I : ideal R) : hProp :=
-    (∀ a b, I (a * b) ⇒ I a ∨ I b) ∧ (∃ x, ¬ I x).
+  Definition prime_ideal : UU := ∑ p : ideal R, (∀ a b, p (a * b) ⇒ p a ∨ p b) ∧ (∃ x, ¬ p x).
```
DanGrayson, line 134: Why remove `is_prime`?  It's potentially useful.

UniMath/Algebra/RigsAndRings/Ideals.v
```diff
@@ -115,17 +117,73 @@ Proof.
...
 
-  Definition is_prime_ax1 {I : ideal R} (H : is_prime I) :
-    ∀ a b, I (a * b) ⇒ I a ∨ I b := dirprod_pr1 H.
+  Definition make_prime_ideal (p : ideal R) (H1 : ∀ a b, p (a * b) ⇒ p a ∨ p b) (H2 : ∃ x, ¬ p x) :
+    prime_ideal := p ,, (make_dirprod H1 H2).
```
DanGrayson, line 137: Wouldn't something like `p ,, H1 ,, H2` work?

UniMath/Algebra/RigsAndRings/Ideals.v
```diff
@@ -115,17 +117,73 @@ Proof.
...
+
+  Definition prime_ideal_ax1_contraposition :
+    ∀ a b : R, ¬ p a ⇒ ¬ p b ⇒ ¬ p (a * b).
+  Proof.
+    intros a b Ha Hb.
+    apply (negf (prime_ideal_ax1 p a b)), toneghdisj.
+    exact (make_dirprod Ha Hb).
+  Defined.
+
+  Lemma prime_ideal_rigunel2 : ¬ p 1.
```
DanGrayson, line 160: Wouldn't it be simpler to make `¬ p 1` part of the definition of being prime?  (Replacing `∃ x, ¬ I x`.)

UniMath/Algebra/RigsAndRings/Ideals.v
benediktahrens: Right, I see. The coding style is a bit wrong here, since when everything is transparent (closed with `Defined.`), typechecking becomes very slow eventually, up to the point where it is unfeasible.

Now, it is not always easy to say which results can be closed with `Qed`. A guideline is:
- proofs of `isofhlevel (S n) X` can usually be closed with `Qed`.
- proofs of `iscontr X` should *not* be closed with `Qed`.
- proofs of general propositions (as is the case here) can be closed with `Qed`.
- proofs of equality of objects in an algebraic structure (e.g., group) and of morphisms in a category can be closed with `Qed.`

This is a non-exhaustive list.

UniMath/Algebra/RigsAndRings/Ideals.v
timcherganov: When we create `prime_ideal`, we use `make_prime_ideal` that takes two propositions, if we want to get those propositions, we use `prime_ideal_ax1` and `prime_ideal_ax2`. So I thought there is no need in `is_prime`. But if you think it's useful, I will gladly return it.

UniMath/Algebra/RigsAndRings/Ideals.v
timcherganov: My idea was to use `,,` for dependent sums and `make_dirprod` for direct products (I even tried to find smth like `make_hconj` for conjuntions). Maybe it's unnecessary and we can simply use `,,` everywhere.

UniMath/Algebra/RigsAndRings/Ideals.v
timcherganov: Don't see the difference. For me `∃ x, ¬ I x` sounds slightly more general and that is why maybe a little beat easier to provide to define `prime_ideal`.

UniMath/Algebra/RigsAndRings/Ideals.v
timcherganov: Thank you for clarifying this.

UniMath/Algebra/RigsAndRings/Ideals.v
DanGrayson: I think it's useful -- many textbooks will speak of whether an ideal is prime.

UniMath/Algebra/RigsAndRings/Ideals.v
DanGrayson: Elements of direct products are pairs, too, so the notation `(x,,y)` seems quite intuitive.  (Actually, I wish we could replace the double comma by a single comma throughout.)

UniMath/Algebra/RigsAndRings/Ideals.v
DanGrayson: I'm not sure what you mean by "more general".  The advantage of `¬ p 1` over `∃ x, ¬ I x` is that you don't have to wonder what `x` is.

DanGrayson:
Is the pull request ready now for reconsideration?

timcherganov:
Yes

timcherganov:
> Is the pull request ready now for reconsideration?

DanGrayson:
Looks good to me.

@benediktahrens ?

PR #1377, opened 2021-08-05
Multisorted binding signatures revisited

mortberg:
This (draft) PR revisits the construction of a monad from a multisorted binding signature. The previous approach relied on the slice category `Set / X` and I've been generalizing this to the functor category `[X,C]` where `X` is a discrete category and `C` a category with some mild assumptions (details to be figured out).

I've mainly made this draft so that I can discuss the code easier with @rmatthes and @benediktahrens , so there's no need for detailed code review yet.


mortberg:
Thanks to some help from @rmatthes there are no more admits in this PR and it's ready to be reviewed

rmatthes:
I would feel more sure about the approach if there were also files STLC_alt.v and CCS_alt.v based on the new MultiSorted_alt.v. For STLC, it would be good to check in addition that beta-reduction works fine. We did not do this back then in the submission hurry.

mortberg:
> I would feel more sure about the approach if there were also files STLC_alt.v and CCS_alt.v based on the new MultiSorted_alt.v. For STLC, it would be good to check in addition that beta-reduction works fine. We did not do this back then in the submission hurry.

Sure, I can look into this this week. I'm not sure what you mean by "that beta-reduction works"?

rmatthes:
I meant just that the right-hand side of the beta-reduction rule can be naturally constructed from the ingredients of the left-hand side by using the general mechanism for going from an hss to individual substitution. But I see that there is anyway no risk since the sorts do form a set for this example. So, this all should be a mostly mechanical transcription of what we had earlier - and you have already acquired the experience for this specific transcription.

mortberg:
I managed to write `STLC_alt` without any major problems. Something annoying is that UniMath seems to have trouble figuring out which category structure to use when trying to compose morphisms in iterated functor categories... For example, I have to use `functor_composite` explicitly in https://github.com/UniMath/UniMath/pull/1377/files#diff-140c20a339df9e5d69f34efbfc54346e620822b724fac1866aaa10e6cbd86970R185 . This worked nicer for functor categories of slice categories

UniMath/SubstitutionSystems/STLC_alt.v
```diff
@@ -0,0 +1,243 @@
...
+
+Let STLC_alg : algebra_ob STLC_Functor :=
+  InitialObject STLC_Functor_Initial.
+
+(** The variables *)
+Definition var_map : sortToSet2⟦1,STLC⟧ :=
+  BinCoproductIn1 _ (BinCoprodSortToSet2 _ _) · STLC_mor.
+
+
+(* TODO: upstream? *)
```
mortberg, line 135: @benediktahrens What do you think? I find these versions a lot nicer to use as you usually only have to give one argument. The new name means that it won't clash with the already existing functors. Changing the arguments to the existing ones would probably mean a lot of rewriting of code. So having these version means some duplication, but at a cheap price

benediktahrens: I am a bit torn. Having both is surely confusing for someone who starts using UniMath. Also, the concept of precategories has not proved useful, I think.

The version specialized to categories should probably be the default, but, as you point out, it would be a lot of work to change that now. Maybe we should add it to the (already very long) list of things that can be improved in UniMath.

(Sometimes, such improvements do happen, as in the case of @rmatthes ' overhaul of monoidal categories. Thanks again!)

DanGrayson: > ... Also, the concept of precategories has not proved useful, I think.
> 

I'm always confused by the terminology, as it conflicts with that of the HoTT book.

mortberg: How about I add my `post/pre_comp_functor` to `UniMath.CategoryTheory.whiskering` with a comment explaining the duplication as well as adding an issue about replacing all uses of `post/pre_composition_functor`?

benediktahrens: That would be great!

UniMath/SubstitutionSystems/MonadsMultiSorted.v
```diff
@@ -407,25 +407,25 @@ Proof.
...
-  set (rs1 := subst_slice_eqn (mweak_slice _ L) (mexch_slice M) (mweak_slice_ok _ L)).
+  set (rs1 := subst_slice_eqn (mweak_slice _ _ L) (mexch_slice M) (mweak_slice_ok _ L)).
   set (rs2 := subst_slice L N).
   simpl in rs1.
 
-Problem: mweak_slice is not an instance of bind_slice, and rewriting is not possible since also
-mweak_slice_ok appears in the term.
+(* Problem: mweak_slice is not an instance of bind_slice, and rewriting is not possible since also *)
+(* mweak_slice_ok appears in the term. *)
+Admitted.
```
benediktahrens, line 428: This `Admitted` is in a comment, isn't it? (Just to be sure...)

benediktahrens:
@mortberg : should the code be merged to stay in sync with master?

UniMath/SubstitutionSystems/MonadsMultiSorted.v
mortberg: Yes

mortberg:
> @mortberg : should the code be merged to stay in sync with master?

Sure! I'll do the few things that were left to do in another PR

PR #1378, opened 2021-08-23
Avoid warnings on implicit arguments and some other warnings at compilation time

rmatthes:
The last of the three commits also touches on Foundations - it could be applied manually, and then I could undo that last commit.


benediktahrens (approved):
Thanks for preparing this PR!

PR #1380, opened 2021-08-27
pre-strength used more consequently, and more in this spirit

rmatthes:
in particular, not only are the strength laws not necessary for the construction of the bracket operation, but one needs the strength only for fixed second parameter (the pointed Z)
this could have been implemented ages ago but better late than never

Please comment and then merge if it looks fine.

### DIFF
--- a/UniMath/CategoryTheory/Monads/Monads.v
+++ b/UniMath/CategoryTheory/Monads/Monads.v
@@ -102,6 +102,33 @@ Proof.
     + apply nat_trans_eq; try exact hs. exact H3.
 Qed.
 
+Let T0' := (functor_from_functor_with_μ C T): EndC.
+Let η' := η: EndC⟦functor_identity C, T0'⟧.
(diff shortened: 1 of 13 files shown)


rmatthes:
@benediktahrens : Thank you for your review. In view of the nature of the changes, it would not even be necessary to await the outcome of the checking.

benediktahrens (approved):
I don't understand why the left unitor is not required any more. Can you explain briefly?
EDIT: Sorry, I meant  `ρ_functor` - why is that not necessary any more?

rmatthes:
> I don't understand why the left unitor is not required any more. Can you explain briefly?
> EDIT: Sorry, I meant `ρ_functor` - why is that not necessary any more?

This should now be clear from PR #1381. We knew this before: Functor T can be replaced by T dot Id and Id dot T in the source and the target of hom-sets in functor categories, w.r.t. convertibility.

PR #1381, opened 2021-08-30
more clarity for monoidal isos in functor category

rmatthes:
NaturalTransformations.v: take identity instead of pointwise identity in two lemmas
UnitorsAndAssociatorsForEndofunctors.v:
- verify that the types of the isos are convertible with those of identity (formulated a bit more generally)
- explicitly define all those isos as the identity
- show the motivational equality lemmas w/o homset hypotheses (the comments pointed to obsolete files)
- rename these isos to have names of the form XX_functors[_inv] instead of with _functor (those are the monoidal isos for functors, they are not functors themselves)
the other nine files only suffer those identifier changes

### DIFF
--- a/UniMath/CategoryTheory/CategoricalRecursionSchemes.v
+++ b/UniMath/CategoryTheory/CategoricalRecursionSchemes.v
@@ -66,10 +66,10 @@ Definition comp_distr_laws {C C' C'' D D' D'' : precategory}{F : functor C D}{F'
   DistrLaw F F'' (H' ∙ H ) (K' ∙ K).
 Proof.
   red.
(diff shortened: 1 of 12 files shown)


benediktahrens (approved):
LGTM

rmatthes:
@benediktahrens : Thank you. By the way, since you had already merged the previous PR #1380, I could make good use of the new `git switch -c `, with the name of my new branch as argument.

rmatthes:
The failure for largecatmodules will be trivial to fix - it is only the change of identifiers.

rmatthes:
Failure for largecatmodules fixed in [PR#131 there](https://github.com/UniMath/largecatmodules/pull/131).

PR #1384, opened 2021-09-03
WIP: the subcategory of omega-cocontinuous functors

rmatthes:
only first steps to fix definitions and to detect shortcomings of the UniMath library


UniMath/CategoryTheory/limits/graphs/colimits.v
```diff
@@ -687,6 +687,21 @@ Section Reflects.
 
 End Reflects.
 
+Section Creates.
+
+  Context {C D : precategory} (F : functor C D).
+
+  Definition creates_colimit_data {g : graph} (d : diagram g C) : UU :=
```
benediktahrens, line 694: This definition does not align with the one given in terms of displayed categories here:
https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/DisplayedCats/Limits.v#L38

In short, the definition there says that a displayed category creates limits if
1. "above" every limiting cone there is exactly one cone, and
2. that cone is limiting.
"Above" is conveniently expressed in terms of displayed categories, and would need an equality (or an isomorphism) on objects if expressed in terms of functors.

benediktahrens: Addendum: the definition of creation of limits does not immediately say that limits are preserved (whereas reflection of limits is more immediate). Indeed, preservation of limits requires a small argument, given at https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/DisplayedCats/Limits.v#L60.

rmatthes:
I have the bad feeling that in our situation, colimits are not preserved, and so, we should not go for creation of colimits in the sense of displayed categories (that generalizes the situation of full subcategories).

rmatthes:
Commit  [7fe3c26d5](https://github.com/UniMath/UniMath/pull/1384/commits/7fe3c26d5a6f1) will be separated out into a new PR, and the present PR will be paused for further insights from category theory.

PR #1385, opened 2021-09-04
Sheaf of rings on Spec R

timcherganov:
#1316

### DIFF
--- a/UniMath/Algebra/RigsAndRings/Ideals.v
+++ b/UniMath/Algebra/RigsAndRings/Ideals.v
@@ -64,11 +64,19 @@ Section Definitions.
   Definition ideal_isl (I : ideal) : is_lideal I := pr12 I.
 
   Definition ideal_isr (I : ideal) : is_rideal I := pr22 I.
+
+  Lemma isaset_ideal : isaset ideal.
+  Proof.
+    apply isaset_total2.
+    - apply isaset_submonoid.
+    - intro S. apply isasetaprop, propproperty.
+  Qed.
 End Definitions.
 
 Arguments lideal _ : clear implicits.
 Arguments rideal _ : clear implicits.
 Arguments ideal _ : clear implicits.
+Arguments isaset_ideal _ : clear implicits.
 
 (** *** The above notions for commutative rigs *)
 
@@ -152,7 +160,14 @@ Arguments prime_ideal _ : clear implicits.
 Section prime_facts.
   Context {R : commring} (p : prime_ideal R).
 
-  Definition prime_ideal_ax1_contraposition :
+  Lemma isaset_prime_ideal : isaset (prime_ideal R).
+  Proof.
+    apply isaset_total2.
+    - apply isaset_ideal.
+    - intro I. apply isasetaprop, propproperty.
+  Qed.
+
+  Lemma prime_ideal_ax1_contraposition :
     ∀ a b : R, ¬ p a ⇒ ¬ p b ⇒ ¬ p (a * b).
   Proof.
     intros a b Ha Hb.
--- a/UniMath/Algebra/RigsAndRings/Ideals.v
+++ b/UniMath/Algebra/RigsAndRings/Ideals.v
@@ -70,7 +70,7 @@ Section Definitions.
     apply isaset_total2.
     - apply isaset_submonoid.
     - intro S. apply isasetaprop, propproperty.
-  Qed.
+  Defined.
 End Definitions.
 
 Arguments lideal _ : clear implicits.
--- a/UniMath/AlgebraicGeometry/Sheaf.v
+++ b/UniMath/AlgebraicGeometry/Sheaf.v
@@ -1,0 +1,70 @@
+Require Import UniMath.Algebra.RigsAndRings.
+Require Import UniMath.MoreFoundations.Subtypes.
+Require Import UniMath.Topology.Topology.
+Require Import UniMath.CategoryTheory.Core.Categories.
+Require Import UniMath.CategoryTheory.Core.Functors.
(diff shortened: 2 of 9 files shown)


UniMath/AlgebraicGeometry/Sheaf.v
```diff
@@ -0,0 +1,70 @@
...
+  Definition open_po : po (@Open X) :=
+    make_po _ (isporesrel _ _ (subtype_containment_ispreorder X)).
+
+  Definition open_category : category := po_category open_po.
+End open_category.
+
+
+Section sheaf_commring_prop.
+  Context {X : TopologicalSet}
+          (F' : functor_data (open_category X)^op commring_precategory).
```
DanGrayson, line 31: Why is F' just functor_data and not a functor?

UniMath/AlgebraicGeometry/Sheaf.v
```diff
@@ -0,0 +1,70 @@
...
+
+  Definition F (U : Open) : commring := F' U.
+
+  Definition restriction {U V : Open} (H : U ⊆ V) : ringfun (F V) (F U) := #F' H.
+
+  Definition restrict {A : hsubtype Open} (f : F (⋃ A)) (U : A) : F (pr1 U) :=
```
DanGrayson, line 37: I think later code would be more readable if A were an explicit argument.

UniMath/AlgebraicGeometry/Sheaf.v
```diff
@@ -0,0 +1,70 @@
...
+  Context (X : TopologicalSet).
+
+  Definition open_po : po (@Open X) :=
+    make_po _ (isporesrel _ _ (subtype_containment_ispreorder X)).
+
+  Definition open_category : category := po_category open_po.
+End open_category.
+
+
+Section sheaf_commring_prop.
```
DanGrayson, line 29: It would be desirable to handle sheaves of other types of things, too, not just rings.  Can you think of a way to do that?  If not, maybe the name of this file should be `SheavesOfRings.v` instead.

UniMath/AlgebraicGeometry/Spec.v
```diff
@@ -101,35 +113,37 @@ Section spec.
...
+  Definition spec : TopologicalSpace :=
+    mkTopologicalSpace (make_hSet (prime_ideal R) isaset_prime_ideal)
+                       zariski_topology
+                       zariski_topology_union
+                       zariski_topology_htrue
+                       zariski_topology_and.
 End spec.
 
-Arguments spec_top _ : clear implicits.
+Arguments spec _ : clear implicits.
```
DanGrayson, line 124: Maybe take advantage of the opportunity to capitalize `Spec` here.

UniMath/AlgebraicGeometry/Topology.v
```diff
@@ -0,0 +1,85 @@
+(* Additional definitions and facts from topology. Maybe this should go in UniMath.Topology. *)
+
+Require Import UniMath.MoreFoundations.Subtypes.
+Require Import UniMath.Topology.Topology.
+
+Local Open Scope subtype.
+
+Section top.
+  Definition TopologicalSpace : UU := ∑ X : hSet, isTopologicalSet X.
```
DanGrayson, line 9: Good.  Later on we should phase out "TopologicalSet".

DanGrayson (changes requested):
Thank you for the contribution, it is appreciated.

In this review I make some minor comments.

UniMath/Algebra/RigsAndRings/Ideals.v
```diff
@@ -64,11 +64,19 @@ Section Definitions.
   Definition ideal_isl (I : ideal) : is_lideal I := pr12 I.
 
   Definition ideal_isr (I : ideal) : is_rideal I := pr22 I.
+
+  Lemma isaset_ideal : isaset ideal.
+  Proof.
+    apply isaset_total2.
+    - apply isaset_submonoid.
+    - intro S. apply isasetaprop, propproperty.
+  Qed.
```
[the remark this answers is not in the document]
timcherganov: No particular reason. Do you think it's better to use `Defined` here?

UniMath/AlgebraicGeometry/Sheaf.v
timcherganov: There is no need in functoriality for the definition of locality and gluing.

UniMath/Algebra/RigsAndRings/Ideals.v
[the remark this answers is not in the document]
DanGrayson: I always prefer "Defined" so things compute.  @benediktahrens may have a different opinion.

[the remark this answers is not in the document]
benediktahrens: 1. Sometimes making things opaque is necessary for type-checking to terminate in a reasonable amount of time. 
2. I would be surprised if this particular proof would ever participate in a computation. Also, the particular shape of the proof does not make it seem particularly well-behaved in a computation.

Because of 2, I think it probably does not matter so much whether the proof is opaque or transparent. I would make it opaque myself, but I am fine with either choice.

DanGrayson:
@benediktahrens --

Look at the error: **The term "C" has type "precategory" while it is expected to have type "category".**

benediktahrens:
> @benediktahrens --
> 
> Look at the error: **The term "C" has type "precategory" while it is expected to have type "category".**

This is the error message from compiling the satellite repository https://github.com/UniMath/largecatmodules. I am in the process of fixing the satellites.

Compilation of UniMath/UniMath works fine with this PR.

DanGrayson:
Thank you for the contribution!

PR #1387, opened 2021-09-06
Names for formulas about cones and cocones

rmatthes:
makes the definitions of limits and colimits more modular - not an issue for general theory but for applications when the argument terms may be complicated


benediktahrens:
Thanks a lot for this PR!

PR #1389, opened 2021-09-09
More on monoidal categories - to understand action-based  strength

rmatthes:
a better understanding of action-based strength is the aim (first three commits from another PR erroneously also here)


rmatthes:
the biadjunction for currying is not about monoidal categories but behind the shift between the two different ways of defining action-based strength

rmatthes:
@nmvdw : thanks a lot, this reduced the boring part significantly

rmatthes:
@nmvdw : thanks once more for your quick helpful comments - I had to change the type of `disp_2cells_isaprop` to make it useful for me since this should be used on the way of constructing the bicategory. I'll answer later about your second hint.

rmatthes:
@nmvdw : the answer is positive and its proof has been committed - thanks again

rmatthes:
I assume I'll add still more to this line of thought, but I think the PR is sufficiently mature and big enough to warrant inclusion into the master branch by now.

PR #1392, opened 2021-09-20
Some more examples of multisorted binding signatures

mortberg:
This PR does the following:
- Add PCF as a MS sig
- Port CoC à la Streicher as a 2-sorted sig to use `[sort,Set]`
- Instantiate abstract monadic interchange law for STLC
- Polish `MonadsMultiSorted_alt.v`

### DIFF
--- a/UniMath/SubstitutionSystems/MultiSorted_alt.v
+++ b/UniMath/SubstitutionSystems/MultiSorted_alt.v
@@ -121,6 +121,17 @@ Definition arity (M : MultiSortedSig) : ops M → list (list sort × sort) × so
 Definition mkMultiSortedSig {I : hSet}
   (ar : I → list (list sort × sort) × sort) : MultiSortedSig := (I,,ar).
 
+(** Sum of multisorted binding signatures *)
+Definition SumMultiSortedSig : MultiSortedSig → MultiSortedSig → MultiSortedSig.
+Proof.
+intros s1 s2.
+use tpair.
+- apply (setcoprod (ops s1) (ops s2)).
+- induction 1 as [i|i].
+  + apply (arity s1 i).
+  + apply (arity s2 i).
+Defined.
+
 (** * Construction of an endofunctor on [C^sort,C^sort] from a multisorted signature *)
 Section functor.
 
--- a/UniMath/SubstitutionSystems/MultiSorted_alt.v
+++ b/UniMath/SubstitutionSystems/MultiSorted_alt.v
@@ -61,8 +61,6 @@ Require Import UniMath.SubstitutionSystems.BindingSigToMonad.
 Local Open Scope cat.
 
 (* These should be global *)
-Arguments post_composition_functor {_ _ _} _ _ _.
-Arguments pre_composition_functor {_ _ _} _ _ _.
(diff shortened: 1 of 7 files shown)


UniMath/SubstitutionSystems/PCF_alt.v
```diff
@@ -0,0 +1,203 @@
...
+  alg_map _ (InitialObject PCF_Functor_Initial).
+
+Let PCF_M_alg : algebra_ob PCF_Functor :=
+  InitialObject PCF_Functor_Initial.
+
+(** The variables *)
+Definition var_map : typeToSet2⟦Id,PCF_M⟧ :=
+  BinCoproductIn1 _ (BinCoproducts_functor_precat _ _ _ _ _ _) · PCF_M_mor.
+
+(* TODO: extract the other constructors *)
```
benediktahrens, line 201: Should we wait for this TODO to be implemented, or not?

UniMath/SubstitutionSystems/MultiSorted_alt.v
```diff
@@ -123,6 +121,17 @@ Definition arity (M : MultiSortedSig) : ops M → list (list sort × sort) × so
 Definition mkMultiSortedSig {I : hSet}
   (ar : I → list (list sort × sort) × sort) : MultiSortedSig := (I,,ar).
 
+(** Sum of multisorted binding signatures *)
```
benediktahrens, line 124: The map from multisorted signatures to signatures with strength preserves sums, in a suitable sense. Maybe that's something to mention?

UniMath/SubstitutionSystems/PCF_alt.v
```diff
@@ -0,0 +1,203 @@
...
+use mkMultiSortedSig.
+- apply (type + (type × type) + (type × type) + type)%set.
+- intros [[[t|[t s]]|[t s]]|t].
+  * exact ([],,t).                                  (* Bottom *)
+  * exact ((([],,(arr s t)) :: ([],,s) :: nil),,t). (* App *)
+  * exact (((cons t [],,s) :: []),,(arr t s)).      (* Lam *)
+  * exact ((([],,(arr t t)) :: nil),,t).            (* Y *)
+Defined.
+
+Definition PCF_Sig : MultiSortedSig type :=
```
benediktahrens, line 164: Would it be cool to use the sum operator that you implemented to build PCF from STLC + another signature?

UniMath/SubstitutionSystems/PCF_alt.v
mortberg: I don't think it's very important unless one has concrete applications in mind. It shouldn't be too hard to do, but kind of fiddly so I would prefer not to do it now

UniMath/SubstitutionSystems/PCF_alt.v
mortberg: Ah, that could be fun! Let me try it quickly

mortberg: It worked out without any problems

mortberg:
I think this is ready to be merged. I have some ideas how to improve some small details, but that can come in a future PR

PR #1391, opened 2021-09-18
WIP: trying to relate param. distributivities to monoidal functors

rmatthes:
first step: correspondence between nat. trafos into functor cat. with functors into suitable cat.

not fully satisfying since problems with equality reasoning are circumvented by isomorphisms and lead to elaborate statements (not completed because of too big complexity)

PR #1397, opened 2021-09-30
WIP: identifies developments that can be done with CAT

rmatthes:
CAT = the bicategory of categories

This means, without taking univalent categories (UCAT) but just precategories with homsets.

The new files are simple adaptations of the respective files without "WithoutUnivalence" in their name.

Biadjunction.v now does not refer to UCAT but to CAT, this needed more material in BicatOfCatsWithoutUnivalence.v to be taken from the development for UCAT.

Final.v and Initial.v did not even depend on UCAT (a redundant import)
Strictness.v only depends on UCAT for an example (better, a counter-example)
Monads.v: disp_2cells_isaprop_monad does not need the hypothesis is_univalent_2 C
MonadKtripleBiequiv.v: adapted to that change in Monads.v


benediktahrens:
What is the status of this PR? Is the goal to have it merged? Or is it only a basis for discussion?
Also, will it be in conflict with #1391 ?

rmatthes:
It is rather a request for comments. No conflict with #1391 envisaged. 

The main question: is it good to have two separate strands instead of first developing the stuff without univalence and then taking the full sub-bicategory. The current PR shows to what extent one can proceed without univalence. The untouched files did not promise to me a good version without univalence. But experts on the bicategories library will be in a better position to judge.

nmvdw:
> It is rather a request for comments. No conflict with #1391 envisaged.
> 
> The main question: is it good to have two separate strands instead of first developing the stuff without univalence and then taking the full sub-bicategory. The current PR shows to what extent one can proceed without univalence. The untouched files did not promise to me a good version without univalence. But experts on the bicategories library will be in a better position to judge.

This is something we considered in the development of the library of bicategories. Because the bicategory of univalent categories is quite important in this library, we would like that it is convenient to use it. In addition, we want to be able to make use of displayed machinery (displayed bicategories, displayed pseudofunctors and so on) as much as possible. This is useful for proving the univalence of bicategories and for constructing pseudofunctors, biadjunctions, biequivalences and so on.

For this reason, we first of all decided to define the full subbicategory via displayed bicategories. This has consequences on what the actual full subbicategory looks like: if we have a bicategory `B` and a predicate `P` on the objects of `B`, then the full subbicategory (as we defined in UniMath) looks as follows:
- its objects are objects `x` of `B` for which `P` is satisfied
- the morphisms are morphisms in `B` together with an inhabitant of the unit type
- the 2-cells are 2-cells in `B` together with an inhabitant of the unit type
These inhabitants from the unit type come from the fact that we defined the full subbicategory via displayed bicategories. The relevant parts in the code are:
- https://github.com/UniMath/UniMath/blob/2075f7d4f1ab29f7b600beb3663e92e9ffcb9839/UniMath/Bicategories/DisplayedBicats/Examples/FullSub.v#L35
- https://github.com/UniMath/UniMath/blob/2075f7d4f1ab29f7b600beb3663e92e9ffcb9839/UniMath/Bicategories/DisplayedBicats/Examples/DisplayedCatToBicat.v#L43

For this reason, if we define the bicategory of univalent categories using a full subbicategory, the 1-cells are functors together with an inhabitant of the unit type. Hence, if you use the full subbicategory as defined currently in UniMath, then I suspect there are some difficulties in modifying the code. Of course, these can be solved, but using the bicategory of univalent categories will be more tedious in this case. 

However, it is also possible to define the full subbicategory directly and without using displayed machinery. If this approach is used, then I think there won't be many technical difficulties. As a result, there will be 2 definitions of the full subbicategory though.

rmatthes:
Good explanation. I cannot decide on that matter. Is the code copying acceptable in the present form - hence with rather long file names? Would more of the developments involving UCAT as I called it above be possible with CAT, maybe with some extra effort? I am thinking in particular of Bicategories/DisplayedBicats/Examples/MonadKtripleBiequiv.v .

rmatthes:
I am trying to resolve the conflicts still today.

rmatthes:
Should be fine again.

arnoudvanderleer:
@rmatthes What is the status of this PR? (Or what was the status 2 years ago?)

rmatthes:
A difficult question that I cannot attack in the next seven days.

rmatthes:
The PR was a request for comments, and there were important such comments by @nmvdw and @benediktahrens. The PR proposed a simple-minded solution with code duplication. Solving the current conflicts alone would not make it useful again. The main means to avoid code duplication would be to develop the bicategory of categories and then have more material on the bicategory of univalent categories, defined from the bicategory of categories by a form of subbicategory. The current definition of the bicategory of univalent categories does not follow that path but "hard-wires" the univalence into the construction right in the [beginning](https://github.com/UniMath/UniMath/blob/master/UniMath/Bicategories/Core/Examples/BicatOfUnivCats.v#L34). This is not modular but convenient. Defining it as subbicategory through the displayed mechanism would effectively change the structure (adding unit stuff), defining it through an elementary notion of subbicategory would require interfaces between instances of such an elementary notion and those from the displayed mechanism. I was not ready to invest deeper thoughts into this without having a mathematical challenge for which this would make the difference. I would still be happy if everyone on this thread would take a moment to think if the situation in 2024 is different from 2021 in their mind. If the situation is considered as stable, then I would rather not pursue this line of thought and close the PR.

PR #1402, opened 2021-10-25
Refactor categories

benediktahrens:
Work with categories was made more complicated by the strange definition of univalent categories, which sometimes led to having multiple proofs of `has_homsets` around. (Observed, in particular, by @rmatthes )

This PR aims to refactor the definition of categories into
- precategory
- `category = \Sigma (C : precategory), has_homsets C`
- `univalent_category = \Sigma (C : category), is_univalent C`

The previous definitions made sense when taking `precategory` seriously. But there are few instances of `precategory` that are not `category`, and they are not worth the mess the previous definitions generated.

My changes:
- change the definitions as outlined above
- make many definitions dependent on `category` instead of `precategory`
- adapt files to those changes

Not all files compile yet. It would be great if someone could help in making files compile, by filing PRs against my branch `refactor-univalent-cats` (https://github.com/benediktahrens/UniMath/tree/refactor-univalent-cats).

In particular, 
- the proof that UnivCat is a univalent bicategory might need to be adapted (@nmvdw : could you help here?)
- many instances (in subdir `CategoryTheory/categories`) need to be adapted

PR #1405, opened 2021-11-04
Adapting material on bicategories in terms of enrichment to changes introduced in refactoring of cats

benediktahrens:
follow-up to #1402


benediktahrens:
Fixes most of https://github.com/UniMath/UniMath/pull/1402#issuecomment-960308709

PR #1408, opened 2021-11-06
Pullbacks in bicategories

nmvdw:
- Added the definition of iso-comma category
- Added a proof that the product of univalent categories is again univalent
- Added definition of pullbacks in bicategories, and showed that 1-types and Cat have pullbacks

In the file `Pullback.v`, lines 401-619 could be moved elsewhere.

### DIFF
--- a/UniMath/Bicategories/Colimits/Pullback.v
+++ b/UniMath/Bicategories/Colimits/Pullback.v
@@ -1,0 +1,960 @@
+(****************************************************************
+
+ Pullbacks in bicategories
+
+ In this file we define the notion of pullback square in arbitrary
+ bicategories. For this definition, there are 2 possibilities. One
+ could either write universal properties, which expresses the
+ existence of a morphism up to a unique 2-cell. Alternatively, one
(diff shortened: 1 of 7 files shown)


UniMath/Bicategories/Colimits/Pullback.v
```diff
@@ -0,0 +1,960 @@
...
+           (B : bicat)
+  : UU
+  := ∏ (b₁ b₂ b₃ : B)
+       (f : b₁ --> b₃)
+       (g : b₂ --> b₃),
+     ∑ (p : pb_cone f g),
+     has_pb_ump p.
+
+
+(** MOVE ??!??!???!??? *)
```
benediktahrens, line 401: It would indeed be good to move this material somewhere else, e.g., in `MoreFoundations`.

benediktahrens (approved):

UniMath/Bicategories/Colimits/Pullback.v
nmvdw: Good point, I didn't notice that.

PR #1411, opened 2021-11-13
Established an equivalence between two definitions of bicategories

nmvdw:
Also, moved a lemma that doesn't belong to `Strictness.v`.

### DIFF
--- a/UniMath/Bicategories/WkCatEnrichment/hcomp_bicat.v
+++ b/UniMath/Bicategories/WkCatEnrichment/hcomp_bicat.v
@@ -755,6 +755,76 @@ Proof.
   - exact (bicat_to_hcomp_bicat_laws B).
 Defined.
 
+Definition hcomp_bicat_to_bicat_to_hcomp_bicat
+           (B : hcomp_bicat)
+  : bicat_to_hcomp_bicat (hcomp_bicat_to_bicat B) = B.
+Proof.
+  use total2_paths_f.
+  - do 13 (use total2_paths_f ; [ apply idpath | ] ; cbn).
+    use funextsec ; intro x.
+    use funextsec ; intro y.
+    use funextsec ; intro z.
+    use funextsec ; intro f₁.
+    use funextsec ; intro f₂.
+    use funextsec ; intro g₁.
+    use funextsec ; intro g₂.
+    use funextsec ; intro α.
+    use funextsec ; intro β.
+    cbn.
+    etrans.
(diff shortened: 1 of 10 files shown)


UniMath/Bicategories/WkCatEnrichment/hcomp_bicat.v
```diff
@@ -0,0 +1,824 @@
...
+Defined.
+
+Definition hcomp_bicat_weq_bicat
+  : hcomp_bicat ≃ bicat.
+Proof.
+  use make_weq.
+  - exact hcomp_bicat_to_bicat.
+  - use gradth.
+    + exact bicat_to_hcomp_bicat.
+    + intros B.
```
benediktahrens, line 765: Would it be easy to move this subgoal to a separate lemma, to be closed with `Qed`? (I have not looked at the form of the goal myself.)

Same question for the other subgoal below.

UniMath/Bicategories/WkCatEnrichment/hcomp_bicat.v
```diff
@@ -0,0 +1,824 @@
...
+  - exact (λ f, hb_runit f).
+  - intros f₁ f₂ α.
+    apply (pr2 (pr222 (pr222 (pr222 B)))).
+Defined.
+
+Definition hcomp_bicat_to_prebicategory_data
+           (B : hcomp_bicat)
+  : prebicategory_data.
+Proof.
+  simple refine (_ ,, _ ,, _ ,, _).
```
benediktahrens, line 405: I like this subtle reminder that we should have builders for these structures :-)

benediktahrens (approved):

UniMath/Bicategories/WkCatEnrichment/hcomp_bicat.v
nmvdw: That's possible, but perhaps we should first focus our attention to making suitable builders for displayed bicategories. The notion of `prebicategory_data` (so, where hom-categories are used) and `hcomp_bicat` are not the main goals of this library. Outside of here, they should be avoided, so that the stuff we developed, can be reused.

UniMath/Bicategories/WkCatEnrichment/hcomp_bicat.v
nmvdw: Yes. Both are moved into separate lemmas now.

PR #1417, opened 2021-11-26
Augmented Simplex Category

patrick-nicodemus:
I defined the augmented simplex category of finite ordinals, as well as non monotonic maps, and the category of finite cardinals, together with the standard monoidal structure on these two. I also defined the obvious forgetful functor from finite ordinals to finite cardinals and proved it was monoidal.

### DIFF
--- a/AugmentedSimplexCategory.v
+++ b/AugmentedSimplexCategory.v
@@ -1,0 +1,1057 @@
+Require Import UniMath.Foundations.PartD.
+Require Import UniMath.Foundations.NaturalNumbers.
+Require Import UniMath.Foundations.Sets.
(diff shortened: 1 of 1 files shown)


patrick-nicodemus (commented):
Sorry, I meant to upload this specifically to the directory /CategoryTheory/Monoidal, but I am new to github.
This file defines the augmented simplex category and the category of finite cardinals together with the standard monoidal structure on each. It also defines the standard forgetful functor from one to the other and proves it is monoidal.

DanGrayson:
Just use `git mv` to rename the file.  Add its name in the appropriate place to the file `.package/files`, and push to your branch.  Then we will see what you intended.

PR #1426, opened 2021-12-08
treats leftovers of the transition from precats to cats

rmatthes:
in particular, names are adapted accordingly, comments are put in sync with the contents, and intermediate constructions of precats and proof of hom_set_property are omitted


benediktahrens (approved):
LGTM

PR #1362, opened 2021-06-10
Update Mac OS install instructions

peterlefanulumsdaine:
Addresses #1360 : bring the Mac OS install instructions up-to-date.  Several changes:

- remove instructions to install packages that are no longer needed, in particular `camlp5` and the CoqIDE packages
- add note about how to switch OCaml versions if needed
- include mention of VSCode as an alternative to Emacs

Another minor change, while I was working on `INSTALL.md`: fixed spelling of Proof General throughout (it’s two words not one: https://proofgeneral.github.io)

PR #1437, opened 2022-01-04
start fixing the build process

DanGrayson:
Start with Mac OS X, where homebrew now comes with coq version 8.14.1, and that
version can compile UniMath.  So, we make the building of coq *not* the default
in the makefile, and fix the instructions, too.

Next -- see what's happening under ubuntu and under opam.


UniMath/Foundations/Preamble.v
```diff
@@ -138,7 +138,7 @@ Notation "1000" := (10 * 100) : nat_scope.
 (** Identity Types *)
 
 Inductive paths {A:UU} (a:A) : A -> UU := paths_refl : paths a a.
-Hint Resolve paths_refl : core .
+#[export] Hint Resolve paths_refl : core .
```
benediktahrens, line 141: What is the purpose of this change (and similar ones in other files of `Foundations`)? Is this change accidental?

DanGrayson: No, it's intentional: version 8.14.1 of coq, which is what homebrew provides, prints out gigantic warning messages about those lines, and says that in a future version of coq, that change (or another similar one) will be mandatory.  (The syntax looks strange to me.)

benediktahrens:
This looks great! 

One minor detail: the building of the satellites (see https://github.com/UniMath/UniMath/blob/master/.github/workflows/build-libs.yml) still uses Ubuntu and the Coq version that is bundled in the submodule. There, the flag to build Coq would now need to be set to "yes" explicitly.
Alternatively, we could switch that build to happen on MacOS instead of Ubuntu.

benediktahrens:
@DanGrayson : It seems that coq-community is maintaining a Docker image that can be used on Github Action, see, e.g., the HoTT/HoTT repository: https://github.com/HoTT/HoTT/blob/master/.github/workflows/ci.yml
How about we use this to check UniMath on Ubuntu?

benediktahrens:
@DanGrayson @peterlefanulumsdaine I think this MacOS installation route should become the default - can we adapt #1362 ?

peterlefanulumsdaine:
I’ve tested this build route now, and it worked pretty smoothly for me.  (I had minor problems caused by some leftover files from a previously-installed version of Coq, but homebrew’s troubleshooting hints sufficed for dealing with it.)

So it seems to me like #1362 is essentially obsolete — that it’d be easiest to just close that, and merge this PR instead (once its failing checks are fixed)?

benediktahrens:
> I’ve tested this build route now, and it worked pretty smoothly for me. (I had minor problems caused by some leftover files from a previously-installed version of Coq, but homebrew’s troubleshooting hints sufficed for dealing with it.)
> 
> So it seems to me like #1362 is essentially obsolete — that it’d be easiest to just close that, and merge this PR instead (once its failing checks are fixed)?

This present PR is about setting up the CI on MacOS; it does not modify the installation instructions. But, essentially, yes, it seems to me that simply copying the CI instructions into the MacOS installation manual is the best solution.

We then have very little intersection between the installation for Linux and the installation for MacOS. I will thus create two separate installation manuals for the two OSs.

peterlefanulumsdaine:
Oh, my mistake — I thought this one was covering the instructions as well as the CI.  At least the change in the `makefile` to `BUILD_COQ ?= no` overlaps with what’s needed for fixing the Mac OS build instructions.

PR #1438, opened 2022-01-06
More on cartesian 1-cells

nmvdw:
This PR contains:
- Some results on cartesian 1-cells
- Comprehension bicategories from (op)fibrations of categories

@benediktahrens The file `CleavingOfBicat.v` is getting quite big. Do you think it should be split up? Any suggestions for that?


benediktahrens:
> 

Yes, splitting it at section 4 or 6 (or anywhere, really) could be a good idea.

PR #1441, opened 2022-01-11
splits monoidal_cat_eq into two components

rmatthes:
they are monoidal_cat_triangle_eq and monoidal_cat_pentagon_eq

this only corrects an unelegant design decision and is mathematically trivial and was triggered by PR #1432

### DIFF
--- a/UniMath/CategoryTheory/Monoidal/MonoidalCategories.v
+++ b/UniMath/CategoryTheory/Monoidal/MonoidalCategories.v
@@ -191,12 +191,9 @@ Definition monoidal_cat_right_unitor : right_unitor (pr1 (pr2 M)) (pr1 (pr2 (pr2
 
 Definition monoidal_cat_associator : associator (pr1 (pr2 M)) := pr1 (pr2 (pr2 (pr2 (pr2 (pr2 M))))).
 
-Definition monoidal_cat_eq :
-  (λ α' : associator (pr1 (pr2 M)),
-          triangle_eq (pr1 (pr2 M)) (pr1 (pr2 (pr2 M))) (pr1 (pr2 (pr2 (pr2 M)))) (pr1 (pr2 (pr2 (pr2 (pr2 M))))) α'
-                      × pentagon_eq (pr1 (pr2 M)) α')
-    (pr1 (pr2 (pr2 (pr2 (pr2 (pr2 M))))))
-  := pr2 (pr2 (pr2 (pr2 (pr2 (pr2 M))))).
+Definition monoidal_cat_triangle_eq : triangle_eq (pr1 (pr2 M)) (pr1 (pr2 (pr2 M))) (pr1 (pr2 (pr2 (pr2 M)))) (pr1 (pr2 (pr2 (pr2 (pr2 M))))) (pr1 (pr2 (pr2 (pr2 (pr2 (pr2 M)))))) := pr1 (pr2 (pr2 (pr2 (pr2 (pr2 (pr2 M)))))).
+
+Definition monoidal_cat_pentagon_eq : pentagon_eq (pr1 (pr2 M)) (pr1 (pr2 (pr2 (pr2 (pr2 (pr2 M)))))) := pr2 (pr2 (pr2 (pr2 (pr2 (pr2 (pr2 M)))))).
 
 End Monoidal_Cat_Accessors.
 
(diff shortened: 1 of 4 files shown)


UniMath/CategoryTheory/Monoidal/MonoidalCategories.v
```diff
@@ -191,12 +191,9 @@ Definition monoidal_cat_right_unitor : right_unitor (pr1 (pr2 M)) (pr1 (pr2 (pr2
 
 Definition monoidal_cat_associator : associator (pr1 (pr2 M)) := pr1 (pr2 (pr2 (pr2 (pr2 (pr2 M))))).
 
-Definition monoidal_cat_eq :
-  (λ α' : associator (pr1 (pr2 M)),
-          triangle_eq (pr1 (pr2 M)) (pr1 (pr2 (pr2 M))) (pr1 (pr2 (pr2 (pr2 M)))) (pr1 (pr2 (pr2 (pr2 (pr2 M))))) α'
-                      × pentagon_eq (pr1 (pr2 M)) α')
-    (pr1 (pr2 (pr2 (pr2 (pr2 (pr2 M))))))
-  := pr2 (pr2 (pr2 (pr2 (pr2 (pr2 M))))).
+Definition monoidal_cat_triangle_eq : triangle_eq (pr1 (pr2 M)) (pr1 (pr2 (pr2 M))) (pr1 (pr2 (pr2 (pr2 M)))) (pr1 (pr2 (pr2 (pr2 (pr2 M))))) (pr1 (pr2 (pr2 (pr2 (pr2 (pr2 M)))))) := pr1 (pr2 (pr2 (pr2 (pr2 (pr2 (pr2 M)))))).
```
benediktahrens, line 194: Can those chains of projections be replaced by a meaningful function?

rmatthes: I do not know if they can now. They could not in July 2021, as evidenced by the Coq comment above:
`(** it is important that no new coercions are used in the given types in the following projections *)`

rmatthes: I am now testing if the coercions can be used in the typing of those accessors.

benediktahrens:
I don't understand the comment 
```
 (** it is important that no new coercions are used in the given types in the following projections *)
```
In any case, the definition
```
Definition monoidal_cat : UU :=
  ∑ C : category, ∑ tensor : C ⊠ C ⟶ C, ∑ I : C,
  ∑ λ' : left_unitor tensor I,
  ∑ ρ' : right_unitor tensor I,
  ∑ α' : associator tensor,
         (triangle_eq tensor I λ' ρ' α') × (pentagon_eq tensor α').
```
does not look like what I would expect, and how complicated structures are built up elsewhere in UniMath.

Specifically, I would expect intermediate Sigma types `category_with_tensor`, `category_with_tensor_and_unit`, etc, (left and right unitor and associator would be added in one step) and from each such type a coercion down to categories. This approach works and scales.

rmatthes:
Unfortunately, this was the legacy code of 2018.

benediktahrens:
This PR is an improvement, and I am happy to merge it once @rmatthes signals that their investigations are completed or will be continued in a separate PR.

Without having taken a close look, it seems to me that the formalization of monoidal categories is odd, and needs to be overhauled - no deep mathematical changes, but reorganizing the components will, I believe, improve the usability of the definition.

benediktahrens:
> Unfortunately, this was the legacy code of 2018.

Yes, I know - it should not have been merged back then.

UniMath/CategoryTheory/Monoidal/MonoidalCategories.v
rmatthes: After putting "meaningful" types, `UniMath/CategoryTheory/Monoidal/ActionBasedStrengthOnHomsInBicat.v` takes over 20' to compile on my machine. I'll make a comparison with the definitions in this PR.

rmatthes: Okay, with this PR, the above-mentioned file needed 3'28'' to compile. The overall compilation time for the part depending on `MonoidalCategories.v` went down from 27' to 11' and the total CPU core time went down from 53' to 36', which is the gain for the above-mentioned file. Hence, with the current structure of the definition of monoidal categories, we better stay with the unintelligible typing of the accessors.

PR #1443, opened 2022-01-19
Univalence of display map bicategories

nmvdw:
Added a proof of the univalence of the diplayed bicategory associated to a display map bicategory


benediktahrens:
This PR removes a file. Is the content of that file (or equivalent) retained elsewhere? (I am asking because the total linecount goes down with this PR.)

nmvdw:
The file that was removed, was the file `CodomainFibs.v`. That file existed, because at first, the displayed bicategory of internal Street Fibrations was defined separately. However, now we have a more unified approach (via display map bicategories), and for that reason, that file has become obsolete. The reason why it remained, was because some of the material in that file was not adapted yet to display map bicategories. Now all the material has been adapted.

benediktahrens:
> The file that was removed, was the file `CodomainFibs.v`. That file existed, because at first, the displayed bicategory of internal Street Fibrations was defined separately. However, now we have a more unified approach (via display map bicategories), and for that reason, that file has become obsolete. The reason why it remained, was because some of the material in that file was not adapted yet to display map bicategories. Now all the material has been adapted.

Thanks for the explanation!

PR #1444, opened 2022-01-19
Remove distributivity as a condition for a couple of lemmas in lattices

giorgio93p:
Currently, lemmas `Lmax_le_case` and `Lmax_ge_case` in lattices have `isrdistr (Lmax is) (Lmin is)` as a condition. However, the resulting property holds for all lattices, not only distributive ones, hence it should be available without the extra condition.

This pull request proves the property without using `isrdistr (Lmax is) (Lmin is)` and changes the type of `Lmax_le_case` and `Lmax_ge_case`.

I searched UniMath for occurrences of these two lemmas and I found no more. However, I don't know how good it would be to change the type of a defined entity that has been "out there" for years. Maybe it would be better to add the more generic lemma with a new name and mark the existing one as deprecated?


DanGrayson:
Looks good to me.

benediktahrens:
@giorgio93p : thanks a lot for your pull request. Thanks also for your consideration about compatibility with other libraries. Note that as part of our continuous integration, we test some other libraries on top of UniMath/UniMath, notably
- https://github.com/UniMath/TypeTheory
- https://github.com/UniMath/SetHITs
- https://github.com/UniMath/largecatmodules

If you are aware of other libraries on top of UniMath/UniMath, please let us know.

giorgio93p:
Thank you for the info. No, I am not aware of any more libraries; in fact, I was not even aware of the ones you mentioned.

Issue #1396, opened 2021-09-30
CategoryTheory package depends on Bicategories


benediktahrens:
@rmatthes has introduced a dependency of package CategoryTheory on (some files of) the package Bicategories.

While I agree that it is conceptually undesirable to separate the two, this dependency means that Bicategories has to be compiled in other projects, at least in the CI.

How about the following suggestion: move the definition of bicategory, as well as a few other files (in particular, the bicategory of (univalent) categories), into the CategoryTheory package?

@rmatthes : what do you think? How many files would be moved? (Of course, a lot of Import statements would need to be changed - hopefully that could be automated.)

rmatthes:
I have not yet thought about the global implications, but I would like to add an aspect to this issue: the instantiation of the concept of bicategory with CAT should not create new data but only take lemmas one could also use without the bicategorical infrastructure. So, the fields of the definition of CAT should all be lemmas that are readily available outside. And this probably even when the definition of bicategories is moved into the package CategoryTheory.

The packages CategoryTheory, Bicategories AND SubstitutionSystems are now intertwined because of file [ActionBasedStrengthOnHomsInBicat.v](https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/Monoidal/ActionBasedStrengthOnHomsInBicat.v).

rmatthes:
It is even worse: CategoryTheory depends on SubstitutionSystems as well. (A) CategoryTheory/Monoidal/ActionBasedStrengthOnHomsInBicat.v should be moved into SubstitutionSystems. (B) CategoryTheory/Monoidal/ActionsFormBicategory.v should be moved into Bicategories/DisplayedBicats/Examples, but the documentation of action-based strength should point to that further result at a distant place. (C) Bicategories/Core/Examples/BicatOfCatsWithoutUnivalence.v should go into CategoryTheory with a telling name such as CAT.v and a telling name such as CAT for the main construction. (D) CAT should receive sufficient attention in the library - see PR #1397. (E) Bicategories/Core/BicategoryLaws.v should go into CategoryTheory along with Bicategories/Core/Bicat.v, and these laws should be spelt out in the situation of CAT as well (as far as that is interesting), which is intimately related to my point above that CAT should be created with decent lemmas beforehand - call this point (F).

I hope I did not overlook other unwanted dependencies. There is certainly a lot of interference with PR #1391.

benediktahrens:
[...]

PR #1445, opened 2022-01-23
Restore linearity of package dependency

peterlefanulumsdaine:
This is a quick fix to #1396, restoring ordering of packages by dependency, by removing imports of `Bicategories` within `CategoryTheory` plus a couple of other small such points.

This is NOT the thorough reorganisation we considered in the discussion #1396.  This is a quicker fix, and slightly less satisfactory in a couple of ways — in particular, the material on monoidal categories that makes use of results about bicategories now lives in `Bicategories`, which probably isn’t right in the long-term.  But I think the quick fix is useful in the short-term — in particular, for trying to get the dependency-order check in the `Makefile` working again (it currently doesn’t seem to work, at least on my system — I intend to investigate further and open an issue shortly), as that will be very helpful for doing the larger reorganisation.

This PR interacts substantially with recent material by @rmatthes — in reorganising such material, I’ve tried to respect the intent as best as I could, but Ralph, please do let me know if anything I’ve done seems unfortunate for you.  If there’s anything that’s easy to fixable, I’m happy to add it to the PR; if there’s anything that’s non-trivial, I can keep it in mind for the bigger reorganisation afterwards.


benediktahrens (approved):
LGTM. It would be good to have @rmatthes' feedback as well.

rmatthes:
I agree that this looks acceptable as a quick fix. The material on action-based strength will anyway be split into the traditional approach and a bicategorical setting for the Janelidze/Kelly variant. So, the traditional material could then be moved back into package CategoryTheory.

I only wonder if the README.md files should not better at least mention that more material is found in the other package - even for notions that have nothing to do with bicategories (ontologically - before being organized into one).

benediktahrens:
Thanks for your feedback, @rmatthes !

> I only wonder if the README.md files should not better at least mention that more material is found in the other package - even for notions that have nothing to do with bicategories (ontologically - before being organized into one).

Agreed! Merging this PR now, documentation can be improved separately.

Issue #1420, opened 2021-11-29
Broken pipe in sanity checks on Github CI


benediktahrens:
The Github Action CI shows the following lines:
```
--- checking the ordering prescribed by the files UniMath/*/.packages/files ---
sed: couldn't write 387 items to stdout: Broken pipe
grep: write error: Broken pipe
touch .check-prescribed-ordering.okay
```
Is that reason for concern?

DanGrayson:
Yes, that's worrisome.  It indicates it may not be checking the ordering of the files.  I can't reproduce the problem at home.  Why not restart the jobs and see if it happens again?  Which pull request were you looking at?

benediktahrens:
> Yes, that's worrisome. It indicates it may not be checking the ordering of the files. I can't reproduce the problem at home. Why not restart the jobs and see if it happens again? Which pull request were you looking at?

I think it happens in all the recent PRs, see, e.g., here: https://github.com/UniMath/UniMath/runs/4359935245?check_suite_focus=true#step:5:9

peterlefanulumsdaine:
I’ve put together a proposed fix for this issue (at least as far as I’ve tested it on my own system) — you can review it in the branch [`fix-sanity-checks`](https://github.com/peterlefanulumsdaine/UniMath/tree/fix-sanity-checks) (fixing just the check itself) or, better, [`test-fixed-check`](https://github.com/peterlefanulumsdaine/UniMath/tree/test-fixed-check) (fixing the library to pass the check).

I would issue a PR based on [`test-fixed-check`](https://github.com/peterlefanulumsdaine/UniMath/tree/test-fixed-check), but that branch also incorporates the open PR #1445 so it shouldn’t be pulled until #1445 is accepted.  Just in case #1445 is rejected for some reason, then we can make a PR based on [`fix-sanity-checks`](https://github.com/peterlefanulumsdaine/UniMath/tree/fix-sanity-checks) instead.

peterlefanulumsdaine:
On second thoughts, I think there’s no real reason to wait on making a PR for this — I’ll just add a note to wait on #1445 before merging.

benediktahrens:
I think this issue is fixed, closing.

PR #1446, opened 2022-01-26
Fix and pass file dependency ordering check (and tweak other sanity checks)

peterlefanulumsdaine:
This PR gives several related small contributions:

- fixes the Makefile target `check-prescribed-ordering`, which checks that file dependency follows the (linear) ordering prescribed in the package file lists.   This check was reported broken in #1420, but has presumably been broken longer, since the ordering was violated already in #1396.

- gets the library to pass this check again.  The non-trivial part of this (removing interdependency between packages) was already done in #1445 ; beyond that, no files needed relocating, but a little reordering was needed in the package file lists.

- for all the `sanity-checks`, adds explicit success messages, so they’re less likely to break silently in future.

A few caveats:

- this PR builds on (and includes) #1445, so probably shouldn’t be merged until that is accepted.  If that isn’t accepted, then this PR can be replaced by one based on the [`fix-sanity-checks`](https://github.com/peterlefanulumsdaine/UniMath/tree/fix-sanity-checks) branch, which includes the just fixes to the checks themselves, as discussed in #1420 .

- the fix to `check-prescribed-ordering` has been tested on my local machine (Mac OS Big Sur), on both passing and failing versions of the library.  If it also works in the CI (with the library’s current passing state), I guess that’s safe to merge — but it wouldn’t hurt if people on other systems could test it a bit as well.  (E.g. try re-ordering the first few lines of `UniMath/Foundation/.package/files`, then run `make sanity-checks`, and see if this check throws an error as intended.)

- previously, the makefile had two separate versions of `check-prescribed-ordering`, depending on whether the Coq version was ≥8.8 or <8.8.   Since I don’t think the library works on Coq <8.8 any more (and we certainly don’t use that in the CI), I didn’t bother to test/fix that version — I replaced it in that case with a message that the check is skipped (the same behaviour that the check gives for old versions of `bash`).


benediktahrens:
@peterlefanulumsdaine : Thanks a lot for preparing this (and the previous) PR!
@DanGrayson : Could you take a look at the changes in the `Makefile`?

DanGrayson:
Looks good to me.

Peter, thanks for figuring out that issue with the Makefile!

PR #1447, opened 2022-01-27
Determine some fibers

nmvdw:
Contents:
- Definition of slice bicategories (includes fibrational slices in arbitrary bicategories and for cat)
- Determination of the fiber of trivial displayed bicategory, codomain bicategory, and displayed bicategories arising from display map bicategories. This shows that the strict fiber bicategory in these cases are equvialent to something simpler.
- Change of base functors between slices
- Reindexing gives rise to a pseudofunctor.

Note: this branch might create conflicts with https://github.com/UniMath/UniMath/pull/1446, so I suggest merging this PR after that one.


benediktahrens (approved):
LGTM

PR #1450, opened 2022-02-08
More limits and colimits in bicategories

nmvdw:
This PR contains:
- Refactor the files on limits and colimits
- Definition of comma objects and coproducts
- Added examples


benediktahrens:
It seems to me that in the last commit (https://github.com/UniMath/UniMath/pull/1450/commits/b357be54a082b933b857d0b19312ee1fa4fe4820), you removed the "wrong" copies of 
```
ArrowCategory.v
RightKanExtension.v
```
from file `CategoryTheory/.package/files`.

I suggest moving these two further up again, so that it reads
``` 
PointedFunctorsComposition.v
ArrowCategory.v
RightKanExtension.v
slicecat.v
```
i.e., they should go between `PointedFunctorsComposition` and `slicecat`.

nmvdw:
> It seems to me that in the last commit ([b357be5](https://github.com/UniMath/UniMath/commit/b357be54a082b933b857d0b19312ee1fa4fe4820)), you removed the "wrong" copies of
> 
> ```
> ArrowCategory.v
> RightKanExtension.v
> ```
> 
> from file `CategoryTheory/.package/files`.
> 
> I suggest moving these two further up again, so that it reads
> 
> ```
> PointedFunctorsComposition.v
> ArrowCategory.v
> RightKanExtension.v
> slicecat.v
> ```
> 
> i.e., they should go between `PointedFunctorsComposition` and `slicecat`.

These files depend on `CommaCategories`. Before, it didn't depend on `DisplayedCats`, but now it does. So, the dependencies got screwed up a bit. Perhaps the constant and the general case should be in different files

UniMath/CategoryTheory/ArrowCategory.v
```diff
@@ -87,30 +89,261 @@ Section Defn.
...
+        intro ; apply homset_property.
+      }
+      cbn.
+      use pathsdirprod.
+      + apply id_right.
+      + apply id_right.
+  Qed.
+
+  Definition arrow_category_equiv_comma_category
+    : adj_equivalence_of_cats arrow_category_to_comma_category.
```
benediktahrens, line 332: Just a question: do we have proofs of univalence for both categories?

UniMath/CategoryTheory/.package/files
```diff
@@ -203,6 +201,10 @@ DisplayedCats/ComprehensionC.v
 DisplayedCats/StreetFibration.v
 
 IsoCommaCategory.v
+CommaCategories.v
+
+ArrowCategory.v
+RightKanExtension.v
```
benediktahrens, line 207: These should be moved up, see my earlier comment.

UniMath/CategoryTheory/ArrowCategory.v
nmvdw: For the comma category: yes

For the arrow category: there is no proof of univalence in the file in which it's defined.

benediktahrens:
> These files depend on `CommaCategories`. Before, it didn't depend on `DisplayedCats`, but now it does. So, the dependencies got screwed up a bit. Perhaps the constant and the general case should be in different files

How about we move the whole block of files on displayed categories much earlier?

nmvdw:
> > These files depend on `CommaCategories`. Before, it didn't depend on `DisplayedCats`, but now it does. So, the dependencies got screwed up a bit. Perhaps the constant and the general case should be in different files
> 
> How about we move the whole block of files on displayed categories much earlier?

Sure. I think it should be fixed now.

PR #1451, opened 2022-02-08
Add opam file for the development version of UniMath.

maggesi:
Opam file for installing the development version of package coq-unimath.
Instructions in the README will be added with a subsequent PR once everything is double-checked.
So far, it has been tested from a private repository.

### DIFF
--- a/coq-unimath.opam
+++ b/coq-unimath.opam
@@ -1,0 +1,14 @@
+opam-version: "2.0"
+maintainer: "The UniMath Development Team"
+homepage: "https://github.com/UniMath/UniMath"
+dev-repo: "git+https://github.com/UniMath/UniMath.git"
+bug-reports: "https://github.com/UniMath/UniMath/issues"
+license: "Kind of MIT"
+authors: ["The UniMath Development Team"]
+build: [make "BUILD_COQ=no" "-j%{jobs}%"]
+install: [make "BUILD_COQ=no" "install"]
+depends: [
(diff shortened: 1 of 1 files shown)


coq-unimath.opam
```diff
@@ -0,0 +1,14 @@
+opam-version: "2.0"
+maintainer: "The UniMath Development Team"
+homepage: "https://github.com/UniMath/UniMath"
+dev-repo: "git+https://github.com/UniMath/UniMath.git"
+bug-reports: "https://github.com/UniMath/UniMath/issues"
+license: "Kind of MIT"
```
DanGrayson, line 6: What does "kind of MIT" mean?  It's probably better simply to say "see the file LICENSE.md"

benediktahrens: I would imagine that this file should be somewhat self-contained?

maggesi: Yes, I think it is meant to be self-contained.
Of course, we can change this description if we find something more appropriate.
We can also remove all the metadata (license, description etc), which is meaningful only inside a package collection.
However, it would be useful to keep here a complete opam file ready to be copied in the opam-coq-archive whenever we want to propose an update (e.g., for the future releases of the Coq Platform distribution).

benediktahrens, line 6: I suggest the following slight reformulation:
```suggestion
license: "Similar to MIT license"
```

PR #1452, opened 2022-02-10
Extensive bicategories

nmvdw:
- Definition of extensive bicategories
- 1-types is extensive
- The bicategory of univalent categories is extensive


nmvdw:
In `OneTypesColimits.v`, there currently is the definition `empty_hlevel` which says that the empty is of every positive hlevel. What is a better place for this statement?

benediktahrens:
> In `OneTypesColimits.v`, there currently is the definition `empty_hlevel` which says that the empty is of every positive hlevel. What is a better place for this statement?

In `MoreFoundations/PartA` there are already quite a few results about h-levels. Your definition would fit in there.

PR #1457, opened 2022-02-13
Univalence of slices

nmvdw:
- Slice bicategory is univalent
- Adjoint equivalences in slice bicategory
- Slices from display map bicategory is univalent
- Adjoint equivalences in the slices coming from display map bicategories


benediktahrens (approved):
LGTM

PR #1465, opened 2022-02-28
Created basic components of the definition of a curried monoidal cate…

Kfwullaert:
…gory. Maybe have to add coercions for the associator and unitors for readability? The CONTENTS.md file was not supposed to change, not sure how this happened.

### DIFF
--- a/UniMath/CategoryTheory/Monoidal/DisplayedMonoidalCurried.v
+++ b/UniMath/CategoryTheory/Monoidal/DisplayedMonoidalCurried.v
@@ -37,10 +37,6 @@ Local Open Scope cat.
 Local Open Scope mor_disp_scope.
 
 
-  (* These notations come from 'MonoidalCategoriesCurried' *)
-  Notation "x ⊗_{ T } y" := (tensoronobjects_from_tensor_cat T x y) (at level 31).
-  Notation "f ⊗^{ T } g" := (tensoronmorphisms_from_tensor_cat T  _ _ _ _ f g) (at level 31).
-
   Definition displayed_tensor {C : category} (T : tensor C) (D : disp_cat C) : UU :=
     ∑ (tensor_ob : ∏ (x y : C), (D x) → (D y) -> (D (x ⊗_{T} y))),
     ∏ (x x' y y' : C), ∏ (f : C⟦x,x'⟧) (g : C⟦y,y'⟧), ∏ (a : D x) (a' : D x') (b : D y) (b' : D y'),
@@ -78,6 +74,16 @@ Local Open Scope mor_disp_scope.
   Definition displayed_leftunitor_naturality {C : category} {T : tensor C} {I : C} {D : disp_cat C} {TD : displayed_tensor T D} {i : D I} {lu : left_unitor T I} (dlu : displayed_leftunitor_on_objects TD i (pr1 lu)) : UU :=
     ∏ (x y : C), ∏ (a : D x) (b : D y) (f : C⟦x,y⟧) (f' : a -->[f] b), (((id_disp i) ⊗^{{TD}} f') ;; (dlu y b)) = transportb _ ((pr2 lu) x y f) ((dlu x a) ;; f').
 
+(* TEST *)
+(*
+Definition ayed_leftunitor_naturality {C : category} {T : tensor C} {I : C} {D : disp_cat C} {TD : displayed_tensor T D} {i : D I} {lu : left_unitor T I} (dlu : displayed_leftunitor_on_objects TD i (pr1 lu)) : ∏ (x y : C), ∏ (a : D x) (b : D y) (f : C⟦x,y⟧) (f' : a -->[f] b), nat.
+Proof.
+  intros x y a b f f'.
+  Check (((id_disp i) ⊗^{{TD}} f') ;; (dlu y b)).
+  Check (dlu x a) ;; f'.
+  = transportb _ ((pr2 lu) x y f) ((dlu x a) ;; f').
+  *)
+
   Definition displayed_leftunitor {C : category} {T : tensor C} {I : C} {D : disp_cat C} (TD : displayed_tensor T D) (i : D I) (lu : left_unitor T I) : UU :=
     ∑ (dlu : displayed_leftunitor_on_objects TD i (pr1 lu)), displayed_leftunitor_naturality dlu.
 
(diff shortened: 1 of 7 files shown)


UniMath/CategoryTheory/Monoidal/DisplayedMonoidalCurried.v
```diff
@@ -0,0 +1,243 @@
...
+Require Import UniMath.CategoryTheory.DisplayedCats.Constructions.
+Require Import UniMath.CategoryTheory.DisplayedCats.Total.
+
+Local Open Scope cat.
+
+Section DisplayedCurriedMonoidalCategories.
+
+  (* These notations come from 'MonoidalCategoriesCurried' *)
+  Notation "x ⊗_{ T } y" := (tensoronobjects_from_tensor_cat T x y) (at level 31).
+  Notation "f ⊗^{ T } g" := (tensoronmorphisms_from_tensor_cat T  _ _ _ _ f g) (at level 31).
```
benediktahrens, line 42: @nmvdw : could you point us to best practices regarding notations? I believe that putting them into a submodule is a good idea, and importing the notations by opening the submodule. Can you point us to an example?

UniMath/CategoryTheory/Monoidal/DisplayedMonoidalCurried.v
```diff
@@ -0,0 +1,243 @@
...
+    tensor (total_category D).
+  Proof.
+    unfold tensor.
+    split with (λ xa yb, (pr1 xa) ⊗_{T} (pr1 yb) ,, (pr2 xa) ⊗_{{TD}} (pr2 yb)).
+    intros xa yb xa' yb' f g.
+    simpl.
+    split with (pr1 f ⊗^{T} pr1 g).
+    apply (pr2 TD).
+    + exact (pr2 f).
+    + exact (pr2 g).
```
benediktahrens, line 80: This construction can probably be made quite a bit shorter, by removing `unfold...`, `simpl`.

UniMath/CategoryTheory/Monoidal/DisplayedMonoidalCurried.v
```diff
@@ -0,0 +1,243 @@
...
+  Defined.
+
+  (* Remark::A unit for a displayed category D consists of the unit in C and a term in (D I), hence we do not need any special construction right now *)
+
+  Definition displayed_leftunitor_on_objects {C : category} {T : tensor C} {I : C} {D : disp_cat C} (TD : displayed_tensor T D) (i : D I) (lu : left_unitor_on_objects T I) :=
+    ∏ (x : C), ∏ (a : D x), ((i ⊗_{{TD}} a)-->[(lu x)] a).
+
+  (* The following notation comes from 'DisplayedCats.Core' *)
+  Notation "ff ;; gg" := (comp_disp ff gg) (at level 50, left associativity, format "ff ;; gg").
```
benediktahrens, line 89: This notation is already defined somewhere. See, e.g., `DisplayedCats/Equivalences.v` for how to bring it in, in particular, see `Local Open Scope mor_disp_scope.`

UniMath/CategoryTheory/Monoidal/DisplayedMonoidalCurried.v
```diff
@@ -0,0 +1,243 @@
...
+      - exact ((pr1 dlu) x a).
+    +  simpl.
+       intros x y f.
+       induction x as [x a].
+       induction y as [y b].
+       induction f as [f f'].
+
+       use total2_paths_b.
+        - exact ((pr2 lu) x y f).
+        - exact ((pr2 dlu) x y a b f f').
```
benediktahrens, line 114: Data and property should be constructed separately: we want the data to be transparent, the proof of the property to be opaque.
Here, one could put the proof of the property into an `abstract(...)` block.

The various `induction foo as [bar foobar]` tactics don't seem necessary.

UniMath/CategoryTheory/Monoidal/DisplayedMonoidalCurried.v
```diff
@@ -0,0 +1,243 @@
...
+    +  simpl.
+       intros x y f.
+       induction x as [x a].
+       induction y as [y b].
+       induction f as [f f'].
+       cbn.
+
+       use total2_paths_b.
+        - exact ((pr2 ru) x y f).
+        - exact ((pr2 dru) x y a b f f').
```
benediktahrens, line 145: As above.

UniMath/CategoryTheory/Monoidal/DisplayedMonoidalCurried.v
```diff
@@ -0,0 +1,243 @@
...
+      induction y as [y b].
+      induction y' as [y' b'].
+      induction z as [z c].
+      induction z' as [z' c'].
+      induction f as [f f'].
+      induction g as [g g'].
+      induction h as [h h'].
+      use total2_paths_b.
+      - exact ((pr2 α) x x' y y' z z' f g h).
+      - exact ((pr2 dα) x x' y y' z z' a a' b b' c c' f g h f' g' h').
```
benediktahrens, line 184: As above.

UniMath/CategoryTheory/Monoidal/DisplayedMonoidalCurried.v
```diff
@@ -0,0 +1,243 @@
...
+    : triangle_identity (total_category_has_tensor T D TD)  (I,,i) (total_category_has_leftunitor dlu) (total_category_has_rightunitor dru) (total_category_has_associator dα).
+  Proof.
+    unfold triangle_identity.
+    intros x y.
+    induction x as [x a].
+    induction y as [y b].
+    use total2_paths_b.
+    + exact (tri x y).
+    + exact (dtri x y a b).
+  Defined.
```
benediktahrens, line 200:
```suggestion
  Qed.
```

UniMath/CategoryTheory/Monoidal/DisplayedMonoidalCurried.v
```diff
@@ -0,0 +1,243 @@
...
+    unfold pentagon_identity.
+    intros w x y z.
+    induction w as [w e].
+    induction x as [x a].
+    induction y as [y b].
+    induction z as [z c].
+    use total2_paths_b.
+    + exact (pen w x y z).
+    + exact (dpen w x y z e a b c).
+  Defined.
```
benediktahrens, line 217:
```suggestion
  Qed.
```

UniMath/CategoryTheory/Monoidal/DisplayedMonoidalCurried.v
```diff
@@ -0,0 +1,243 @@
...
+
+  Definition displayed_monoidal_category {M : monoidal_category} (D : disp_cat M) :=
```
benediktahrens, line 219: Put type annotations everywhere.

```suggestion
  Definition displayed_monoidal_category {M : monoidal_category} (D : disp_cat M) : UU :=
```

benediktahrens (changes requested):
Good stuff! Some small changes are necessary, see my comments.

UniMath/CategoryTheory/Monoidal/DisplayedMonoidalCurried.v
Kfwullaert: Concerning the seperation of the data and the property: If I understand it correctly, this abstract block makes the proof of the construction "invisible" in the same sense that "Defined" (resp. "Qed") makes the proof visible (resp. invisible). But don't we need this construction of the left unitor in order to proof the unitality of the forgetful functor (to show it is monoidal)?

benediktahrens: Yes, exactly. This is why the **proof part only** should be made opaque, that is, the part from line 106 to 114. The data part should remain transparent.

rmatthes:
I have an ideological remark: I find it unfortunate that the order of composition is turned around by using the notation `∘` instead of the usual `·` - usual for UniMath (not the applicative order of textbooks). This will ask for heavier efforts for a later migration to the new format of monoidal categories.

DanGrayson:
> ∘

Can't you just choose to use the operator appropriate for your context?

benediktahrens:
> I have an ideological remark: I find it unfortunate that the order of composition is turned around by using the notation `∘` instead of the usual `·` - usual for UniMath (not the applicative order of textbooks). This will ask for heavier efforts for a later migration to the new format of monoidal categories.

I don't anticipate any problems. This notation decodes to the normal composition. Furthermore, it is parse-only, so people not modifying the code here directly won't ever be bothered by it. But I might be missing something, of course.

rmatthes:
I agree that the user of the library will not be bothered with the opposite order. I still thought that such "parse-only" notation is meant to deal with legacy code.

rmatthes:
A non-ideological question: How does it come that the functor laws for the tensor are not part of the notion of a monoidal category in the curried style? I also do not see what `tensor_cat_struct` could be good for.

benediktahrens:
> I agree that the user of the library will not be bothered with the opposite order. I still thought that such "parse-only" notation is meant to deal with legacy code.

I agree that this code should eventually be written using `\cdot`, also to be in line with the corresponding code for displayed categories.

benediktahrens:
> The CONTENTS.md file was not supposed to change, not sure how this happened.

This file is built automatically using the information in the files `.package/files`.

UniMath/CategoryTheory/Monoidal/DisplayedMonoidalCurried.v
nmvdw: Putting notations in a submodule can have advantages. The user of the code has then more control over what notations are in scope. We used it here:
- https://github.com/UniMath/UniMath/blob/master/UniMath/Bicategories/Limits/Products.v#L2070
- https://github.com/UniMath/UniMath/blob/master/UniMath/Bicategories/Core/Bicat.v#L1655

Note: if the notation is only for the specific section, then it should be declared `Local`: https://coq.github.io/doc/v8.10/refman/user-extensions/syntax-extensions.html#coq:cmd.local-notation. We used it here:
- https://github.com/UniMath/UniMath/blob/master/UniMath/Bicategories/Core/Bicat.v#L38
Local Notation only exists within the section in which it is declared, while without the `Local`-declaration, it persists even after the section.

I think all global notations should be put in a scope (which did not happen in these two lines of code). If they are not put in a scope, then the notations will always be present if the file is imported, and that can cause clashes.

benediktahrens: A small addendum: I think that even non-local notations within sections are local. Try this code:
```
Section foo.
  Notation "'bar'" := nat.
  Check bar.
End foo.
Check bar.
```

benediktahrens: Is it fair to say that a scope is like a namespace for notations?

benediktahrens:
@Kfwullaert : please address the requested changes, in particular those regarding notation, in a subsequent PR. I am merging this now.

Issue #1473, opened 2022-03-18
CI fails with missing package in Ubuntu repos


benediktahrens:
See, e.g., https://github.com/UniMath/UniMath/pull/1472 and https://github.com/UniMath/UniMath/pull/1471:
```
Get:96 http://azure.archive.ubuntu.com/ubuntu focal/universe amd64 camlp5 amd64 7.11-1build1 [3695 kB]
Fetched 141 MB in 13s (10.5 MB/s)
E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/libx/libxml2/libxml2-utils_2.9.10+dfsg-5ubuntu0.20.04.1_amd64.deb  404  Not Found [IP: 40.81.13.82 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
Error: Process completed with exit code 100.
```
I am investigating. @DanGrayson @maggesi : could you also have a look?

benediktahrens:
solved by #1474, closing

PR #1474, opened 2022-03-18
add to workflow

DanGrayson:
try to fix issue #1473


benediktahrens:
It looks to me as if there are tabs instead of spaces in your added lines. This is not valid yaml syntax, I fear.

benediktahrens:
Thanks for the quick fix, @DanGrayson - I had not noticed that the `update` was missing!

PR #1478, opened 2022-03-25
make assert into separate lemma

Tobias-Schmude:
The proof of cartesian_factorisation_unique in CategoryTheory/DisplayedCats/Fibrations.v asserts something that might reasonably be a separate lemma.


benediktahrens:
@Tobias-Schmude : shouldn't the goal `goal'` be removed, then?

benediktahrens (changes requested):
Subgoal needs to be removed

Tobias-Schmude:
I removed the assertion. Should I click on "re-request review" now?

benediktahrens:
Sure, that sounds fun!

PR #1479, opened 2022-03-25
Some limits and colimits in slice bicategories

nmvdw:
Content:
- Coproducts in slice bicategory
- Initial and terminal object in slices arising from display map bicategories (eg, internal Street (op)fibrations)
- Properties of morphism from strict biinitial objects


UniMath/Bicategories/Colimits/Examples/DisplayMapSliceColimits.v
```diff
@@ -0,0 +1,77 @@
+Require Import UniMath.Foundations.All.
+Require Import UniMath.MoreFoundations.All.
+Require Import UniMath.CategoryTheory.Core.Categories.
+Require Import UniMath.CategoryTheory.DisplayedCats.Core.
+Require Import UniMath.Bicategories.Core.Bicat.
+Import Notations.
```
benediktahrens, line 6: Could we use qualified imports here, so that it is clear which "Notations" are imported?

benediktahrens (commented):
You are using large `abstract`ed blocks. These are a bit annoying to repair if they break. It could be better - in the future, not necessarily in this PR - to replace a definition with a large abstracted block by a definition, a lemma, and another definition.

UniMath/Bicategories/Colimits/Examples/DisplayMapSliceColimits.v
nmvdw: Sure. These unqualified imports happened at more points, and I fixed those as well.

benediktahrens:
@nmvdw : Thanks a lot for factoring out the abstract parts!

benediktahrens (approved):

PR #1492, opened 2022-04-15
Fix build process

DanGrayson:


.github/workflows/build-libs.yml
```diff
@@ -30,17 +30,16 @@ jobs:
...
+          type coqc
+          coqc --version
 
       # Change into UniMath directory and build UniMath
       - name: Build UniMath and libraries TypeTheory, Largecatmodules, and SetHITs
         run: |
           pushd $GITHUB_WORKSPACE
-          make build-coq
           make -j2
           make install
```
benediktahrens, line 42: It is possible that this command requires `sudo` now
```suggestion
          sudo make install
```

DanGrayson: I'm working on it.  I'll use "sudo" to make the directory writable by all.

DanGrayson: But now I'm puzzled that the makefile seems to recompile the *.vo needlessly.

benediktahrens: > But now I'm puzzled that the makefile seems to recompile the *.vo needlessly.

What do you mean by that? At this point, no *.vo files exist yet, I believe.

DanGrayson: I mean, I reproduced the problem at home.  The command "make install" built all the *.vo files and then failed trying to put them in "user-contrib".   Now I fixed that little problem and it's making all the *.vo files again.  Users won't like that.

Makefile
```diff
@@ -88,10 +91,15 @@ else
 EFFECTIVE_MEMORY_LIMIT = unlimited
 endif
 
-all html install uninstall $(VOFILES): build/CoqMakefile.make
+install: build/CoqMakefile.make
+	sudo chmod a+w /usr/lib/coq/user-contrib
```
benediktahrens, line 95: Will this break local installations of Coq? 

That is, what if one compiles Coq using the option `$ make BUILD_COQ=yes`? Then `$ make install` should not require `sudo`, and the directory `/user/lib/coq/user-contrib` won't exist.

DanGrayson: Good point!  I'll fix it.

PR #1495, opened 2022-04-18
More on preservation

nmvdw:
- Right biadjoints preserve equifiers
- Left biadjoints preserve coproducts


UniMath/Bicategories/Colimits/Coproducts.v
```diff
@@ -1,8 +1,8 @@
 (****************************************************************
 
- Products in bicategories
+ Coroducts in bicategories
```
benediktahrens, line 3:
```suggestion
 Coproducts in bicategories
```

benediktahrens (approved):

PR #1494, opened 2022-04-17
Fix build process

DanGrayson:

PR #1496, opened 2022-04-19
Fix build process

peterlefanulumsdaine:
A few tweaks following #1494 ; I’d been meaning to suggest these in review there, but it got merged before I’d noted them.  

- added `.dir-locals.el` to `.gitignore`, since it’s now auto-generated following 6285892
- a bit of rewriting in the build instructions for Mac OS X, which had become a bit unclear

Marking as draft for now since I have a couple of other documentation improvements I’d like to make — but in case I don’t get back to this promptly, it can also be merged as-is.


peterlefanulumsdaine:
This finishes all the small tweaks I had on my list.

benediktahrens (approved):

PR #1506, opened 2022-05-11
Adapt w.r.t. coq/coq#16004.

ppedrot:


Alizter:
@benediktahrens That "sanity check" is just complaining that the foundations files have changed right?

benediktahrens:
> @benediktahrens That "sanity check" is just complaining that the foundations files have changed right?

@Alizter : Yes, that's right.

@ppedrot : Thanks a lot for preparing this PR.

PR #1509, opened 2022-05-19
Definition of and constructions for displayed displayed categories

Tobias-Schmude:
For several parts of the last proof I haven't yet found nice ways to write them.


benediktahrens:
@Tobias-Schmude : Thanks a lot for this PR.

Could you explain what this PR does? In particular, 

- why the need for rewriting the construction of `sigma_disp_cat_axioms` in `DisplayedCats/Constructions.v`?
- why the need to delete `UniMath/CategoryTheory/DisplayedCats/MoreFibrations/FiberDisplayedCategory.v`?
- what is the aim of `UniMath/CategoryTheory/DisplayedCats/MoreFibrations/DisplayedDisplayedCats.v`?

Tobias-Schmude:
Sure!
- The sigma construction acts on a displayed displayed category, that's why it appears in that file again. In principle I'd like this to replace the original occurrence in DisplayedCats/Constructions.v, but this would mean renaming stuff.
- I feel like the name creates confusion. This file contains the construction of a displayed category on a category `C` from a functor into `C`, so this is different from the fiber displayed category construction. So far it isn't relevant to the rest of the project either.
- The end goal is a characterization of compositions of fibrations `P: C -> B, Q: D -> C`. In the displayed setting these are given by a displayed category `P'` on `C` and a displayed category `Q'` on the total category of `P'`. The composition of the functors corresponds to what was already called the sigma construction, while fibers of `Q` wrt an object `c` of `C` (used in the desired characterization) correspond to the fiber displayed categories defined in this PR.

Tobias-Schmude:
I've seen that the build tests fail. On my machine it has been compiling for hours. I'll have another look tomorrow.

Tobias-Schmude:
Pulling the changes somehow broke something: when trying to step through code in emacs I get the error message `Searching for program: No such file or directory, coqtop`.

Tobias-Schmude:
Is it normal that the MacOS build tests fail?

DanGrayson:
> Is it normal that the MacOS build tests fail?

No, but it was just a download failure, hopefully transient.  I've restarted them.

DanGrayson:
> > Pulling the changes somehow broke something: when trying to step through code in emacs I get the error message `Searching for program: No such file or directory, coqtop`.
> 
> We recently changed how the file `UniMath/.dir-locals.el` is handled, to take care of the case where the user uses a pre-installed version of coq rather than building it. Now the file is created from `UniMath/.dir-locals.el.in` with one of two possibilities: the line is `(setq coq-prog-name (concat unimath-topdir "sub/coq/bin/coqtop"))` commented out, or not. In your case, it must be commented out, so it looks for coqtop on your PATH. Is it there?

PS: The default now is not to build Coq, so if you haven't installed it, you should build with `make BUILD_COQ=yes`.

Tobias-Schmude:
After installing Coq directly, I get the following error message in Emacs:
```
Error:
File /home/tobias/Documents/Science/Proof Assistants/Coq/My UniMath Fork/UniMath/Foundations/Sets.vo
has bad magic number 1131376929 (expected 81100). It is corrupted or was
compiled with another version of Coq.
```
My only idea is to pull a fresh copy and `make` that. Or is there an easier way?

DanGrayson:
> After installing Coq directly, I get the following error message in Emacs:
> 
> ```
> Error:
> File /home/tobias/Documents/Science/Proof Assistants/Coq/My UniMath Fork/UniMath/Foundations/Sets.vo
> has bad magic number 1131376929 (expected 81100). It is corrupted or was
> compiled with another version of Coq.
> ```
> 
> My only idea is to pull a fresh copy and `make` that. Or is there an easier way?

Running `make clean` should work.  (That will remove all the *.vo files, forcing them all to be remade.)

Tobias-Schmude:
And I guess after that I will have to re`make` everything?

Tobias-Schmude:
That sounds likely. Thank you very much for your help and patience!

Tobias-Schmude:
For some reason cleaning and remaking didn't solve the problem. I just cloned another copy, and now it works!

DanGrayson:
> For some reason cleaning and remaking didn't solve the problem. I just cloned another copy, and now it works!

We do have a more dramatic cleaning operation, which should be equivalent to making a new clone:

```sh
make git-clean
```

benediktahrens:
@DanGrayson @Tobias-Schmude : Thanks a lot for the PR and the discussion around it.

For the future, it would be great to have discussions around compilation difficulties in https://unimath.zulipchat.com/ rather than here - in that way, other users can profit from the discussion and find it more easily.

DanGrayson:
> @DanGrayson @Tobias-Schmude : Thanks a lot for the PR and the discussion around it.
> 
> For the future, it would be great to have discussions around compilation difficulties in https://unimath.zulipchat.com/ rather than here - in that way, other users can profit from the discussion and find it more easily.

If you want people to look there, we should advertise it better.  For example, there could be a link from the Unimath github organization to it.

Tobias-Schmude:
I've noticed that one check seems to be hanging. Is there something to be done about that?

DanGrayson:
I see no way to tell why that one build is hung up.  

You could merge the master branch into your branch.  That's needed, anyway, and the update should restart the builds.

benediktahrens:
This is the name of a build that does not exist any more, I believe. It is only listed because it was listed as a "necessary" check in the branch protection rules.

I have changed the settings to that this "faux" build rule is not listed any more.

UniMath/CategoryTheory/DisplayedCats/Constructions.v
```diff
@@ -529,20 +529,25 @@ Definition sigma_disp_cat_axioms
...
+      apply maponpaths_2. apply homset_property.
+  - intros x y z w f g h [xx xxx] [yy yyy] [zz zzz] [ww www] [ff fff] [gg ggg] [hh hhh].
+    simpl in ff, fff, gg, ggg, hh, hhh.
+    use total2_reassoc_paths'.
+    + simpl. apply assoc_disp.
+    + apply (pathscomp0 (assoc_disp fff ggg hhh)).
+      apply maponpaths_2. apply homset_property.
+  - intros. apply isaset_total2.
+    + apply homsets_disp.
+    + intros. apply homsets_disp.
```
benediktahrens, line 550: Why are you replacing a shorter construction by a longer one here?

PR #1511, opened 2022-05-25
Some renamings

nmvdw:
This is a renaming for comprehension bicategories and it adds some definitions.

I also added a proof that (local (op))cleavings form a proposition if the involved (displayed) bicategories are univalent.


benediktahrens (approved):
This is great, thanks!

PR #1514, opened 2022-06-14
Limits of structured categories

nmvdw:
I added:
- Every pointwise pseudonatural equivalence is an equivalence in the bicategory of pseudofunctors
- I removed a line in `DisplayedDisplayedCats.v` that gave unnecessary printing during compilation
- I defined some bicategories of structured categories (eg, categories with products, categories with finite limits)
- I also showed that these bicategories have some limits (terminal object, products, pullbacks)


nmvdw:
@benediktahrens There's an error in `./TypeTheory/Auxiliary/Pullbacks.v`. If I see it correctly, then this is not related to my PR but instead, to the replacing of `iso` by `z_iso`. Is this correct?

benediktahrens:
> t

Yes, that's correct. TypeTheory is not yet compatible with latest UniMath, unfortunately. I don't have time at the moment to fix this :-(

nmvdw:
> > t
> 
> Yes, that's correct. TypeTheory is not yet compatible with latest UniMath, unfortunately. I don't have time at the moment to fix this :-(

Sure. That's no problem.

Thank you for the clarification

benediktahrens:
This is great stuff!

PR #1515, opened 2022-06-15
cartesian monoidal categories as examples of monoidal categories

rmatthes:
This is a rather shallow example. The proofs are not elegant and cry for some automation.

Not ready to be merged since the example should be carried over to monoidal functors, etc.


rmatthes:
This branch was merged into PR #1516.

PR #1516, opened 2022-06-16
Kw displayedmonoidalcats examples

ghost:
Added a construction of SET as a (cartesian) monoidal category, constructed the (cartesian) monoidal category of pointed sets (resp. sets with an associative binary operations) as a displayed monoidal category on the (cartesian) monoidal category SET.


UniMath/CategoryTheory/Monoidal/Examples/SetCartesianMonoidal.v
```diff
@@ -0,0 +1,104 @@
...
+      + intro a.
+        exact (a,,tt).
+      + use tpair.
+        * apply funextsec ; intro a ; induction a as [a t] ; induction t ; apply idpath.
+        * apply funextsec ; intro ; apply idpath.
+    - intros x y.
+      apply funextsec ; intro ; apply idpath.
+    - intros x y z w.
+      apply idpath.
+      (* Qed. <- I can't make this a Qed, otherwise, I later have to show how the inverse looks *)
```
nmvdw, line 97: This is rather surprising. In this definition, you are proving the laws of a monoidal category. Where do you need to compute these?

rmatthes: Even though the monoidal laws constitute a property, they contain the inverse z_isos for associator and unitors which makes it harder to use the library because this mixes the phases of construction and verification.

nmvdw: Are there plans to remove the inverses of the associator and the unitor from the laws and put them in the data?

benediktahrens: This is how it is done for bicategories, abd it would be good to keep this aspect in monoidal cats

rmatthes:
I wonder if it is necessary to have the proof that HSET is a monoidal category also executed for the specific situation of HSET, not just for a category with finite products. To open for experimenting, I added this example to my PR #1515.

UniMath/CategoryTheory/Monoidal/Examples/SetCartesianMonoidal.v
benediktahrens: I reinforce Niels' message. The definition should be changed to have the inverses as oart if the structure

UniMath/CategoryTheory/Monoidal/Examples/BinopCartesianMonoidal.v
```diff
@@ -0,0 +1,231 @@
...
+
+Local Open Scope cat.
+
+(* Import BifunctorNotations.
+Import MonoidalNotations. *)
+
+Section BinopCategory.
+
+  Definition Binop_data (X : hSet) : UU := X -> X -> X.
+  Definition Binop_law {X : hSet} (m : Binop_data X) : UU
```
rmatthes, line 29: In `Algebra.BinaryOperations`, no associativity is imposed. So, it would be better to change the names here to indicate associativity. Does associativity of the ops even play a role for this example? Then, this piece of library would become very much related to `CategoryTheory.category_binops` (where univalence is shown but no displayed categories are used).

Kfwullaert:
> I wonder if it is necessary to have the proof that HSET is a monoidal category also executed for the specific situation of HSET, not just for a category with finite products. To open for experimenting, I added this example to my PR #1515.

Its probably not necessary, I was just already busy with it before I saw your message. Even the construction of pointed sets and binary operations can be generalized to the setting of arbitrary cartesian categories.

UniMath/CategoryTheory/Monoidal/Examples/BinopCartesianMonoidal.v
rmatthes: If associativity is irrelevant, then better keep it out and use the notions in the UniMath library.

rmatthes:
> > I wonder if it is necessary to have the proof that HSET is a monoidal category also executed for the specific situation of HSET, not just for a category with finite products. To open for experimenting, I added this example to my PR #1515.
> 
> Its probably not necessary, I was just already busy with it before I saw your message. Even the construction of pointed sets and binary operations can be generalized to the setting of arbitrary cartesian categories.

If this is so, you could merge my PR into your development branch and use that definition in your two interesting examples.

rmatthes:
The development in SetCartesianMonoidal.v should receive a comment that this is a shorter proof of an instance. In the two other developments involving displayed cats, there is a lot of repetition in the proofs. The techniques of my PR #1517 should avoid some of the "boilerplate".

rmatthes:
To further my last comment: I mean in particular the convenience construction `UniMath.CategoryTheory.DisplayedCats.Constructions.disp_struct` and a `try` directly after spawning multiple new goals with the default tactic for solving most of them.

PR #1517, opened 2022-06-16
provides whiskered monoidal category of pointed endofunctors

rmatthes:
goes through displayed monoidal categories, inspired by the most recent examples by Kobe Wullaert
Here, the small changes to WhiskeredMonoidalFromBicategory.v are crucial.

This should demonstrate the usefulness of displayed monoidal categories for getting better code for already known structures - compare this with the previous implementation that does not go through dispalyed notions.


rmatthes:
This is a small PR, and I think it would be good to have it in master before the TYPES'22 talk on Tuesday.

rmatthes:
This is a small PR, and I think it would be good to have it in master before the TYPES'22 talk on Tuesday.

PR #1518, opened 2022-06-21
examples of monoidal categories

rmatthes:
two alternative approaches when the monoidal category is cartesian monoidal and comes from a displayed category
(1) through displayed monoidal categories, where the displayed tensor has to be constructed
(2) through displayed binary products that have to be constructed


benediktahrens:
Does that mean that the present PR makes #1516 obsolete?

Kfwullaert:
> The branch on which that PR was based has been merged and then evolved constructively.

As a confirmation: Then I agree!

PR #1521, opened 2022-06-28
constructs the displayed cartesian monoidal category

rmatthes:
requiring displayed binary products and displayed terminal object

The proofs of the tensor laws look clumsy. I would be glad to see advice for a slicker development - numerous displayed monoidal laws are treated the same tedious way.


rmatthes:
The extra benefit of this development should be to have most of the tools for binary products now also for displayed binary products (and analogously for terminal objects).

UniMath/CategoryTheory/DisplayedCats/Binproducts.v
```diff
@@ -198,7 +412,9 @@ Section FixDispCat.
           (* we proceed as follows: *)
           cbn.
           etrans.
-          2: { apply dispBinProductArrowEta. }
+          2: { assert (aux := dispBinProductArrowEta (Ps c d) cc dd (dPs _ _ cc dd) aa fgfg).
+               apply transportf_transpose_left in aux.
+               exact aux. }
           apply (maponpaths (fun z => transportf (mor_disp aa (dispBinProductObject (Ps c d) (dPs c d cc dd))) z fgfg)).
           apply C.
   Defined.
```
benediktahrens, line 420: I am anticipating performance problems using this construction, since it is entirely transparent. Is there a way to separate existence and uniqueness?

UniMath/CategoryTheory/Monoidal/DisplayedCartesianMonoidalCategoriesWhiskered.v
```diff
@@ -0,0 +1,1044 @@
...
+             etrans.
+             2: { apply pathsinv0, mor_disp_transportf_prewhisker. }
+             rewrite id_right_disp.
+             rewrite transport_f_b.
+             apply transportf_comp_lemma.
+             apply transportf_comp_lemma_hset; try apply homset_property; apply idpath.
+      }
+      apply transportf_comp_lemma.
+      apply transportf_comp_lemma_hset; try apply homset_property; apply idpath.
+      Unshelve.
```
benediktahrens, line 348: I wonder which tactic caused a goal to be shelved. Often this happens with `refine`, in which case `simple refine` or `use` might be better to use.

benediktahrens (commented):
LGTM, but I think that one construction should be split for performance reasons.

UniMath/CategoryTheory/Monoidal/DisplayedCartesianMonoidalCategoriesWhiskered.v
rmatthes: I made a further commit to illustrate what happens to the shelf. The only tactics that create new goals on the shelf are now 'etrans' (which is shorthand for 'eapply pathscomp0') and 'apply dispBinProduct_endo_is_identity'. Both do not come with 'refine'. Both tactics applications often are followed by a proof that resolves the shelf goal, but for both not always. What would be a good practice to deal with those goals as early as possible?

UniMath/CategoryTheory/DisplayedCats/Binproducts.v
rmatthes: I made an effort towards separation of data and proofs. Unfortunately, the definition of `isBinProduct` does not provide any abbreviation for intermediary steps.

UniMath/CategoryTheory/Monoidal/DisplayedCartesianMonoidalCategoriesWhiskered.v
benediktahrens: The shelved goal from `etrans` is always solved automatically, since the proof of the shelved goal (the intermediate object) is automatically inferred from the type of the two arguments (paths) produced in the explicit goals.

For `dispBinProduct_endo_is_identity` it is different - this construction has a few hypotheses which do not occur in the types of later arguments, so these hypotheses won't be inferred automatically. Possibly the shelving happens because an argument is declared implicit when it should be declared explicit? Only arguments that can be inferred from (the type of) later arguments should be declared implicit.

PR #1523, opened 2022-07-02
Kw bicat of univ monoidal cats is univalent

Kfwullaert:
Added all files of the (almost finished) univalence proof. 
TODO:: Have to show one equality (as part of an equivalence) in the tensor layer, in the associator,left and rightunitor layers, have to show composition. 
ADVICE NEEDED: Is it possible to have coercions between the type of objects of (bi)categories?


nmvdw:
> ADVICE NEEDED: Is it possible to have coercions between the type of objects of (bi)categories?

What do you mean? 

I interpret it as follows: you have two bicategories and you want to have a coercion from one to the other?

Kfwullaert:
> I interpret it as follows: you have two bicategories and you want to have a coercion from one to the other?
@nmvdw 

Yes indeed. For general categories I can see why this is not possible. But I ask this more specifically for displayed (bi)categories. Lets say the underlying bicategory has some notations. If you build another bicategory on top of that, using displayed bicategories, you would have to define the notations again. But this is cumbersome. So it would be nice if there would be a way  to have the notation inhereted in a certain way.

Kfwullaert:
This pull request should be ready to get merged.

rmatthes:
I am ready to have a look. But I also expect more work to do when merging the current master into your branch.

Kfwullaert:
> I am ready to have a look. But I also expect more work to do when merging the current master into your branch.

@rmatthes what do you mean precisely? I've just merged the master branch into this branch without any issues?

rmatthes:
@Kfwullaert I recall a discussion about how much to put into the data part of a monoidal category (in whiskered form). And that this would be relevant for the planned univalence proof. Now, the change in the definition of what the data is appears to be transparent to the univalence argument. All the better!

Kfwullaert:
> I have to retract my appreciation of the transparence mentioned in my last comment. Looking at the file `Tensorlayer.v`, I do not even see shared code between the notion of whiskered monoidal category and what the tensor layer introduces. The final result of this PR is about some formulation of a bicategory of univalent monoidal categories. What notion of monoidal categories is in its objects? If the development is not related to the notions in the library, then changes in that library naturally do not have an influence on the result.

You are indeed right in that I've not literally copy pasted everything from the library, I did used it and changed it a little but. 
But isn't this the same with bicategories, there are multiple different formalisation, but I think this is positive since this allows you to use whatever you want to use?

rmatthes:
We should discuss this outside of this specific PR. But I had to correct my previous comment to this PR in the same place.

UniMath/wc_count_monoidal.l
```diff
@@ -0,0 +1 @@
+../../monoidal/wc_count_monoidal.sh
```
DanGrayson, line 1: What's this about?

rmatthes: This must have inadvertently been committed. The shell script in question quantifies recent contributions made to UniMath that are described in an upcoming workshop talk.

PR #1527, opened 2022-07-25
Added proof of univalence of the bicategory of univalent monoidal categories

Kfwullaert:
This pull request now makes use of the sigma construction of displayed bicategories.


Kfwullaert:
> Would there be a way to define a bicategory of monoidal categories on the basis of categories instead of univalent categories and only restrict to univalent categories for the purpose of getting the univalence result? But maybe this would be more of a hassle than to copy the code to a version without univalence in the ingredients and accordingly without univalence results.

@rmatthes
One needs to be able to restrict a displayed bicategory to a subbicategory (in our case a subtype of the type of objects in the sense of just adding a property to the objects) and in particular a result that the total category is a subcategory as well. 
@nmvdw What do you think?

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/TensorLayer.v
```diff
@@ -0,0 +1,1077 @@
...
+      rewrite transport_f_f.
+      rewrite pathsinv0l.
+      apply idpath_transportf.
+    - intro q.
+      apply homset_property.
+    - intro q.
+      apply homset_property.
+  Defined.
+
+  Lemma transport_of_bifunctor_map_is_pointwise {C : univalent_category}
```
rmatthes, line 714: I see that there is also a price to be paid for the currying approach.

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/TensorLayer.v
```diff
@@ -0,0 +1,1077 @@
...
+
+  Definition tensor_eqi'''_to_eq {C : univalent_category} (TC TD : tensor C) : tensor_eqi''' TC TD -> TC = TD.
+  Proof.
+    intro te.
+    use total2_paths_f.
+    - use total2_paths_f.
+      + (* apply funextsec ; intro ; apply funextsec ; intro ; apply (pr1 te). *)
+        exact (tensor_eqi'''_to_eq1 TC TD te).
+      + repeat (apply funextsec ; intro).
+        use pathscomp0.
```
rmatthes, line 745:
```suggestion
        etrans.
```

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/TensorLayer.v
```diff
@@ -0,0 +1,1077 @@
...
+        3: apply (pr2 te).
```
rmatthes, line 746:
```suggestion
        2: apply (pr2 te).
```
`use pathscomp0` rarely serves a purpose

rmatthes (commented):
This is a preliminary review only on the two first important parts of the construction. Overall, it looks very good.

rmatthes (commented):
Another intermediary feedback.

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/AssociatorLayer.v
```diff
@@ -0,0 +1,332 @@
...
+              ** exact tt.
+              ** use tpair.
+                 --- apply isapropunit.
+                 --- apply isapropunit.
+           ++ use tpair.
+              ** exact tt.
+              ** use tpair.
+                 --- apply isapropunit.
+                 --- apply isapropunit.
+  Defined.
```
rmatthes, line 215: This proof-checking takes excessively long time. Is it better when following my advice on the treatment of unitors?

rmatthes (commented):
all that is said about the left unitor also hold for the right unitor and the associator - they are developed in very close correspondence

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/TensorLayer.v
```diff
@@ -0,0 +1,1077 @@
...
+          rewrite id_left in t.
+          rewrite tensor_id in t.
+          rewrite id_left in t.
+          unfold compositions_preserves_tensor_data in t.
+          exact t.
+    - intros x1 x2 y1 y2 f g.
+      exact (ptnat x1 x2 y1 y2 f g).
+  Defined.
+
+  Lemma swap_nat_along_zisos {C : category} {x1 x2 y1 y2 : C}
```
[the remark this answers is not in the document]
Kfwullaert: That is a good question, I have no idea why I called it like this.

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/TensorLayer.v
Kfwullaert: @rmatthes Yes indeed, we always need some transport lemma. For the whiskered approach, we would need another transportation lemma, something like transportation distributes over a pair. So in the end, it wouldn't have made such a big difference by using a whiskered approach.

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/AssociatorLayer.v
```diff
@@ -0,0 +1,332 @@
+(*
+This is one file which leads to showing that the bicategory of univalent monoidal categories is again univalent.
+In this file we construct one side of the second displayed layer above the bicategory of univalent categories, more precisely:
```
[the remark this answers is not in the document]
Kfwullaert: @rmatthes  With this I mean that the second layer consists of 2 sides, the associator and the unitors (although the left and right unitor are treated separately.

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/FinalLayer.v
```diff
@@ -0,0 +1,574 @@
...
+  Proof.
+    apply disp_locally_groupoid_sigma.
+    - apply univalent_cat_is_univalent_2.
+    - apply disp_tensor_unit_unitors_associator_disp_2cells_isaprop.
+    - apply disp_2cells_isaprop_fullsubbicat.
+    - apply disp_tensor_unit_unitors_associator_disp_locally_groupoid.
+    - apply disp_locally_groupoid_fullsubbicat.
+  Qed.
+
+  Definition bicat_univlaxmon : bicat := total_bicat disp_univlaxmon.
```
rmatthes, line 100: This bicategory is a bit peculiar: the 1-cells are lax as expected, but also the objects are lax monoidal categories. Of course, that is sufficient to define lax monoidal functors, but it would even be so for the definition of (strong) monoidal functors. So, it this really a bicategory that should be put forward?

rmatthes (changes requested):
comments on the final layer, including a question if the lax bicategory is desirable
the remaining `Admitted` must be addressed before a merge

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/FinalLayer.v
```diff
@@ -0,0 +1,574 @@
...
+    ∏ (C D E : bicat_univlaxmon) (F : bicat_univlaxmon⟦C,D⟧) (G : bicat_univlaxmon⟦D,E⟧),
+      P_strong_preserving _ _ F → P_strong_preserving _ _ G -> P_strong_preserving _ _ (F·G).
+  Proof.
+    intros C D E F G sF sG.
+    use tpair.
+    - apply is_z_isomorphism_comp.
+      + apply sG.
+      + cbn in *.
+
+  Admitted.
```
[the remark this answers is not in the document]
Kfwullaert: I guess I forgot about this. I'll check what happened here.

rmatthes:
The second layer for the unitors and the associator also redoes a generic construction several times. For example, one could just put
`Definition bidisp_lu_disp_prebicat : disp_prebicat tu_cat := disp_cell_unit_prebicat disp_lu_disp_cat_data.`
after having imported `UniMath.Bicategories.DisplayedBicats.Examples.DisplayedCatToBicat.`.
This can save three times 5 definitions and bring more clarity.

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/FinalLayer.v
Kfwullaert: @rmatthes I understand your comment. I've been thinking about this as well. There is a "clean" solution for this, just add an extra displayed bicategory on top of this where you require that the unitors and associator become invertible and just using that you a full displayed bicategory of something univalent. However, then you don't have a nice separation between data and properties. If one wants to separate it, then one could do the following (which is less efficient):
Copy paste the files on the unitors/associator and where you add the inverse and on then add a displayed bicategory on the product displayed bicategory of e.g. the left unitor and the left unitor inverse where you then take a full subbicategory where you require them to really be inverses.

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/AssociatorLayer.v
Kfwullaert: I've changed this as you proposed, which definitely looks a lot cleaner, but there is no difference in time.

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/FinalLayer.v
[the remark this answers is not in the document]
Kfwullaert: It is finished now.

Kfwullaert:
> The second layer for the unitors and the associator also redoes a generic construction several times. For example, one could just put `Definition bidisp_lu_disp_prebicat : disp_prebicat tu_cat := disp_cell_unit_prebicat disp_lu_disp_cat_data.` after having imported `UniMath.Bicategories.DisplayedBicats.Examples.DisplayedCatToBicat.`. This can save three times 5 definitions and bring more clarity.

@rmatthes That is indeed an excellent point and worth changing. I'll try to do that.

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/FinalLayer.v
rmatthes: For future uses, it seems that the ugly copying process should be preferred.

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/FinalLayer.v
```diff
@@ -0,0 +1,578 @@
...
+             (tuuaC : disp_univlaxmon C)
+             (tuuaD : disp_univlaxmon D)
+    : tuuaC -->[F] tuuaD
+            -> functor_tensor_unit_unitors_associator F (assunitors_from_layer C (pr1 tuuaC)) (assunitors_from_layer D (pr1 tuuaD)).
+  Proof.
+    intro ptuua.
+    use tpair.
+    - apply equality_functor_assunitors_with_layer.
+      apply (pr1 ptuua).
+    - repeat (use tpair).
```
rmatthes, line 159:
```suggestion
    - repeat split.
```

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/AssociatorLayer.v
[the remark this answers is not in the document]
rmatthes: This is a useful view, but then the header of LeftUnitorLayer.v also claims to construct one side. Maybe better mention that three elements are treated in parallel and even rather similarly.

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/AssociatorLayer.v
rmatthes: When playing your proofs, sometimes the huge goals took some time to show. Maybe you could save time by less unfolding. Of course, this would be pure experimentation.

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/TensorLayer.v
rmatthes: A somewhat sad conclusion.

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/TensorLayer.v
```diff
@@ -0,0 +1,1077 @@
...
+  Qed.
+
+  Definition bidisp_tensor_disp_locally_groupoid : disp_locally_groupoid bidisp_tensor_disp_bicat.
+  Proof.
+    unfold disp_locally_groupoid.
+    intros C D F G α TC TD ptC ptD ptc.
+    unfold is_disp_invertible_2cell.
+    use tpair.
+    - intros x y.
+
```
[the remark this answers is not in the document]
rmatthes: What do you think about by proposal? It shows also fiddly type-checking problems.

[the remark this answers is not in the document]
Kfwullaert: I'm not able to go through it now in detail, so I'll come back to this.

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/TensorLayer.v
```diff
@@ -0,0 +1,1077 @@
...
+      + intros.
+        apply isaprop_bidisp_tensor_disp_2cell_struct.
+  Defined.
+
+
+  (** We now show that this displayed bicategory is globally univalent **)
+
+  Definition tensor_iso {C : univalent_category} (TC TD : tensor C) : UU
+    := ∑ α : ∏ x y : C, z_iso (x ⊗_{TD} y) (x ⊗_{TC} y),
+          ∏ {a1 a2 b1 b2 : C} (f : C⟦a1,a2⟧) (g : C⟦b1,b2⟧),
```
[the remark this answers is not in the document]
rmatthes: There are still 20 more warnings of this kind.

rmatthes:
I see that the blocking problem is solved, but I do not know how to mark this in the system.

Kfwullaert:
@rmatthes  Concerning disp_cell_unit_bicat, you are completely right, not only are the definitions more easy. Also the univalence proofs become even more easy. I didn't have to construct a factorization just a lemma. It is not even worth anymore to have 3 different files for the unitors and associators since it is now so short.

Kfwullaert:
@rmatthes I've changed the unitors and associator layers. Much shorter and easier now. I've changed the "message/overview of the file (i.e. the comment in the beginning of the file)" in the associator layer. Do you think that this is better? If so, I'll do the same change for the unitors. And if you agree I can also put the associators and unitors in the same file since they are so short now.

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/AssociatorLayer.v
```diff
@@ -0,0 +1,123 @@
+(*
+The second displayed layer we lay over the bicategory of univalent categories has the purppose of adding the unitors and the associator. In this file we construct the displayed bicategory which adds all the data of the associator.
```
rmatthes, line 2:
```suggestion
The second displayed layer we lay over the bicategory of univalent categories has the purpose of adding the unitors and the associator. In this file we construct the displayed bicategory which adds all the data of the associator.
```

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/AssociatorLayer.v
```diff
@@ -0,0 +1,123 @@
...
+More precisely:
+The total category corresponding to this displayed layer is the univalent bicategory defined as followed:
```
rmatthes, line 4:
```suggestion
The total category corresponding to this displayed layer is the univalent bicategory defined as follows:
```
This is  a recurrent problem.

rmatthes:
@Kfwullaert I did not see those further simplifications coming. Very well observed. Yes, I think you should merge the three files of the second layer, even if they can be processed in parallel in their current form.

Kfwullaert:
> @Kfwullaert I did not see those further simplifications coming. Very well observed. Yes, I think you should merge the three files of the second layer, even if they can be processed in parallel in their current form.

It might be possible that other lemmas in that file, e.g. the locally groupoid and disp 2cells is a prop, also become a one liner and so that I don't need seperate lemmas for them.

rmatthes:
Still plenty of compilation warnings for a known cause.

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/FinalLayer.v
```diff
@@ -0,0 +1,579 @@
+(*
+This is the last file which concludes that the bicategory of univalent monoidal categories is again univalent.
+In this file we conclude that both the bicategory of univalent monoidal categories with lax monoidal functors is univalent
+and that the result still holds when we replace lax by strong monoidal functors.
```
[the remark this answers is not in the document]
rmatthes: But I fear that the lax monoidal categories are not replaced by monoidal categories!

rmatthes (commented):
I should now have seen all the code.

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/FinalLayer.v
```diff
@@ -0,0 +1,579 @@
...
+
+Section StrongMonoidalCatLayer.
+
+  Definition P_sass : bicat_univlaxmon -> UU
+    :=  λ C : bicat_univlaxmon,
+        strong_associator (pr2 (assunitors_from_layer (pr1 C)  (pr12 C))).
+
+   Definition P_slu : bicat_univlaxmon -> UU
+    :=  λ C : bicat_univlaxmon,
+        strong_lunitor (pr2 (assunitors_from_layer (pr1 C)  (pr12 C))).
```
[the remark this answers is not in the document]
Kfwullaert: @rmatthes They indeed give the same output, but I assume that this is because of the coercions you have from unitors_associator to the left unitor, right unitor and the associator. But I'll temporary remove the coercions as a test to be sure of this.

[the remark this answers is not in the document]
Kfwullaert: @rmatthes It is indeed because of the coercions. If you want, I can remove the coercions, I only had to change something in the curriedmonoidalcategories file, in the other files nothing should have to be adapted.

Kfwullaert:
> Still plenty of compilation warnings for a known cause.

@rmatthes Should I remove the implicit binder declarations (all in the tensorlayer) in order to get rid of these warnings?

rmatthes:
@Kfwullaert Yes, please remove these curly variable binders.

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/FinalLayer.v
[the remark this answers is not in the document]
rmatthes: It is better to avoid this very unwanted effect - thank you.

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/TensorLayer.v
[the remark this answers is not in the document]
rmatthes: What is so difficult about this little piece of code?

Kfwullaert:
@rmatthes I think everything what should be in there is (if I did not forget anything). There is 1 lemma that I've admitted. The proof is commented out because, although the proof runs without a problem, the Qed is stuck in a loop. So could you have a look at that to see what is the issue?

rmatthes:
Where is file `UniMath/CategoryTheory/Monoidal/MonoidalCategoriesReordered.v`?

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/CurriedMonoidalCategories.v
```diff
@@ -947,71 +953,75 @@ Section PentagonTriangle.
...
     := ∏ (x : C), is_z_isomorphism (pr1 ru^{ua} x).
 
-  Definition strong_associator {C : category}
+  Definition associator_invertible {C : category}
              {tu : tensor_unit C}
              (ua : unitors_associator tu) : UU
     := ∏ (x y z : C), is_z_isomorphism (pr1 ass^{ua} x y z).
 
-  Definition strong {C : category}
+  Definition invertible_data {C : category}
```
rmatthes, line 977: The name suggests that all inside is data, but there are also all the proofs of `is_inverse_in_precat`.

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/FinalLayer.v
```diff
@@ -96,19 +123,186 @@ Section LaxLayer.
     - apply disp_locally_groupoid_fullsubbicat.
   Qed.
 
-  Definition bicat_univlaxmon : bicat := total_bicat disp_univlaxmon.
+  Definition UMONCAT : bicat := total_bicat disp_univmon.
```
rmatthes, line 126: It would be good to indicate in a comment that this is the bicategory of monoidal categories (over univalent categories) and lax monoidal functors, both in a form that follows the constructions process that made it possible to carry out a modular proof of univalence.

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/FinalLayer.v
```diff
@@ -96,19 +123,186 @@ Section LaxLayer.
...
+    - apply disp_univstrongfunctor_disp_2cells_isaprop.
+    - apply disp_sub1cell_univalent_2.
+      + apply UMONCAT_is_univalent_2.
+      + intro ; intros ; apply isaprop_P_strong_preserving.
+    - apply disp_univmon_disp_locally_groupoid.
+    - apply disp_univstrongfunctor_disp_locally_groupoid.
+    - apply disp_bicat_univstrongfunctor_is_univalent_2.
+  Qed.
+
+  Definition UMONCAT_strong : bicat := total_bicat disp_univstrongfunctor.
```
rmatthes, line 290: Similar to above, it would be good to indicate in a comment that this is the bicategory of monoidal categories (over univalent categories) and monoidal functors (non-lax, but not requiring functorial strength), both in a form that follows the constructions process that made it possible to carry out a modular proof of univalence.

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/CurriedMonoidalCategories.v
Kfwullaert: What do you think of "unitorsassociator_are_invertible"?

rmatthes: Maybe `unitorsassociator_are_z_isos` - this would be more technical but lets the insiders know that there is data involved.

UniMath/Bicategories/MonoidalCategories/UnivalenceMonCat/FinalLayer.v
```diff
@@ -0,0 +1,317 @@
...
+    - apply disp_univstrongfunctor_disp_2cells_isaprop.
+    - apply disp_sub1cell_univalent_2.
+      + apply UMONCAT_is_univalent_2.
+      + intro ; intros ; apply isaprop_P_strong_preserving.
+    - apply disp_univmon_disp_locally_groupoid.
+    - apply disp_univstrongfunctor_disp_locally_groupoid.
+    - apply disp_bicat_univstrongfunctor_is_univalent_2.
+  Qed.
+
+  (* UMONCAT is the bicategory defined as follows:
```
rmatthes, line 299:
```suggestion
  (* UMONCAT_strong is the bicategory defined as follows:
```

rmatthes:
@Kfwullaert 
> @rmatthes I think everything what should be in there is (if I did not forget anything). There is 1 lemma that I've admitted. The proof is commented out because, although the proof runs without a problem, the Qed is stuck in a loop. So could you have a look at that to see what is the issue?

I have solved this by playing a bit around with it - see PR no. 9 of your fork.

rmatthes:
I maybe have lost track of a previous statement on the following matter: the file CurriedMonoidalCategories.v is not dependent on the notion of bicategory and should be moved into UniMath/CategoryTheory/Monoidal in a subsequent PR that is confined to the renaming. But how is that positioned w.r.t. the already existing MonoidalCategoriesCurried.v in that directory?

rmatthes (changes requested):
Monoidal/EquivalenceWhiskeredNonCurriedMonoidalCategories.v
in CategoryTheory/.package/files
needs to be moved downwards - for example right after 
Monoidal/MonoidalCategories.v

rmatthes (approved):

PR #1528, opened 2022-07-26
Added an example of a monoidal section where an object in the total c…

Kfwullaert:
…ategory consists of a set together with a subset.

### DIFF
--- a/UniMath/CategoryTheory/Monoidal/Examples/SetWithSubset.v
+++ b/UniMath/CategoryTheory/Monoidal/Examples/SetWithSubset.v
@@ -1,0 +1,490 @@
+(*
+In this file we construct the category whose objects are pairs, consisting of a set and a subset of that set, as a displayed category.
+Furthermore, we show that this displayed category is monoidal
+and we construct a monoidal section which maps a set X to (X,X) (where the second X is considered to be maximal subset of X).
+*)
+
+Require Import UniMath.Foundations.All.
+Require Import UniMath.MoreFoundations.All.
+
+Require Import UniMath.CategoryTheory.Core.Categories.
+Require Import UniMath.CategoryTheory.Core.Functors.
+Require Import UniMath.CategoryTheory.Core.Isos.
+Require Import UniMath.CategoryTheory.DisplayedCats.Core.
+Require Import UniMath.CategoryTheory.DisplayedCats.Constructions.
+
+Require Import UniMath.CategoryTheory.Monoidal.WhiskeredBifunctors.
+Require Import UniMath.CategoryTheory.Monoidal.MonoidalCategoriesWhiskered.
+Require Import UniMath.CategoryTheory.Monoidal.MonoidalFunctorsWhiskered.
+Require Import UniMath.CategoryTheory.Monoidal.WhiskeredDisplayedBifunctors.
+Require Import UniMath.CategoryTheory.Monoidal.DisplayedMonoidalWhiskered.
+Require Import UniMath.CategoryTheory.Monoidal.MonoidalSectionsWhiskered.
+
+
+Require Import UniMath.CategoryTheory.categories.HSET.All.
+Require Import UniMath.CategoryTheory.Monoidal.CartesianMonoidalCategoriesWhiskered.
+
+Section Subsets.
+
+  Definition subset (X : hSet) : UU := X -> hProp.
+
+  Definition subset_to_elements' {X : hSet} (U : subset X)
+    : UU := ∑ x : X, U x.
+
+  Definition subset_to_elements {X : hSet} (U : subset X) : hSet.
+  Proof.
+    exists (subset_to_elements' U).
+    apply isaset_total2.
+    - apply X.
+    - intro.
+      apply isasetaprop.
+      apply U.
+  Defined.
+
+  Definition subset_to_elements_in_entireset
+             {X : hSet} {U : subset X}
+             (x : subset_to_elements U)
+    : X := pr1 x.
+
+  Definition subset_in_product {X Y : hSet}
+             (U : subset X) (V : subset Y)
+    : subset (setdirprod X Y).
+  Proof.
+    intro xy.
+    exists (U (pr1 xy) × V (pr2 xy)).
+    apply isapropdirprod. apply U. apply V.
+  Defined.
+
+  Definition subset_in_product' {X Y : hSet} (UV : subset (setdirprod X Y))
+    : subset X × subset Y.
+  Proof.
+    split.
+    - intro x.
+      exists (∃ y : Y, UV (x,,y)).
+      apply ishinh.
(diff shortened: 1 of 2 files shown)


rmatthes:
I have started studying the file. There are definitions that already appear in UniMath.Foundations.Sets or UniMath.MoreFoundations.Sets. For example, the first three definitions could be shrunk to
`Definition subset (X : hSet) : UU := hsubtype X.
 Definition subset_to_elements {X : hSet} (U : subset X) : hSet := Sets.subset U.`
Unfortunately, this reveals that the naming scheme is not compatible with the library (maybe write setsubtype instead of subset).  
Also the next concept is already in the library - one could write:
`Definition subset_in_product {X Y : hSet}
             (U : subset X) (V : subset Y)
    : subset (setdirprod X Y) := subtypesdirprod U V.`
However, the file also contains new elements that might be worth adding upstream (with a PR rather involving MoreFoundations than Foundations). But `subset_in_product'` is not used further, and it is not validated by a lemma that the definition captures the expected concept.
Only the last 200loc are specific to monoidal categories.

rmatthes (changes requested):
This is a nice example for the announced purpose. However, I would hope to see a better integration with the already existing elements in the library (by reusing it or by contributing to it upstream). I laid this out in my comment to the PR.

Kfwullaert:
@rmatthes I've integrated the work. I've created a section Subtype_AUX where I've put certain things which could be set in the foundations (or more foundations). So we maybe should discuss about where we could put that code before merging this.

benediktahrens:
@rmatthes : have your change requests been addressed?

UniMath/CategoryTheory/Monoidal/Examples/SetWithSubset.v
```diff
@@ -0,0 +1,415 @@
...
+
+  Definition SS_disp_monoidal_laws : disp_monoidal_laws (SS_disp_monoidal_data).
+  Proof.
+    repeat split ; try (intro ; intros) ; apply isaprop_hsubtype_preserving.
+  Qed.
+
+  Definition SS_disp_monoidal : disp_monoidal SS_disp_cat SET_cartesian_monoidal
+    := SS_disp_monoidal_data,, SS_disp_monoidal_laws.
+
+  Definition entire_section_monoidal_data : smonoidal_data SET_cartesian_monoidal SS_disp_monoidal total_section .
```
rmatthes, line 388: The names contain "entire" here. Why not the same as for the underlying category? (However, "total" without further qualification seemed too little, see above.)

rmatthes (approved):
The file looks mostly okay. Please take into account my suggestions. In their majority, they concern organization.

UniMath/CategoryTheory/Monoidal/Examples/SetWithSubset.v
```diff
@@ -0,0 +1,415 @@
...
+    : subtype_containedIn (binary_intersection U V) V.
+  Proof.
+    intros ? xinUV.
+    apply xinUV.
+  Qed.
+
+  Definition intersection_contained {X : UU} {U U' V V' : hsubtype X}
+             (uu : subtype_containedIn U U')
+             (vv : subtype_containedIn V V')
+    : subtype_containedIn (binary_intersection U V) (binary_intersection V' U').
```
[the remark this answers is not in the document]
Kfwullaert: This I do because the only lemma where I need it, I need precisely this form. I know it seems weird, I did it at forehand in the 'natural way, but I changed it to this for this reason.

UniMath/CategoryTheory/Monoidal/Examples/SetWithSubset.v
Kfwullaert: Excellent point, no specific reason I remember. Do you think entire is better suited then total or do you have any other suggestion?

UniMath/CategoryTheory/Monoidal/Examples/SetWithSubset.v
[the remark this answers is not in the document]
rmatthes: Are you sure? I changed it locally without distant effects.

UniMath/CategoryTheory/Monoidal/Examples/SetWithSubset.v
rmatthes: All with "entire" would at least exclude that misunderstanding. Or something with an extra word, such as "total_subset".

rmatthes (approved):
While it is okay to merge the PR, it would still be better if the parts of general nature were moved upstream (as indicated in my previous comments).

Kfwullaert:
> While it is okay to merge the PR, it would still be better if the parts of general nature were moved upstream (as indicated in my previous comments).

Sorry for the delay, I'll do the moving in the afternoon!

UniMath/MoreFoundations/Subtypes.v
```diff
@@ -196,3 +196,206 @@ Proof.
...
+Lemma total_hsubtype_preserving {X Y : UU} (f : X → Y)
+  : hsubtype_preserving (totalsubtype X) (totalsubtype Y) f.
+Proof.
+  exact (λ _ _, tt).
+Qed.
+
+Section SetSubtype.
+
+  Definition setsubtype (X : hSet) : UU := hsubtype X.
+  Identity Coercion setsubtype_to_subtype : setsubtype >-> hsubtype.
```
benediktahrens, line 380: Why is it necessary to introduce this synomym?

UniMath/MoreFoundations/Subtypes.v
```diff
@@ -196,3 +196,206 @@ Proof.
...
+
+  Definition singletonsetsubtype {X : hSet} (x : X) : setsubtype X.
+  Proof.
+    intro x'.
+    exists (x'=x).
+    apply X.
+  Defined.
+
+  Definition emptysetsubtype {X : hSet} : setsubtype X := emptysubtype X.
```
benediktahrens, line 389: The argument `X` should probably not be implicit here.

UniMath/MoreFoundations/Subtypes.v
```diff
@@ -196,3 +196,206 @@ Proof.
...
+
+  Definition setsubtype_in_product {X Y : hSet}
+             (U : hsubtype X) (V : hsubtype Y)
+    : setsubtype (setdirprod X Y) := subtypesdirprod U V.
+
+  Definition union {X I : hSet} (U : I -> setsubtype X) :  setsubtype X := subtype_union U.
+
+  Definition totalsetsubtype (X : hSet) : setsubtype X := totalsubtype X.
+
+  Definition contained {X : hSet} (U V : setsubtype X) : UU := subtype_containedIn U V.
```
benediktahrens, line 399: Why introduce all these synomyms?

PR #1532, opened 2022-08-01
identifies inserter category as abstract construction behind CPP'22

rmatthes:
(final section of paper with Benedikt and Anders) 

The basic situation is now developed in CommaCategories.v, with the equation being a special case of the one for comma categories - which is turned around w.r.t. to the previous definition. Therefore, in the application, some equations were turned around in the statements, but mostly, there are now more uses of pathsinv0.

desirable next step: do the construction of inserter category analogously for monoidal categories with strong monoidal functors and profit from displayed monoidal categories and displayed sections (maybe even establish the construction as the inserter in the corresponding bicategory)

Is there a chance to simplify the application by yet more abstraction?


nmvdw (commented):
Note: for bicategories, we defined inserters (https://github.com/UniMath/UniMath/blob/master/UniMath/Bicategories/Limits/Inserters.v) and we showed that `Cat` has inserters (https://github.com/UniMath/UniMath/blob/master/UniMath/Bicategories/Limits/Examples/BicatOfUnivCatsLimits.v). There are some alternative names for the inserter category
- Subequalizer (see the paper 'Subequalizers' by Lambek)
- Dialgebras (https://ncatlab.org/nlab/show/dialgebra)
The second name comes from the fact that inserters generalize both algebras and coalgebras on a functor.

In UniMath, we have a formalization of the category of dialgebras and a proof that it's the inserter. It is also constructed using displayed categories. See
https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/categories/Dialgebras.v
I think it might be good to add the new constructions to that file.

rmatthes:
@nmvdw Thanks a lot for this valuable feedback. It is a bit unfortunate that dialgebras have their characteristic equality oriented differently from the comma category of which they are a special case. The new constructions are now in the file for dialgebras. The proof that they form the inserter will be the model for the desired construction on monoidal categories.

nmvdw:
> It is a bit unfortunate that dialgebras have their characteristic equality oriented differently from the comma category of which they are a special case.

Good point. Do you think it is difficult to make this change?

Note: as far as I know, inserters are not really a special case of comma categories. If we have `F, G : C -> D`, then the comma category of `F` and `G` would have as objects triples `x1, x2 : C` together with a morphism `F x1 -> G x2. However, the inserter would have only one object in `C`.

It is the case that inserters can be constructed if one has enough limits. For example, I think you can construct from comma categories and iso-comma categories or from comma categories and equifiers.

rmatthes:
@nmvdw I am sorry for my bad use of English grammar: I had intended to say that the equality for dialgebras is a special case of the one for comma categories, not that dialgebras are a special case of comma categories. I'll try out if turning around the equality requires much effort later.

benediktahrens:
Is this PR ready for review/merging?

benediktahrens:
The PR contains a construction for not-necessarily univalent categories that already exists for univalent categories. Why are we interested in the construction on non-univalent categories?

rmatthes:
@benediktahrens This PR has grown from a better connection within the UniMath library to (hopefully) new research. I might have identified the inserter in the bicategory of monoidal categories, but will probably need to have the most elementary definition of it to get a proof in the first place (not the construction that was motivated by getting univalence). Please bear with me for the few lines of code. I'll open another PR for the monoidal dialgebras (already done on another branch based on the one for this PR). I hope that the usefulness of all this will be seen there.

rmatthes:
I do not intend to contribute more material to this PR, another one is in preparation. So, please merge if it seems sound.

PR #1541, opened 2022-08-20
Update to Coq8.15.2

benediktahrens:
This PR does not work yet. I don't know what to do to fix it.

- `ltac` is not a make target any more, it seems. I temporarily make `world` to compile everything needed
- the `make install` seems to be necessary when compiling Coq 8.15.2 manually. Here, it seems to create a loop.

@DanGrayson : can you help?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -229,8 +229,8 @@ distclean::          ; rm -f build/Makefile-configuration
 #############################################################################
 # building coq:
 export PATH:=$(shell pwd)/sub/coq/bin:$(PATH)
-CONFIGURE_OPTIONS := -coqide "$(COQIDE_OPTION)" -with-doc no -local -no-custom
-BUILD_TARGETS := coqbinaries tools states ltac
+CONFIGURE_OPTIONS := -coqide "$(COQIDE_OPTION)" -with-doc no -prefix $(shell pwd)
+BUILD_TARGETS := coqbinaries tools states world
 ifeq ($(DEBUG_COQ),yes)
 CONFIGURE_OPTIONS += -annot
 BUILD_TARGETS += byte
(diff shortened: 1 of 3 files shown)


DanGrayson:
Okay, I'll take a look.

DanGrayson:
okay, what I just pushed should make it work under ubuntu

DanGrayson:
And it did.  Now for the macos build ...

DanGrayson:
That worked.

Makefile
```diff
@@ -229,8 +229,8 @@ distclean::          ; rm -f build/Makefile-configuration
 #############################################################################
 # building coq:
 export PATH:=$(shell pwd)/sub/coq/bin:$(PATH)
-CONFIGURE_OPTIONS := -coqide "$(COQIDE_OPTION)" -with-doc no -local -no-custom
-BUILD_TARGETS := coqbinaries tools states ltac
+CONFIGURE_OPTIONS := -coqide "$(COQIDE_OPTION)" -with-doc no -prefix $(shell pwd)
+BUILD_TARGETS := coqbinaries tools states world
```
benediktahrens, line 233: @DanGrayson : Do you know how to get rid of the `world` target? It would be great to not compile all of Coq all the time.

DanGrayson: If we're supposed to build coq, as indicated by BUILD_COQ=yes, then I don't see a way to avoid the target "world".  I think it just means "all".

What part of coq do you think does not have to be built?

DanGrayson: I'll take a look ...

benediktahrens, line 233:
```suggestion
BUILD_TARGETS := coqbinaries tools states coq
```

DanGrayson:
Thanks!

PR #1551, opened 2022-09-22
Univalent Kleisli categories

nmvdw:
At every UniMath school, there was an exercise, which was about showing that the usual definition of the Kleisli category does not give rise to a univalent category.

In this pull request, I define a univalent version of the Kleisli category and show that it satisfies the universal property for Kleisli objects as given in the formal theory of monads by Street.

One interesting thing to note: to deduce the universal property, I first show that the bicategory of (not necessarily univalent) categories has Kleisli objects. Then I construct a fully faithful and essentially surjective functor between the two variants of the Kleisli category, and then I can deduce the desired universal property (basically Theorem 9.9.4 in the HoTT book).

Contents:
- Definition of Kleisli objects
- If a bicategory `B` has Kleisli objects, then `B^op` has Eilenberg-Moore objects
- The bicategory of categories (not necessarily univalent) has Kleisli objects. We use the usual definition of Kleisli category
- The bicategory of univalent categories has Kleisli objects. We define these as the image of the free algebra functor. Note: one cannot use the Rezk completion, because that would increase the universe level (we need to stay within the same universe).


peterlefanulumsdaine:
This is a very neat observation — obvious once you say it, but I’ve not seen it noticed in the context of univalent categories before.  Is it noted in the literature somewhere, by you or otherwise?

This definition of the Kleisli category as a full subcategory of the EM-category is certainly out there to some extent: [the n-lab takes it as their primary definition](https://ncatlab.org/nlab/show/Kleisli+category), and I feel like I’ve seen it mentioned somewhere at least as an alternative definition.  On the other hand, checking a few standard textbooks on my shelf (Mac Lane, Adámek–Rosičky, Lawvere–Schanuel, Johnstone, Leinster, Riehl…), and the original Kleisli 1965 paper, none use it or even suggest it explicitly as an alternative — the discussion in Mac Lane comes closest.

benediktahrens (approved):

nmvdw:
The construction is mentioned in https://arxiv.org/pdf/2102.06275.pdf (page 115), and there they also say that this construction is univalent. However, they view the construction from the perspective of the formal theory of monads here as described by Street.

Issue #1552, opened 2022-09-22
Not all sanity checks can be run locally


nmvdw:
If I run `make sanity checks` on my laptop (Mac OS 12.2.1), I get
```
--- checking the ordering prescribed by the files UniMath/*/.packages/files ---
make: *** skipping checking the linear ordering of packages, because 'bash' is too old
--- checking every proof file is listed in one of the packages ---
make: *** skipping checking the listing of proof files, because 'bash' is too old
```
As such, I can only see whether the sanity checks are satisfied after doing a pull request.

benediktahrens:
@DanGrayson : could you take a look at this?

DanGrayson:
Yes, bash is too old on mac os x.  On my system it's stuck at version 3.2.57, which is why I use homebrew to install a modern version, 5.1.16.  Since /usr/local/bin is ahead of /bin on my PATH, the modern one gets used instead.

nmvdw:
That helped! Thank you

benediktahrens:
This issue seems to be resolved, closing.

PR #1553, opened 2022-09-27
Recommend installing recent bash on Mac OS, for sanity checks (addresses #1552)

peterlefanulumsdaine:
Added a note in `INSTALL.md`, to document the fix for #1552 noted in the comments on that issue.


DanGrayson:
Thanks!

PR #1554, opened 2022-09-27
Removed long-empty file `UniMath/Folds/aux_lemmas.v`

peterlefanulumsdaine:
The file [`UniMath/Folds/aux_lemma.v`](https://github.com/UniMath/UniMath/blob/master/UniMath/Folds/aux_lemmas.v) has been empty since commit f26b873 in 2019.  This PR deletes it.

(It can sometimes be useful to keep this sort of file around, even when all its lemmas have been upstreamed, in order to have a place available to put more such lemmas.  However, those reasons typically apply (a) when a package is under active development, and (b) when the auxiliary file provides some useful *context* — imports, notation scope setup, etc — that would be non-trivial to recreate.  Neither of those applies here: the package has been stable for several years, and the file is completely empty.)

### DIFF
--- a/UniMath/Folds/.package/files
+++ b/UniMath/Folds/.package/files
@@ -1,5 +1,4 @@
 UnicodeNotations.v
-aux_lemmas.v
 folds_precat.v
 from_precats_to_folds_and_back.v
 folds_isomorphism.v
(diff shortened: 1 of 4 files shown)


benediktahrens:
I liked this file for its fast compilation... But I see your point. Thanks for taking care of this!

Issue #1555, opened 2022-09-28
`.dir-locals.el` should be generated by more `make` targets (not just `make all`), and removed by `make distclean` or similar


peterlefanulumsdaine:
Since #1494, the Emacs local variables file `UniMath/.dir-locals.el` is now autogenerated by the build, as suggested + motivated in #810 .  This issue is a couple of points about it.

Firstly: for now, `.dir-locals.el` only gets created when `make all` is invoked — it’s only requested in the Makefile via the line
```make
   all: UniMath/.dir-locals.el
```

But a user building only a subset of packages (by e.g. `make Algebra`) will still typically want `.dir-locals.el`.  So presumably `.dir-locals.el` should at least be a dependency of the package targets.  Is that enough, or are there other earlier-under-dependency targets that should also request it?

Secondly: presumably either `make clean` or `make distclean` should remove `.dir-locals.el` — I guess `make distclean`?

DanGrayson:
Good points.

benediktahrens:
Is this issue solved completely by #1556 ?

peterlefanulumsdaine:
I think so.  #1556 certainly addresses all the issues I raised in this issue, but I wasn’t very confident it was the best way to do so — I’m not so familiar with the full design intent of the Makefile structure (is there an intended conceptual distinction behind exactly what should be compiled in which target?).  But if you and Dan are happy with #1556 then I think this can be closed!

DanGrayson:
> is there an intended conceptual distinction behind exactly what should be compiled in which target?

I think you can just let common sense be your guide there.

PR #1556, opened 2022-09-28
Tweak management of .dir-locals.el from Makefile: build more often, and remove on distclean

peterlefanulumsdaine:
This PR implements a suggested solution to the issues raised in #1555 :

- add `.dir-locals.el` as a dependency of the individual package targets
- add `rm -f UniMath/.dir-locals.el` as part of the `distclean` target


peterlefanulumsdaine:
(rebased this PR to over other recently merged PR’s)

PR #1559, opened 2022-10-03
add Niels' GrpdHITs to CI

benediktahrens:
solves #1533


benediktahrens:
I am seeing this curious error in the building of the satellite repositories:
```
File "./UniMath/Foundations/Init.v", line 186, characters 17-30:
Error: Illegal application (Non-functional construction): 
The expression "is_iso_qinv ?f ?g ?i ?c ?y" of type
 "iscontr (hfiber (precomp_with ?f) ?y)"
cannot be applied to the term
 "?y0" : "?T"

make[1]: *** [Makefile:678: prelude/grpd_bicat.vo] Error 1
make: *** [Makefile:327: all] Error 2
Error: Process completed with exit code 2.
```
This is curious because the error seems to be in UniMath/Foundations, but arises when compiling @nmvdw 's GrpdHITs.

The whole log can be viewed at https://github.com/UniMath/UniMath/actions/runs/3177339156/jobs/5177663170.

DanGrayson:
The error seems to occur when making prelude/grpd_bicat.vo.  And there are oodles of warning messages above the error message.  Maybe the package redefines some core symbols?

nmvdw:
> I am seeing this curious error in the building of the satellite repositories:
> 
> ```
> File "./UniMath/Foundations/Init.v", line 186, characters 17-30:
> Error: Illegal application (Non-functional construction): 
> The expression "is_iso_qinv ?f ?g ?i ?c ?y" of type
>  "iscontr (hfiber (precomp_with ?f) ?y)"
> cannot be applied to the term
>  "?y0" : "?T"
> 
> make[1]: *** [Makefile:678: prelude/grpd_bicat.vo] Error 1
> make: *** [Makefile:327: all] Error 2
> Error: Process completed with exit code 2.
> ```
> 
> This is curious because the error seems to be in UniMath/Foundations, but arises when compiling @nmvdw 's GrpdHITs.
> 
> The whole log can be viewed at https://github.com/UniMath/UniMath/actions/runs/3177339156/jobs/5177663170.

Interestingly enough, I don't get such errors when compiling the files locally.

I don't redefine core symbols. Interestingly enough, those warnings occur when you compile a project using UniMath and you create the makefile using a `_CoqProject` file. In `TypeTheory` and `LargeCatModules` a different way is used to compile the whole project and there, none of those warning are given.

benediktahrens:
Similar warnings also arise when compiling SetHITs (as opposed to GrpdHITs), but the error does not. So maybe we should study the error separately from the warnings?

benediktahrens:
I still don't understand what is wrong here. However, I would like to suggest some changes to the repositories SetHITs and GrpdHITs:

1. Can we use `-Q` instead of `-R`? The former is better behaved, while the latter is too liberal.
2. I believe that point 1 requires a change to the `Import` statements, which then all will have to start like `Require Import sem.foo.`. 

@nmvdw : would you agree to these changes?

benediktahrens:
Things to possibly fix:
- Import statements in GrpdHITs are outdated, @nmvdw is working on that
- Possibly replace `-R` option by `-Q`, requires simple change in many files

nmvdw:
Rerunning the satellites check and address some errors.

nmvdw:
> Things to possibly fix:
> 
> * Import statements in GrpdHITs are outdated, @nmvdw is working on that
> * Possibly replace `-R` option by `-Q`, requires simple change in many files

The continuous integration is accepted now. I didn't use `-Q` yet.

PR #1560, opened 2022-10-06
Mark gradth as deprecated

m-lindgren:
This PR changes gradth to be a Notation for isweq_iso and marks it as [deprecated](https://coq.inria.fr/refman/using/libraries/writing.html#coq:attr.deprecated). This means that coqc / coqtop will issue a warning and a suggestion to use isweq_iso whenever gradth is used. This addresses #792 and also replaces all instances of gradth added since #848.

Any satellite repos that use gradth _should_ still work, but will now trigger warnings.

As this PR contains a change to Foundations I expect `make sanity-checks` to fail.

### DIFF
--- a/.github/workflows/build-unimath.yml
+++ b/.github/workflows/build-unimath.yml
@@ -68,7 +68,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: Install build dependencies
         run: |
-           brew install coq
+           brew install coq ocaml-findlib
            brew config --verbose
            type coqc
            coqc --version
--- a/UniMath/Bicategories/Core/Discreteness.v
+++ b/UniMath/Bicategories/Core/Discreteness.v
@@ -63,7 +63,7 @@ Definition is_univalent_2_1_discrete_bicat
   : is_univalent_2_1 (discrete_bicat C).
 Proof.
   intros x y f g.
-  use gradth.
+  use isweq_iso.
   - exact (λ p, pr1 p).
   - abstract
       (intro p ; cbn ;
@@ -212,7 +212,7 @@ Definition discrete_left_adj_equiv_weq_z_iso
 Proof.
   use make_weq.
   - exact (λ f, _ ,, z_iso_to_discrete_left_adj_equiv HB (pr2 f)).
-  - use gradth.
+  - use isweq_iso.
     + exact (λ f, make_z_iso' _ (discrete_left_adj_equiv_to_z_iso HB f)).
     + abstract
         (intro Hf ;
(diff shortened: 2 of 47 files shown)


m-lindgren:
I have no idea why Build-UniMath-on-MacOS fails. It doesn't seem related to the changes in the PR since it fails as early as on Foundations/Init.v with

> COQC UniMath/Foundations/Init.v
> Error: Anomaly
> "Uncaught exception Failure("Config file not found - neither /usr/local/etc/findlib.conf nor the directory /usr/local/etc/findlib.conf.d")."

but I might be wrong. Maybe someone with MacOS and coq 8.16.0 installed with brew can check if it builds/fails locally.

DanGrayson:
To fix the MacOS build, try this change:
```diff
diff --git a/.github/workflows/build-unimath.yml b/.github/workflows/build-unimath.yml
index 9783b0be..2b4e889d 100644
--- a/.github/workflows/build-unimath.yml
+++ b/.github/workflows/build-unimath.yml
@@ -68,7 +68,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: Install build dependencies
         run: |
-           brew install coq
+           brew install coq ocaml-findlib
            brew config --verbose
            type coqc
            coqc --version
```

peterlefanulumsdaine:
> To fix the MacOS build, try this change:

@DanGrayson Wouldn’t it be better to make that build fix a separate PR?  It’s a very quick + clear one to merge, and wanted by all branches.  This PR looks to me probably a good thing, but worth a little discussion/consideration before merging — it touches Foundations, and it sets a precedent of using new Coq pragmas which we hadn’t been using before.

peterlefanulumsdaine:
I’ve created a separate PR for the CI Mac OS build fix — I think merging it upstream (presuming it works!) and then pulling it into this shouldn’t cause any merge problems.

peterlefanulumsdaine:
On the actual substance of this PR: I like it.  It seems in line with the consensus of the discussions in #792 and #848 — at that point, the `deprecated` attribute didn’t exist yet (iirc), and so we just noted the intended deprecation in comments — but presumably if we could have done this in the code then, we would have.

One implication that seems worth reflecting on: This is a “non-core” feature of Coq that isn’t yet used in UniMath.  But of our two main reasons for being conservative — type-theoretical, and pedagogical — this seems OK to me on both.  Type-theoretically, this doesn’t affect the fragment we stay within; and pedagogically, it makes things clearer (pushing users to the single `isweq_iso`, avoiding haphazard variation between usage of `isweq_iso` and `gradth`) not worse.

So overall, lgtm.

benediktahrens:
Lgtm

peterlefanulumsdaine:
Side note: at the moment, only notations and tactics can be marked `#[deprecated]` — but there’s an open Coq feature request coq/coq#12266 to allow definitions/lemmas/etc to be `#[deprecated]`, and it sounds hopeful to be added soon.  But it seems to me that changing `gradth` to a notation for `iweq_iso` is good in itself anyway (less code duplication, better searchability), so no reason to wait on that feature.

PR #1562, opened 2022-10-07
Rename TopologicalSet to TopologicalSpace

### LINKED ISSUE #1403: change "TopologicalSet"
We should change "TopologicalSet" to "TopologicalSpace" (and require X to be a set), to conform with standard terminology.

Screen Shot 2021-11-01 at 10 10 42 AM

m-lindgren:
This also requires the underlying type of a topological space to be an hSet. Fixes #1403.

### DIFF
--- a/UniMath/AlgebraicGeometry/SheavesOfRings.v
+++ b/UniMath/AlgebraicGeometry/SheavesOfRings.v
@@ -17,7 +17,7 @@ Local Open Scope open.
 (* The category of open subsets of a topological space. *)
 
 Section open_category.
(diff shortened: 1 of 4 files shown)


DanGrayson:
lgtm

benediktahrens (approved):

benediktahrens:
@m-lindgren Thanks for taking care of this!

DanGrayson:
I think the point was that Vladimir wanted to consider the concept where the underlying type of points was not a set.  I don't know why he thought there was some value in doing that.

PR #1564, opened 2022-10-12
Every set is a colimit of its finite subsets. (Resolves #139)

### LINKED ISSUE #139: Colimits in HSET
1. To construct colimits of arbitrary functors C -> HSET (using setquot)
2. To construct, for any h-set X an isomorphism between X and the colimit of its finite subsets (a finite subset is defined as S : X -> hProp such that isfinite ( carrier S ) ) .

I have looked through it on paper and it seems to work.

m-lindgren:
A short proof that every set is a colimit of its finite subsets, as requested in #139.

### DIFF
--- a/UniMath/CategoryTheory/categories/HSET/Colimits.v
+++ b/UniMath/CategoryTheory/categories/HSET/Colimits.v
@@ -10,6 +10,7 @@
   - Pushouts from colimits [PushoutsHSET_from_Colims]
   - Initial object [InitialHSET]
     - Initial object from colimits [InitialHSET_from_Colims]
+  - Every set is the colimit of its finite subsets [is_colimit_finite_subsets_cocone]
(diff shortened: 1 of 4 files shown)


benediktahrens (approved):
Beautiful!

DanGrayson:
> > Is there some reason to make a graph out of the finite subsets, when they more naturally form a poset, and hence a category?
> 
> Yes: (co)limits, as per our definition, are relative to a graph morphism into the category. (One might argue that that's not a good design choice, but that's how it is right now.)

Okay.

lgtm

peterlefanulumsdaine:
lgtm — thanks @m-lindgren , this is nice to have!

On a mathematical note, though, I think this actually isn’t quite the “right” result to have — I’ll write more at #139.

PR #1549, opened 2022-09-13
WIP: towards bicategorical scenario for actions

rmatthes:
tries to build on the code that was there for technical purposes of proof management only
the code is unduly scattered for the moment

the bicategorical scenario will be compared with an elementary definition of actegories

PR #1569, opened 2022-10-14
Kw omegacontinuousfunctors

Kfwullaert:
Fixed definition of omega-continuous functor, added properties of those functors. Most of these properties were proven using the already existing code on omega-cocontinuous functors. 
Added a folder about opposite categories (in opp_precat, both the opposite category as the opposite functor are defined w.r.t. precategories). Added some general lemmas that I needed on opposite categories, e.g. opposite of a functor category is again a functor category and in particular showed that the opposite of the precomposition functor is (up to equivalence of categories) natural isomorphic to the precomposition functor of the opposite functor , an adjunction induces an adjunction between the opposite categories, limits as colimits in the opposite category.


rmatthes:
@Kfwullaert Since also PR #1549 fixes the def. of omega-continuous functor (to accommodate your other development there), what would you think is the best order for merging the two PR so as to minimize merge conflicts?

Kfwullaert:
> @Kfwullaert Since also PR #1549 fixes the def. of omega-continuous functor (to accommodate your other development there), what would you think is the best order for merging the two PR so as to minimize merge conflicts?

The definition is exactly the same as the one given in PR #1549, I've just copy pasted it. I think merging this PR first would not give any issues.

Kfwullaert:
> @Kfwullaert Good to know. But in the list of the files of package CategoryTheory, `LimitsAsColimits.v` comes too early, which makes the sanity check fail.

Oké thanks, I'll fix it!

Kfwullaert:
I'm not sure why the the sanity checks now fail: the error is the following: make: 
*** [Makefile:435: check-for-changes-to-CONTENTS.md] Error 1
The category theory folder compiles perfectly locally.

nmvdw:
You should push the `CONTENTS.md` file as well.

Kfwullaert:
> You should push the `CONTENTS.md` file as well.

Should I change anything in here manually?

nmvdw:
> > You should push the `CONTENTS.md` file as well.
> 
> Should I change anything in here manually?

When you do `make` in the terminal, the `CONTENTS.md` file is updated automatically. That is the one you should push.

UniMath/CategoryTheory/OppositeCategory/OppositeAdjunction.v
```diff
@@ -0,0 +1,71 @@
...
+Local Definition opp_cat (C : category)
+  : category.
+Proof.
+  exists (opp_precat C).
+  apply has_homsets_op.
+Defined.
+
+(* functor_opp is defined as a functor between precategories,
+   so in order to avoid always explicit casting, the following is introduced: *)
+Local Definition functor_op {C D : category} (F : functor C D)
```
rmatthes, line 20: is also defined in Core.v in this directory

UniMath/CategoryTheory/OppositeCategory/LimitsAsColimits.v
```diff
@@ -0,0 +1,172 @@
...
+  Definition diagram_op
+             {C : category}
+             {g : graph} (d : diagram g C)
+    : diagram (graph_op g) (opp_cat C).
+  Proof.
+    exists (λ v, pr1 d v).
+    exact (λ v1 v2 e, pr2 d v2 v1 (edge_op e)).
+  Defined.
+
+  Definition cone_op
```
rmatthes, line 35: is this file meant as an addition to or a replacement of module `co` in `CategoryTheory/limits/graphs/limits.v`?

Kfwullaert: @rmatthes I've looked at CategoryTheory/limits/graphs/limits.v, however they define the notion of a limit as a colimit. So if I understand the code correctly, in order to use that definition, one also has to adapt e.g. the definition of continuous functors using the definition of limits given in there in order to be able to say that if something is cocontinuous, then the opposite is also continuous.
What I've done is shown/gave constructions that given a colimit, you get a limit in the opposite category. So the two files are differently, because I use the explicit definition of a limit instead of a "work around".

UniMath/CategoryTheory/Chains/OmegaContFunctors.v
```diff
@@ -0,0 +1,408 @@
+(** * ω-cocontinuous functors
```
rmatthes, line 1: a "co" is remaining in this comment

rmatthes (commented):
Looks good to me - a quite complete library, obtained with limited effort through dualization of the results, not the concepts. Only very minor points and the sanity check remain.

PR #1579, opened 2022-10-25
fix issue 1577

### LINKED ISSUE #1577: why do we need .glob files for installing UniMath?
I ran by accident into the following observation: if the .vo files are present but not all .glob files, then make install fails but make sees nothing to be done. I would install UniMath so that I can compile the satellite repositories. I do not see in what sense I could profit from the .glob files for that compilation. Maybe I overlook some subtlety (the .glob files appeared already in [issue #341](https://github.com/UniMath/UniMath/issues/341)). Maybe one could have an installation target that did not care about the .glob files. BTW, these files weigh about 75MB now.

DanGrayson:
resolves #1577

We add the build target "html" when building "all" so the *.glob files will be built. Without that, the install step will fail.
See the issue at https://github.com/UniMath/UniMath/issues/1577. It's a little bit sad that we have to make the html files always, but the dependencies in Coq's makefile, build/CoqMakefile.make, are not right: the install target doesn't depend on the glob files, even though it needs them.


DanGrayson:
Yes, I think it's building everything twice.  I'll track it down.

PR #1583, opened 2022-10-27
ensures focus in tactic applications for all that depends on bifunctors

rmatthes:
exception: DisplayedCartesianMonoidalCategoriesWhiskered.v works with shelving and so there are commands to deactivate that checking for two subsequent lemmas


nmvdw:
This PR causes problems at Type Theory, because one file uses `CategoryTheory.All`

It might be better to do CategoryTheory later

rmatthes:
@nmvdw You were way too modest: I see 19 files in TypeTheory that import `CategoryTheory.All`. To my mind, it is rather bad style to import approximately 270 files into developments of only a few hundred lines.

rmatthes:
I only kept the changes that make the proofs respect the guideline, and the code for checking it is now commented, so this PR cannot possibly have adverse effects.

PR #1586, opened 2022-10-27
Filtered categories and Kuratowski finiteness

m-lindgren:
This PR includes among other things

- Kuratowski finiteness (as types admitting a surjection from some standard finite set [n])
- Filtered categories (two equivalent definitions are given)
- Explicit construction of filtered colimits in SET as a set quotient


m-lindgren:
I don't know why SetHITs did not build. Something about group axioms, but nothing I touched has anything to do with groups.

nmvdw:
Rerunning, now it should work.

nmvdw (approved):

Issue #1522, opened 2022-06-30
Modernize build system?


benediktahrens:
In a discussion during TYPES, @ppedrot suggested we could/should change our build system to allow the building in "stages". I have not understood what this means, and what we would need to do precisely in order to implement this change.

The purpose of this issue is to continue the conversation and gather feedback from everyone involved. We would be grateful for any pointers and advice to improve our build system and to relieve the Coq dev team of some of the burden of checking UniMath.

@ppedrot mentioned @ejgallego in the discussion, so I am tagging them here as well.

ejgallego:
Dune's Coq support could indeed help you, several options are available in terms of splitting Coq developments: theories and packages.

Building a theory can be just done with `dune build $theory_path`.

Docs are here, let me know folks if you have any question: https://dune.readthedocs.io/en/stable/coq.html

DanGrayson:
> Dune's Coq support could indeed help you, ...

Help us do what?  Does something need to be done?

ejgallego:
Apart from all the features Dune provides as built-in (caching, "sounder" builds, separate object dir, easier checking vs different Coq versions, finer verbosity control) Dune could help you with:
- you can organize theories in a more fine-grained way, using `(coq.theory )`  and the `(theories ...)` field, similarly to the way OCaml libraries are organized.
- you can build subset of targets easily too, or define your own aliases to do so.

arnoudvanderleer:
@benediktahrens Is this still relevant?

benediktahrens:
@arnoudvanderleer : thanks for bringing this back to my attention. I think this has been addressed - closing.

PR #1588, opened 2022-11-01
Build UniMath with Dune

m-lindgren:
This enables building UniMath with dune, as discussed in #1522 . This is a simple setup where the entire code base is treated as one project/package. 

I would like to point out that I am no expert on dune at all, and this PR is only a suggestion and example on what a configuration could look like, and something to experiment with and build on.

I wrote a bit on Zulip about my experiences with dune and UniMath and I'll try to elaborate a bit here. Hopefully I can answer some of the questions of #1522 from the perspective of someone who has used UniMath. So far my experience with dune has been great, most things just works out of the box (notice the very short configuration in UniMath/dune). The biggest selling points for me right now is that it is smarter about when files need to be rebuilt and when they don't. Together with the ability to cache compiled files this addresses one of the more painful parts of working with unimath: long rebuild times. With this there's no more rebuilding simply because you switched git branch and no more triggering a 4h rebuild because you accidentally hit 'save' when you were looking at something in Foundations/ (....this happens to me a lot). This really makes life a lot easier  working in the UniMath code base.

After this PR you can: 
- build the entire code base with ```dune build```
- build specific parts (with dependencies): ```dune build UniMath/Algebra```
- build HTML documentation with ```dune build @doc``` and tex ```dune build @doc-latex```

There's room to do much more by extending this simple configuration. One can easily organize parts of the library into separate packages by writing .opam- and dune-files. This could make it easier to publish more fine-grained UniMath libraries to opam and other package systems, among other things.

I'm sure there are benefits with using this in CI as well but I have not looked into it. With the compositional nature of dune and its cache functionality i find it likely that there's some benefit over the current CI workflows at least.

An example of what I mean by dune being compositional: I have a clone of UniMath where I put all the satellite repos in the root and I can build them all in one go (or individually). Any changes I make to UniMath then trigger rebuilds only of those parts of the satellites that needs to be rebuilt, as if they were all part of one project. Setting this up was only a matter of dropping one dune-file in each of the satellites (with the appropriate modifications to name and flags of course), and then dune simply builds all of it. (.... though it takes a while)

### Usage
To use this you need the latest version of dune which is 3.5.0. I recommend using opam for installing both dune and coq. (I simply recommend opam overall. It is the only sane package manager out there.)

**Caveats**
- Your working tree needs to be fairly clean before you build. Run a ```make clean``` before you try to do ```dune build``` the first time, and if you still get strange errors it might be because there's still stuff in the tree that dune doesn't know how to deal with. 
- Dune by default does not give a lot of feedback when it's compiling. I recommend you do ```dune build --display=short```, or put ```(display short)``` in your ```~/.config/dune/config```.
- Dune's cache is disabled by default. To enable it add the flag ```--cache=enabled``` to ```dune build```, or ```(cache enabled)``` to your dune config.
- Dune stores the compiled files in ```_build/```, so you need to give coqtop a flag like 
```
-R /your/path/to/UniMath/_build/default/UniMath UniMath
```
   There's a nice way to call coqtop with all the correct flags for any Coq project that is built by dune, by calling ```dune coq top``` instead of coqtop, however PG currently does not support it. (It is possible to get it to work with a workaround and I'd be happy to explain how for anyone interested -- it really isn't difficult -- but this PR message is long enough as it is).


m-lindgren:
Oh. Right. I forgot, this adds Foundations/All.v and MoreFoundations/All.v to git because coqdep needs them (they are the only .All -files actually Require Import:ed in the library I think)

There's probably a way to have dune automatically generate them, but I don't know how right now.

benediktahrens:
@m-lindgren : thanks for preparing this PR!

How about adding a file, e.g., `MAKE_DUNE.md`, that mentions the most important explanations you are giving above? This file could be extended over time, together with the dune functionality.

rmatthes:
On the `All.v` files: in the [discussion of PR 1583](https://github.com/UniMath/UniMath/pull/1583#issuecomment-1295920764), uses of the corresponding file of package `CategoryTheory` in the satellite `TypeTheory` were mentioned. Are those uses unproblematic for the dune setting that integrates the satellites?

In UniMath, I see more uses of import of `All`:
- UniMath/Bicategories/Core/Strictness.v:Require Import UniMath.CategoryTheory.categories.HSET.All.
- UniMath/Bicategories/DisplayedBicats/Examples/CwF.v:Require Import UniMath.CategoryTheory.categories.HSET.All.
- and more of those
- UniMath/CategoryTheory/Inductives/LambdaCalculus.v:Require Import UniMath.CategoryTheory.Chains.All.
- UniMath/CategoryTheory/Inductives/Lists.v:Require Import UniMath.CategoryTheory.Chains.All.
- UniMath/CategoryTheory/Inductives/Trees.v:Require Import UniMath.CategoryTheory.Chains.All.
- and more of those
Such uses seem to correspond to the initial purpose of having those convenience files, but those `All.v` are **not** generated (with all the expected problems: in `Chains` and `HSET`, some newer files are missing from `All.v`).

Should the style guide for UniMath not anyway disallow the other (all the generated ones after merging the present PR) `All.v` files in developments (and leave them for testing purposes) ? But it would probably be preferable if more generated `All.v` could coexist with the installation scripts so that a moderately fine-grained import through `All` was possible (the file-by-file import is maximally fine-grained but verbose and can even hide information - that the imports were just copied from somewhere and are not all needed).

m-lindgren:
> uses of the corresponding file of package CategoryTheory in the satellite TypeTheory were mentioned. Are those uses unproblematic for the dune setting that integrates the satellites?

Ah, good catch, yes they are problematic! Indeed those parts of TypeTheory that imports CategoryTheory/All.v fails, I missed that dune reported this error. The other cases you mentioned (Chains/HSET) do not cause a problem because those files are part of the git-repo, so dune finds them when checking dependencies. 

I also found other things that doesn't compile like 'lBsystems_work.v' in TT and a lot of files in largecatmodules that import UniMath.CategoryTheory.Categories. I assume this is material that are intentionally left out of the Makefiles, but dune tries to compile as much as possible.

> But it would probably be preferable if more generated All.v could coexist with the installation scripts so that a moderately fine-grained import through All was possible

I agree, maybe it would be good to have all 'All.v' generated for consistency, but have at least some of them added to the repo? Or are there any drawbacks to having all such files in the repository? Besides that one might forget to commit updates, as with CONTENTS.md. 

I wonder if those Import CategoryTheory.All in TypeTheory could be replaced by CategoryTheory.Core.All ... It's possible that when CT was refactorized to move the essentials into Core/ that TypeTheory still imported the old files.

.... nevermind I just now realized we don't have a CategoryTheory/Core/All.v!

rmatthes:
@m-lindgren What we do not have can be added if a positive consensus is obtained. 

Importing all of the CategoryTheory package should not appear in developments with an official support by the UniMath organization. This is my view, but I hope others will agree.

Issue #1575, opened 2022-10-21
Improve CI for satellite repos


benediktahrens:
- put compilation of each satellite repo in a separate "step"
- continue CI upon failure of a step (see https://stackoverflow.com/questions/58858429/how-to-run-a-github-actions-step-even-if-the-previous-step-fails-while-still-f)

rmatthes:
I looked into the [YAML file](https://github.com/UniMath/UniMath/blob/master/.github/workflows/build-libs.yml). It does not mention GrpdHIT in the step description (trivial to add). Why are the satellites not compiled with `make -j2`? Way over 3h running time look extreme. And there are 16kloc of output, mostly with warnings. Is there a way to avoid those warnings? Maybe, at least to assess the time consumption, the make commands could be timed with `\time -v`?

benediktahrens:
- Instead of editing the step description, it would be better to have separate steps for each compilation.
- `make`ing with several threads would be good, as long as the error messages are still readable (but I think that they would be), that is, that output of different threads is not mixed too much
- I don't know how to suppress the warnings, I am afraid.
- `time` is a great idea!

m-lindgren:
I've been experimenting with using dune cache in CI, and I think UniMath can benefit from it. I'll see If I can build the satellites with it as well.

m-lindgren:
In #1590 the satellites finish in 15-25 minutes, and in my own tests they finish in about 5min if their dependencies are not changed, i.e. if they have everything cached.

rmatthes:
Concerning the warnings I mentioned three days ago: compiling `GrpdHITs` according to the instructions produces an enormous number of warnings of the same type that can be suppressed with the simple line `-arg "-w -notation-overridden"` in file `_CoqProject`. Other warnings will still appear (and could be dealt with by easy code changes). I filed a PR.

benediktahrens:
@arnoudvanderleer : yes, this is obsolete, closing - thanks!

PR #1590, opened 2022-11-04
Use Dune cache in CI

m-lindgren:
This is a proof of concept for using dune in CI, related to #1575 but also enables cache for the main build. This PR is based on #1588, but makes additional changes to CI to use the dune cache functionality.

When satellites are not affected by changes they currently finish in roughly 8 minutes.

This disables the BUILD_COQ=yes builds, because they were to slow to test, but they can be enabled. They will however be the slowest to build by far since they do not use the cache (not sure how to make that happen at the moment).

Also note that the satellites in this patch do not pull from the main repositories but from my own, because I add a patch that enables dune building. This can be fixed by either adding those patches to the repositories, or having CI pull the main repos and then rebasing these patches before compiling.

Currently enabled satellites:
- SetHITs
- TypeTheory
- largecatmodules

This also schedules building of everything each monday, to keep the caches fresh.


rmatthes:
I have a question that could evolve into an issue. Since dune with caching reduces compilation times in CI so massively, it could be interesting to profit from the generated .vo files in our local installations as basis for further development. This would be a first step into CD (continuous delivery). But I think, instead of having full-blown CD, we could do just with the creation of a tar/zip file of all those .vo files as an artifact. Is the current GitHub scheme of UniMath generous enough to allow massive storage of those archives of .vo files? I already do not know if the caching will have to be paid for (on the long run).

m-lindgren:
> Since dune with caching reduces compilation times in CI so massively, it could be interesting to profit from the generated .vo files in our local installations as basis for further development.

I think this is orthogonal to this pr, but think it is interesting. Are you saying that even if one use dune cache locally a build after a pull can still be very slow. For example if the pull includes changes to `Foundations` or `CategoryTheory/Core/Categories.v` a lot of files will need to be rebuilt. In this case having access to the output of githubs build could help. Is this the use case you had in mind? I like your idea about producing artifacts to solve this, and would be happy to discuss it further in a separate issue. I think your idea using artifacts is doable (and not very difficult to implement) if the CI is using dune.

m-lindgren:
In the documentation on [speeding up workflows using cache](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows) GitHub writes the following:

> GitHub will remove any cache entries that have not been accessed in over 7 days. There is no limit on the number of caches you can store, but the total size of all caches in a repository is limited to 10 GB.

This is the reason for adding a scheduled rebuild once a week, which should take less than 10 mins unless the cache for some reason has become invalid.

>If you exceed the limit, GitHub will save the new cache but will begin evicting caches until the total size is less than the repository limit.

You can find the currently used cache files in [Actions -> Caches](https://github.com/UniMath/UniMath/actions/caches). Currently UniMath use less than 2GB, and I don't think it should increase too much, but we should monitor it and see if things needs to be tweaked. 

At the moment I'm trying to fix the problems with TypeTheory/CI (#1591) so that UniMath/TypeTheory#226 can be merged first, because the Dune changes in this PR changes CategoryTheory/All.v (to actually include everything) and that breaks some things in TypeTheory.

benediktahrens:
This is a great PR, thanks a lot!
I think the following changes should be made before merging:
- [ ]  Switch to the satellite repositories in the UniMath organization
- [ ] Maybe retain one of the long builds, as a check of the old `Makefile`?

m-lindgren:
Yes, it kind of makes sense to also build the old version, as long as building that way is something that is recommended in INSTALL.md etc. 

One way of avoiding this would be to have the Makefile call `dune build` behind the scenes. That way the build instructions to users would still be "run `make`", but that could be the subject of another issue.

m-lindgren:
Another idea, not sure if it's good, would be to remove the slow build from the `on pull_request` build and have it in a scheduled job that runs once a week or something. One risks losing the knowledge about which PR actually broke the build, and would have to fix it afterwards. But on the other hand if it broke the slow build it probably involved messing with the makefiles or something, in which case a reviewer could trigger the job manually.

benediktahrens:
> Another idea, not sure if it's good, would be to remove the slow build from the `on pull_request` build and have it in a scheduled job that runs once a week or something.

I think that is a good idea!

m-lindgren:
> Maybe this can be changed in a separate PR?

Absolutely! I have a patch ready that moves the slow build from the `on: pull_request` workflow to a weekly, but an advantage of having it in a separate PR would be that it gives others the opportunity to provide feedback on this topic specifically.

benediktahrens:
If nobody speaks out against merging this PR by today 23:59 GMT I am going to merge it.

rmatthes:
If have a question about how to control the number of threads Dune uses. Currently, it spawns 12 compilation threads on my machine with 6 cores (and the useless feature hyperthreading - useless for the compilation of UniMath as of before the move to Dune at least). This makes the system unresponsive. Is there anything in the current setting that suggests to Dune to take the number of virtual cores? Or this is just default behaviour? If yes, how to do less?

m-lindgren:
> Is there anything in the current setting that suggests to Dune to take the number of virtual cores? Or this is just default behaviour? If yes, how to do less?

Dune tries to choose a setting based on your system. My guess is your system gets unresponsive because it runs out of RAM. You can specify the maximum number of jobs dune should run in parallel with `dune build -j 8` for example.

rmatthes:
@m-lindgren I have lots of RAM. I'll try your hint. So, there is a blank between j and 8, unlike when starting a makefile.

Another question: what should the users of Proof General do for the moment? I copied the file `.dir-locals.el.in` at the root of the UniMath code into `.dir-locals.el`, deleted the line starting with `@LOCAL` and added the extra option to `coqtop` just by way of three further strings in variable `coq-prog-args`: the `"-R"`, the hard-coded path according to the instructions in BUILD_DUNE.md and the final `"UniMath"`, again according to those instructions.

m-lindgren:
@rmatthes So what I ended up using is something slightly different for Proof General. The command `dune coq top` would in theory set up everything for us (no need for dirlocals or _CoqProject), but currently PG does not [support it.](https://github.com/ProofGeneral/PG/issues/477)

The very hacky solution that I use is based on [this comment](https://github.com/ocaml/dune/pull/5457#issuecomment-1050683785), and specifically I did this:
- Create an executable shell script somewhere in your path with something like
```
#!/bin/sh

opam exec -- dune coq top "$2" --display=short -- -emacs -w -notation-overridden -noinit -quiet -indices-matter -type-in-type
```
- Set your `coq-prog-name` to that shell script.

With this setup you don't have to care about `-R`  at all (because dune takes care of that), all you have to do is set the other flags in your script. It's unfortunate that you have to hardcode all the flags currently, the idea with `dune coq top` is that it should take care of both the `-R` and the other flags, but there's this [bug](https://github.com/ocaml/dune/issues/6366) (a fix is available but not yet merged I believe)

rmatthes:
How am I meant to work with the satellites in Proof General? In the past, I installed the .vo files of UniMath into the Coq installation, and then I did not have to care about UniMath in the satellites.

m-lindgren:
> Is this covered by the shell script solution?

Yes! My setup is the following: In the UniMath folder I have a opam switch linked with coq 8.15.2 (this is important because TypeTheory breaks with 8.16.0 (if you really want to use 8.16 as much as possible you can install a 8.15-switch in TT and have an 8.16 in the base repository, dune will then use 8.16 for everything except TT)) installed and dune 3.5.0, and in the same folder I have cloned the satellites. So it looks something like this: 
```
UniMath$ tree
.
├── SetHITs
├── TypeTheory
└── UniMath
```
After building it should just work. Dune treats it as one big project essentially. The cache is system-wide.

rmatthes:
I am sorry, but the script above does not work for me in UniMath if I I start Emacs at the root of the Git repo with the name of the file. Error messages when trying to step through a file in CategoryTheory/Core are:
``Error: cannot find file: UniMath/CategoryTheory/Core
Hint: is the file part of a stanza?
Hint: has the file been written to disk?``

PR #1595, opened 2022-11-08
Cartesian and cocartesian objects in bicategories

nmvdw:
In this PR, I define cartesian and cocartesian objects internal to arbitrary bicategories. I give two definitions and prove that they are equivalent.

I also added some examples of cartesian and cocartesian objects.


m-lindgren:
I see "8 successful and 2 expected checks" and the PR is marked yellow instead of green. I suspect some job, probably 'sanity-checks', was marked as required before, and I renamed it in my latest commit. It's probably something like that that's blocking it.

benediktahrens:
> I see "8 successful and 2 expected checks" and the PR is marked yellow instead of green. I suspect some job, probably 'sanity-checks', was marked as required before, and I renamed it in my latest commit. It's probably something like that that's blocking it.

Yes, renaming the checks means that the repository has be reconfigured. I'll take care of this right now - thanks for pointing out the problem.

PR #1598, opened 2022-11-17
Monoidal rezk completion

Kfwullaert:
In this formalization, it is shown that monoidal categories have a Rezk completion in a suitable way.
More precisely: if a monoidal category is weakly equivalent to a univalent category, then one can construct a monoidal structure on that univalent category such that the weak equivalence has the structure of a strong monoidal functor. 
Furthermore, the universal property of the Rezk completion for categories extends to lax (and strong) monoidal functors.


DanGrayson:
We have a package called "Monoidal"?  That's an adjective.  Maybe we should rename it to a suitable noun.

nmvdw:
In Bicategories, there is also a directory with results about monoidal categories, and that one is called MonoidalCategories. Perhaps MonoidalCategories is a better name for that directory as well

PR #1599, opened 2022-11-17
CI: Remove the BUILD_COQ=yes job and use docker-coq.

m-lindgren:
This pull request removes the slow `make BUILD_COQ=yes` job from the workflow that's triggered by pull requests. Instead this is moved to a scheduled workflow that runs every two weeks, on the 1st and 15th of every month. In #1590 we discussed having this job run once a week, but I suggest a less frequent build since what this job really is testing is if the submodule syncs alright and the Makefile is still working. 

This PR also reorganizes and changes the CI to use docker-coq, and move the satellites into the same workflow as the rest of UniMath. Nothing is really changed as to _what_ happens when a PR is made, but there are some important changes to how jobs perform their tasks. So when a PR is made the following happens (in parallel):
- sanity-checks is run
- UniMath is built on Linux (NEW: Uses docker-coq, will always use the latest stable release of Coq, and with this change we can easily test multiple versions of Coq at the same time. Everything is still cached so the builds will be incremental)
- UniMath is built on MacOS (using 8.16.0 as before, but caching is a bit improved)
- The satellites are built using Coq 8.16.x (except TypeTheory, which is built with 8.15). (NEW: Uses docker-coq) 

On the surface not a lot changes, but the change to using docker-coq will make things easier to maintain in the long run.  The [docker-coq](https://github.com/coq-community/docker-coq/wiki) images are built directly from the Coq GitLab. They should provide a consistent build environment, and enables UniMath to be up to date with the latest stable release of Coq, and ensure changes added to UniMath doesn't break Coq-CI. 

And also:
- The main UniMath linux build is now in a matrix, so if we decide we want to test each PR against several different versions of coq all one needs to do is change `coq-version`. Currently it is set to `[latest]` but when `8.17` is released we might want to set it to something like `[latest, 8.16, 8.15]`.
- The satellites are now also run in a matrix, this simplifies adding / removing new satellites to the build.
- The satellite workflow is now included in the main workflow (but it is still a separate job). This way contributors only need to look at one workflow when they make a PR.


benediktahrens:
@m-lindgren : thanks a lot for all your work on CI - this is extremely valuable!

I will adapt the branch protection settings after merging this PR (a name mismatch is responsible for the "1 expected checks").

PR #1601, opened 2022-11-30
Update CI

m-lindgren:
This PR addresses one major issue and two minor issues that I've observed over the past weeks.

- First of all this PR adds [concurrency](https://docs.github.com/en/actions/using-jobs/using-concurrency) management and use the `cancel-in-progress` feature of Github Actions. This should have the effect that jobs in progress on a  branch are cancelled when new commits are added to the same branch before starting a new workflow. So for example if you need to add a commit to a PR with a workflow running, pushing would previously start a new CI-workflow in parallel with the first one. With this change the first workflow is cancelled before a new CI-workflow is started.
- ~~Improve caching / fix cache glitch on the MacOS build.~~ (Will submit bug report to ocaml/setup-ocaml instead)
- Remove `-verbose` from the set of arguments to Coq. When using `dune coq top` this flag is passed on to `coqtop` which does not recognize it and fails. We remove it and instead add the `:standard` flags.
- Use Coq 8.16.1 on macOS. Since the changes above will cause a full rebuild of everything we might as well change the macOS build to use the latest version of Coq.


benediktahrens:
@m-lindgren : thanks a lot for this PR. 

I have a question regarding point 2: pinning a version means that someone should update the version regularly. So it seems to me that there is a trade-off between (a) more efficient caching and (b) less manual configuring of the CI. How often will ocaml/ocaml-setup need to be moved to a new version? What are the implications of staying with an older version?

m-lindgren:
> How often will ocaml/ocaml-setup need to be moved to a new version? What are the implications of staying with an older version?

So my original reasoning was that it would never have to change if it works now, unless GH makes breaking changes. However I realize there is a thing that potentially can cause problems down the line, and that is if opam makes changes that requires updates to ocaml/setup-ocaml (I don't think the script pins a version of opam). 

Since this is basically a hack to work around an issue that should be fixed upstream and not in UniMath I think the best solution is if I remove the change in (2) and create a bug report with ocaml/setup-ocaml instead. --- I'll do so and push a commit undoing the change in (2) once I see that there are no other issues with this CI-run.

benediktahrens (approved):

PR #1603, opened 2022-12-02
Housekeeping PR october

m-lindgren:
Some minor housekeeping commits I collected in October that I forgot to create a PR for.

Mostly removing duplicate definitions and upstreaming some useful definitions. Also removes the scope [cat_deprecated]. This will cause TypeTheory to fail until [232](https://github.com/UniMath/TypeTheory/pull/232) is merged.

### DIFF
--- a/UniMath/MoreFoundations/Sets.v
+++ b/UniMath/MoreFoundations/Sets.v
@@ -110,10 +110,6 @@ Proof.
   exact (setquotuniv4prop R (λ x1 x2 x3 x4, make_hProp (P x1 x2 x3 x4) (H x1 x2 x3 x4)) ps).
 Defined.
 
-Definition setcoprod (X Y : hSet) : hSet :=
-  make_hSet (X ⨿ Y) (isasetcoprod X Y (pr2 X) (pr2 Y)).
-
-
 (** ** The equivalence relation of being in the same fiber *)
 
 Definition same_fiber_eqrel {X Y : hSet} (f : X → Y) : eqrel X.
--- a/UniMath/MoreFoundations/Sets.v
+++ b/UniMath/MoreFoundations/Sets.v
@@ -129,12 +129,6 @@ Defined.
 
 (** ** Subsets *)
 
-Definition subset {X : hSet} (Hsub : hsubtype X) : hSet :=
-  make_hSet (carrier Hsub) (isaset_carrier_subset _ Hsub).
-
-Definition makeSubset {X : hSet} {Hsub : hsubtype X} (x : X) (Hx : Hsub x) : subset Hsub :=
-  x,, Hx.
-
 Definition pi0 (X : UU) : hSet := setquotinset (pathseqrel X).
 
 Section Pi0.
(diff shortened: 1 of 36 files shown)


benediktahrens (approved):

UniMath/MoreFoundations/Sets.v
```diff
@@ -110,10 +110,6 @@ Proof.
   exact (setquotuniv4prop R (λ x1 x2 x3 x4, make_hProp (P x1 x2 x3 x4) (H x1 x2 x3 x4)) ps).
 Defined.
 
-Definition setcoprod (X Y : hSet) : hSet :=
```
DanGrayson, line 113: This looks potentially useful.  What's the point in removing it?

UniMath/MoreFoundations/Sets.v
```diff
@@ -129,12 +125,6 @@ Defined.
 
 (** ** Subsets *)
 
-Definition subset {X : hSet} (Hsub : hsubtype X) : hSet :=
-  make_hSet (carrier Hsub) (isaset_carrier_subset _ Hsub).
-
```
DanGrayson, line 134: These also look useful.

UniMath/MoreFoundations/Subtypes.v
```diff
@@ -119,9 +119,6 @@ Definition subtype_binaryunion_leq1 {X} (A B : hsubtype X) : A ⊆ (A ∪ B)
 Definition subtype_binaryunion_leq2 {X} (A B : hsubtype X) : B ⊆ (A ∪ B)
   := fun x => hdisj_in2.
 
-Definition carrier_set {X : hSet} (S : hsubtype X) : hSet :=
-  make_hSet (carrier S) (isaset_carrier_subset _ S).
-
```
DanGrayson, line 124: This one, too.

UniMath/MoreFoundations/Sets.v
nmvdw: https://github.com/UniMath/UniMath/blob/master/UniMath/Foundations/Sets.v#L105
https://github.com/UniMath/UniMath/blob/master/UniMath/Foundations/Sets.v#L292

The removed definitions already appear in Foundations.

UniMath/MoreFoundations/Subtypes.v
m-lindgren: This and  [subset] are duplicates of [carrier_subset], defined in Foundations/Sets.v

UniMath/MoreFoundations/Sets.v
DanGrayson: Oh, thanks!

DanGrayson:
lgtm

But that failing check with "TypeTheory" shows that "TypeTheory" needs to be updated.

DanGrayson:
Okay, I've restarted the failed job.  But why is that last job (Ubuntu) taking so long?

m-lindgren:
All the jobs are completed, it's just waiting on a status report for a job marked as required, but that job has been renamed in an earlier PR. It should be set to `build-Unimath-ubuntu` if I'm not mistaken.

PR #1604, opened 2022-12-05
Added isomorphism of (monoidal) functor categories induced by the monoidal rezk completion

Kfwullaert:


UniMath/Bicategories/MonoidalCategories/MonoidalRezkCompletion.v
```diff
@@ -0,0 +1,289 @@
...
+          (Euniv : is_univalent E)
+          (H_eso : essentially_surjective H)
+          (H_ff : fully_faithful H)
+          (TC : functor (C ⊠ C) C)
+          (TE : functor (E ⊠ E) E).
+
+
+  Definition precomp_tensor_catiso
+    : catiso (total_category (functor_tensor_disp_cat (TransportedTensor Duniv H_eso H_ff TC) TE))
+             (total_category (functor_tensor_disp_cat TC TE)).
```
m-lindgren, line 62: This and most other definitions in this file doesn't seem specific to Bicategories, but looks like they should apply to monoidal categories in general, or am I missing something?

m-lindgren (commented):
Looks good, but If I'm not mistaken maybe most of `MonoidalRezkCompletion.v` added in this PR belongs in `CategoryTheory/` somewhere.

UniMath/Bicategories/MonoidalCategories/MonoidalRezkCompletion.v
Kfwullaert: You are right, it is eventually the purpose to put it in CategoryTheory/Monoidal.
However, the construction that says that an adjoint equivalence of univalent categories is automatically an isomorphism of categories is currently only found in the bicategories package, not in the category package. Therefore, this file is now in the bicategory package.

benediktahrens:
@m-lindgren : Thanks a lot for your feedback. I have opened issue #1605 to continue the discussion there.

UniMath/Bicategories/MonoidalCategories/MonoidalRezkCompletion.v
m-lindgren: Ah, I see what you mean, the stuff on adjunctions of categories in the `Bicategories` library should be upstreamed. However, the only proof in this file that rely on these definitions can actually be rewritten to use only `CategoryTheory` definitions like this I think:

```diff
diff --git a/UniMath/Bicategories/MonoidalCategories/MonoidalRezkCompletion.v b/UniMath/Bicategories/MonoidalCategories/MonoidalRezkCompletion.v
index 845534e94..a4db30771 100644
--- a/UniMath/Bicategories/MonoidalCategories/MonoidalRezkCompletion.v
+++ b/UniMath/Bicategories/MonoidalCategories/MonoidalRezkCompletion.v
@@ -12,2 +12,3 @@ Require Import UniMath.CategoryTheory.Adjunctions.Core.
 Require Import UniMath.CategoryTheory.Equivalences.Core.
+Require Import UniMath.CategoryTheory.Equivalences.FullyFaithful.
 
@@ -29,5 +30,2 @@ Require Import UniMath.CategoryTheory.Monoidal.RezkCompletion.MonoidalRezkComple
 
-Require Import UniMath.Bicategories.Core.EquivToAdjequiv.
-Require Import UniMath.Bicategories.Core.Examples.BicatOfUnivCats.
-
 Local Open Scope cat.
@@ -40,8 +38,6 @@ Definition adj_equivalence_of_cats_to_cat_iso
 Proof.
-  use (invweq (cat_iso_to_adjequiv (_,,Cuniv) (_,,Duniv))).
-  use equiv_to_adjequiv.
-  2: {
-    apply adj_equivalence_to_left_equivalence.
-    exact Fa.
-  }
+  exists F.
+  split.
+  - apply fully_faithful_from_equivalence, Fa.
+  - apply(weq_on_objects_from_adj_equiv_of_cats C D Cuniv Duniv F Fa).
 Defined.
```

PR #1606, opened 2022-12-06
Displayed universal arrows

Kfwullaert:
In this PR, the notion of a "displayed universal arrow" (of a displayed pseudo-functor) is introduced.
In order to show how such a displayed universal arrow defines a universal arrow for the corresponding total pseudo-functor, it is shown how (the previously defined notion of) displayed adjunctions (resp. adjoint equivalences) induce an adjunction (resp. adjoint equivalence)  between the total categories.


UniMath/Bicategories/PseudoFunctors/DisplayedUniversalArrow.v
```diff
@@ -0,0 +1,183 @@
...
+      intros [x xx].
+      exact (η x ,, ηη _ xx).
+    }
+    intros [x xx] [y yy].
+
+    use nat_iso_adj_equivalence_of_cats.
+    { exact (total_functor (disp_left_universal_arrow_functor xx yy)). }
+    { apply nat_trans_id. }
+    { use is_nat_z_iso_nat_trans_id. }
+    exact (total_adj_equivalence_of_cats (dLURu x xx y yy)).
```
m-lindgren, line 153: An aesthetic comment: Wouldn't proof structuring with `-` look better than focusing three goals with `{ ... }`?

UniMath/CategoryTheory/DisplayedCats/TotalAdjunction.v
```diff
@@ -0,0 +1,513 @@
...
+  Definition total_adjunction_forms_equivalence2
+    :  ∏ b : total_category D2, Isos.is_z_isomorphism (adjcounit (total_adjunction FF) b).
+  Proof.
+    intro x.
+    use is_z_iso_total.
+    - use Isos.is_z_iso_comp_of_is_z_isos.
+      { apply Isos.identity_is_z_iso. }
+      use Isos.is_z_iso_comp_of_is_z_isos.
+      2: { apply Isos.identity_is_z_iso. }
+      apply (pr2 F).
```
m-lindgren, line 428: Why not import `Isos`? 

Also, if you use `+ ... ` instead of `{...}` for proof structuring I think it would look a lot neater / more readable. Something like: 
```
    - apply is_z_iso_comp_of_is_z_isos.
      + apply identity_is_z_iso.
      + apply is_z_iso_comp_of_is_z_isos.
        * apply(pr2 F).
        * apply identity_is_z_iso.
```
(I'm sure there's some function to replace `pr2 F` with but I couldn't find one right now)

UniMath/CategoryTheory/DisplayedCats/TotalAdjunction.v
```diff
@@ -0,0 +1,513 @@
...
+  Definition total_adjunction_forms_equivalence1
+    : ∏ a : total_category D1, Isos.is_z_isomorphism (adjunit (total_adjunction FF) a).
+  Proof.
+    intro x.
+    use is_z_iso_total.
+    - use Isos.is_z_iso_comp_of_is_z_isos.
+      { apply Isos.identity_is_z_iso. }
+      use Isos.is_z_iso_comp_of_is_z_isos.
+      2: { apply Isos.identity_is_z_iso. }
+      apply (pr2 F).
```
m-lindgren, line 362: Proof structuring?

m-lindgren (commented):
Good material overall, I mostly have comments on some style issues.

UniMath/CategoryTheory/DisplayedCats/TotalAdjunction.v
Kfwullaert: > Proof structuring?

As from your previous comments, you prefer the use of +. 
I might have written {} too much, but as in this scenario, you quickly dive into a very "ugly aligned" code.
The code would look like this:
 - use Isos.is_z_iso_comp_of_is_z_isos.
     + apply Isos.identity_is_z_iso. 
      + use Isos.is_z_iso_comp_of_is_z_isos.
             * apply (pr2 F)
             * apply Isos.identity_is_z_iso. 
I personally find this "more ugly".

UniMath/CategoryTheory/DisplayedCats/TotalAdjunction.v
```diff
@@ -0,0 +1,513 @@
...
+          rewrite transport_f_f.
+          rewrite id_right_disp.
+          rewrite id_left_disp.
+          rewrite transport_b_b.
+          unfold transportb.
+          rewrite mor_disp_transportf_postwhisker.
+          rewrite transport_f_f.
+          use transportf_set.
+          apply homset_property.
+  Qed.
```
benediktahrens, line 485: Do not `Qed.` this.

Kfwullaert:
@m-lindgren Thanks for all your feedback!
Regarding make_left_universal_arrow.
There are currently already two such make statements, but they are not usable in this case.
As you said, they are easy to define, but I have not done it now, instead, I did it in the following form (1):
exists ..
exists ..
exact ..
The definition of left_universal_arrow is a type of the form \sum _, \sum _, _.
So the constructor is exactly the same as the form (1). If you think it would be better to have it, I'll include it and use that, but since I could just apply exists twice, I don't think it does anything about readability (correct me if you don't agree!).

m-lindgren:
Ah, you're right, using `exists` looks a lot better here, I'm happy with these changes.

PR #1607, opened 2022-12-07
Inserted improvement provided by m-lindgren

Kfwullaert:
In PR  #1604 , some lemmas about the rezk completion and isomorphism of categories have been added. 
But this was defined in the Bicategories package. 
@m-lindgren noticed that this was actually not necessary.
This PR uses @m-lindgren 's solution.


Kfwullaert:
@m-lindgren You have asked in PR #1604 about the multiple 4: { ... }
The reason I did this was because the lemma that is used, has 4 constructors.
However, constructor 3 is "derivable" from 4. 
In LiftedTensor and LiftedTensorUnit, I've changed this, but then we are forced to write:
1: ..
1: ..
2: ..
I personally prefer (as was done previously and is still in e.g. LiftedUnitors):
4: ...
- ...
- ...

m-lindgren:
Thank you! I see what you mean, it kind of forces you to either give all the arguments in the first `use` or do what you did. I looked at it a bit and I think if you change order of the arguments in the definition of  `adj_equivalence_of_cats_to_catiso` so that the `is_univalent ...` arguments comes after `(Fa : adj_equivalence_of_cats F)` I _think_ that you should be able to replace patterns like
```
use adj_equivalence_of_cats_to_catiso.
4: exact something.
1: ...
```
with 
```
apply(adj_equivalence_of_cats_to_catiso something).
- ...
- ...
```

m-lindgren (approved):
Thanks for fixing this! Looks good to me.

PR #1613, opened 2023-01-24
Towards Topos theory

CalosciMatteo:
This Pull Request contains the theorems from sections IV.1 and IV.2 of the book "Sheaves in Geometry and Logic" by Mac Lane & Moerdijk. The main files are UniMath/CategoryTheory/Subobjects.v, UniMath/CategoryTheory/SubobjectClassifier.v, UniMath/CategoryTheory/PowerObject.v and UniMath/CategoryTheory/elementary_topos.v but there are intermediate result and minor changes in other files.

The main result formalized are:
- The definition of [SubObj_Functor], a functor from C^op to hset_category which maps c to the set of the subobject (module z_isomorphism) of c and maps morphism "by pullback"
- The derivation of [SubobjectClassifier_nat_z_iso], the natural (z-)isomorphism from the functor above to Hom(-,Omega) induced by the subobject classifier Omega
- The definition of PowerObject as a map [P : ob C -> ob C] and a morphism from b x (P b) to the subobject classifier such that a certain proposition hold;
    - The derivation of [PowerObject_functor], a functor extending P;

    - The derivation of [PowerObject_nat_z_iso], the natural (in a and b) (z-)isomorphism from Hom(b x a , Omega) to Hom(a,P b);
    - The derivation of [PowerObject_charname_nat_z_iso], the natural (z-)isomorphism from Hom(- , Omega) to Hom(1,P(-)) obtained from the one above choosing a = T (T is the Terminal Object);
- The definition of [Topos] as a category with Terminal Object, Binary Pullbacks, Power Object and Subobject Classifier
    - The derivation [Exponentials_from_Topos] of Exponentials in a Topos

These results are respectively in:
- UniMath/CategoryTheory/Subobjects.v
- UniMath/CategoryTheory/SubobjectClassifier.v
- UniMath/CategoryTheory/PowerObject.v
- UniMath/CategoryTheory/elementary_topos.v

Some other result and minor changes in these files:
- Subobjects.v
    - I made the [C] argument in [Subobjectscategory] implicit ([Subobjectscategory] was only used in this and in grothendieckTopos.v) and I deleted the " C " where needed
    - I added the accessor for objects and morphisms of a [Subobjectscategory]
    - I added a constructor for [z_iso] in [Subobjectscategory]
    - I added [z_iso_from_z_iso_in_Subobjectcategory] an "accessor" to the underlying (z-)isomorphismo of a (z-)isomorphism in a [Subobjectscategory] (maybe it's possible to define this one as a composition of analogous results for [slice_cat] and [sub_precategories])
    - I changed [: HSET] to [:hSet] in the definition of [SubObj] in order to make compile term of the form "x:SubObj c"
    - I added an accessor for [SubObj] and proved [PullbackSubObj] which define a [SubObj] "by pullback" given another [SubObj] and a morphism
- SubobjectClassifier.v
    - I proved [subobject_classifier_balanced] which states that a mono and epi in a category with subobject classifier is a (z-)isomorphism

During development I also proved some lemmas and I made some minor changes which I explain below:

- in UniMath/CategoryTheory/categories/HSET/Core.v I added [cov_homSet_functor] and [contra_homSet_functor], these are the same of [cov_hom_functor] and [cov_hom_functor] defined in .../Type/Core.v but for categories, so the target is hset_category instead of type_precat.
I need these to define the functors Hom(-,Omega), Hom(-- x - , Omega), Hom(-,P(-)), etc...

- in UniMath/CategoryTheory/categories/Type/Core.v I added [contra_hom_functor], which was missing due to [functor_fix_snd_arg] being defined only for functor between categories (and not between precategories as [functor_fix_fst_arg])

- in UniMath/CategoryTheory/Core/NaturalTransformations.v I added the accessor [pr2_nat_z_iso]

- in UniMath/CategoryTheory/limits/binproducts.v
    - I proved some lemma about [BinProductOfArrows] involving identities
    - Defined [BinProduct_assoc] the (z-)isomoprhism from a x (b x c) to (a x b) x c and proved a lemma involving it and [BinProductOfArrows]
	
- in UniMath/CategoryTheory/limits/equalizers.v I added [z_iso_Equalizer_of_same_map], a lemma stating that the equalizer of two morphism connected by a path is a (z-)isomorphism

- in UniMath/CategoryTheory/limits/pullbacks.v
    - I added [pullback_glue_pullback] which is a reformulation of [isPullbackGluedSquare] with data already packaged in [Pullback]
    - I made implicit some of the arguments of [isPullbackGluedSquare]
    - I prooved [Pullback_of_z_iso] which states that the pullback of a (z-)iso is a (z-iso)
    - I added [Pullback_z_iso_of_morphisms] which is a reformulation of [isPullback_z_iso_of_morphisms] with data already packaged in [Pullback]
    - I added [isPullback_mor_paths'] which is a reformulation of [isPullback_mor_paths] not requesting the assumption [H2], which can be derived from the other
    - I added [Pullback_mor_paths] which is a reformulation of [isPullback_mor_paths] with data already packaged in [Pullback]

- in UniMath/CategoryTheory/GrothendieckTopos.v I deleted some " _ ", because I made implicit an argument of [Subobjectscategory] (which was only used implicitly anyway)

- in UniMath/CategoryTheory/Monics.v  I proved a lemma on (z-)isomorphism in  [subcategory_of_monics]

- in UniMath/CategoryTheory/opp_precat.v I proved a lemma which states that [op_nt] is a natural (z-isomorphis) if his argument is a natural (z-isomorphis) too

- in UniMath/CategoryTheory/PrecategoryBinProduct.v
    - I defined [nat_z_iso_fix_fst_arg] analogously to [nat_trans_fix_fst_arg]
    - I changed [functor_fix_snd_arg], defining it for precategories as [functor_fix_fst_arg]
    - I defined the functor [category_op_binproduct] from ((C^op) × (D^op)) to (C × D)^op (Probably there is a weq between these categories)


benediktahrens:
@CalosciMatteo : thanks a lot for the detailed description of your PR - this makes reviewing it much easier.

UniMath/CategoryTheory/.package/files
```diff
@@ -314,3 +314,6 @@ Monoidal/RezkCompletion/LiftedUnitors.v
 Monoidal/RezkCompletion/LiftedAssociator.v
 Monoidal/RezkCompletion/LiftedMonoidal.v
 Monoidal/RezkCompletion/MonoidalRezkCompletion.v
+
+PowerObject.v
+elementary_topos.v
```
benediktahrens, line 319: I believe it is now more common to use CamelCase file names rather than names with underscores. So maybe rename `elementary_topos.v` into `ElementaryTopos.v`?

UniMath/CategoryTheory/Monics.v
```diff
@@ -156,6 +156,27 @@ Section monics_subcategory.
     subcategory_of_monics⟦subprecategory_of_monics_ob c', subprecategory_of_monics_ob c⟧ :=
     tpair _ f isM.
 
+  (*A morphism (f,,s) in subcategory_of_monics is a z_iso if the underlying morphism is also a z_iso,
+  because the inverse must be monic.
+  Here p is redundant (it can be derived by H)*)
+  Lemma is_z_iso_in_subcategory_of_monics_from_is_z_iso {a b : subcategory_of_monics}
```
benediktahrens, line 162: I am not sure it is a good idea to have `a` and `b` implicit, since only their first components appear in the types of later arguments.

benediktahrens, line 162: Does the converse of this statement also hold?

UniMath/CategoryTheory/Subobjects.v
```diff
@@ -189,3 +287,102 @@ Definition SubObjPoset (c : C) : Poset :=
   (SubObj c,,SubObj_rel c,,ispreorder_SubObj_rel c,,isantisymm_SubObj_rel c).
 
 End subobj.
+
+(*Definition of the functor C^op -> HSET which maps c on SubObj c and mpas morphism "by pullback"*)
```
benediktahrens, line 291:
```suggestion
(*Definition of the functor C^op -> HSET which maps c on SubObj c and maps morphism "by pullback"*)
```

UniMath/CategoryTheory/SubobjectClassifier.v
```diff
@@ -93,6 +109,231 @@ Coercion subobject_classifier_object : subobject_classifier >-> ob.
...
+          intro.
+          use homset_property. }
+        simpl.
+        use (PullbackArrowUnique' _ _ _ (subobject_classifier_pullback O
+        f_asMonic)).
+        * exact t_tri.
+        * use TerminalArrowUnique.
+  Defined.
+
+  Let path_from_fepi : (characteristic_morphism O f_asMonic) = (TerminalArrow T c · (true O)).
```
benediktahrens, line 164: I have not seen lemmas introduced by `Let`. What is the effect? Is that result available outside this module (file)?

UniMath/CategoryTheory/elementary_topos.v
```diff
@@ -0,0 +1,597 @@
...
+Require Import UniMath.CategoryTheory.exponentials.
+Require Import UniMath.CategoryTheory.Adjunctions.Core.
+Require Import UniMath.CategoryTheory.Subobjects.
+Require Import UniMath.CategoryTheory.opp_precat.
+Require Import UniMath.CategoryTheory.categories.HSET.MonoEpiIso.
+
+Local Open Scope cat.
+
+(*
+An elemnetary topos is a category which has:
```
benediktahrens, line 40:
```suggestion
An elementary topos is a category which has:
```

UniMath/CategoryTheory/elementary_topos.v
```diff
@@ -0,0 +1,597 @@
...
+  use left_adjoint_from_partial.
+  + intro c.
+    exact (Subobject_dom (G0 b c)).
+  + intro c.
+    exact (ev b c).
+  + intro c.
+    use (universality b c).
+Defined.
+
+End Exponenitals.
```
benediktahrens, line 595:
```suggestion
End Exponentials.
```

UniMath/CategoryTheory/elementary_topos.v
```diff
@@ -0,0 +1,597 @@
...
+  use characteristic_morphism.
+  { exact B. }
+  use make_Monic.
+  + exact (SingletonArrow B).
+  + exact (SingletonArrow_isMonic B).
+Defined.
+End KroneckerDelta.
+
+(*In this section we show that an elemntary topos has exponentials*)
+Section Exponenitals.
```
benediktahrens, line 281:
```suggestion
Section Exponentials.
```

UniMath/CategoryTheory/.package/files
CalosciMatteo: Ok, just committed the renaming. Will keep that in mind for the future.

UniMath/CategoryTheory/Monics.v
CalosciMatteo: I do agree, committed.

CalosciMatteo: I think the converse always hold for a general subcategory of C. If there is an inverse (g,,sg) of (f,,sf) we have g is an inverse of f in C because of how composition is subcategory is defined.

A quick proof would be:

Lemma name (C:category) (C':sub_precategories C)(a b : C')
(f : C'⟦ a , b ⟧)
(H: is_z_isomorphism f)
: is_z_isomorphism (pr1 f).
Proof.
  induction H as (g,(gl,gr)).
  induction g as (g_und,?).
  use make_is_z_isomorphism.
  + exact g_und.
  + split.
    - exact (maponpaths pr1 gl).
    - exact (maponpaths pr1 gr).
Defined.

Should I add this to .../Subcategory/Core.v ?

UniMath/CategoryTheory/SubobjectClassifier.v
CalosciMatteo: What I wanted to obtain using [Let](https://coq.inria.fr/refman/language/core/sections.html#coq:cmd.Let) was to make the result not available outside of the section (and the module) in which it's used. My motivation in this particular case is that the result is essentially `subobject_classifier_square_commutes` and the definition of epi, so I didn't want to "flood" the namespace.
On the other hand I also can see why one could want every intermediate result available.
I faced a similar dilemma when deriving exponentials in the file now named `ElementaryTopos`. I used `Let` because I think those intermediate result/definition can be later accessed by accessor in `UniMath/CategoryTheory/Adjunctions/Core.v`

UniMath/CategoryTheory/elementary_topos.v
CalosciMatteo: Fixed

UniMath/CategoryTheory/elementary_topos.v
CalosciMatteo: Fixed

UniMath/CategoryTheory/elementary_topos.v
CalosciMatteo: Fixed

UniMath/CategoryTheory/ElementaryTopos.v
```diff
@@ -0,0 +1,597 @@
...
+  { exact (PowerObject_on_ob P b). }
+  { use Subobjectscategory_ob.
+    - exact T.
+    - exact (name_true b c).
+    - use from_terminal_isMonic.
+  }
+  - use u.
+Defined.
+
+Let G0_Sqr (b c : C) : (Subobject_mor (G0 b c) · (u b c) = (TerminalArrow T _ ) · (name_true b c)).
```
nmvdw, line 331: Interestingly, if you finish a `Let` with a `Qed`, then outside of the section this definition is still unfolded. It think it would be better to make this a `Lemma` or `Proposition`.

UniMath/CategoryTheory/Monics.v
```diff
@@ -156,6 +156,40 @@ Section monics_subcategory.
...
+      - use eq_in_sub_precategory.
+        exact H2.
+  Defined.
+
+  (*The converse of the previous result is always true, see [is_z_iso_from_is_z_iso_in_subcategory]*)
+  Definition is_z_iso_in_subcategory_of_monics_weq (a b : subcategory_of_monics)
+  (f : a --> b)
+  : (is_z_isomorphism (pr1 f)) ≃ (is_z_isomorphism f).
+  Proof.
+    use weqiff.
```
nmvdw, line 186: Maybe `weqimplimpl` might be more convenient to use here, because then the `split` on the next line would not be needed.

UniMath/CategoryTheory/PowerObject.v
```diff
@@ -0,0 +1,356 @@
...
+      BinProductOfArrows_comp,
+      id_right,
+      id_left.
+    apply idpath. }
+  apply pathsinv0.
+  rewrite <-(PowerObject_transpose_tri f), <-BinProductOfArrows_idxcomp, !assoc'.
+  use cancel_precomposition.
+  use pathsinv0.
+  use PowerObject_transpose_tri.
+Defined.
```
nmvdw, line 255: `Qed` can be used here.

UniMath/CategoryTheory/SubobjectClassifier.v
```diff
@@ -93,6 +109,231 @@ Coercion subobject_classifier_object : subobject_classifier >-> ob.
 Coercion true : subobject_classifier >-> Monic.
 
 (** The arrow Goldblatt calls [true! := (! : X -> T) · true] *)
-Definition const_true {C : category} {T : Terminal C} {X : ob C}
+Definition const_true {C : category} {T : Terminal C} (X : ob C)
            (O : subobject_classifier T) : X --> subobject_classifier_object O :=
   TerminalArrow T X · true O.
+
+(*A category with subobjectclassifier is balnced: if a morphism is mono and epi then it is iso*)
```
nmvdw, line 116:
```suggestion
(*A category with subobjectclassifier is balanced: if a morphism is mono and epi then it is iso*)
```

UniMath/CategoryTheory/limits/equalizers.v
```diff
@@ -245,6 +247,35 @@ Section def_equalizers.
     exact (make_Monic C (EqualizerArrow E) (EqualizerArrowisMonic E)).
   Defined.
 
+  (*
+  The equalizer is a z-isomorphism if f = g
+  *)
+  Lemma z_iso_Equalizer_of_same_map {y z : C} {f g : y --> z} (E : Equalizer f g) (p:f = g) : is_z_isomorphism (EqualizerArrow E).
+  Proof.
+    induction p.
+    transparent assert (E_id : (Equalizer f f) ). {
```
nmvdw, line 256: I think you can turn this one into its own separate definition

UniMath/CategoryTheory/Subobjects.v
```diff
@@ -189,3 +287,102 @@ Definition SubObjPoset (c : C) : Poset :=
...
+    - cbn.
+      fold pb.
+      fold pbpb.
+      assert (H : PullbackPr1 pb_glued = PullbackPr1 pbpb). { apply idpath. }
+      induction H.
+      apply pathsinv0.
+      use (PullbackArrow_PullbackPr1 pb_glued).
+Defined.
+
+Definition SubObj_Functor : C^op ⟶ hset_category.
```
nmvdw, line 381: Something interesting here, is the following. If one works with categories without the requirement that they univalent, then one needs to take a quotient on the subobject in order to get this functor. However, if one works with univalent categories instead, then this quotient can be avoided.

nmvdw (approved):
Interesting work!

I guess that the reason why the sanity checks failed, is because the latest version of `CONTENTS.md` still needs to be pushed.

UniMath/CategoryTheory/SubobjectClassifier.v
nmvdw: `Let` indeed does not make the result available outside of the section. This can be quite nice to introduce some naming that is only needed within a specific section. However, for proofs, I think it's better to use something like `Local Lemma` (https://coq.inria.fr/refman/language/core/definitions.html). That is because outside of the section, the specific definition will be transparent, which often is not desired.

benediktahrens: Re: "outside of the section, the specific definition will be transparent which often is not desired." Just to be extra sure I understand, this refers to definitions given with `Let`?

nmvdw: > Re: "outside of the section, the specific definition will be transparent which often is not desired." Just to be extra sure I understand, this refers to definitions given with `Let`?

Yes.

UniMath/CategoryTheory/Subobjects.v
benediktahrens: That's indeed very interesting - thanks for pointing this out.
We should have a list of such observations about univalent categories somewhere.

CalosciMatteo:
> Interesting work!
> 
> I guess that the reason why the sanity checks failed, is because the latest version of `CONTENTS.md` still needs to be pushed.

Thanks for the help, you were right. I just had to delete the file and launch `make UniMath/CONTENTS.md`. Now the sanity check succeeded.

UniMath/CategoryTheory/ElementaryTopos.v
```diff
@@ -0,0 +1,597 @@
...
+  := (BinProductOfArrows _ (BinProd _ _) (BinProd _ _) f g) (at level 10).
+  (*\timesbar*)
+
+Section KroneckerDelta.
+(*The [KroneckerDelta] predicate is defined as
+the chraracteristic morphis of the [diagonalMap],
+The [SingletonArrow] is defined as
+the [PowerObject_transpose] of the [KroneckerDelta]*)
+
+Local Definition diagonalMap (B:C) : Monic C B (B ⨉ B).
```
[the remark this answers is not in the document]
CalosciMatteo: I agree. I also moved the definition to `UniMath/CategoryTheory/limits/binproducts.v`

UniMath/CategoryTheory/ElementaryTopos.v
CalosciMatteo: Ok, should i make it also `Local`? or should I change the name to something like `Exponentials_from_Topos_G0sqr`?

UniMath/CategoryTheory/Monics.v
CalosciMatteo: I agree, committed. Thank you.

UniMath/CategoryTheory/PowerObject.v
CalosciMatteo: Ok, committed.

UniMath/CategoryTheory/ElementaryTopos.v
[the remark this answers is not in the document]
CalosciMatteo: I agree. I also moved this definition to `UniMath.CategoryTheory.limits.binproducts.v`

UniMath/CategoryTheory/SubobjectClassifier.v
CalosciMatteo: I understand, but I also noticed (through experimentations) that a `Local Lemma` can still be accessed outside of his section and, for example, you can't define something else outside said section in the same file with the same name of the lemma.
Is this ok for us? Is there a way to define something opaque and "delimited" to a section?

UniMath/CategoryTheory/limits/equalizers.v
CalosciMatteo: Ok, done!

UniMath/CategoryTheory/SubobjectClassifier.v
benediktahrens: It is not a problem if something can be accessed outside a module or section, I think.

benediktahrens:
If I have understood correctly, this PR is almost ready to be merged. The only remaining change seems to be to replace `Let` by something suitable, e.g., `Local Lemma`.
Have I forgotten something?
@CalosciMatteo : are you willing to make this change?

CalosciMatteo:
> If I have understood correctly, this PR is almost ready to be merged. The only remaining change seems to be to replace `Let` by something suitable, e.g., `Local Lemma`. Have I forgotten something? @CalosciMatteo : are you willing to make this change?

Yes, I do. I was only waiting for an answer to this:

> I understand, but I also noticed (through experimentations) that a `Local Lemma` can still be accessed outside of his section and, for example, you can't define something else outside said section in the same file with the same name of the lemma. Is this ok for us? Is there a way to define something opaque and "delimited" to a section?

I will start making the changes right now.

benediktahrens:
> 
> Yes, I do. I was only waiting for an answer to this:
> 
> > I understand, but I also noticed (through experimentations) that a `Local Lemma` can still be accessed outside of his section and, for example, you can't define something else outside said section in the same file with the same name of the lemma. Is this ok for us? Is there a way to define something opaque and "delimited" to a section?
> 
> I will start making the changes right now.

I see, sorry for missing this. Yes, I think this is ok for us.

Issue #1611, opened 2023-01-16
Confusion about CI sanity checks


benediktahrens:
I am confused by the output of the sanity check to a (draft) PR: https://github.com/UniMath/UniMath/actions/runs/3903770065/jobs/6668606428
It says that the CI run has failed, but I don't see which of the sanity checks has failed. I am probably missing something obvious...

JasonGross:
https://github.com/UniMath/UniMath/actions/runs/3903770065/jobs/6668606428#step:4:36
make[1]: *** No rule to make target 'UniMath/CategoryTheory/ZigZag.v', needed by '..coq_makefile_output.d'.

Issue #1567, opened 2022-10-13
Add failure messages for sanity checks?


peterlefanulumsdaine:
Currently, all the sanity checks print messages when they start and if they succeed, but only `check-listing-of-proof-files` gives a message on failure.  Adding failure messages would have a couple of advantages:

- users getting a failed check will get a better pointer for what they need to fix
- if running the sanity checks in parallel locally, with eg `make -j4`, this makes it possible to see which check failed (currently very unclear)

peterlefanulumsdaine:
Note for reference: the discussion on #1615 suggests further tweaks to make sure the sanity check reports appear prominently at the end of a build, even under `-k` — not implemented there, but if someone’s working over the sanity checks for this issue, they would be good to consider too.

PR #1615, opened 2023-01-27
call sanity checks in CI in a way that makes error appear last

benediktahrens:
see #1611


DanGrayson:
There may be multiple errors.  This will move only the first error into prominence.

benediktahrens:
> There may be multiple errors. This will move only the first error into prominence.

For me, this would be an improvement over the current situation. A better, but also more complicated, mechanism could save the errors in an intermediate file and print that file at the end.

DanGrayson:
I think it's pretty easy to search the output for the string " *** ", thereby finding all the errors.

m-lindgren:
Maybe a solution would be to grep the output for '***' if there is an error? If we also implement what Peter suggested in #1567 the output should be pretty informative.

peterlefanulumsdaine:
Is there any reason not to merge this?  I agree with @benediktahrens that it’s an incremental improvement on the current situation.  The suggested further improvements could be great, but it’s been 3 months without progress on them, so I suggest merging this for now, and noting this discussion in #1567 (or another issue) for possible future improvements.

m-lindgren (approved):

PR #1610, opened 2023-01-12
Two-sided displayed categories

nmvdw:
- Notion of two-sided displayed category
- Some basic theory
- Two-sided fibration
- Examples

PR #1618, opened 2023-01-27
CI: Build with latest dev version of Coq.

### LINKED ISSUE #1617: CI: testing with Coq master
Can we set up CI so that it also tests with Coq master? This would allow us to catch problems such as #1614 before merging.
@m-lindgren : would you be willing to look into this?

m-lindgren:
This builds PR's with the latest relased version of Coq and the latest development version. Should fix #1617.


m-lindgren:
I expect the -dev build to fail on the changes in #1610.

Issue #1614, opened 2023-01-27
Broken in coq CI


SkySkimmer:
eg https://gitlab.com/coq/coq/-/jobs/3668196302
Seems to be due to https://github.com/UniMath/UniMath/pull/1610

nmvdw:
Interestingly enough, I don't have this error if I compile the code locally and the error did not occur in the UniMath CI.

Could it be related to the version of Coq?

benediktahrens:
Thanks for filing this issue!

The build fails with
```
COQC UniMath/CategoryTheory/LeftKanExtension.v
File "./UniMath/CategoryTheory/LeftKanExtension.v", line 153, characters 6-29:
Error:
Found no subterm matching "colimIn (lan_colim ?x) ((?M80,, ?M82),, ?M81)
                           · lan_mor ?M79" in the current goal.
Command exited with non-zero status 1
UniMath/CategoryTheory/LeftKanExtension.vo (real: 0.63, user: 0.48, sys: 0.14, mem: 378124 ko)
build/CoqMakefile.make:815: recipe for target 'UniMath/CategoryTheory/LeftKanExtension.vo' failed
make[3]: *** [UniMath/CategoryTheory/LeftKanExtension.vo] Error 1
make[3]: *** Waiting for unfinished jobs....
```
@nmvdw : could you look into this?
@SkySkimmer : which version of Coq does the failed build use? I cannot seem to find the version information.

m-lindgren:
Should we revert #1610 for now, unless we have a solution ready/in the works? So we don't keep coq's CI pipeline broken longer than necessary.

benediktahrens:
Maybe it would be sufficient to remove the offending file on left Kan extensions?

m-lindgren:
Removing `LeftKanExtension.v` hopefully fixes Coq CI. On our side the build will also fail [here](https://github.com/UniMath/UniMath/actions/runs/4026525308/jobs/6921149836#step:5:1981):
```
          coqc UniMath/Bicategories/OtherStructure/Examples/StructureOneTypes.{glob,vo} (exit 1)
  File "./UniMath/Bicategories/OtherStructure/Examples/StructureOneTypes.v", line 604, characters 7-35:
  Error:
  Found no subterm matching "pair_2cell (one_types,, has_binprod_one_types)
                               ?M65 ?M66 (?M67,, ?M68)" in the current goal.
  
[...]

PR #1619, opened 2023-01-27
remove file on left kan extension

benediktahrens:
see #1614


nmvdw:
`CONTENTS.md` needs to be pushed

benediktahrens:
@nmvdw : Thanks for fixing the error. Is a similar change possible for the file on left Kan extensions?

nmvdw:
This PR fixes the Coq CI, and it deletes the file on Left Kan extensions. However, that's not a big problem, so for that reason, I merge it.

nmvdw:
> @nmvdw : Thanks for fixing the error. Is a similar change possible for the file on left Kan extensions?

I dunno. I would need to look at that, but I will do that later (at some point in the future).

PR #1621, opened 2023-02-02
Univalent enriched categories

nmvdw:
This PR contains:
- The notion of enriched category. Note that the definition I use here, is different from what's usually given in the literature. 
- The bicategory of univalent enriched categories and a proof that this bicategory is univalent.
- Some accessors/lemmas for monoidal categories. None of this is new, but I gave stuff more convenient names.
- A random functor about the arrow 2-sided displayed category that is not related at all to enriched categories, but which still is in this PR.

@benediktahrens @rmatthes I think you two should take a look at this PR, because it touches monoidal categories. There's no rush by the way, so just look at it when you have time.


benediktahrens:
Problem with the ordering of files

nmvdw:
> Problem with the ordering of files

Fixed

UniMath/Bicategories/DisplayedBicats/Examples/EnrichedCats.v
```diff
@@ -0,0 +1,895 @@
+(*****************************************************************
+
+ The bicategory of univalent enriched categories
+
+ We construct the bicategory of univalent enriched categories and
+ we prove that this bicategory is univalent. Note that in order to
+ prove the univalence, we need to assume that the involved
```
benediktahrens, line 7: “need” is a bit ambiguous: you mean that (V univalent) => (Enriched(V) 2-univalent), which could be verbalized as saying “it is sufficent for V to be univalent to obtain that (Enriched V) is univalent.
“need” sounds like “necessary”, which is the other implication.

UniMath/Bicategories/DisplayedBicats/Examples/EnrichedCats.v
```diff
@@ -0,0 +1,895 @@
...
+         use disp_univalent_2_0_enriched_cats_help_path ;
+         apply idpath).
+  Defined.
+
+  Definition disp_univalent_2_0_enriched_cats
+             (HV : is_univalent V)
+    : disp_univalent_2_0 disp_bicat_of_enriched_cats.
+  Proof.
+    intros C₁ C₂ p E₁ E₂.
+    induction p.
```
benediktahrens, line 874: I have not looked at the proof, but the “induction p” step could/should probably be encapsulated into an intermediate result about something being “fiberwise univalent”, similar to the 1-dimensional case. It is a one-liner, but it would be good to explicitly formulate it.

UniMath/CategoryTheory/Enriched/Enrichment.v
```diff
@@ -0,0 +1,1585 @@
...
+ In this file, we define enrichments of categories, functors, and
+ natural transformations. Note that we define these enrichments as
+ categories/functors/transformations with extra data and
+ properties, whereas the standard definition of enriched category
+ does not do so.
+
+ There are a couple of reasons for this choice:
+ - It will help use prove the univalence of the bicategory of
+   univalent enriched categories. That is because in the whole
+   prove, we don't have to prove equality of the type of objects.
```
benediktahrens, line 14:
```suggestion
   proof, we don't have to prove equality of the type of objects.
```

benediktahrens: Another reason for using this def is that you can reuse the proof of the univalence of the bicategory of univalent categories.

benediktahrens (commented):
preliminary comments, review not finished.
Since the definition is not standard, an equivalence with the standard definition should eventually be constructed.

UniMath/Bicategories/DisplayedBicats/Examples/EnrichedCats.v
nmvdw: Fixed

UniMath/Bicategories/DisplayedBicats/Examples/EnrichedCats.v
nmvdw: Indeed. I forgot to use these statements.

nmvdw:
Now there is a proof that the two notions of enriched category coincide.

UniMath/CategoryTheory/Enriched/Enrichment.v
```diff
@@ -0,0 +1,1587 @@
+(*****************************************************************
+
+ Enrichments of categories
+
+ In this file, we define enrichments of categories, functors, and
+ natural transformations. Note that we define these enrichments as
+ categories/functors/transformations with extra data and
+ properties, whereas the standard definition of enriched category
+ does not do so.
```
benediktahrens, line 9: Maybe point to the paper by McDermott & Uustalu here, even though their definition is a bit different?

benediktahrens (approved):

UniMath/CategoryTheory/Enriched/Enrichment.v
nmvdw: Sure. I also added more comments in the file on the displayed bicategory of enriched categories.

PR #1623, opened 2023-02-02
CI: Build satellites with dev version of Coq.

### LINKED ISSUE #1620: CI should check satellites with dev version of Coq
Currently, the satellites are not checked by the Coq CI, and we only check them with released versions of Coq. This means that we catch problems with Coq only when the problems has entered a released version.

Instead, we should also check the satellites with Coq dev, so that problems can be caught early.

@m-lindgren : could you implement this?

m-lindgren:
This change builds all satellites except TypeTheory using two different versions of Coq: the latest released version (currently 8.16.1 but soon 8.17) and current head of the master branch. Fixes #1620.

### DIFF
--- a/.github/workflows/build-unimath.yml
+++ b/.github/workflows/build-unimath.yml
@@ -138,7 +138,7 @@ jobs:
       fail-fast: false
       matrix:
         satellite: [SetHITs, largecatmodules, GrpdHITs]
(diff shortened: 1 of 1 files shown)


benediktahrens:
@m-lindgren : great, thanks a lot!

PR #1624, opened 2023-02-04
CI: Enable caching on MacOS build.

### LINKED ISSUE #1622: CI: MacOS build does not use caching
I open this issue so that I don't forget I need to fix this. 

It seems the script that manages the cache trims the cache too much, which results in rebuilding most of UniMath och each build.

m-lindgren:
With this PR the CI builds UniMath on macOS using Coq installed with Homebrew, and manages the cache manually. Fixes #1622.


benediktahrens:
@m-lindgren : thanks a lot!

PR #1626, opened 2023-02-07
More on enriched categories

nmvdw:
This PR contains:
- Refactor the file on enrichments, so that there smaller files
- Added two examples of enriched categories: the unit and the full subcategory
- Constructed a terminal object and equifiers in the bicategory of enriched categories
- Characterization of faithful 1-cells of enriched categories
- Fully faithful enriched functors between enriched categories, and a proof that they give rise to fully faithful 1-cells


UniMath/CategoryTheory/EnrichedCats/Enrichment.v
```diff
@@ -0,0 +1,462 @@
...
+ 3. Faithfulness
+ *)
+Definition faithful_moncat
+           (V : monoidal_cat)
+  : UU
+  := ∏ (x y : V)
+       (f g : x --> y),
+     (∏ (a : 𝟙 --> x), a · f = a · g)
+     →
+     f = g.
```
benediktahrens, line 462: Interesting. I seem to remember that such a property is called "well-pointed" in topos theory.

benediktahrens (approved):

UniMath/CategoryTheory/EnrichedCats/Enrichment.v
rmatthes: This difference also shows up in pointed objects in a monoidal category: in lack of an established terminology I know of, I called them monoidal-pointed objects to stress that they go out of 1 which is not necessarily terminal - my example were the endofunctors.

nmvdw: That terminology makes sense, and I think it is good to consistently call maps `1 -> X` points. For this particular definition, I like faithful though, because that directly expresses what is being defined.

nmvdw:
I added inserters of enriched categories and some documentation.

UniMath/CategoryTheory/EnrichedCats/Examples/SliceEnriched.v
```diff
@@ -0,0 +1,188 @@
+(**********************************************************************
+
+ Enriched slice categories
+
+ In this file, we define enriched slice categories. The approach that
+ we take, is based on the fact that slice categories can be defined
```
benediktahrens, line 6:
```suggestion
 we take is based on the fact that slice categories can be defined
```

UniMath/CategoryTheory/EnrichedCats/Examples/SliceEnriched.v
```diff
@@ -0,0 +1,188 @@
...
+
+ Let's be more specific. Suppose that we have a category `C` and an
+ object `x` in `C`. To construct the slice category `C/x`, we take the
+ category of dialgebras between the identity and the functor that is
+ constantly `x`. As such, the objects of this category are pairs of an
+ object `a` in `C` together with a morphism `a --> x`. As such, this
+ corresponds to objects in the slice category `C/x`. The same can be
+ said for morphisms.
+
+ Note that we assume that the monoidal category `V` has equalizers and
```
benediktahrens, line 19: I would like to understand this better: the definition of slice requires (chosen) equalizers? Can you explain why that is?

nmvdw: Basically, one wants to take the same definition of the slice category. As such, morphisms from `a --> x` to `b --> x` are morphisms `a --> b` that satisfy a commutativity requirement. With the definition of enriched category, we can talk about morphisms `a --> b`, but we also want to be able to require that a certain diagram commutes. That is why we use equalizers: we are looking at the elements that 'equalize' two arrows.

I added an expanded version of this in the comments there.

Note that requirements like having equalizers are rather common in enriched category theory. For example:
- To define functor categories, one requires `V` to be complete
- To define Eilenberg-Moore categories, one requires `V` to have split equalizers (in the formalization, I used equalizers in general, though, but I don't think that's a big issue)
- To define limits in the bicategory of enriched categories, one requires `V` to be complete
- To define colimits in the bicategory of enriched categories, one requires `V` to be cocomplete
- To define the tensor of enriched categories, one requires `V` to be symmetric
- To define the opposite of enriched categoies, one requires `V` to be symmetric
- To define self-enrichments, one requires `V` to be symmetric closed
It is very common to lay down some requirements on `V` in order to do category theory: this is captured in the notion of Bénabou cosmos (https://ncatlab.org/nlab/show/Bénabou+cosmos). There are other notions as well, such as Street cosmoi (https://ncatlab.org/nlab/show/cosmos), and such ideas have also been employed in higher category (https://ncatlab.org/nlab/show/infinity-cosmos).

nmvdw: I think that in a univalent setting, 'cartesian' is also a very natural requirement for `V`. If we have two enriched categories `E1, E2`, then their tensor is defined as follows:
- Objects: pairs of objects in `E1` and `E2`
- Morphisms from `(x1, y1)` to `(x2, y2)`: the tensor of `E1[[ x1, x2]]` and `E2[[y1, y2]]`

However, to guarantee univalence of the tensor, we need that isomorphisms from `(x1, y1)` to `(x2, y2)`, are the same as pairs of isomorphisms from `x1` to `x2` and isomorphisms from `y1` to `y2` (this is forced by how we chose the objects). This requirement does not hold in general. If we would look at `V` the monoidal category of abelian groups with the tensor product, then I think this requirement would not hold, and thus the usual definition of the tensor of enriched categories, does not give rise to a univalent category. 

For that reason, we want the tensor of the monoidal category `V` to be like a product for this construction, which would be the case if `V` is cartesian. While some examples (abelian groups with the tensor product) would not be captured by this, other examples (enrichments over domains or enrichments over presheaf categories) satisfy these requirements.

PR #1633, opened 2023-02-28
representation of inductive and coinductive syntax with actegories

rmatthes:
This work expands on the CPP'22 paper "Implementing a category-theoretic framework for typed abstract syntax" about the construction of inductive syntax for multisorted binding signatures in two ways: 
- coinductive syntax for multisorted binding signatures is taken into account (whose existence is based on $\omega$-continuity of the generated signature functor, while the inductive case is based on $\omega$-cocontinuity)
- the constructions are first abstractly done on the level of a monoidal category (using the strength notion coming from actegories - for the action of (monoidal-)pointed objects on all objects) then instantiated to suitable functor categories over a base category with suitable properties, finally the base category is instantiated to the category of sets

The material is complete but not yet polished. It contains plenty of upstream material but could do even more service to the UniMath library. Comments are welcome. The new material should not interfere with the packages of UniMath other than SubstitutionSystems and the parts on monoidal categories in CategoryTheory and Bicategories.


rmatthes:
I do not know what the failure of the sanity checks means. Locally, they run fine. Is there a problem if one normally runs Dune but does the sanity checks with `make`?

benediktahrens:
Re sanity checks: CI seems to say that CONTENTS.md is not up to date. Could you make that file and commit it?

rmatthes:
> Re sanity checks: CI seems to say that CONTENTS.md is not up to date. Could you make that file and commit it?

When I check it with `make sanity-checks`, all is fine. When I want to build it with Dune, nothing is to be done. So, I just have no idea how to proceed.

m-lindgren:
Try the command `make UniMath/CONTENTS.md`, it should generate a new `CONTENTS.md` for you (I hope) that's up to date.

rmatthes:
> Try the command `make UniMath/CONTENTS.md`, it should generate a new `CONTENTS.md` for you (I hope) that's up to date.

I removed a line from that file, and still this make command did nothing. But motivated by your suggestion, I deleted the file, and then it was generated with the next make command.

rmatthes:
Thanks to @benediktahrens and @m-lindgren, the first hurdle is overcome.

UniMath/MoreFoundations/Tactics.v
```diff
@@ -62,3 +62,24 @@ Proof. intros. exact (f ig). Defined.
...
+Ltac split13 := split; [| split12].
+Ltac split14 := split; [| split13].
+Ltac split15 := split; [| split14].
+Ltac split16 := split; [| split15].
+Ltac split17 := split; [| split16].
+Ltac split18 := split; [| split17].
+Ltac split19 := split; [| split18].
+Ltac split20 := split; [| split19].
+Ltac split21 := split; [| split20].
+(** this allows to decompose a goal for [prebicat_laws] *)
```
nmvdw, line 85: I tend to use `repeat split` for such goals. This way, I don't need to know how many goals there will be, which seems to be necessary for these tactics.

UniMath/CategoryTheory/BicatOfCatsElementary.v
```diff
@@ -0,0 +1,314 @@
...
+
+Definition lassociator_CAT {C D E F : category} (X : [C, D]) (Y : [D, E]) (Z : [E, F]) :
+    [C, F] ⟦functor_compose X (functor_compose Y Z), functor_compose (functor_compose X Y) Z⟧
+    := nat_trans_id (functor_composite X (functor_composite Y Z)).
+
+Definition rassociator_CAT {C D E F : category} (X : [C, D]) (Y : [D, E]) (Z : [E, F]) :
+    [C, F] ⟦functor_compose (functor_compose X Y) Z, functor_compose X (functor_compose Y Z)⟧
+    := nat_trans_id (functor_composite (functor_composite X Y) Z).
+
+Definition vcomp2_CAT {C D : category} (F G H : [C, D]) : [C,D]⟦F, G⟧ -> [C,D]⟦G, H⟧ -> [C,D]⟦F, H⟧.
```
nmvdw, line 51: Do you want `F, G, H` as implicit or as explicit arguments? For the bicategorical `vcomp2`, they are implicit. For `nat_trans_comp`, they are explicit, but not that this causes there to be more `_` in the code (see line 53 here).

UniMath/CategoryTheory/BicatOfCatsElementary.v
```diff
@@ -0,0 +1,314 @@
+(** the constituents of the bicategory of categories without using the package [Bicategories];
```
nmvdw, line 1: In UniMath, `CategoryTheory` does not depend for good reasons on `Bicategories`. My question is: why is this file needed? What stuff depends on it? What is the advantage of keeping that in `CategoryTheory` instead of categories? This could also be good to add in the comments of this file.

UniMath/Bicategories/MonoidalCategories/MonadsAsMonoidsWhiskered.v
```diff
@@ -0,0 +1,66 @@
...
+Require Import UniMath.CategoryTheory.Monoidal.CategoriesOfMonoidsWhiskered.
+Require Import UniMath.Bicategories.MonoidalCategories.EndofunctorsWhiskeredMonoidal.
+Require Import UniMath.CategoryTheory.Monads.Monads.
+
+Local Open Scope cat.
+
+Section MonoidToMonad.
+
+  Context {C : category}.
+  Let ENDO := monoidal_of_endofunctors C.
```
nmvdw, line 25: While not needed, I like to add type declarations to `Let` statements. Like `Let ENDO : monoidal_cat := monoidal_of_endofunctors C.`

UniMath/Bicategories/MonoidalCategories/MonadsAsMonoidsWhiskered.v
```diff
@@ -0,0 +1,66 @@
+(** In this file, we show how any monoid in the monoidal category of endofunctors is a monad **)
```
nmvdw, line 1: Would it be a lot of work to construct an equivalence of types? This could also be left for later

UniMath/Bicategories/MonoidalCategories/ActionOfEndomorphismsInBicatWhiskered.v
```diff
@@ -82,10 +183,176 @@ Section Instantiation_To_Bicategory_Of_Categories.
...
+  (* Transparent action_from_precomp_CAT_laws. *)
+
+  Definition actegoryfromprecomp_actions_equal_statement : UU :=
+    actegory_action (monendocat_monoidal C) (actegory_from_precomp_CAT C D) =
+      actegory_action (monendocat_monoidal C) actegoryfromprecomp. (* slow *)
+
+Lemma actegoryfromprecomp_actions_equal : actegoryfromprecomp_actions_equal_statement.
+Proof.
+  cbn.
+  use total2_paths_f.
```
nmvdw, line 201: I think you can use `subtypePath` here instead of `total2_paths_f`.

UniMath/CategoryTheory/FunctorCoalgebras.v
```diff
@@ -164,14 +167,180 @@ Proof.
   apply (coalgebra_mor_commutes F f).
 Defined.
 
-Lemma terminalcoalgebra_isiso : is_iso α.
+Lemma terminalcoalgebra_is_z_iso : is_z_isomorphism α.
 Proof.
-  apply (is_iso_qinv α α').
+  exists α'.
```
nmvdw, line 172: `make_is_z_isomorphism` could be of use here

UniMath/CategoryTheory/FunctorCoalgebras.v
```diff
@@ -164,14 +167,180 @@ Proof.
...
   split.
   - exact αα'_idA.
   - exact α'α_idFA.
 Defined.
 
-Definition terminalcoalgebra_iso : iso A (F A) := make_iso α terminalcoalgebra_isiso.
+Definition terminalcoalgebra_z_iso : z_iso A (F A) := α,, terminalcoalgebra_is_z_iso.
+
+Definition terminalcoalgebra_iso : iso A (F A) := z_iso_to_iso terminalcoalgebra_z_iso.
```
nmvdw, line 180: Will the `iso` be needed or is the `z_iso` sufficient?

UniMath/CategoryTheory/limits/terminal.v
```diff
@@ -188,6 +188,18 @@ use make_Terminal.
...
 
+Definition Terminal_functor_precat_alt : Terminal [C,D].
+Proof.
+use make_Terminal.
+- exact (constant_functor _ _ ID).
+- intros F.
+  use tpair.
+  + use make_nat_trans; simpl.
+    * intro a; apply TerminalArrow.
+    * intros a b f; apply TerminalArrowEq.
```
nmvdw, line 199: This line could be made abstract

UniMath/CategoryTheory/limits/binproducts.v
```diff
@@ -613,9 +616,104 @@ Proof.
...
+  - rewrite assoc'.
+    etrans.
+    { apply maponpaths.
+      apply BinProductPr2Commutes. }
+    etrans.
+    { apply BinProductOfArrowsPr1. }
+    apply idpath.
+Qed.
+
+Definition BinProduct_of_functors_commutes :
```
nmvdw, line 703: I think this might be an instance of that fact that `X * Y \cong Y * X`, but then instantiated to functor categories.

UniMath/CategoryTheory/limits/binproducts.v
```diff
@@ -744,18 +842,28 @@ Proof.
     + exact (id_right _ @ !id_left _).
 Defined.
 
+Lemma terminal_binprod_unit_l {C : category}
+      (T : Terminal C) (BC : BinProducts C) :
+  ∏ x : C, is_iso (BinProductPr2 C (BC T x)).
```
nmvdw, line 847: Why do you use `is_iso` instead of `is_z_isomorphism`?

UniMath/CategoryTheory/limits/binproducts.v
```diff
@@ -766,6 +874,68 @@ Proof.
     + apply TerminalArrowEq.
 Defined.
 
+Lemma terminal_binprod_unit_r {C : category}
+      (T : Terminal C) (BC : BinProducts C) :
+  ∏ x : C, is_iso (BinProductPr1 C (BC x T)).
```
nmvdw, line 879: Would an `is_z_isomorphism` also be fine here?

UniMath/CategoryTheory/limits/binproducts.v
```diff
@@ -766,6 +874,68 @@ Proof.
...
+Defined.
+
+Lemma terminal_BinProduct_of_functors_unit_r_law :
+  is_nat_trans _ _ terminal_BinProduct_of_functors_unit_r_data.
+Proof.
+  intros c c' f.
+  apply BinProductOfArrowsPr1.
+Qed.
+
+Definition terminal_BinProduct_of_functors_unit_r  :
```
nmvdw, line 927: This also seems to be an instance of a general fact, namely that `X * 1 \cong X`, but then instantiated to functor categories.

UniMath/CategoryTheory/Monoidal/CategoriesOfMonoidsWhiskered.v
```diff
@@ -0,0 +1,212 @@
...
+Require Import UniMath.CategoryTheory.DisplayedCats.Core.
+Require Import UniMath.CategoryTheory.DisplayedCats.Total.
+
+Local Open Scope cat.
+
+Section Category_of_Monoids.
+
+  Context {C : category} (M : monoidal C).
+
+  Notation "x ⊗ y" := (x ⊗_{M} y).
```
nmvdw, line 22: Would this notation be usable: https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/Monoidal/MonoidalCategories.v#L490 (the scope is `moncat`)? How much of this notation can be put in the file on monoidal categories, so that it is usable outside of this file as well?

nmvdw (commented):
I am not finished yet with the review, but here are some comments already. The material looks very interesting!

nmvdw:
For `SetHITs`, the error is:
```
File "./UniMath/Combinatorics/Lists.v", line 223, characters 4-26:
[614](https://github.com/UniMath/UniMath/actions/runs/4298031600/jobs/7491700247#step:6:620)
  Error:
[615](https://github.com/UniMath/UniMath/actions/runs/4298031600/jobs/7491700247#step:6:621)
  Found no subterm matching "foldr1_map ?M77 ?M78 ?M79 nil" in the current goal.
```
The Coq development version was added to the CI, when the file on Left Kan Extensions led to a problem with it. This was also caused by a rewrite, which worked fine on my version of Coq, while on the development version it did not work. I don't have a general solution for it, but in this case, the lines that cause problems with the development version, could basically use `apply` instead of `rewrite` (after using `etrans` and `maponpaths` so that it is used in the right location of the term).

UniMath/CategoryTheory/Monoidal/ProductActegory.v
```diff
@@ -0,0 +1,419 @@
...
+      + red; split; apply id_left.
+  Defined.
+
+End OneProduct.
+
+Section Power.
+
+  Context (I : UU) {C : category} (Act : actegory Mon_V C).
+
+  Definition actegory_power : actegory Mon_V (power_category I C) :=
```
nmvdw, line 347: Personally, I am not too fond of the terminology `power_category` for the product (same for `actegory_power`). Powers are some kind of weighted limit and they occur frequently in enriched category theory and bicategory theory (https://ncatlab.org/nlab/show/powering). In the bicategory of categories, they correspond to functor categories, but that does not hold in general. 

The names 'tensor' (refers to the copower, which is a colimit) and 'cotensor' (refers to the power, which is a limit) have also been used for these concepts, but they come with their own confusions.

If at some point we add the notion of powers in bicategories, then these names are going to cause a clash. For that reason, I am in favor of renaming these concepts.

UniMath/CategoryTheory/coslicecat.v
```diff
@@ -91,6 +92,8 @@ Proof.
   * apply assoc'.
 Defined.
 
+(** that the previous Lemma is defined was done on purpose in 2018 - is it still appropriate? *)
```
nmvdw, line 95: Does the code still compile if you make it `Qed`?

UniMath/CategoryTheory/Monoidal/ConstructionOfActegoryMorphisms.v
```diff
@@ -0,0 +1,1212 @@
...
+  Definition composedlifteddistributivity_data: lifteddistributivity_data (v1 ⊗_{Mon_V} v2).
+  Proof.
+    red; intros.
+    exact (αinv_{Mon_V} _ _ _ · δ1 w ⊗^{Mon_V}_{r} v2 · α_{Mon_V} _ _ _
+             · v1 ⊗^{Mon_V}_{l} δ2 w · αinv_{Mon_V} _ _ _).
+  Defined.
+
+  Lemma composedlifteddistributivity_nat: lifteddistributivity_nat composedlifteddistributivity_data.
+  Proof.
+    do 2 red; intros; unfold composedlifteddistributivity_data; cbn.
```
nmvdw, line 520: Why do you need `do 2 red`? Would `red` have been sufficient?

UniMath/CategoryTheory/Monoidal/ConstructionOfActegoryMorphisms.v
```diff
@@ -0,0 +1,1212 @@
...
+         apply pathsinv0, (functor_comp (leftwhiskering_functor Mon_V v1)). }
+    cbn.
+    repeat rewrite assoc.
+    apply cancel_postcomposition.
+    apply pathsinv0, monoidal_associatornatleftright.
+  Qed.
+
+  Lemma composedlifteddistributivity_tensor: lifteddistributivity_tensor composedlifteddistributivity_data.
+  Proof.
+    do 2 red; intros; unfold composedlifteddistributivity_data; cbn.
```
nmvdw, line 565: Same here: why is `do 2 red` needed?

nmvdw (approved):
I looked at all files, except for those in `SubstitutionSystems`, and I think that part looks good.

UniMath/MoreFoundations/Tactics.v
rmatthes: In the past, I mostly used `repeat split`. The downside is that it flattens out the whole substructure of goals, e.g., when there are several laws for each unitor and the associator in whiskered monoidal categories. `split` does not only apply the splitting of a pair into the components but can also solve other goals that are inductive with only one constructor (such as `unit`). This means that the bulleted list of proofs that come may no longer reflect the logical structure of the goal as iterated product (since `unit` may only be the type thanks to an instantiation of the treated abstract goal).

UniMath/CategoryTheory/BicatOfCatsElementary.v
rmatthes: Good point. Currently, `vcomp2_CAT` is not used, but it should better have the same number of implicit arguments as `vcomp2`. I changed this in my local copy.

UniMath/CategoryTheory/BicatOfCatsElementary.v
rmatthes: I have the comments now in my local copy. I'll move the file to `categories` in a second commit.

benediktahrens:
> For `SetHITs`, the error is:
> 
> ```
> File "./UniMath/Combinatorics/Lists.v", line 223, characters 4-26:
> [614](https://github.com/UniMath/UniMath/actions/runs/4298031600/jobs/7491700247#step:6:620)
>   Error:
> [615](https://github.com/UniMath/UniMath/actions/runs/4298031600/jobs/7491700247#step:6:621)
>   Found no subterm matching "foldr1_map ?M77 ?M78 ?M79 nil" in the current goal.
> ```
> 
> The Coq development version was added to the CI, when the file on Left Kan Extensions led to a problem with it. This was also caused by a rewrite, which worked fine on my version of Coq, while on the development version it did not work. I don't have a general solution for it, but in this case, the lines that cause problems with the development version, could basically use `apply` instead of `rewrite` (after using `etrans` and `maponpaths` so that it is used in the right location of the term).

This might be an issue similar to #1614, a regression in Coq dev compared to Coq 8.16. It would be good to let @SkySkimmer know about this, possibly by filing a bug report against Coq with a pointer to this PR.

UniMath/Bicategories/MonoidalCategories/MonadsAsMonoidsWhiskered.v
rmatthes: I agree in general. In my local copy, there are now types for all `Let` of this file.

SkySkimmer:
Can you try on https://github.com/coq/coq/pull/17294 ?

UniMath/Bicategories/MonoidalCategories/ActionOfEndomorphismsInBicatWhiskered.v
rmatthes: right so, now in my local copy

UniMath/CategoryTheory/FunctorCoalgebras.v
rmatthes: ok

UniMath/CategoryTheory/limits/terminal.v
rmatthes: I now also make use of `abstract` in the "legacy" `Terminal_functor_precat`.

UniMath/CategoryTheory/Monoidal/ConstructionOfActegories.v
```diff
@@ -112,8 +116,8 @@ Defined.
 
 Lemma lifted_actegory_laws: actegory_laws Mon_W lifted_actegory_data.
 Proof.
-  split; [| split; [| split]]. (* splits into the 4 main goals *)
-  - repeat split.
+  split4. (* splits into the 4 main goals *)
```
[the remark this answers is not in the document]
m-lindgren: How about having a `make_actegory_laws` that takes the four main goals as arguments (and similar functions in other situations), to avoid some of these `splitxx` and `repeat split`?

UniMath/CategoryTheory/limits/binproducts.v
rmatthes: I did not see commutativity of binary products in the library - did I overlook it? I agree it would be better to work from first principles. The commutativity for the binary product of functors was only needed for a side question, therefore I did not treat it with due respect.

UniMath/CategoryTheory/limits/binproducts.v
rmatthes: As far as I remember, the `is_iso` versions had been there already, and I did not want to break anything downstream (maybe in the satellites).

UniMath/CategoryTheory/limits/binproducts.v
rmatthes: see above for `unit_l`

UniMath/CategoryTheory/Monoidal/ConstructionOfActegories.v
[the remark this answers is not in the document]
nmvdw: That is better!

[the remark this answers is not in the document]
nmvdw: Note: this approach would not be convenient in all situations. The tactic `use` goes up to 15 arguments (https://github.com/UniMath/UniMath/blob/master/UniMath/Foundations/Init.v#L188), whereas bicategories have 21 laws (https://github.com/UniMath/UniMath/blob/master/UniMath/Bicategories/Core/Bicat.v). However, this is not a problem for actegories, so there it is suitable to use this approach.

UniMath/CategoryTheory/limits/binproducts.v
rmatthes: I only played it back to the pointwise situation. Sorry for this laziness. And there is a small difference that is maybe really ugly: I prefer the constantly terminal functor over `Terminal_functor_precat`. But I take the opportunity to change this completely.

UniMath/CategoryTheory/Monoidal/CategoriesOfMonoidsWhiskered.v
nmvdw: I see. That is indeed that case if you have `Context {C : category} (M : monoidal C).` However, in a definition like `monoidal_cat`, where everything is put together, it is not a problem. It is something worth to ponder about, but then not relevant for this PR.

UniMath/CategoryTheory/Monoidal/ConstructionOfActegories.v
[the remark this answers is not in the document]
rmatthes: Cf. my answer to your comment on the definition of those `splitn` tactics. `split5` does not even apply here since the splitting does not go into the side branches.

UniMath/CategoryTheory/Monoidal/CoproductsInActegories.v
```diff
@@ -0,0 +1,401 @@
...
+Section FixAMonoidalCategory.
+
+  Context {V : category} (Mon_V : monoidal V). (** given the monoidal category that acts upon categories *)
+
+
+Section BinaryCoproduct.
+
+  Context {C : category} (BCP : BinCoproducts C) (Act : actegory Mon_V C).
+
+  Definition actegory_bincoprod_antidistributor := bifunctor_bincoprod_antidistributor BCP BCP Act.
```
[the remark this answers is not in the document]
rmatthes: Okay, this is long but useful.

UniMath/CategoryTheory/Monoidal/CoproductsInActegories.v
```diff
@@ -0,0 +1,401 @@
...
+
+  Lemma actegory_bincoprod_antidistributor_nat_left (v : V) (cd1 cd2 : category_binproduct C C) (g : category_binproduct C C ⟦ cd1, cd2 ⟧) :
```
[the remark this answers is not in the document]
rmatthes: Thanks, I now do so in the entire file.

UniMath/CategoryTheory/Monoidal/CoproductsInActegories.v
```diff
@@ -0,0 +1,401 @@
...
+    red; repeat split.
+    - red.
+      intros v cd1 cd2 g.
+      apply (z_iso_inv_to_right _ _ _ _ (_,,bincoprod_functor_lineator_strongly δ v _)).
+      rewrite assoc'.
+      apply (z_iso_inv_to_left _ _ _ (_,,bincoprod_functor_lineator_strongly δ v _)).
+      apply bincoprod_antidistributor_nat_left.
+    - red.
+      intros v1 v2 cd f.
+      apply (z_iso_inv_to_right _ _ _ _ (_,,bincoprod_functor_lineator_strongly δ _ _)).
```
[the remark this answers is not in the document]
rmatthes: we encounter the recurrent `Illegal application (Non-functional construction)`

UniMath/CategoryTheory/Monoidal/Examples/ActionOfEndomorphismsInCATWhiskeredElementary.v
```diff
@@ -0,0 +1,274 @@
...
+
+  (*
+  Import BifunctorNotations.
+  Import MonoidalNotations.
+  Import ActegoryNotations.
+   *)
+
+  Context (C D : category).
+
+Local Definition Mon_endo: monoidal [C, C] := monendocat_monoidal C.
```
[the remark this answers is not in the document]
rmatthes: I am not sure if that helps so much if `C` still has to be given.

UniMath/CategoryTheory/Monoidal/Examples/ActionOfEndomorphismsInCATWhiskeredElementary.v
```diff
@@ -0,0 +1,274 @@
...
+
+Definition action_from_precomp_CAT_data : bifunctor_data [C, C] [C, D] [C, D].
+Proof.
+  use make_bifunctor_data.
+  - intros v f. exact (functor_composite v f).
+  - intros v f1 f2 β. exact (lwhisker_CAT v β).
+  - intros f v1 v2 α. exact (rwhisker_CAT f α).
+Defined.
+
+(* (** we explicitly do not opacify the following definition: *) *)
```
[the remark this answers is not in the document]
rmatthes: This is why the coqdoc comment is again commented. If there was some spare time, I would still like to understand if there is a solution to that nasty problem that the lifted lineators cannot be transferred between actions of endomorphisms on endomorphisms (by instantiating to the same category) and the self-action of the endomorphisms.

UniMath/CategoryTheory/Monoidal/Examples/ActionOfEndomorphismsInCATWhiskeredElementary.v
```diff
@@ -0,0 +1,274 @@
...
+Section TheHomogeneousCase.
+
+  Context (C : category).
+
+(* (** requires [action_from_precomp_CAT] with known proofs of the laws to be even convertibility *) *)
+Definition action_in_actegory_from_precomp_CAT_as_self_action :
+  actegory_action (Mon_endo C) (actegory_from_precomp_CAT C C) =
+    actegory_action (Mon_endo C) (actegory_with_canonical_self_action (Mon_endo C)).
+Proof.
+  use total2_paths_f.
```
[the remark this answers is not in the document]
rmatthes: Okay.

UniMath/CategoryTheory/Monoidal/Examples/ActionOfEndomorphismsInCATWhiskeredElementary.v
```diff
@@ -0,0 +1,274 @@
...
+  Let BCPCD : BinCoproducts [C, D] := BinCoproducts_functor_precat C D BCP.
+
+  Definition actegory_from_precomp_CAT_bincoprod_distributor_data :
+    actegory_bincoprod_distributor_data (Mon_endo C) BCPCD (actegory_from_precomp_CAT C D).
+  Proof.
+    intro F.
+    apply precomp_bincoprod_distributor_data.
+  Defined.
+
+  Goal ∏ F G1 G2 c, pr1 (actegory_from_precomp_CAT_bincoprod_distributor_data F G1 G2) c = identity _.
```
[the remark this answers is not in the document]
rmatthes: Okay.

UniMath/CategoryTheory/Monoidal/Examples/MonadsAsMonoidsWhiskeredElementary.v
```diff
@@ -0,0 +1,94 @@
...
+Require Import UniMath.CategoryTheory.Monoidal.MonoidalCategoriesWhiskered.
+
+
+Require Import UniMath.CategoryTheory.Monoidal.CategoriesOfMonoidsWhiskered.
+Require Import UniMath.CategoryTheory.Monoidal.Examples.EndofunctorsWhiskeredMonoidalElementary.
+Require Import UniMath.CategoryTheory.Monads.Monads.
+
+Local Open Scope cat.
+
+Section A.
```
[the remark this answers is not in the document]
rmatthes: Okay for the renaming. It will be an equivalence if it is in the abstract situation, but it might be easier to prove here.

UniMath/CategoryTheory/Monoidal/Examples/MonoidalPointedObjects.v
```diff
@@ -446,28 +467,29 @@ Section PointedObjectFixpointMonoidal.
               ]).
   Defined.
 
+
   Definition ptdob_to_ptdptdob_fmonoidal_stronglaws
     : fmonoidal_stronglaws (fmonoidal_preservestensordata ptdob_to_ptdptdob_fmonoidal_lax)
                            (fmonoidal_preservesunit ptdob_to_ptdptdob_fmonoidal_lax).
   Proof.
```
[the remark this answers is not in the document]
rmatthes: For the time being, I marked it as TODO in the code.

UniMath/CategoryTheory/Monoidal/Examples/SelfActionInCATWhiskeredElementary.v
```diff
@@ -0,0 +1,194 @@
...
+      apply functor_id.
+    - intros ?; intros.
+      apply (nat_trans_eq C).
+      intro c.
+      cbn.
+      rewrite id_left.
+      apply functor_id.
+  Qed.
+
+  (** the following definition may not be usable because of its tigght typing *)
```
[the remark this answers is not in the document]
rmatthes: Okay.

UniMath/CategoryTheory/Monoidal/ProductActegory.v
```diff
@@ -0,0 +1,419 @@
+(** binary and I-indexed product of actegories w.r.t. the same acting monoidal category
```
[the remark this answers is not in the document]
rmatthes: Well observed. This figures on an internal list of issues I made.

UniMath/CategoryTheory/coslicecat.v
rmatthes: Even if it would compile, @DanGrayson might not like it to lose duality again (at least, this was the purpose of that transparent form).

UniMath/CategoryTheory/Monoidal/ConstructionOfActegoryMorphisms.v
rmatthes: No, not sufficient.

UniMath/CategoryTheory/Monoidal/ConstructionOfActegoryMorphisms.v
rmatthes: Same as before: one is not enough.

m-lindgren:
> Can you try on [coq/coq#17294](https://github.com/coq/coq/pull/17294) ?

@SkySkimmer I compiled part of this PR (`make Combinatorics`) with the version of coq you linked, and it compiled fine.

I also verified again that `make Combinatorics` fails with current master of coq, and it does:
```
COQC UniMath/Combinatorics/Lists.v
File "./UniMath/Combinatorics/Lists.v", line 223, characters 4-26:
Error: Found no subterm matching "foldr1_map ?M77 ?M78 ?M79 nil" in the current goal.

make[1]: *** [build/CoqMakefile.make:816: UniMath/Combinatorics/Lists.vo] Fel 1
```

Note, the version of this PR that I tested is: 28136081130b0aa607115e86f1527d9754103c89

UniMath/CategoryTheory/Monoidal/ConstructionOfActegories.v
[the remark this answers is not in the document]
rmatthes: > Note: this approach would not be convenient in all situations. The tactic `use` goes up to 15 arguments (https://github.com/UniMath/UniMath/blob/master/UniMath/Foundations/Init.v#L188), whereas bicategories have 21 laws (https://github.com/UniMath/UniMath/blob/master/UniMath/Bicategories/Core/Bicat.v). However, this is not a problem for actegories, so there it is suitable to use this approach.

This is why I proposed split tactics up to `split21`.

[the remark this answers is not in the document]
rmatthes: > How about having a `make_actegory_laws` that takes the four main goals as arguments (and similar functions in other situations), to avoid some of these `splitxx` and `repeat split`?

I find the dedicated `makexxx` helpful for dependent sums, but to my mind, this would be too much boilerplate for simple n-ary conjunctions of propositions.

UniMath/CategoryTheory/coslicecat.v
rmatthes: I checked that the main UniMath repository is unaffected by putting `Qed`.

UniMath/CategoryTheory/Monoidal/ConstructionOfActegoryMorphisms.v
```diff
@@ -0,0 +1,1212 @@
...
+          {W : category} (Mon_W : monoidal W)
+          {F : W ⟶ V} (U : fmonoidal Mon_W Mon_V F).
+
+
+Section LiftedLaxLineator.
+
+  Context {C D : category} (ActC : actegory Mon_V C) (ActD : actegory Mon_V D) {H : functor C D}
+    (ll : lineator_lax Mon_V ActC ActD H).
+
+  Definition lifted_lax_lineator_data : lineator_data Mon_W (lifted_actegory Mon_V ActC Mon_W U)
```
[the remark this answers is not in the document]
rmatthes: See [new commit](https://github.com/UniMath/UniMath/pull/1633/commits/9b94877c997a633a5447680e6cd74487af2a3bf5)

UniMath/Bicategories/PseudoFunctors/Examples/LiftingActegories.v
```diff
@@ -0,0 +1,82 @@
...
+                | apply actbicat_disp_locally_groupoid]).
+  Defined.
+
+  Lemma lifting_actegories_disp_psfunctor_laws :
+    is_disp_psfunctor dBV dBW (id_psfunctor _) lifting_actegories_disp_psfunctor_data.
+  Proof.
+    split7; red; intros; apply actbicat_disp_2cells_isaprop.
+  Qed.
+
+  Definition lifting_actegories_disp_psfunctor : disp_psfunctor dBV dBW (id_psfunctor _)
```
nmvdw, line 76: Cool! For this definition, it might be more convenient to use https://github.com/UniMath/UniMath/blob/master/UniMath/Bicategories/DisplayedBicats/DispBuilders.v#L30.

UniMath/CategoryTheory/limits/binproducts.v
nmvdw: It does indeed not seem to be in the library. For the current PR, I don't think it's necessary to refactor it, but it can be left as future work.

UniMath/SubstitutionSystems/ActionScenarioForGenMendlerIteration_alt.v
```diff
@@ -0,0 +1,273 @@
...
+    apply is_omega_cocont_BinCoproduct_of_functors.
+    - apply is_omega_cocont_constant_functor.
+    - exact HH.
+  Qed.
+
+  Let μF_Initial : Initial AF := colimAlgInitial IC HF (CC chnF).
+  Let μF : C := alg_carrier _ (InitialObject μF_Initial).
+  Let α : Const_plus_H c0 μF --> μF := alg_map (Const_plus_H c0) (pr1 μF_Initial).
+
+  Let η : C⟦constant_functor C C c0 μF, μF⟧ := BinCoproductIn1 (CP _ _) · α.
```
nmvdw, line 84: Is the type of `\eta` equal to `c0 --> \muF`? The reason why you wrote the type in this way, is because it functions as some kind of algebra map, right?

UniMath/SubstitutionSystems/ActionScenarioForGenMendlerIteration_alt.v
```diff
@@ -0,0 +1,273 @@
...
+                                     BinCoproductArrow (CP _ _) f (θ v μF · #H h · ρ).
+
+  Context (δ : actegory_bincoprod_distributor Mon_V CP Act).
+
+  Definition instance_SpecialGenMendlerIterationWithActegory (h : C ⟦v ⊗_{Act} μF, y⟧) : UU :=
+    v ⊗^{Act}_{l} α · h = δ v c0 (H μF) · BinCoproductOfArrows _ (CP _ _) (CP _ _) f (θ v μF) ·
+                            #(Const_plus_H y) h · BinCoproductArrow (CP _ _) (identity y) ρ.
+
+  Local Lemma charpropequivsingleeq (h : C ⟦v ⊗_{Act} μF, y⟧) :
+    charprop_SpecialGenMendlerIterationWithActegoryAndStrength h <-> singleeq_SpecialGenMendlerIterationWithActegoryAndStrength h.
```
nmvdw, line 116: Would `\simeq` (using `weqimplimpl` in the proof) be better here? What does @benediktahrens think?

UniMath/SubstitutionSystems/BindingSigToMonad_actegorical.v
```diff
@@ -0,0 +1,424 @@
...
+        + intro; intros. apply (nat_trans_eq C).
+          intro c'.
+          cbn.
+          repeat rewrite id_right.
+          apply idpath.
+        + intro; intros. apply (nat_trans_eq C).
+          intro c'.
+          apply idpath.
+    Defined.
+    (* should be better opacified *)
```
nmvdw, line 159: Each of the goals after `split4` can be put in an `abstract`

nmvdw (commented):
I have gone through 109 out of the 132 files now. So, I think that a review of this PR should be ready within one or two days the most.

nmvdw:
The error is:
```
error: *** UniMath/Bicategories/PseudoFunctors/Examples/LiftingActegories.vo should not require UniMath/Bicategories/DisplayedBicats/DispBuilders.vo
```
`LiftingActegories.v` should be put later in the file list (after `DispBuilders.v`).

UniMath/SubstitutionSystems/EquivalenceSignaturesWithActegoryMorphisms.v
```diff
@@ -422,4 +456,57 @@ End AA.
...
  Defined.
 
+ Lemma weqSignatureLaxMorphismActegories_alt :
+   Signature C D D' ≃ hom(C:=actbicat Mon_ptdendo) (hom(C:=bicat_of_cats) C D',,actegoryPtdEndosOnFunctors D') (hom(C:=bicat_of_cats) C D,,actegoryPtdEndosOnFunctors D).
+ Proof.
+   apply (weqcomp weqSignatureLaxMorphismActegories).
+   apply weqfibtototal.
+   intro H.
+   cbn.
+   apply idweq. (* very slow *)
```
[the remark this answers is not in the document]
rmatthes: The gain is only fractions of a second on my machine. But I'll make this small change.

UniMath/SubstitutionSystems/ActionScenarioForGenMendlerIteration_alt.v
rmatthes: I do not see any longer the advantage of displaying the hom-set. I changed all occurrences in the file (locally).

UniMath/SubstitutionSystems/BindingSigToMonad_actegorical.v
rmatthes: I put the entire split4 construction into `abstract`.

rmatthes:
> The error is:
> 
> ```
> error: *** UniMath/Bicategories/PseudoFunctors/Examples/LiftingActegories.vo should not require UniMath/Bicategories/DisplayedBicats/DispBuilders.vo
> ```
> 
> `LiftingActegories.v` should be put later in the file list (after `DispBuilders.v`).

Every time I check the dependencies, I have to remove the All.v files in order to run Dune again, so I do not do this systematically. Sorry.

UniMath/SubstitutionSystems/ContinuitySignature/InstantiateHSET.v
```diff
@@ -0,0 +1,218 @@
...
+
+Section OmegaLimitsCommutingWithCoproductsHSET.
+
+  Definition HSET_ω_limits : Lims_of_shape conat_graph HSET.
+  Proof.
+    intro coch.
+    apply LimConeHSET.
+  Defined.
+
+  Lemma test {I : HSET} (ind : pr1 I → cochain SET)
```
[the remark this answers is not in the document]
rmatthes: I applied your suggestion, but @Kfwullaert should agree with it.

UniMath/SubstitutionSystems/ContinuitySignature/GeneralLemmas.v
```diff
@@ -0,0 +1,197 @@
...
+           (I : UU) (C : category) (a : I -> C) (CC : Coproduct I C a)
+           (c : I -> C) (CC' : Coproduct I C c) (f : ∏ i : I, C⟦a i, c i⟧)
+  : (∏ i : I, is_z_isomorphism (f i)) -> is_z_isomorphism (CoproductOfArrows I C CC CC' f).
+Proof.
+  intro fi_iso.
+  use make_is_z_isomorphism.
+  - use CoproductOfArrows.
+    exact (λ i, pr1 (fi_iso i)).
+  - split.
+    + etrans. { apply precompWithCoproductArrow. }
```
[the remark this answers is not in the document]
rmatthes: ok

UniMath/SubstitutionSystems/ContinuitySignature/GeneralLemmas.v
```diff
@@ -0,0 +1,197 @@
...
+Lemma coproduct_of_functors_nat_trans
+      {I : UU} {C D : category} (CP : Coproducts I D) {F G : I → C ⟶ D}
+      (α : ∏ i : I, nat_trans (F i) (G i))
+  : nat_trans (coproduct_of_functors I C D CP F) (coproduct_of_functors I C D CP G).
+Proof.
+  use make_nat_trans.
+  - intro c.
+    use CoproductOfArrows.
+    exact (λ i, α i c).
+  - intros c1 c2 f.
```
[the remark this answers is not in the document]
rmatthes: ok

UniMath/SubstitutionSystems/ContinuitySignature/GeneralLemmas.v
```diff
@@ -0,0 +1,197 @@
...
+        etrans.
+        1: apply maponpaths, pathsinv0, (pr21 (nat_z_iso_inv i)).
+        etrans.
+        1: apply assoc.
+        apply maponpaths_2, (pr2 f v).
+      }
+
+      exact (base_paths _ _ (pr2 t c')).
+    + apply (impred_isaprop) ; intro ; apply homset_property.
+Defined.
```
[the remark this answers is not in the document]
rmatthes: in principle, there is computational content in $\omega$-continuity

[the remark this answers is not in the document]
benediktahrens: Limits and preservation of limits contain an existence part and a uniqueness part. The former should be transparent, the latter opaque.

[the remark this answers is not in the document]
rmatthes: To be concrete: every result concluding some continuity has to be defined, but internally organized not to expose the uniqueness proofs. Although I do not know where we profit from those existence proofs. Is there an example in UniMath where this exposure is used?

rmatthes:
Does someone know why there is no more CI triggered on this PR?

UniMath/SubstitutionSystems/SimplifiedHSS/Lam.v
```diff
@@ -0,0 +1,544 @@
...
+    + abstract
+        (intro c; rewrite id_right ;
+         apply BinCoproductIn1Commutes_right_dir;
+         apply idpath) .
+Defined.
+
+(** this iso does nothing, but is needed to make the argument to [fbracket] below well-typed *)
+(* maybe a better definition somewhere above could make this iso superfluous *)
+(* maybe don't need iso, but only morphism *)
+Local Definition bla : iso (ptd_from_alg_functor CC LamE_S LamE_algebra_on_Lam) (ptd_from_alg_functor CC _ Lam).
```
nmvdw, line 194: Because it is a local definition `bla` (and `bla'` will be in the namespace afterwards). A renaming might be good here.

benediktahrens:
@rmatthes : your latest commit to this branch triggered CI to be launched. Very mysterious...

UniMath/SubstitutionSystems/ContinuitySignature/InstantiateHSET.v
```diff
@@ -0,0 +1,218 @@
...
+    : propcoproducts_commute_binproducts HSET BinProductsHSET (λ p, CoproductsHSET p (isasetaprop (pr2 p))).
+  Proof.
+    intros p x y.
+    use make_is_z_isomorphism.
+    - intros [ix iy].
+      exists (pr1 ix).
+      exact (pr2 ix,,pr2 iy).
+    - split.
+      + apply funextsec ; intro ixy.
+        use total2_paths_f ; apply idpath.
```
nmvdw, line 57: I think `apply idpath` might have worked here, so the `use total2_paths_f` is not needed.

UniMath/SubstitutionSystems/ContinuitySignature/InstantiateHSET.v
```diff
@@ -0,0 +1,218 @@
...
+      + apply funextsec ; intros [ix iy].
+        use total2_paths_f.
+        * apply idpath.
+        * use total2_paths_f.
+          -- apply (pr2 p).
+          -- cbn.
+             induction (pr1 (pr2 p (pr1 ix) (pr1 iy))).
+             rewrite idpath_transportf.
```
nmvdw, line 65: `idpath_transportf` should hold by `idpath`, so I think that on this line, `apply idpath` could have been used.

rmatthes:
Maybe the CI is stalled if there were too many canceled checks - due to new commits to the branch on which the PR is based.

UniMath/SubstitutionSystems/EquivalenceSignaturesWithActegoryMorphisms.v
```diff
@@ -422,4 +456,57 @@ End AA.
    apply weqABStrengthLaxMorphismActegories.
  Defined.
 
+ Lemma weqSignatureLaxMorphismActegories_alt :
+   Signature C D D' ≃ hom(C:=actbicat Mon_ptdendo) (hom(C:=bicat_of_cats) C D',,actegoryPtdEndosOnFunctors D') (hom(C:=bicat_of_cats) C D,,actegoryPtdEndosOnFunctors D).
+ Proof.
+   apply (weqcomp weqSignatureLaxMorphismActegories).
+   apply weqfibtototal.
+   intro H.
+   cbn.
```
nmvdw, line 465: If you remove the `cbn`, does that help?

UniMath/SubstitutionSystems/MultiSortedMonadConstruction_actegorical.v
```diff
@@ -0,0 +1,285 @@
...
+(* Assumptions on [C] used to construct the functor *)
+(* Note that there is some redundancy in the assumptions *)
+Context (TC : Terminal C) (IC : Initial C)
+          (BP : BinProducts C) (BC : BinCoproducts C)
+          (PC : forall (I : UU), Products I C) (CC : forall (I : UU), isaset I → Coproducts I C).
+
+Local Notation "'1'" := (TerminalObject TC).
+Local Notation "a ⊕ b" := (BinCoproductObject (BC a b)).
+
+(** Define the discrete category of sorts *)
```
nmvdw, line 99: Now that `Hsort` is of hlevel 3, this category is not discrete anymore. So, I'd remove 'discrete' in this comment.

UniMath/SubstitutionSystems/ConstructionOfGHSS.v
```diff
@@ -0,0 +1,489 @@
...
+      etrans.
+      2: { apply pathsinv0, precompWithBinCoproductArrow. }
+      apply maponpaths_12.
+      + cbn. rewrite id_left. exact Hyp2.
+      + cbn. exact Hyp1.
+    - intros [h Hyp].
+      use total2_paths_f.
+      + apply idpath.
+      + apply isapropdirprod; apply V.
+  Defined.
```
nmvdw, line 332: Would `Qed` be more appropriate here?

nmvdw (commented):
I'm at 129/139

UniMath/SubstitutionSystems/SimplifiedHSS/BindingSigToMonad.v
```diff
@@ -0,0 +1,351 @@
...
+Definition SignatureToHSSisInitial
+  (IC : Initial C) (CLC : Colims_of_shape nat_graph C)
+  (H : Presignature C C C) (Hs : is_omega_cocont H) :
+  isInitial _ (SignatureToHSS IC CLC H Hs).
+Proof.
+now unfold SignatureToHSS; destruct InitialHSS.
+Qed.
+
+(* Redefine this here so that it uses the arguments above *)
+Let Monad_from_hss (H : Signature C C C) : HSS H → Monad C.
```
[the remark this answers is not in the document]
rmatthes: I think it was only a convenience abbreviation not worthy being used elsewhere. The entire file is rather a summary than the code to work with.

UniMath/SubstitutionSystems/SimplifiedHSS/Lam.v
rmatthes: Was there for >7 years in the file from where this is copied, but I now changed it - and this is better (in both files).

UniMath/SubstitutionSystems/ContinuitySignature/InstantiateHSET.v
rmatthes: true

UniMath/SubstitutionSystems/EquivalenceSignaturesWithActegoryMorphisms.v
rmatthes: Same time consumption (~7s for the slow lines each), therefore `cbn` removed

UniMath/SubstitutionSystems/MultiSortedMonadConstruction_actegorical.v
rmatthes: ok - done in four files (the assumption evolved)

UniMath/SubstitutionSystems/ContinuitySignature/InstantiateHSET.v
rmatthes: ok

UniMath/Bicategories/MonoidalCategories/MonadsAsMonoidsWhiskered.v
rmatthes: I created the correspondence both ways, including for morphisms. But I leave the rest for later.

UniMath/CategoryTheory/Monoidal/Examples/EndofunctorsWhiskeredMonoidalElementary.v
```diff
@@ -0,0 +1,110 @@
...
+    - intros a b. exact (functor_compose a b).
+    - intros a b1 b2 β. exact (lwhisker_CAT _ β).
+    - intros b a1 a2 α. exact (rwhisker_CAT _ α).
+  Defined.
+
+  (* (** we explicitly do not opacify the following definition: *) *)
+  Definition monendocat_tensor_laws : is_bifunctor monendocat_tensor_data.
+  Proof.
+    split5.
+    - intros ?; apply lwhisker_id2_CAT.
```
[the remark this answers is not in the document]
rmatthes: ok

UniMath/CategoryTheory/Monoidal/ProductActegory.v
rmatthes: While I see your point, it is not this PR that introduced the use of `power`. It would be good to open an issue for that and deal with it separately - so that others not interested in the present PR could have their say.

nmvdw: That's a valid point.

UniMath/SubstitutionSystems/ContinuitySignature/MultiSortedSignatureFunctorEquivalence.v
```diff
@@ -0,0 +1,803 @@
...
+Require Import UniMath.CategoryTheory.limits.Preservation.
+
+Local Open Scope cat.
+
+Section B.
+
+  Lemma make_isBinProduct'
+        {C : category} {x y : C}
+        (Pxy : BinProduct _ x y)
+    : ∏ z : C, ∏ (zx : C⟦z,x⟧) (zy : C⟦z,y⟧),
```
nmvdw, line 51:
```suggestion
    : ∏ (z : C) (zx : C⟦z,x⟧) (zy : C⟦z,y⟧),
```

UniMath/SubstitutionSystems/MultiSortedMonadConstruction_coind_actegorical.v
```diff
@@ -0,0 +1,512 @@
...
+        {C : category} {I : UU}
+        {i : I -> C} {x : C}
+        (j : ∏ i0 : I, C ⟦ i i0, x ⟧)
+    : isaprop (isCoproduct I C i x j).
+  Proof.
+    repeat (apply impred; intro).
+    apply isapropiscontr.
+  Qed.
+
+  Definition z_iso_from_coproduct_to_coproduct
```
nmvdw, line 108: The `Isos.` can be removed if `CategoryTheory.Core.Isos` is imported.

nmvdw (approved):
I think the code looks good and is ready to be merged. The only comments that I have, are very minor.

benediktahrens:
At some point, this PR exposed a regression in Coq, between 8.16 and current dev version. This does not seem to be the case any more, that is, the code that exhibited the regression is gone from the PR. Why is that?

rmatthes:
All commits in March come from me, and I never tried to address the problems with Coq dev. And all CI runs with Coq dev are still failing.

m-lindgren:
> At some point, this PR exposed a regression in Coq, between 8.16 and current dev version. This does not seem to be the case any more, that is, the code that exhibited the regression is gone from the PR. Why is that?

It's still there: https://github.com/UniMath/UniMath/actions/runs/4331261734/jobs/7562982947#step:5:2008

benediktahrens:
Thanks, @rmatthes and @m-lindgren. I was confused...

m-lindgren:
We should probably revert this, so we don't upset the Coq developers. :-)

rmatthes:
@m-lindgren Please see my PR #1639 concerning the compilation error you referred to above.

PR #1636, opened 2023-03-02
Dagger category theory

Kfwullaert:
This PR is the start of the formalization of theory of dagger categories.
The following material is currently provided:
- Definitions of dagger categories, dagger functors, dagger adjoint, dagger isos and univalence
- Definitions of weak and strong dagger equivalences between dagger categories (no results, only definitions)
- A proof that any fully faithful functor into, a dagger category, 'creates' a unique dagger structure on the domain such that the functor is a dagger functor
- A construction of the functor-dagger category and a proof that it is dagger-univalent if the codomain is dagger-univalent.

Furthermore, two examples of dagger categories together with a univalence result:
1. The category of relations: This category is introduced, it is shown that this is both univalent and dagger univalent.
2. Any groupoid becomes a dagger category and a proof that it is dagger univalent if and only if it is univalent.


m-lindgren (commented):
Looks like good material. I'll look at it more closely tomorrow. 

Maybe the `Core` folder isn't necessary? The imports get very long currently. What do you think about having the files currently in `Core` directly in `DaggerCategories` instead, and renaming them `Categories.v`, `Functors.v` etc?

UniMath/CategoryTheory/DisplayedCats/Examples/Relations.v
```diff
@@ -0,0 +1,457 @@
...
+Section Isos.
+
+  Lemma inverse_swap_relation
+             {X Y : hSet} {r : hrel X Y}
+             (i : is_z_isomorphism (C := REL) r)
+    : ∏ (x : X) (y : Y), r x y -> pr1 i y x.
+  Proof.
+    intros x y p.
+
+    set (t := toforallpaths _ _ _ (toforallpaths _ _ _ (pr12 i) x) x).
```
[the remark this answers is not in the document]
Kfwullaert: Thanks a lot for this, this will make the code a lot nicer.

UniMath/CategoryTheory/DisplayedCats/Examples/Relations.v
```diff
@@ -0,0 +1,457 @@
...
+Require Import UniMath.CategoryTheory.Core.Isos.
+Require Import UniMath.CategoryTheory.Core.Univalence.
+
+Require Import UniMath.CategoryTheory.categories.HSET.All.
+
+Local Open Scope cat.
+
+Section Relations.
+
+  Definition hrel (X Y : hSet) : UU := X -> Y -> hProp.
```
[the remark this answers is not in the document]
Kfwullaert: I have missed that for some reason. If I have a proposal for a better name, I will post it here before making any changes.

UniMath/CategoryTheory/DisplayedCats/Examples/Relations.v
```diff
@@ -0,0 +1,457 @@
+(* The category REL of (homotopy) relations:
```
nmvdw, line 1: Why is this file in this folder? It doesn't seem to be a displayed category, or am I missing something?

UniMath/CategoryTheory/DaggerCategories/Examples/groupoids.v
```diff
@@ -0,0 +1,118 @@
+(* Any groupoid becomes a †-category by defining f^† := f^-1 *)
```
nmvdw, line 1: Can you call this file Groupoids.v (with a capital)? The same for `relations.v` in DaggerCategories.

nmvdw (commented):
(Only some small comments on file names and locations)

nmvdw:
> > Looks like good material. I'll look at it more closely tomorrow.
> > Maybe the `Core` folder isn't necessary? The imports get very long currently. What do you think about having the files currently in `Core` directly in `DaggerCategories` instead, and renaming them `Categories.v`, `Functors.v` etc?
> 
> This is now done.

You missed CONTENTS.md

UniMath/CategoryTheory/DaggerCategories/FunctorCategory.v
```diff
@@ -0,0 +1,481 @@
...
+  + apply uip.
+    do 2 (apply impred_isaset ; intro).
+    apply funspace_isaset.
+    apply C'.
+Defined.
+
+Local Lemma equality_of_composition
+      {C : category} {x y z : C} (f1 f2 : C⟦x,y⟧) (g1 g2 : C⟦y,z⟧)
+  : f1 = f2 -> g1 = g2 -> f1 · g1 = f2 · g2.
+Proof.
```
m-lindgren, line 58: I'm surprised we don't have this in the library already, at least specific to composition of morphisms in a category. 

The general case would be `maponpaths_12` I think, and for composition it would be `maponpaths_12 compose`. In fact I suspect there are many instances of `etrans; apply maponpaths_12` that could/should be replaced with this.

But this seems like a useful function to have since it's the general case of `cancel_postcomposition` and such, and I would be strongly in favor of upstreaming this.

UniMath/CategoryTheory/DaggerCategories/Examples/Groupoids.v
```diff
@@ -0,0 +1,118 @@
...
+
+Require Import UniMath.CategoryTheory.DaggerCategories.Categories.
+Require Import UniMath.CategoryTheory.DaggerCategories.Unitary.
+Require Import UniMath.CategoryTheory.DaggerCategories.Univalence.
+
+Local Open Scope cat.
+
+Local Lemma z_iso_inv_eq {C : category} {x y : C}
+      (f : C⟦x,y⟧) (g h : C⟦y,x⟧)
+  : is_inverse_in_precat f g -> is_inverse_in_precat f h -> g = h.
```
m-lindgren, line 21: I think this already exists. `inverse_unique_precat` seems to have the correct type.

m-lindgren (changes requested):
Not quite done with the review yet, but only some minor comments so far.

UniMath/CategoryTheory/DaggerCategories/Examples/Groupoids.v
```diff
@@ -0,0 +1,118 @@
...
+    repeat split ; intro ; intros ; use z_iso_inv_eq.
+    - exact (identity x).
+    - apply (pr2 C).
+    - apply is_inverse_in_precat_identity.
+    - exact (f · g).
+    - apply (pr2 C).
+    - apply is_inverse_in_precat_comp ; apply (pr2 C).
+    - apply (pr2 C x y f).
+    - apply (pr2 C).
+    - apply is_inverse_in_precat_inv ; apply (pr2 C).
```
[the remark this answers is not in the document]
benediktahrens: Yes, I strongly second this suggestion.

UniMath/CategoryTheory/DaggerCategories/FunctorCategory.v
benediktahrens: Two or three years ago, @peterlefanulumsdaine replaced many uses of `cancel_postcomposition` by uses of `maponpaths_12`. I don’t have a strong opinion either way, but we should not spend effort switching between the general and the specialized lemma.

UniMath/CategoryTheory/DaggerCategories/Examples/Groupoids.v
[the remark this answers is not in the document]
Kfwullaert: There is already a lemma available: groupoid_is_pregroupoid.
I should have looked more at the code in groupoids.v.

UniMath/CategoryTheory/DaggerCategories/FunctorCategory.v
peterlefanulumsdaine: @m-lindgren : What was your reason for preferring this specific lemma over the general `mapopaths_12`?  I find the general lemmas almost always just as easy to apply (`maponpaths_12` for both arguments at once, or `maponpaths`/`maponpaths_2` for one argument at a time), so I don’t see the advantage of the specialised one.

Overall I have a mild preference for using the general lemma (I don’t object to other people use a specialised one, but I prefer to use the general myself unless there are compelling reasons otherwise) — but I have a very strong preference for keeping their names unified as some variant of `maponpaths`.  The point is that `maponpaths` and its specialisations are a pieces of completely general, mathematically near-trivial algebraic reasoning, that applies to all functions; so (a) that trivialness should be kept as clear as possible, (b) people shouldn’t have to learn a different set of lemma names for algebraic reasoning with different operations, and (c) non-unified names means people don’t discover it, so they re-invent it under different names.  When I replaced a lot of instances by `maponpaths_12` a few years ago, it was because there were about half a dozen different copies of this lemma under different names scattered through the library.  I don’t mind having one specialised version, but I’m strongly against having multiple ones.  So if we keep this, please could we rename it to `maponpaths_compose` or similar?

m-lindgren: @Kfwullaert Yes, that sounds good, and maybe write the proof using maponpaths_12 so that it's obvious that it is an intended specialization of the general lemma?

@benediktahrens Ah, interesting! In this case I think it fits in well with the current use of `cancel_postcomposition` etc and would be a good addition for new code entering CategoryTheory, but not something to spend time on changing old code.

m-lindgren: @peterlefanulumsdaine So for one, this lemma is opaque and `maponpaths_12` is not. Another reason is discoverability: This lemma is domain specific and expressed using the notation that is common when working with categories, and it's input is explicitly morphisms of a category, making it easy to find. I think of this as providing an API to working with/in categories, where the underlying structure and implementation is abstracted away and these helper functions provides a nice interface to use, which also provides somewhat more readable code.

m-lindgren: @peterlefanulumsdaine I have also seen duplications of the general lemma, before your cleanup, and I would of course strongly object to anyone adding more duplication like that. :-) But instances like this one is not duplication, it's specialization for the sake of readability, usability and discoverability, IMO.

UniMath/CategoryTheory/DaggerCategories/Univalence.v
```diff
@@ -0,0 +1,88 @@
...
+  Qed.
+
+End DaggerUnivalence.
+
+Section Lemmas.
+
+  Lemma idtodaggeriso_is_dagger_of_idtodaggeriso {C : category} (dagC : dagger C)
+        {x y : C} (p : x = y)
+    : pr1 (idtodaggeriso dagC x y p)
+      = dagC y x (pr1 (idtodaggeriso dagC y x (! p))).
```
m-lindgren, line 66: Maybe have a coercion so these (and similar) `pr1` can be removed? Or at least have an accessor function with a descriptive name.

I think in particular in type declarations one should avoid using `prX`, unless absolutely necessary.

UniMath/CategoryTheory/DaggerCategories/Univalence.v
```diff
@@ -0,0 +1,88 @@
...
+        (x : C)
+    : pr1 (idtodaggeriso dagC x x (idpath x))
+      = identity x.
+  Proof.
+    apply idpath.
+  Qed.
+
+  Lemma idtoiso_of_pathsinv_is_dagger_of_idtoiso {C : category} (dagC : dagger C)
+        {x y : C} (p : x = y)
+    : pr1 (Univalence.idtoiso (! p)) = dagC x y (pr1 (Univalence.idtoiso p)).
```
m-lindgren, line 82: Same as above, but in this case I think you can just turn the identity around so you have `dagC x y (Univalence.idtoiso p)) = ...`, and the coercions should trigger.

Also, maybe import Univalence?

UniMath/CategoryTheory/categories/Relations.v
```diff
@@ -0,0 +1,457 @@
...
+
+  Definition hrel (X Y : hSet) : UU := X -> Y -> hProp.
+  Identity Coercion idhrel : hrel >-> Funclass.
+
+  Local Definition eq_set {X : hSet} (x y : X)
+    : hProp.
+  Proof.
+    exists (x = y).
+    apply (pr2 X).
+  Defined.
```
m-lindgren, line 34: Just reminding my self that `hrel` needs another name, and just noticed that `eq_set` here is a duplicate of `eqset` defined in Foundations.Sets.

UniMath/CategoryTheory/Core/Categories.v
```diff
@@ -278,6 +278,13 @@ Proof.
   apply maponpaths.
 Defined.
 
+Lemma maponpaths_compose
+      {C : category} {x y z : C} (f1 f2 : C⟦x,y⟧) (g1 g2 : C⟦y,z⟧)
+  : f1 = f2 -> g1 = g2 -> f1 · g1 = f2 · g2.
+Proof.
```
m-lindgren, line 284: Actually what I liked about this function the most was the name you gave it before, it was very category theory-specific, which I think will promote readable code.

UniMath/CategoryTheory/DaggerCategories/FunctorCategory.v
```diff
@@ -0,0 +1,472 @@
...
+                       (toforallpaths (λ _ : pr11 C, pr11 D) F0 G0 gamma b)
+                       (F1 a b f).
+  Proof.
+    induction gamma.
+    apply idpath.
+  Qed.
+
+  Lemma double_transport_idtodaggeriso (a a' b b' : pr111 D)
+        (p : a = a') (q : b = b')  (f : a --> b) :
+    Univalence.double_transport p q f = pr1 (idtodaggeriso dagD _ _ (! p)) · f · pr1 (idtodaggeriso dagD _ _ q).
```
m-lindgren, line 351: Maybe replace pr1 with accessors or suitable coercion?

m-lindgren (changes requested):
Overall nice material, I learned things i didn't know about dagger categories reviewing this. I'm mostly concerned about finding a better name for `hrel`, the rest is style comments.

UniMath/CategoryTheory/DaggerCategories/Univalence.v
Kfwullaert: In this case, you mean something like idtomor (in the non-dagger/just categories case)?

UniMath/CategoryTheory/DaggerCategories/Univalence.v
Kfwullaert: That would indeed be nicer, not sure why I chose this form.
Regarding the import of Univalence, I should have mentioned this as a comment in this PR (and even in the file),
this time it was not that I forgot to include it. I choose to not include it because you are not interested in univalence, but in dagger univalence and therefore, I think it is better practice to not include it. I'm aware this not standard, so I can change this.
The (only) reason I have used Univalence.idtoiso is because this form allows me to use lemmas of about idtoiso that have are already been formalized. I could avoid this and redefine it for idtodaggeriso. However, a lot things had to defined yet again, very analogous to the non-dagger case. So I used it more as a 'trick', but strictly speaking, it is not necessary.

UniMath/CategoryTheory/DaggerCategories/Unitary.v
```diff
@@ -0,0 +1,130 @@
...
+      1: apply maponpaths, dagger_to_law_comp.
+      etrans.
+      1: apply assoc.
+      etrans.
+      1: apply maponpaths_2, assoc'.
+      etrans.
+      1: apply maponpaths_2, maponpaths, gg.
+      etrans.
+      1: apply maponpaths_2, id_right.
+      apply ff.
```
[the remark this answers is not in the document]
Kfwullaert: I could replace them by {} (instead of 1: ),
e.g. etrans. { apply  assoc. }

Kfwullaert:
@m-lindgren Regarding the name "hrel". I have no "good" idea, but I can only think of hrel_2. The 2 denotes that it is a relation between 2 sets. So, hrel, would be considered to be hrel_1 as it is a relation between 1 set.

benediktahrens:
> @m-lindgren Regarding the name "hrel". I have no "good" idea, but I can only think of hrel_2. The 2 denotes that it is a relation between 2 sets. So, hrel, would be considered to be hrel_1 as it is a relation between 1 set.

I am not seeing the context of this discussion, but a relation between two things is sometimes called a "binary relation". Maybe that could be used to find a suitable name?

Kfwullaert:
> > @m-lindgren Regarding the name "hrel". I have no "good" idea, but I can only think of hrel_2. The 2 denotes that it is a relation between 2 sets. So, hrel, would be considered to be hrel_1 as it is a relation between 1 set.
> 
> I am not seeing the context of this discussion, but a relation between two things is sometimes called a "binary relation". Maybe that could be used to find a suitable name?

@benediktahrens The context of this discussion is as follows:
Currently, there is already another definition of hrel. But this is given as the type of relations on X, i.e. hrel X := X -> X -> hProp.
But the definition of hrel that I've now provided, is hrel X Y := X -> Y -> hProp.
Thus, the problem currently is that I've introduced a type whose name already exists. Therefore, I have to change the name of the latter.

benediktahrens:
Mmh, possibly the type of the original `hrel` should be generalized - I don’t see the benefit of restricting to the special case of the two types being the same. We can discuss this at the UniMath User Meeting.

For the purpose of this PR, I suggest you use `bin_hrel` or similar.

Kfwullaert:
I think I have gone through every comment.
What I think currently remains (or at least of those parts that already have been reviewed), is the decision of the name for the above mentioned lemma (f1=f2 -> g1=g2 -> f1g1 =f2g2). Currently it is maponpaths_compose (as proposed by @peterlefanulumsdaine), but @m-lindgren prefers equality_of_composition.

rmatthes:
I find both names well justified. Not helping with the decision...

m-lindgren:
I'm fine with merging this as is, having dagger categories in the library will be nice. By the way, @Kfwullaert any chance we'll see monoidal  dagger categories as well?

Kfwullaert:
> I'm fine with merging this as is, having dagger categories in the library will be nice. By the way, @Kfwullaert any chance we'll see monoidal dagger categories as well?

Yes, but before starting, I wait for PR #1644 to be merged.

PR #1639, opened 2023-03-04
a small simplification that might make a Coq regression go away

rmatthes:


m-lindgren:
lgtm, let's merge this when the current CI run for master is done.

m-lindgren (approved):

benediktahrens:
In general I don't think UniMath should change to work around regressions in Coq. In this particular case, though, the changes are certainly desirable.

rmatthes:
Clearly, my small changes simplify the file. But the former rewrites were a test bed for the identified equations, and their applicability is now even more questionable.

m-lindgren:
This fixes part of it, but there are other problems: https://github.com/UniMath/UniMath/actions/runs/4331627549/jobs/7563581989#step:5:2134

m-lindgren:
It worked! Thanks @rmatthes

PR #1642, opened 2023-03-04
CI: Save cache even if build fails.

m-lindgren:
Currently when a job fails no cache will be saved. This often causes time consuming rebuilds of files that really needn't be rebuilt. At the end of last year github released a beta-version of a script that allows the cache action to save the cache even if a job failed.

Merging this would give it some real world use. Hopefully it'll save us some rebuild time.


rmatthes:
I cannot comment on the .yml script changes to make this happen, but the overall idea looks very good to me.

benediktahrens:
Thanks a lot for preparing this!

PR #1467, opened 2022-03-06
Set-enriched categories

umedaikiti:
This PR includes the following.

- #1031 (`auto` and `autorewrite` are removed)
- the bicategory of enriched categories
- pseudofunctors between the bicategory of set-enriched categories and the bicategory of categories (without a proof that they constitute a biequivalence, which gives the valiadation suggested in #1453)
- the pseudofunctor that takes enriched categories and returns underlying categories

I believe that the proof of biequivalence is straightforward but I would like to finish it later because it requires a lot of efforts.

PR #1643, opened 2023-03-05
Remove Coq.Init.Logic export almost everywhere.

m-lindgren:
This is a partial fix of #827. 

Currently Foundations exports Coq.Init.Logic. As a result of this it sometimes happens that `rewrite` is performed using `eq_refl` instead of `paths_refl`. It also imports datatypes that duplicate those of UniMath, but live in `Set`. This caused the inital problems with #1467, where the wrong coproduct eliminator was used, which caused some types to end up in `Set` (this was unfortunately hidden by `-type-in-type` in this reposity, but thankfully caught by the `SetHITs` satellite).

This PR removes Coq.Init.Logic from foundations and patches things up again. Tactics affected by this were mostly some rewrites, `trivial` and `replace`. To restore the full functionality of these I think one needs to register `core.eq.type` etc, but I know too little about Coq's internals to do this. But the fact that so little actually depended on this functionality shows that maybe this functionality isn't that needed.

Only two files remains to fix:
- GaussianElimination/Elimination.v
- HSET/Slice.v

Most of the changes were oneline-fixes, except for `FinOrd(Co)Products` where I struggled to patch it up and eventually gave up and rewrote most of it.


m-lindgren:
Sanity-checks fails because of the change to Foundations.

For a patch of TypeTheory see UniMath/TypeTheory#234.

UniMath/CategoryTheory/categories/HSET/Slice.v
```diff
@@ -18,6 +18,9 @@ October 2015 - January 2016
 
 *)
 
+(** *NOTE: This file depends on Coq.Init.Logic. *)
```
benediktahrens, line 21:
```suggestion
(** * TODO NOTE: This file depends on Coq.Init.Logic. *)
```

UniMath/Algebra/GaussianElimination/Elimination.v
```diff
@@ -6,6 +6,9 @@ Primary Author: Daniel @Skantz (November 2022)
   With much help from, and thanks to, Peter LeFanu Lumsdaine
 *)
 
+(** * NOTE: This file depends on Coq.Init.Logic. *)
```
benediktahrens, line 9:
```suggestion
(** * TODO NOTE: This file depends on Coq.Init.Logic. *)
```

nmvdw:
Nice! I think it would be good to mention in the style guide that one should use `change` instead of `replace` and that one should not use `trivial`.

UniMath/Foundations/Init.v
```diff
@@ -1,6 +1,5 @@
 (** Initial setup unrelated to Univalent Foundations *)
 
-Require Export Coq.Init.Logic.
 (* this fixes the advanced forms of the 'rewrite' tactic, but we want to eliminate it eventually *)
 (* Unfortunately, it also defines some notations, which we need to override later, hence the necessity to suppress such warnings. *)
```
benediktahrens, line 4: Should the comment also be deleted?

benediktahrens (approved):

benediktahrens:
If this partial fix - which is very good - gets merged, we should open an issue regarding the fixing of the remaining files that do depend on `Coq.Init.Logic`. We could add this to the list of things to discuss at the UniMath User Meeting.

m-lindgren:
> Nice! I think it would be good to mention in the style guide that one should use `change` instead of `replace` and that one should not use `trivial`.

I don't think it's necessary because with this change Coq won't let you do anything problematic (related to Coq.Init.Logic) with them, it will just fail instead. Are there other reasons to avoid these tactics? Because I think they can still be useful in many instances.

benediktahrens (approved):

benediktahrens:
This PR looks good to me. The change to Foundations is not significant in the mathematical sense, and is a very welcome one for the users of the library.

Skantz:
Hi,

I had a quick try to patch the remaining files (since I wrote one of them).

https://gist.github.com/Skantz/f8345c8040ad744d1b35475f6c35c1d1

But compilation now fails again in OmegaCocontFunctors.v, line 1007

m-lindgren:
@Skantz Thank you for looking into this! Feel free to submit a PR with any changes that compiles cleanly. Is it possible that OmegaCocontFunctors fails because of the changes in `HSET/Slice.v`?

Skantz:
@m-lindgren thank you. My guess was that removing the transitive import of Coq.Init.Logic from Slice.v exposed problematic a rewrite downstream in OmegaCocontFunctors.v (and later in MonadsMultiSorted.v)

Admitting these two proofs make the whole project compile: https://gist.github.com/Skantz/970c795f9346f1c0813ce2607ff3b373

benediktahrens:
If I understand correctly what you are saying, @Skantz , we can simply push the import of `Coq.Init.Logic` further down, fixing every use of the undesired rewrite tactic individually.
Could you file a PR that fixes your file and adds the import of `Coq.Init.Logic` to `OmegaCocontFunctors` and `MonadsMultiSorted`?

rmatthes:
https://github.com/UniMath/UniMath/pull/1653#issue-1620510677 now describes where the last obstacle resides. https://github.com/UniMath/UniMath/blob/5ae18b08113ab682fef9def7d88d3e8b0627bae0/UniMath/CategoryTheory/Chains/OmegaCocontFunctors.v#L1010 needs `Coq.Init.Logic` (at `rewrite h2` in the proof for `map_to_K_commutes`), which seems really intricate to replace by induction on the equality assumed by `h2`.

Maybe the situation could be seen more clearly by replacing `destruct` by `induction` already in the definition of `map_to_K`, so that the full-blown `match` `as` `in` `return` form of the pattern-matching construct would not be seen directly but be hidden in occurrences of `paths_rect`.

Skantz:
Thanks @rmatthes.

I came up with something like the following small patch

```
diff --git a/UniMath/CategoryTheory/Chains/OmegaCocontFunctors.v b/UniMath/CategoryTheory/Chains/OmegaCocontFunctors.v
index 5bd1ba1b9..7567f08e5 100644
--- a/UniMath/CategoryTheory/Chains/OmegaCocontFunctors.v
+++ b/UniMath/CategoryTheory/Chains/OmegaCocontFunctors.v
@@ -51,8 +51,6 @@ This file contains proofs that the following functors are (omega-)cocontinuous:
 Written by: Anders Mörtberg and Benedikt Ahrens, 2015-2016
 *)
 
-(** * TODO NOTE: This file depends on Coq.Init.Logic. *)
-Require Import Coq.Init.Logic.
 
@@ -1006,8 +1004,9 @@ destruct (natlthorgeh i j) as [h|h].
         apply cancel_postcomposition, maponpaths_12; try apply idpath.
         now rewrite <- (chain_mor_right h1 h2). }
       { destruct h; unfold fun_gt; simpl.
-        generalize h1; clear h1.
-        rewrite h2; intro h1.
+        destruct (!h2).
+        assert (eq: h2 = (idpath _)). { apply isasetnat. }
+        rewrite eq.
         apply cancel_postcomposition.
         apply maponpaths_12; try apply idpath; simpl.
         destruct (natlehchoice4 j j h1); [destruct (isirreflnatlth _ h)|].
```
If that's an acceptable proof, I'd be happy to make the PR.

m-lindgren:
@Skantz that does look like a simple solution - a PR is of course welcome!

PR #1644, opened 2023-03-06
Refactor monoidal categories

nmvdw:
I am refactoring the directory on monoidal categories to make it more structured and hopefully nicer to use.

I made the following changes:
- The previous code is put in the directory `MonoidalOld`
- I added a file `MonoidalCategories.v`, which is a copy of `MonoidalCategoriesWhiskered.v`, and `WhiskeredBifunctors.v`
- Cosmetic changes in `MonoidalCategories.v` to shorten lines, change a name, and add scopes
- Added monoidal categories (as pairs of a category with a monoidal structure) and the same accessors as I added before for monoidal categories. There are some admits remaining in the code
- Added comments 
- I added a separate folder for actegories
- The file on cartesian monoidal categories is moved to Examples. I also added comments in it
- Renamed `MonoidalFunctorsWhiskered.v` to `MonoidalFunctors`
- Cosmetic changes to `MonoidalFunctors` to shorten lines.
- Added bundled approach to monoidal functors
- Added comments in the file on monoidal functors
- The stuff on enriched categories now mostly makes use of Monoidal instead of MonoidalOld. This only required minor changes. In addition, all opaqueness commands for monoidal unitors/associators should be gone.
- Cosmetic changes to `WhiskeredDisplayedBifunctors.v` to shorten lines 
- I renamed `DisplayedMonoidalWhiskered` to `DisplayedMonoidal`. I added comments and I shortened lines
- Renamed `TotalDisplayedMonoidalWhiskered` to `TotalDisplayedMonoidal`, some lines are shortened and comments are added
- Renamed `MonoidalSectionsWhiskered` to `MonoidalSections`
- Renamed `BraidedMonoidalCategoriesWhiskered.v` to `SymmetricMonoidalCategories`
- Added comments in `SymmetricMonoidalCategories` and `ClosedMonoidalCategories`
- Renamed `CategoriesOfMonoidsWhiskered` to `CategoriesOfMonoids`
- Renamed `MonoidalCategories` (using tensor as the operation) to `MonoidalCategoriesTensored`. Removed the accessors in that file.
- Renamed `MonoidalFunctors` (using tensor as the operation) to `MonoidalFunctorsTensored`. Removed accessors in that file.
- Renamed `DisplayedMonoidal` to `DisplayedMonoidalTensored`.

I also noticed the following issues:
- ~~The file on cartesian monoidal categories only constructs cartesian monoidal categories. It does not provide a definition for when a monoidal category is cartesian. The notion of being cartesian is relevant in several applications of monoidal categories. For example, it is used in the definition of Markov category.~~ SOLVED
- ~~In addition, the notion of semi-cartesian monoidal category must be defined~~ SOLVED
- There is no notion of oplax monoidal functor
- There is a notion of strict monoidal functor, which preserves the tensor/unit strictly (up to equality). However, within an intentional setting such a notion will only be useful if the involved monoidal categories are strict (ie, their objects form an hSet). I think it would make sense to add this restriction to their definition.
- The local helper lemmas in MonoidalFunctors.v need to be moved
- Currently, monoidal natural transformations are in the same file as monoidal functors. I think it would be good to create a separate file for monoidal natural transformations.
- No make_disp_functor exists in Displayedcats.Functors
- is_iso_disp is defined using iso instead of z_iso
- In `MonoidalSections`: some equalities might be reproven (equality_for_rightunitality, equality_for_leftunitality, equality_for_associativity)
- For symmetric monoidal categories, relatively little infrastructure is defined (nothing for functors or displayed monoidal categories). In addition, there should be a constructor for symmetric monoidal categories using only one hexagon.
- Monoidal dialgebras currently use a lax functor and a strong functor. The assumption of strongness should be relaxed to oplaxness.
- Define the category of comonoids
- The augmented simplex category is not defined in the whiskered style
- The MonoidalCategory folder in Bicategories is very disjoint from Bicategories: files are separated from the rest, and some files could get a more logical place.
- The file `Actions` (in Bicategories) does not use the whiskered style


nmvdw:
Note:
- In the folder `MonoidalOld`, all old files are kept. 
- Stuff in `SubstitutionSystems` still depends on `MonoidalOld`. That is because I didn't fix it yet
- The monoidal Rezk completion still needs to be done, because I didn't have time for it yet
- Some parts on enriched categories still depends on `MonoidalOld`, and those need to be fixed
- In some accessors, `Admitted` is still used. This should be fixed.

nmvdw:
The sanity checks fail, but that problem will be solved later. It has to do with the fact that parts about enriched categories have not been fixed yet.

Kfwullaert:
@nmvdw Currently, the file "CategoriesOfMonoids.v" is in the top directory. Wouldn't it be better to have it in the Examples directory? Even though this is currently not formalized, this also has a (displayed) monoidal structure.

nmvdw:
> @nmvdw Currently, the file "CategoriesOfMonoids.v" is in the top directory. Wouldn't it be better to have it in the Examples directory? Even though this is currently not formalized, this also has a (displayed) monoidal structure.

I considered putting it in `Examples`, but since it does not seem to be an example of a monoidal category, I didn't put it there.

Kfwullaert:
@nmvdw Currently, there is a file on symmetric mon. cats and closed in the top folder. But CartesianMonoidal.v sits in the Examples directory. Shouldn't they be all at the same level? 
I do not think of these as "examples" of monoidal categories. I think they could even deserve their own subfolder, e.g. "StructuredMonoidal" (which then contains Cartesian, Symmetric and Closed).

rmatthes:
I have a unicode problem in ProofGeneral now: `unicode  𝟙` on my terminal shows `U+1D7D9 MATHEMATICAL DOUBLE-STRUCK DIGIT ONE`, but the symbol for the monoidal unit is not displayed in ProofGeneral.

UniMath/CategoryTheory/Monoidal/ClosedMonoidalCategories.v
```diff
@@ -0,0 +1,209 @@
+(*********************************************************************************
+
+ Closed monoidal categories
+
+ A closed monoidal category is one for which tensoring with a fix object has a
```
rmatthes, line 5:
```suggestion
 A closed monoidal category is one for which tensoring with a fixed object has a
```

UniMath/CategoryTheory/Monoidal/ClosedMonoidalCategories.v
```diff
@@ -0,0 +1,209 @@
...
+ right adjoint. Depending on whether the monoidal category is symmetric or not,
+ there are different variations of this definition. If the monoidal category is
+ not symmetric, then there are two variations, namely left closed and right
+ closed, and there difference is whether we look at an adjoint for tensoring with
```
rmatthes, line 9:
```suggestion
 closed, and their difference is whether we look at an adjoint for tensoring with
```

UniMath/CategoryTheory/Monoidal/MonoidalCategories.v
```diff
@@ -1,514 +1,1321 @@
...
+ The main definition in this file, takes a so-called displayed approach.
+ More specifically, we define the notion of a monoidal structure on a
+ category. For this notion, we define suitable accessors and we prove the
+ laws. Finally, we also provide a bundled notion of monoidal category, which
+ is a category together with a monoidal structure for it. The necessary
+ accessors and laws are derived from the other notion.
+
+ In this file, we use a whiskered approach. This means that we have two
+ operations to tensor with morphisms: a left and a right whiskering. Both of
+ these take an object and a morphism and input and they return a morphism as
```
rmatthes, line 17:
```suggestion
 these take an object and a morphism as input and they return a morphism as
```

UniMath/CategoryTheory/Monoidal/MonoidalCategories.v
```diff
@@ -1,514 +1,1321 @@
...
+    exists I_{Mon_V}.
+    exists (λ v, ru_{Mon_V} v).
+    exists (λ v, ruinv_{Mon_V} v).
+    exists (λ v, lu_{Mon_V} v).
+    exists (λ v, luinv_{Mon_V} v).
+    exists (λ v1 v2 v3, αinv_{Mon_V} v3 v2 v1).
+    exact (λ v1 v2 v3, α_{Mon_V} v3 v2 v1).
+  Defined.
+
+  Definition monoidal_swapped_laws {V : category} (Mon_V : monoidal V)
```
rmatthes, line 1210:
```suggestion
  Lemma monoidal_swapped_laws {V : category} (Mon_V : monoidal V)
```

UniMath/CategoryTheory/Monoidal/MonoidalFunctors.v
```diff
@@ -1,239 +1,1140 @@
...
+            rewrite assoc.
+            apply cancel_postcomposition.
+            apply (pr22 (fmonoidal_preservesunitstrongly Gm)). }
+          rewrite id_left.
+          rewrite <- functor_comp.
+          etrans.
+          { apply maponpaths.
+            apply (pr22 (fmonoidal_preservesunitstrongly Fm)). }
+          apply functor_id.
+  Defined.
```
rmatthes, line 1024: Unfortunately, it has quite big proof parts added to the inverses - my bad. Maybe there should be a specific code used for comments to mark such problems that become visible with the refactoring process - no need to treat them immediately, but the elements should be marked in place and not w.r.t. to line numbers that could change with the ongoing refactoring.

nmvdw:
> @nmvdw Currently, there is a file on symmetric mon. cats and closed in the top folder. But CartesianMonoidal.v sits in the Examples directory. Shouldn't they be all at the same level?
> I do not think of these as "examples" of monoidal categories. I think they could even deserve their own subfolder, e.g. "StructuredMonoidal" (which then contains Cartesian, Symmetric and Closed).

The definition of cartesian monoidal category is: 'A monoidal category is cartesian if the tensor is a product and the unit is terminal'. However, the file `CartesianMonoidal.v` currently only contains the result that every category with binary products and a terminal gives rise to a monoidal category, and not the definition of when a monoidal structure is cartesian. As such, the current `CartesianMonoidal.v` is different from symmetric monoidal categories, and closed monoidal categories, because the former contains a construction, while the other specify further properties that a monoidal category could have.

As such, I propose the following changes:
- Create a directory 'Structures', which contain symmetric and closed. A new file `CartesianMonoidal.v` should be added that contains definitions of when a monoidal structure is cartesian, semicartesian, cocartesian, and semicocartesian.
- The current `CartesianMonoidal.v` should be renamed to something like `MonoidalFromTerminalAndProd.v`

What about this?

> I have a unicode problem in ProofGeneral now: `unicode 𝟙` on my terminal shows `U+1D7D9 MATHEMATICAL DOUBLE-STRUCK DIGIT ONE`, but the symbol for the monoidal unit is not displayed in ProofGeneral. I am using GNU Emacs 27.1 for ProofGeneral, which is the current version on Ubuntu 22.04. I tried to find out more about this symbol in Emacs with `C-u C-x =`. It tells me that no font is available for display, see the attached screenshot.
> Screenshot from 2023-03-06 17-56-57
> I would like to see what you have in the line `display` on your computer so that I can search for this missing font.

I used the notation `𝟙` before, but now I am switching to `I_{ _ }`. I will delete that notation, and that problem should not occur anymore.

rmatthes:
@m-lindgren suggested to take the singular form `Structure` for the new directory. In object-oriented programming, it is regularly recommended to use the singular form for names of classes since the class name then appears as type, which makes a nice reading. The downside in UniMath would be that our directories currently mostly are in plural form (if they do not designate a mathematical topic), and we better stay consistent with the already existing naming scheme.

nmvdw:
> @m-lindgren suggested to take the singular form `Structure` for the new directory. In object-oriented programming, it is regularly recommended to use the singular form for names of classes since the class name then appears as type, which makes a nice reading. The downside in UniMath would be that our directories currently mostly are in plural form (if they do not designate a mathematical topic), and we better stay consistent with the already existing naming scheme.

Oh good point, I misread and I actually read 'Structures'.

What name do we prefer? 'Structures' or 'Structure'? If we prefer the singular form, should we accordingly modify other folder names?

UniMath/CategoryTheory/Monoidal/MonoidalCategories.v
```diff
@@ -1,514 +1,1321 @@
...
+Import BifunctorNotations.
 
-Context {C : category} (tensor : C ⊠ C ⟶ C) (I : C).
+(**
+ 1. Monoidal structures
+ *)
+Section A.
+
+(** Data **)
+Definition tensor (C : category) : UU :=
```
rmatthes, line 55: The tensor still enters all of this data of a monoidal category - together with the proofs of the tensor laws (in fact, the laws of a bifunctor in whiskered form). I expected having only `bifunctor_data` here for this big refactoring effort.

rmatthes (commented):
I have only looked into a very small portion yet.

UniMath/CategoryTheory/Monoidal/MonoidalCategories.v
nmvdw: Yes. I did not change that yet.

rmatthes:
> > @m-lindgren suggested to take the singular form `Structure` for the new directory. In object-oriented programming, it is regularly recommended to use the singular form for names of classes since the class name then appears as type, which makes a nice reading. The downside in UniMath would be that our directories currently mostly are in plural form (if they do not designate a mathematical topic), and we better stay consistent with the already existing naming scheme.
> 
> Oh good point, I misread and I actually read 'Structures'.
> 
> What name do we prefer? 'Structures' or 'Structure'? If we prefer the singular form, should we accordingly modify other folder names?

I have very little experience in tracking file history in GitHub or plain Git after name changes. This is why I also raised my doubts if it were good to rename `Monoidal` into `MonoidalCategories`. But we could be free to follow a nicer naming scheme for any new directory.

rmatthes (commented):
To my mind, there is too much spacing in this unit notation (in https://github.com/UniMath/UniMath/pull/1644/commits/28c6855f8a15d49f1ae27152a1c38d656b57a5ae). For hom-notation, the spacing could be debated - I prefer generally less spacing, but there is no need to do it consistently. For the unit, I would better like it without any blanks.

UniMath/CategoryTheory/Monoidal/MonoidalFunctors.v
nmvdw: This is fixable and the latest commit contains a fix for it. Thanks for noting it.

m-lindgren:
> But we could be free to follow a nicer naming scheme for any new directory.

Agree, one can use the singular in this instance and others where it makes sense, without having to update the entire library to follow the same scheme. It's a practice we adhere to in other similar cases anyway.

nmvdw:
> To my mind, there is too much spacing in this unit notation (in [28c6855](https://github.com/UniMath/UniMath/commit/28c6855f8a15d49f1ae27152a1c38d656b57a5ae)). For hom-notation, the spacing could be debated - I prefer generally less spacing, but there is no need to do it consistently. For the unit, I would better like it without any blanks.

I inherited my usage of spaces from Agda, where not adding spaces between two characters could lead to parsing issues. I will change it in this PR.

nmvdw:
> > But we could be free to follow a nicer naming scheme for any new directory.
> 
> Agree, one can use the singular in this instance and others where it makes sense, without having to update the entire library to follow the same scheme. It's a practice we adhere to in other similar cases anyway.

Sure. I will call the folder Structure.

nmvdw:
Screenshot 2023-03-06 at 19 17 09
One thing to be noted: if the folder is called `Structure`, then the `Structure` will be highlighted in the import since `Structure` is a keyword in Coq.

nmvdw:
The latest commit contains the definition of cartesian and cocartesian monoidal categories, which I would consider a structure.

Do we agree that the file `Examples/CartesianMonoidal.v` should be renamed to something like `Examples/MonoidalFromTerminalAndProduct.v`?

rmatthes (commented):
Did you also treat the occurrences of I_{ V₁ }?

nmvdw:
> Did you also treat the occurrences of I_{ V₁ }?

Not yet. I will look at them soon.

m-lindgren:
Also in the screenshot you posted you had 
```Require Import UniMath.CategoryTheory.Monoidal.Structure.SymmetricMonoidalCategories.```. 
What do you think about calling the last file only `Symmetric`, since the rest is given by the context? (And same with `Closed`, `Cartesian`, ...)

**Comment by nmvdw** [comment]
> Also in the screenshot you posted you had
> `Require Import UniMath.CategoryTheory.Monoidal.Structure.SymmetricMonoidalCategories.`.
> What do you think about calling the last file only `Symmetric`, since the rest is given by the context? (And same with `Closed`, `Cartesian`, ...)

Good point!

**Comment by rmatthes** [comment]
> Also in the screenshot you posted you had `Require Import UniMath.CategoryTheory.Monoidal.Structure.SymmetricMonoidalCategories.`. What do you think about calling the last file only `Symmetric`, since the rest is given by the context? (And same with `Closed`, `Cartesian`, ...)

This touches on another important global question: do we accept multiple files of the same base name? Depending on the tools one uses, this can be worrying. Emacs is very good at informing the user to which directory a buffer is associated in case of such name clashes.

File: UniMath/CategoryTheory/EnrichedCats/Enriched/Enriched.v [hunk 7]
```diff
@@ -289,7 +292,7 @@ Proof.
     refine (maponpaths (fun f => f · _) (enriched_functor_on_comp F _ _ _) @ _).
     refine (_ @ !maponpaths (fun f => f · _) (tensor_comp _ _ _ _ _)).
 
-    (* Get rid of the common prefix *)
+    (* Get ridentity of the common prefix *)
```
rmatthes, line 295:
```suggestion
    (* Get rid of the common prefix *)
```
a false hit of semi-automatic replacement

UniMath/CategoryTheory/Monoidal/MonoidalCategories.v
nmvdw: This change can be done. However, the consequence is that when using laws like `bifunctor_rightcomp`, one needs to indicate the monoidal category like this:
```
rewrite (bifunctor_rightcomp M).
```
Which do you prefer? A strict separation of data and properties or having the rewrites more convenient?

nmvdw: Note: this would be a time-consuming change, so I will only do it if desired and this consequence is accepted. @Kfwullaert @benediktahrens

nmvdw: Personally, I don't find it a big issue. I met the same for enriched categories, and in practice, it is a small hurdle. In addition, it makes the code easier to read (clearer what the relevant arguments are). It also becomes a bit easier to maintain.

nmvdw:
In the latest commit, there's a version in which data and properties are fully separated for monoidal categories. I did not look at displayed monoidal categories.

nmvdw:
I made some progress on making the files in `Bicategories` depend on `Monoidal` instead of `MonoidalOld`.

@rmatthes Can you check how long it takes to type check `ActionOfEndomorphismsInBicatWhiskered.v` on your device? And whether this time got affected?

nmvdw:
Another thing that I noted: the `MonoidalCategory` folder in `Bicategories` is very disjoint from Bicategories (files are separated from the rest, and some files could get a more logical place).

rmatthes:
> I made some progress on making the files in `Bicategories` depend on `Monoidal` instead of `MonoidalOld`.
> 
> @rmatthes Can you check how long it takes to type check `ActionOfEndomorphismsInBicatWhiskered.v` on your device? And whether this time got affected?

What is a good way to carry out measurements while using Dune? When "making" UniMath, all the compilation times are written into dot files that I sometimes analysed with GNUplot.

rmatthes:
> Another thing that I noted: the `MonoidalCategory` folder in `Bicategories` is very disjoint from Bicategories (files are separated from the rest, and some files could get a more logical place).

Uniting those files in one directory (with few exceptions) was mostly out of fear to forget about what had already been formalized. Is there a way to make a "logical" directory in addition to the one of the file system?

nmvdw:
I made some progress on refactoring the files in Bicategories/MonoidalCategories, but I didn't finish yet. I will continue on Wednesday.

> > Another thing that I noted: the `MonoidalCategory` folder in `Bicategories` is very disjoint from Bicategories (files are separated from the rest, and some files could get a more logical place).
> 
> Uniting those files in one directory (with few exceptions) was mostly out of fear to forget about what had already been formalized. Is there a way to make a "logical" directory in addition to the one of the file system?

I will say more concretely what changes I would expect in this regard on Wednesday.

nmvdw:
By the way, in the files `Monoidal/MonoidalCategories.v` and `Monoidal/MonoidalFunctors.v` there are some admitted left. These are about accessing the appropriate laws for monoidal categories and monoidal functors. Any hints on which laws to use here based on those that are already given, would be very welcome @Kfwullaert @rmatthes @benediktahrens .

UniMath/CategoryTheory/Monoidal/MonoidalCategories.v
rmatthes: I need to understand your arguments better. So, you say that while the coercion from properly layered monoidal categories (whose data only depends on `bifunctor_data`) to bifunctors can be declared, the rewrite mechanism will just not come up with the monoidal structure, although that very monoidal structure is printed out in the formula. You do not only say it but it is in your commit https://github.com/UniMath/UniMath/commit/98f93b13dccdc9929d73d583b41fe7399a1f996d . But you did not say why having to give the monoidal structure is easy ("not a big issue"). Is it because of that presence in the goal formula? Or is it because, in the abstract development, that monoidal structure is always present in a context variable and hence can be very easily added by hand? Looking at the said commit, I second you for the readability of the code. And the price to pay for the clear separation into data and laws does not seem to be too high, even though it seems that giving the extra argument is now even needed when applying those equations as lemmas instead of only rewriting with them. The real benefit will come when the data of a monoidal structure and the actegory structure of its self-action are convertible. Hence, I approve this move.

nmvdw:
> I'll try to contribute a bit to this PR. Please tell me if other work on it is ongoing.

Thank you. Everything is pushed now, so at the current moment no other work on it is ongoing. I am in the process of doing the files in the Bicategories folder though, but I will continue that tomorrow.

rmatthes:
@nmvdw I have stopped working on this project and left a PR on your fork. I would say the new organization of material is good to work with.

nmvdw:
> @nmvdw I have stopped working on this project and left a PR on your fork. I would say the new organization of material is good to work with.

That sounds really good!

rmatthes:
I'll clean up one ugly thing of my commit of yesterday.

nmvdw:
> I'll clean up one ugly thing of my commit of yesterday.

Thank you. Btw, I worked a bit further, and I changed the file `ActionOfEndomorphismsInBicatWhiskered.v` which is much faster now.

rmatthes:
I'll leave this project again for some hours - I have committed everything (with a PR to your fork @nmvdw ).

nmvdw:
> I'll leave this project again for some hours - I have committed everything (with a PR to your fork @nmvdw ).

Thank you. It is merged now.

rmatthes:
> > I'll clean up one ugly thing of my commit of yesterday.
> 
> Thank you. Btw, I worked a bit further, and I changed the file `ActionOfEndomorphismsInBicatWhiskered.v` which is much faster now.

But it seems you did not commit everything you had locally for making it compile. When I check out your PR, I get the following error when compiling with Dune:
```
File "./UniMath/Bicategories/MonoidalCategories/ActionOfEndomorphismsInBicatWhiskered.v", line 70, characters 2-28:
Error:
In environment
C : bicat
c0, d0 : C
The term "action_from_precomp" has type
 "∑ F : bifunctor_data endocat homcat homcat, is_bifunctor F"
while it is expected to have type "action_data homcat".
```

nmvdw:
Yes, I am currently still fixing those errors to make it compile

rmatthes:
The idea of this refactoring is that the directory `MonoidalOld` will become obsolete. Since the newly organized code is essentially the same as before, this would normally mean that the directory would have to be deleted. However, it appears to me that all of the Git history is attached to that directory. And the Git history of the tensored monoidal categories is now in the file representing the whiskered monoidal categories, and not in the file where they reside. And this seems to be similar for all the other alternative developments.

I think the help of a Git wizard would be needed to see how the Git history can be restored - without too much effort on replaying all the commits of the present PR.

nmvdw:
The file `EquivalenceMonCatNonCurried` seems to be broken, and I don't know how to fix it. The problem is that certain lines take too much time. I commented out the whole file for now and I am asking for some help on it @Kfwullaert . This is the only file in Bicategories that remains to be fixed.

nmvdw:
Now everything should compile (sanity checks will still fail).

rmatthes:
> > > I'll clean up one ugly thing of my commit of yesterday.
> > 
> > 
> > Thank you. Btw, I worked a bit further, and I changed the file `ActionOfEndomorphismsInBicatWhiskered.v` which is much faster now.
> 
> But it seems you did not commit everything you had locally for making it compile. When I check out your PR, I get the following error when compiling with Dune:
> 
> ```
> File "./UniMath/Bicategories/MonoidalCategories/ActionOfEndomorphismsInBicatWhiskered.v", line 70, characters 2-28:
> Error:
> In environment
> C : bicat
> c0, d0 : C
> The term "action_from_precomp" has type
>  "∑ F : bifunctor_data endocat homcat homcat, is_bifunctor F"
> while it is expected to have type "action_data homcat".
> ```

This error seems to persist.

nmvdw:
Strangely enough, I didn't have that error. Is it still persisting in the latest commit?

rmatthes:
`pentagon_eq_implies_pentagon_eq_nice` of `UniMath/SubstitutionSystems/ActionBasedStrengthOnHomsInBicat.v` is now stalled at `Qed`.

rmatthes:
I am on Coq8.15.2 on this computer - this might explain compilation problems.

rmatthes:
> The idea of this refactoring is that the directory `MonoidalOld` will become obsolete. Since the newly organized code is essentially the same as before, this would normally mean that the directory would have to be deleted. However, it appears to me that all of the Git history is attached to that directory. And the Git history of the tensored monoidal categories is now in the file representing the whiskered monoidal categories, and not in the file where they reside. And this seems to be similar for all the other alternative developments.
> 
> I think the help of a Git wizard would be needed to see how the Git history can be restored - without too much effort on replaying all the commits of the present PR.

I have to come back to this previous comment of mine. Even if all compiles well, this would at most say that the obtained files are fine now. But it does not speak about their meta-data in Git.

nmvdw:
> I am on Coq8.15.2 on this computer - this might explain compilation problems.

For me, it takes around 100 seconds.

> > The idea of this refactoring is that the directory `MonoidalOld` will become obsolete. Since the newly organized code is essentially the same as before, this would normally mean that the directory would have to be deleted. However, it appears to me that all of the Git history is attached to that directory. And the Git history of the tensored monoidal categories is now in the file representing the whiskered monoidal categories, and not in the file where they reside. And this seems to be similar for all the other alternative developments.
> > I think the help of a Git wizard would be needed to see how the Git history can be restored - without too much effort on replaying all the commits of the present PR.
> 
> I have to come back to this previous comment of mine. Even if all compiles well, this would at most say that the obtained files are fine now. But it does not speak about their meta-data in Git.

I don't know how to solve this. Maybe somebody else does?

rmatthes:
Just after I wrote this comment, my office mate Érik Martin-Dorel entered the office and he even took the time to explain to me what goes wrong here and how to solve the problem without losing much time and the later visibility of the entire process that leads to the new state of master.

The most important part is NOT TO MERGE as it is now.
To be on the safe side, we should rather create a new branch in UniMath from master and modify the present PR so that I can be squash-merged. This would make the history of the present PR invisible to Git but keeps the information in GitHub. It should also allow GitHub to guess properly which files continue the history of the "legacy" files.

rmatthes:
@nmvdw Don't you think that most files from `MonoidalOld` could already be deleted if the remaining ones were made to import the reorganized ones?

nmvdw:
> @nmvdw Don't you think that most files from `MonoidalOld` could already be deleted if the remaining ones were made to import the reorganized ones?

I agree they can be deleted, so I will do that in the next commit.

rmatthes:
I see in the settings that squash merging is allowed for UniMath. When looking through https://github.com/search?q=org%3AUniMath+squash&type=issues, I see that @DanGrayson once gave the advice to use it to clean the commit history. The use case here would be somehow different: the purpose is to guide GitHub to make the right connections between renamed files and their history. Did such a case already arise? @benediktahrens ?

benediktahrens:
> I see in the settings that squash merging is allowed for UniMath. When looking through https://github.com/search?q=org%3AUniMath+squash&type=issues, I see that @DanGrayson once gave the advice to use it to clean the commit history. The use case here would be somehow different: the purpose is to guide GitHub to make the right connections between renamed files and their history. Did such a case already arise? @benediktahrens ?

I don't remember this. I believe that the best way of recording any sort of authorship of code is to write a paper about it.

nmvdw:
I removed monoidal old, and there should be no dependencies on it. I left in the code for the monoidal Rezk completion, because that one still needs to be processed.

UniMath/Bicategories/MonoidalCategories/EquivalenceActegoriesAndActions.v
```diff
@@ -12,11 +12,11 @@ Require Import UniMath.Bicategories.Core.Examples.BicatOfCats.
 Require Import UniMath.Bicategories.DisplayedBicats.DispBicat.
 
 Require Import UniMath.CategoryTheory.Monoidal.WhiskeredBifunctors.
-Require Import UniMath.CategoryTheory.Monoidal.MonoidalCategoriesWhiskered.
-Require Import UniMath.CategoryTheory.Monoidal.MonoidalFunctorsWhiskered.
+Require Import UniMath.CategoryTheory.Monoidal.MonoidalCategories.
+Require Import UniMath.CategoryTheory.Monoidal.MonoidalFunctors.
```
m-lindgren, line 16: This refactoring looks great. What do you think about calling the files `Categories`, `Functors` etc? So that these imports become
```
Require Import UniMath.CategoryTheory.Monoidal.Categories.
```
and maybe same for `Displayed`:
```
Require Import UniMath.CategoryTheory.Monoidal.Displayed.Categories.
Require Import UniMath.CategoryTheory.Monoidal.Displayed.Functors.
```
It's not an important change at all, I just think it looks nice not to repeat 'Monoidal' more than necessary. @nmvdw , @Kfwullaert , @rmatthes, you use this part of the library much more than I do, what do you prefer?

nmvdw: That's a very good point, and I agree with that change. If @Kfwullaert and @rmatthes agree, then I think we should make that change.

nmvdw:
@rmatthes I am a bit unsure about the file `ActionOfEndomorphismsInCATWhiskeredElementary.v`. I forgot whether I renamed it `ActionOfEndomorphismsInCATElementary.v`, and, if I look at them, they seem to be quite similar (I think the same, but not sure). If they are, we can remove `ActionOfEndomorphismsInCATWhiskeredElementary`.

rmatthes:
We might have made parallel developments concerning that file.

nmvdw:
> We might have made parallel developments concerning that file.

Yes, that is possible. Would it be fine to only keep `ActionOfEndomorphismsInCATElementary.v`?

Kfwullaert:
> The file `EquivalenceMonCatNonCurried` seems to be broken, and I don't know how to fix it. The problem is that certain lines take too much time. In addition, the goals in the file are too long/messy to be readable, so it is not easy for me to find an actual fix. I commented out the whole file for now and I am asking for some help on it @Kfwullaert . This is the only file in Bicategories that remains to be fixed.
> 
> At this point, it might also be a good idea to completely refactor this files in the following sense: the standard definition is now the whiskered one, so the desired equivalence is with the whiskered one. Instead of preserving `EquivalenceMonCatNonCurried`, I suggest that it gets replaced by a file that construct the desired equivalence.

@nmvdw I agree with your suggestion. This definition is in the library because it was the bicategory of non-curried monoidal categories has shown to be univalent. I assume that after some time, this (univalence) lemma will be changed to use the whiskered monoidal categories instead (or something else). After that is done, there will be no need for this anyway.

UniMath/Bicategories/MonoidalCategories/EquivalenceActegoriesAndActions.v
Kfwullaert: @nmvdw  I agree with renaming them.

m-lindgren:
> The most important part is NOT TO MERGE as it is now.
> To be on the safe side, we should rather create a new branch in UniMath from master and modify the present PR so that it can be squash-merged.

@rmatthes  : To verify, can you have a look at this branch I created as a proof-of-concept: https://github.com/m-lindgren/UniMath/tree/slices-of-cherry 

Can you tell if the history of the files / code looks like we want it? In that case the method I used was the following:

I created a branch based on `master` that I called `slices-of-cherry`, then I cherry-picked all the commits in @nmvdw's branch `slices-of-bicat` (this PR), then I squashed them to one single commit (of course I got conflicts on `CONTENTS.md` and `.package/files` and I didn't spend any time on resolving them properly). 

If this branch didn't have merge:s with `master` already I believe we could have auto-squashed everything though.

nmvdw:
Current todos for this PR:
- In the files Monoidal/Categories.v and Monoidal/Functors.v, there are still some lemmas left that have admitted
- In Enriched/Examples/HomFunctor.v, some part was commented out that needs to be recovered
- The monoidal Rezk completion

Did I miss any?

m-lindgren:
Did this PR accidentally remove the `sub/coq` submodule by the way?

nmvdw:
> Did this PR accidentally remove the `sub/coq` submodule by the way?

I accidentally committed some stuff with the `sub/coq` submodule, and I don't know how to solve it.

rmatthes:
> > We might have made parallel developments concerning that file.
> 
> Yes, that is possible. Would it be fine to only keep `ActionOfEndomorphismsInCATElementary.v`?

`UniMath/CategoryTheory/Actegories/Examples/SelfActionInCATElementary.v` should be a similar case. I have not yet looked at the meld comparison (for both these cases).

m-lindgren:
@nmvdw Made a PR adding back the coq submodule that I think should work (I don't use submodules much so they are a bit confusing for me).

nmvdw:
I think that now all admits are gone and that the only thing left, is the monoidal Rezk completion.

rmatthes:
@nmvdw Could you please check if anything on my file list linked above (that discusses less than 20 files individually) has been lost in the process? E.g., where did the toy examples `UniMath/CategoryTheory/Monoidal/Examples/PointedSetCartesianMonoidal.v` and `UniMath/CategoryTheory/Monoidal/Examples/SetWithSubset.v` go?

I now think that the command-line options to git log offer enough to study the history for the few files where Git does not connect the new file to the old name before the refactoring - on the basis of @m-lindgren 's hand-squashed commit.

On the doubtful two variants of files: SelfActionInCATWhiskeredElementary.v is the wrong variant, while `ActionBasedStrongFunctorsMonoidal.v` and `ActionBasedStrongFunctorsWhiskeredMonoidal.v` should both be preserved. They could serve as a big example for the usefulness of the move from the tensored to the whiskered style. And the bicategorical analysis is not yet completed on that material either.

nmvdw:
`UniMath/CategoryTheory/Monoidal/AlternativeDefinitions/CategoriesOfMonoids.v` came from https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/Monoidal/CategoriesOfMonoids.v (which was in the previous style)

I indeed seem to have missed `PointedSetCartesianMonoidal.v`, `SetCartesianMonoidal.v`, and `SetWithSubset.v`.

rmatthes:
I assume that the continuity of the history of the three latter files would be guaranteed when @m-lindgren redoes the cherry-picking - once the refactoring would be declared as completed on the side of the Coq files.

@nmvdw Could you please delete `SelfActionInCATWhiskeredElementary.v ` and see if this causes a problem?

nmvdw:
In the latest commit, I added the monoidal Rezk completion. However, I could not fix this: the problem is that the construction makes use of the monoidal functor category, which was defined in a style that's now in `AlternativeDefinitions`. As such, I think we should make an issue about updating it.

nmvdw:
I think this PR is ready. However, we should make numerous of issues for some remaining problems.

rmatthes:
We were now one normal click away from merging this pull request - despite major objections I raised. I would hope for a new version by @m-lindgren that is unproblematic.

nmvdw:
> We were now one normal click away from merging this pull request - despite major objections I raised. I would hope for a new version by @m-lindgren that is unproblematic.

The objections are only related to the git history, right?

m-lindgren:
@rmatthes Updated squash at https://github.com/m-lindgren/UniMath/tree/slices-of-cherry - I don't expect the changes to CONTENTS.md and .package/files to be correct because they caused several merge conflicts, but they should be easy to fix with an extra commit.

m-lindgren:
So to use it in this PR, assuming everything looks good in the git history, @nmvdw could check out my `slices-of-cherry` branch locally and then force-push it to nmvdw:slices-of-bicat, then fix any `CONTENTS` problems.

m-lindgren:
@rmatthes Note that by squashing everything some information will be lost, in particular your authorship of your contributions to this PR.

rmatthes:
I was taken by an urgency. Yes, my objections were only on the meta-data. But this is generally a delicate matter in Git with renaming. There are many discussion threads on this topic. I would once more like to go through the squashed commit - to see if all links that I can reasonably expect (given my extra knowledge since yesterday afternoon) are present. My contribution to this PR was no more than of a helping hand. Authorship is here the least of my concerns. But I also think that there is mathematically no intermediary step to be preserved. @nmvdw should say if he feels something is lost by squashing. Anyway, the GitHub PR incl. discussion can be consulted. Although one might then better not force-push anything but just open a new PR on a new branch.

nmvdw:
I don't think anything is lost with squashing.

nmvdw:
`UniMath/CategoryTheory/Monoidal/Examples/EndofunctorsMonoidalElementary.v,` is removed now and `UniMath/CategoryTheory/Monoidal/Examples/BinopCartesianMonoidal.v` is restored.

rmatthes:
I am sorry, but you removed the wrong file - the one with `Elementary` in its name was the right one.

nmvdw:
> I am sorry, but you removed the wrong file - the one with `Elementary` in its name was the right one.

That should be fixed now

rmatthes:
`UniMath/CategoryTheory/Monoidal/Examples/BinopCartesianMonoidal.v` is now (except for a difference in one line) a copy of `UniMath/CategoryTheory/Monoidal/Examples/EndofunctorsMonoidalElementary.v`

nmvdw:
> `UniMath/CategoryTheory/Monoidal/Examples/BinopCartesianMonoidal.v` is now (except for a difference in one line) a copy of `UniMath/CategoryTheory/Monoidal/Examples/EndofunctorsMonoidalElementary.v`. This cannot be right.

Whoops. That's fixed now.

m-lindgren:
Updated squashed branch @rmatthes.

rmatthes:
I have looked at commit https://github.com/UniMath/UniMath/commit/adb0906bcc3cec147097ac4d4cb828b1fe4e7a24 (prepared by @m-lindgren). Nothing odd springs to my eyes. So, it is probably the commit to put forward in a PR to be merged quite soon. Was there still the question of `CONTENTS.md`? And I could add the info to the ten files where to find the historic Git data on them (according to the list I attached above).

nmvdw:
> So to use it in this PR, assuming everything looks good in the git history, @nmvdw could check out my `slices-of-cherry` branch locally and then force-push it to nmvdw:slices-of-bicat, then fix any `CONTENTS` problems.

I think now this is done.

m-lindgren:
> Although one might then better not force-push anything but just open a new PR on a new branch.

I think this is a good suggestion, then we can link this PR to the new one and keep all the discussions and the development processes intact in this PR.

rmatthes:
I have limited knowledge of Git, in particular when it comes to force-pushing. Yesterday, I already encountered problems in my clones with the force-pushes that happened into the present branch and the one with the squashed commit. Who would now create a PR that is free from the problems that force-pushes can have (in particular on clones that closely followed the entire development)?

rmatthes:
The present PR #1644 seems to be okay concerning the obtained global effect on the Coq source files. But it has a complicated history -- maybe only up to https://github.com/UniMath/UniMath/pull/1644/commits/fc0920273b9e8ef807d94b1af44d630de705587c -- that (when merged) will do harm to the internal structure of the Git repository, in particular in terms of continuity of the history of the individual files. @m-lindgren has prepared an alternative commit https://github.com/UniMath/UniMath/commit/adb0906bcc3cec147097ac4d4cb828b1fe4e7a24 in his fork that squashes all the individual commits into one. My suggestion was to put this commit on top of master in a new PR and do the final changes to `CONTENTS.md` afterwards (and maybe add my header info to the ten files that do not point to their history any longer). The new PR can refer to the present one to see all of the development inside this refactoring process.

It seems that GitHub presents us the sequence of events in the wrong order. Now, it looks like https://github.com/UniMath/UniMath/commit/adb0906bcc3cec147097ac4d4cb828b1fe4e7a24 comes as another commit even before the discussion of whether it would be okay to take it from that fork. So, this is probably a force-push (I understand @nmvdw as having followed the suggestion by @m-lindgren). I do not have the expertise in Git to judge on the consequences of force-pushing (neither for the future master branch in case this PR gets merged nor on the possibility of going into the development of this PR). Because of that lack of expertise, I would much prefer to see a PR with a clean history as presented by GitHub and merge that one into master. One can anyway compare different branches to see if they have the same overall effect on the Coq source files.

m-lindgren:
I think it's better to do what @rmatthes suggested and create a new PR based on the squashed branch (+ CONTENTS-fix and the documentation @rmatthes mentioned) only, and keep this PR (closed) as a historical document on the discussion.

benediktahrens:
Thanks a lot for the explanation.
@m-lindgren : would you be willing to open a new PR?

m-lindgren:
> @m-lindgren : would you be willing to open a new PR?

New PR: #1646.

m-lindgren:
The contents of this PR was squashed and merged into the library with #1646.

nmvdw:
> The contents of this PR was squashed and merged into the library with #1646.

Thanks a lot!

PR #1652, opened 2023-03-12
WIP: Smarter cache management for CI.

m-lindgren:
This is a work in progress trying to make the cache management a bit smarter.

Currently a PR can easily consume all 10 GB available because on each run the cache is duplicated instead of updated. This in and of it self is _not_ a critical problem: GitHub will just evict the oldest cache files until we are below the quota again. It is however a nuisance since it will cause unnecessary rebuilds when a PR is merged, or when another PR is opened. 

This does two things so far: 
- When a PR is closed (or merged into master) all cache files owned by it are removed.
- When a job is rerun for a PR the cache is updated with the new contents, instead of duplicated. Precisely: the old cache file is deleted and the new is saved with the same name.

### DIFF
--- a/.github/workflows/build-unimath.yml
+++ b/.github/workflows/build-unimath.yml
@@ -101,7 +101,7 @@ jobs:
           gh extension install actions/gh-actions-cache
           REPO=${{ github.repository }}
           REF=${{ github.ref }}
-          KEY=${{ github.unimath-cache.key }}
+          KEY=${{ steps.unimath-cache.inputs.key }}
 
           set +e
           echo "Deleting old cache..."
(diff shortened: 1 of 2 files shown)


benediktahrens:
@m-lindgren : this PR is still marked as a draft - is this correct, or is it ready for review?

m-lindgren:
@benediktahrens no this is still a WIP, It needs some work and testing. I'm trying to figure out some permission issues currently.

arnoudvanderleer:
@m-lindgren What is the state of this PR?

m-lindgren:
Defunct. It can be closed.

arnoudvanderleer:
Okay, closing this PR then

PR #1654, opened 2023-03-13
Proof more elegant

CalosciMatteo:
This pull request contains some minor improvement on some of the proof of my last pull request https://github.com/UniMath/UniMath/pull/1613.

It also add two simple results:
- The derivation of `PowerObject` from `Exponentials` in `PowerObject.v`;
- The basic example `is_Topos_HSET` that `HSET` is a Topos in `HSET/Structures.v` using the constructions there defined


UniMath/CategoryTheory/categories/HSET/Structures.v
```diff
@@ -604,3 +608,20 @@ Proof.
       * apply propproperty.
       * apply propproperty.
 Defined.
+
+Definition is_Topos_HSET : is_Topos HSET.
```
benediktahrens, line 612: Being a topos is structure on a category, not a property. It would hence be better to rename `is_Topos` into `Topos_Structure` or similar. Specifically, `is` and `has` should be reserved for predicates.

benediktahrens (approved):
I think some of the constructions here could (should?) be made partially opaque, either by using the `abstract (...)` tactical or by moving some stuff into a separate lemma ended with `Qed.`.

CalosciMatteo:
> Thank you for preparing this!
> 
> Sanity-Checks fails [here](https://github.com/UniMath/UniMath/actions/runs/4404476278/jobs/7714129688#step:4:35) with:
> 
> ```
>  --- checking the ordering prescribed by the files UniMath/*/.packages/files ---
> error: *** UniMath/CategoryTheory/categories/HSET/Structures.vo should not require UniMath/CategoryTheory/ElementaryTopos.vo
> error: *** UniMath/CategoryTheory/categories/HSET/Structures.vo should not require UniMath/CategoryTheory/PowerObject.vo
> ```

Understood, I'm working on fixing this.

UniMath/CategoryTheory/categories/HSET/Structures.v
CalosciMatteo: I understand, I will start working on this refactor as soon as possible

CalosciMatteo: Done!

benediktahrens: I think I missspoke here. If a category C is univalent, then there might be at most one topos structure on it. So the previous wording could have been ok. Sorry for that!

CalosciMatteo: I understand, should I revert the last commit then? Personally I think the new name is ok too, since your consideration above is definitely right if we work in a `category` (in the UniMath sense of the term).

benediktahrens: I think both names are fine.

m-lindgren (approved):
Nice improvements, looks good overall.

PR #1659, opened 2023-03-15
More examples of enriched categories

nmvdw:
This pull request contains:
- A fix of the definition of cartesian and cocartesian monoidal category
- A proof that cartesian monoidal categories are symmetric
- A characterization of enrichments over `Set`
- A characterization of enrichments over `Posets`
- A notion of structure over Sets that gives rise to a monoidal category
- A characterization of categories enriched over such structures. So, if we look at some notion of structure of `hSet`, then we can simplify the definition of what it means to be enriched over that kind of structure.
- Product of enriched categories (assuming that the monoidal category is cartesian)

There are some definitions that need to be moved, namely:
- ~~Lines 41-360 in Posets.v~~
- ~~Lines 47-103 in StructureWithProd.v~~

I don't know yet where to put them.


UniMath/CategoryTheory/DisplayedCats/Examples/Posets.v
```diff
@@ -0,0 +1,532 @@
+(*****************************************************************
+
+ The category of posets
```
benediktahrens, line 3: I often find preorders (without anti-symmetry) more helpful than partial orders. In some sense, the former are simpler, because they don't mention equality of objects. I am wondering if it would be useful to have this intermediate structure of preordered sets.

A "univalent preorder" is anti-symmetric. I am not sure if you are interested in that notion, though.

UniMath/CategoryTheory/DisplayedCats/Examples/StructureWithProd.v
```diff
@@ -0,0 +1,686 @@
...
+ The notion of structure that we consider, consists of:
+ - For every hSet, a set of structures on that set
+ - For every function, a proposition that represents wheter this
+   function preserves the structure.
+ The notion of structure must be closed under product and there
+ must be a structure for the unit set. We also require that
+ structure-preserving maps are closed under identity, composition,
+ constant functions, and pairing. We also require that projections
+ and the map to the unit set are structure preserving. The final
+ requirement is the notion of 'standardness' (see the HoTT book),
```
benediktahrens, line 19: I am not sure if it would be useful here: we do have the notion of structure on a category of the HoTT book formalized somewhere in UniMath, and the construction of a displayed category from it.

That notion does not require many of the properties with respect to limits that you are asking for. Maybe we should have a systematic development of displayed limits etc.

UniMath/CategoryTheory/EnrichedCats/Examples/SetEnriched.v
```diff
@@ -0,0 +1,266 @@
+(*****************************************************************
+
+ Enrichment over the category of sets
+
+ One of the simplest category over which we can study enrichments,
```
benediktahrens, line 5:
```suggestion
 One of the simplest categories over which we can study enrichments
```

UniMath/CategoryTheory/EnrichedCats/Examples/SetEnriched.v
```diff
@@ -0,0 +1,266 @@
...
+ is the category of sets. In this file, we show that enrichments
+ over the category of sets exist and that they are unique. From
+ that, we conclude that the notions of category, functor, and
+ nautlra transformations are equivalent to their enrichment
```
benediktahrens, line 9:
```suggestion
 natural transformations are equivalent to their enrichment
```

UniMath/CategoryTheory/EnrichedCats/Examples/StructureEnriched.v
```diff
@@ -0,0 +1,505 @@
+(*****************************************************************
+
+ Enrichment over structured sets
+
+ In this file, we look at categories enriched over a notion of
+ structured sets. The notion of structured sets in consideration
+ here, is designed so that structures are preserved under product
```
benediktahrens, line 7:
```suggestion
 here is designed so that structures are preserved under product
```

benediktahrens (approved):
Cool stuff! I am looking forward to reading a paper about univalent enriched categories!

benediktahrens:
Regarding the question of where to put the preliminary material: maybe every file in `Foundations` can have a corresponding file (with the same name?) in `MoreFoundations` which collects more material of the same kind?

UniMath/CategoryTheory/DisplayedCats/Examples/StructureWithProd.v
nmvdw: I saw it in here: https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/DisplayedCats/SIP.v, but that file did not have a definition of a structure. That's why I added a new type.

UniMath/CategoryTheory/DisplayedCats/Examples/Posets.v
nmvdw: Posets were chosen for no particular reason. This file is more a preliminary development to find ideas/methods than a 'final result'. For both notions we could use the material in `StructureEnriched.v` (which was inspired by what I did for posets).

nmvdw:
> Regarding the question of where to put the preliminary material: maybe every file in `Foundations` can have a corresponding file (with the same name?) in `MoreFoundations` which collects more material of the same kind?

That's a valid point. However, the stuff about posets could also be put in the Algebra directory, which might be a more sensible place.

UniMath/CategoryTheory/DisplayedCats/Examples/StructureWithProd.v
nmvdw: Note: a more systematic development specialized at structures over `hSet`, would be very welcome, I think. For this particular instance, I believe that one can also develop a nice notion of displayed exponentials.

nmvdw:
I moved the stuff in lines 47-103 of StructureWithProd.v to MoreFoundations, where it fits better.

nmvdw:
> > > Regarding the question of where to put the preliminary material: maybe every file in `Foundations` can have a corresponding file (with the same name?) in `MoreFoundations` which collects more material of the same kind?
> > 
> > 
> > That's a valid point. However, the stuff about posets could also be put in the Algebra directory, which might be a more sensible place.
> 
> There is also `Combinatorics/OrderedSets`.

That's also a good point, but that file is already a bit big. Would it make sense to make a file `Combinatorics/Posets.v` which contains some development of `Posets`?

benediktahrens:
> That's also a good point, but that file is already a bit big. Would it make sense to make a file `Combinatorics/Posets.v` which contains some development of `Posets`?

Yes, absolutely!

nmvdw:
> > That's also a good point, but that file is already a bit big. Would it make sense to make a file `Combinatorics/Posets.v` which contains some development of `Posets`?
> 
> Yes, absolutely!

Sure. I moved it now

PR #1660, opened 2023-03-16
Change displayed functor notation.

m-lindgren:
As @Kfwullaert pointed out over on [zulip](https://unimath.zulipchat.com/#narrow/stream/322787-Help-needed-when-working-with-UniMath.2FCoq/topic/Using.20the.20same.20notation.20for.20different.20purposes/near/314317227) when working with displayed categories one frequently gets in trouble because the notation `# _` is used both for functors applied to morphisms and displayed functors applied to displayed morphisms.

This PR changes the notation in the displayed case to `♯ _`, which can be input by either `\#` or `\sharp`. This also and marks the `# _` notation as deprecated in `mor_disp_scope`. This way users  have a chance to remove it and we can get rid of it completely by next release.

Also `\sharp` is merely a suggestion, if anyone has a better (more readable / distinct?) suggestion I can search-and-replace it in this PR.


benediktahrens:
@m-lindgren : thanks a lot for preparing this! LGTM

benediktahrens (approved):

rmatthes (approved):
a very welcome move

PR #1662, opened 2023-03-17
Self-enriched categories

nmvdw:
This PR contains
- Categories of structured sets with exponentials
- Infrastructure for symmetric monoidal closed categories
- Every cartesian closed category gives rise to a symmetric monoidal category
- Self-enriched categories
- Every strong monad gives rise to an enriched monad on the self-enriched category


m-lindgren (approved):

UniMath/CategoryTheory/exponentials.v
```diff
@@ -18,11 +18,8 @@ Section [ExponentialsCarriedThroughAdjointEquivalence] added by Ralph Matthes in
 
 ************************************************************)
 
-Require Import UniMath.Foundations.PartD.
-Require Import UniMath.Foundations.Propositions.
-Require Import UniMath.Foundations.Sets.
-
-Require Import UniMath.MoreFoundations.Tactics.
+Require Import UniMath.Foundations.All.
+Require Import UniMath.MoreFoundations.All.
```
benediktahrens, line 22: How good is it to import `All` instead of a more fine-grained selection?

benediktahrens (approved):

UniMath/CategoryTheory/exponentials.v
nmvdw: Because `maponpaths2` is not imported then (which I needed in the file).

UniMath/CategoryTheory/EnrichedCats/Examples/SetEnriched.v
```diff
@@ -22,7 +22,6 @@ Require Import UniMath.CategoryTheory.Core.Isos.
 Require Import UniMath.CategoryTheory.Core.Univalence.
 Require Import UniMath.CategoryTheory.Core.Functors.
 Require Import UniMath.CategoryTheory.Core.NaturalTransformations.
-Require Import UniMath.CategoryTheory.Core.
 Require Import UniMath.CategoryTheory.categories.HSET.All.
```
m-lindgren, line 25: Too many times have I imported `UniMath.CategoryTheory.Core` and been surprised that `Definition foo : category` doesn't work.

nmvdw: Yes. The problem here is that `CategoryTheory.Core` contains the core of a category. It should be put in a place where it makes more sense (some folder where we basically put all examples of categories), but as far as I know, we don't have such a folder now.

PR #1665, opened 2023-03-19
Remove some unnecessary imports.

m-lindgren:
Prompted by #1664 I decided to try [this script](https://github.com/HoTT/Coq-HoTT/blob/master/etc/coqcstriprequires.py) created by the maintainers of Coq-HoTT. It worked surprisingly well, I only had to increase the timeout values and run it with `-j1` to avoid some inconsistencies.

If I had done this again I would have configured the script not to remove imports of anything in `Foundations`. Half way through I aborted the run and started it again with this configuration. 

I did not run the script on all of UniMath, in particular I avoided running it on `Bicategories`. That will have to be a separate PR.

I'm not entirely sure that import minimization is effective for build speed improvement. This script doesn't take into consideration transitivity of imports, so many of these removals are effectively no-ops from a compilation standpoint, but some instances I believe this can help prevent some unnecessary rebuilds triggered by changes to dependencies (in particular in DisplayedCats and `SubstitutionSystems`).

See [my comment](https://github.com/UniMath/UniMath/issues/1664#issuecomment-1475279477) in #1664.

PR #1666, opened 2023-03-19
Clean up cache files after merge.

m-lindgren:
This adds a workflow that runs whenever a PR is closed, including if it's merged into master. It then deletes any cache files owned by that PR.


m-lindgren:
If this gets merged before #1665 hopefully we can see if it works as intended when that PR gets merged.

To see current cache files using https://cli.github.com/ you can do `gh actions-cache list` and for example: `gh actions-cache list -B refs/pull/1665/merge`

benediktahrens:
@m-lindgren : thanks a lot!

Issue #648, opened 2017-03-10
file by file type in type


DanGrayson:
VV: " Basics do not require type-in-type and maybe we can make it clear even with the existing Coq."

I'll see if I can set this up in the Makefile.

mortberg:
Just so you know, I switched off type-in-type just to see where building UniMath breaks a while ago. It breaks already in PartA:
```
COQC UniMath/Foundations/PartA.v
File "./UniMath/Foundations/PartA.v", line 971, characters 16-17:
Error:
In environment
X : UU
P : X → UU
Q : X → UU
The term "P" has type "X → UU" while it is expected to have type
 "∏ x : X, ?P x" (unable to find a well-typed instantiation for 
"?P": cannot ensure that "X → Type" is a subtype of
"?X1074@{__:=X; __:=P; __:=Q} → UU").
build/CoqMakefile.make:483: recipe for target 'UniMath/Foundations/PartA.vo' failed
make: *** [UniMath/Foundations/PartA.vo] Error 1
```

On line 971 we find:
```
Definition famhomotfun {X : UU} {P Q : X -> UU}
           (h : P ~ Q) (xp : total2 P) : total2 Q.
```

I'm not exactly sure what VV means by "Basics", but making this change is probably going to be a bit more work than just changing the Makefile.

DanGrayson:
Yes, I encountered the same thing.  We have lots of functions that insist on their arguments being in UU, which is often not going to be the case when universe checking is turned on.

vladimirias:
A bit more work probably.  In this case “famhomotfun” should not be in PartA because it requires the concept of homotopy for functions between types outside of UU and all of PartA should be strictly about what would be, in a type theory with good universe management, about types in a single universe.

Vladimir.

DanGrayson:
Or just don't apply `famhomotfun` to functions whose target is `UU`.

mortberg:
Re: "all of PartA should be strictly about what would be, in a type theory with good universe management, about types in a single universe."

Ok, I think PartA was called uu0 before? I guess this was the reason for that name?

mortberg:
If we are going to switch off type-in-type from parts of UniMath it might be useful to look at the following code from the HoTT library:

https://github.com/HoTT/HoTT/blob/master/theories/Basics/Overture.v#L87
https://github.com/HoTT/HoTT/blob/master/theories/Basics/UniverseLevel.v

[...]

PR #1671, opened 2023-03-24
WIP: Add Alectryon to build system to allow for automated creation of interactive HTML documentation.

Bigstep22:
I would like to add a new make rule that allows one to use [Alectryon](https://github.com/cpitclaudel/alectryon) to generate interactive HTML documentation of UniMath code, such that someone does not need to install Coq to see what the code does.

So far I've got a full build for Alectryon working with a few details of note:
- Alectryon does not allow the use of the -type-in-type compiler flag. There is a simple solution to this by simply adding the following line to the beginning of Foundations/Init.v and removing the use of the -type-in-type flag in the build process.
```
Global Unset Universe Checking.
```
According to this following [coq documentation](https://coq.inria.fr/refman/proof-engine/vernacular-commands.html#coq:flag.Universe-Checking), Unsetting Universe Checking is the same as using the -type-in-type flag.
I am aware that there is existing discussion about the -type-in-type flag in [#648 ](https://github.com/UniMath/UniMath/issues/648). But I do not propose changing its use, only how its use is expressed in the build process to accomodate Alectryon's limitations.
- I am not very familiar with Make and I do not yet fully understand the build process and where I'd place the needed command for Alectryon in a (which one?) makefile. I do understand the basics of how coqdep and coq_makefile work. Alectryon needs all of the files to be compiled passed to it, in the right order.

The second point is the one I would appreciate some help on. Does anyone have greater experience regarding UniMath's build system and could guide me in the right direction regarding integrating Alectryon into it?


m-lindgren:
Cool that you're working on adding alectryon support, that is very welcome!

So UniMath currently has two build systems in use, `dune` and `make` that work independently of each other. Both can be used locally, but only`dune` is used in the CI to build UniMath. To see if the CI compiles with the change you did you should also remove `-type-in-type` from here:

https://github.com/UniMath/UniMath/blob/3d2ff1214a72844dbb259a0e1718e1b1bae2d724/UniMath/dune#L7-L11

If you use `make` to build locally I think you need to change this line: 
https://github.com/UniMath/UniMath/blob/3d2ff1214a72844dbb259a0e1718e1b1bae2d724/Makefile#L106

As to your question on where to add the commands to call alectryon I am not sure currently, I need to look into it. Is the idea that it generates it when the CI is run, so that documentation is available on unimath.org, or are you asking how to invoke it so that the documentation is built locally for users?

m-lindgren:
> * Alectryon needs all of the files to be compiled passed to it, in the right order.

So I think one way of doing this with dune would be to add a custom rule https://dune.readthedocs.io/en/stable/dune-files.html#rule, a simple example of this is in the source tree already:
https://github.com/UniMath/UniMath/blob/3d2ff1214a72844dbb259a0e1718e1b1bae2d724/UniMath/dune#L13-L17

this generates the file `All.v`, by calling the script `utils/generate-exports` with all the files in the source tree as arguments and writing the output to `All.v`. I'm not sure about the order though. 

I think you could do something similar, but using directory targets (because I assume Alectryon generates multiple files, not just one big one, right?) as described here: https://dune.readthedocs.io/en/stable/dune-files.html#directory-targets

Bigstep22:
Thanks for you answer, I'll definitiely look into removing the -type-in-type flag for both build systems.

As to your question at the end:
> As to your question on where to add the commands to call alectryon I am not sure currently, I need to look into it. Is the idea that it generates it when the CI is run, so that documentation is available on unimath.org, or are you asking how to invoke it so that the documentation is built locally for users?
I meant for local building. Integrating Alectryon into the fully automated pipeline would be the next step once everything is working fine locally.

I'll have a look around Coq Zulipchat. Maybe they'll have some insight into how the make process works.

benediktahrens:
What is missing in this PR? I believe that at least a suitable entry in the Makefile is missing - anything else?

I am not sure we need this to work with both `dune` and `make` - one of them should be sufficient? The primary goal could be to build the documentation locally whenever needed. I don't think it would be good to build the alectryon documentation on every build - the output is likely too large to be stored.

benediktahrens:
@DanGrayson : would you be able to assist @Bigstep22 in creating a target in the Makefile?

DanGrayson:
> @DanGrayson : would you be able to assist @Bigstep22 in creating a target in the Makefile?

Okay.  What command should be used for generating the documentation?

nmvdw:
What remains to be done before this can be merged? Are there people working on this?

arnoudvanderleer:
@nmvdw I have not seen any progress on this since last year. However, it seems that
 - [Alectryon does not support -type-in-type](https://github.com/UniMath/UniMath/pull/1671)
 - [There was some discussion about whether and how we should eliminate -type-in-type](https://github.com/UniMath/UniMath/issues/1696)
 - @peterlefanulumsdaine has made [an attempt at eliminating -type-in-type in one of the proposed way](https://github.com/UniMath/UniMath/pull/1697)
 - This last effort has stranded somewhere

nmvdw:
Alectryon does offer some benefits, because it gives the possibility to publish documentation in which the reader can see the goal. This makes the documentation more readable, because readers can see intermediate goals and can understand what happens in the proof. This is definitely affected by the fact that a reviewer mentioned this in a review of my paper, so I would like to put in some effort for Alectryon documentation.

When we were discussing replacing `type-in-type` with `Unset Universe Checking`, the discussion quickly moved to 'can we do this locally', if I recall correctly. This is something that turned out not to work as we wished. 

In addition, a disadvantage of this change was that certain satellites (TypeTheory and SetHITs) did not depend on `type-in-type`, and by moving to `Unset Universe Checking` they would. However, by changes in Coq, both satellites need to move to `type-in-type` by now anyway, but this was not the case at the time of the discussion. So, this disadvantage was relevant at the time, but not anymore.

Alectryon requires us to move from `type-in-type` to `Unset Universe Checking`. The mentioned disadvantages were relevant at the time of the discussion, but due to changes in Coq. In addition, there is a clear advantage and a reason for the change. For these reasons, I think we should reconsider this discussion.

arnoudvanderleer:
Then there was a bit of discussion how global the `Universe Checking` should be. It is reminiscent of discussions about other flags: whether they should be exported (or added separately to every single file), because then any project that uses UniMath automatically inherits those flags and people who use UniMath outside the repository might need to do some work to turn them off again.

nmvdw:
> Then there was a bit of discussion how global the `Universe Checking` should be. It is reminiscent of discussions about other flags: whether they should be exported (or added separately to every single file), because then any project that uses UniMath automatically inherits those flags and people who use UniMath outside the repository might need to do some work to turn them off again.

Yes. However, this is a point that needs to be reconsidered. We see that TypeTheory and SetHITs fail in the CI when importing UniMath, because they don't use `type-in-type`.

arnoudvanderleer:
All right, I will close the old issues (because part of that may have gone stale), and will open a new one for the rest of this discussion.

nmvdw:
I am experimenting a bit with Alectryon, and my impression is that you need to add `Unset Universe Checking` to **every** file in order to get the interaction points to work properly. Otherwise the interaction points, do not get displayed, because Alectryon most likely meets an error in the file (due to not using `type-in-type`). It doesn't seem to remember the `#[global] Unset Universe Checking` from other files, because it does not go through each of them, but just a single one.

arnoudvanderleer:
Oh, that is very unfortunate

nmvdw:
It is also the case that the documentation will take more space. For instance, the file `Bicategories/Monads/Examples/Composition.v` takes around 208 kb if you use coqdoc, but it becomes 2.8 MB is you use Alectryon.

benediktahrens:
@nmvdw : thanks for doing these experiments and recording your findings here. It seems strange that Alectryon does not remember global flags from other files. Maybe it would be good to get help from Alectryon developers?

nmvdw:
I just realized that I missed something in my tests. Alectryon considers a file on its own: if you do `Require Import MyLeastFavoriteFile.`, then it looks for `MyLeastFavoriteFile` in the 'global imports',  So, you must have done `make install` before hand.  I forgot this during my experiments, meaning that I have to try it again....

nmvdw:
> I just realized that I missed something in my tests. Alectryon considers a file on its own: if you do `Require Import MyLeastFavoriteFile.`, then it looks for `MyLeastFavoriteFile` in the 'global imports', So, you must have done `make install` before hand. I forgot this during my experiments, meaning that I have to try it again....

Oh, this seems to be the probleem indeed. So, it does work if you use `#[global] Unset Universe Checking`. However, it is necessary to use `make install` to install UniMath. If you use `make`, then it won't fine the right imports. 

So, to add Alectryon support to UniMath:
- Switch to `#[global] Unset Universe Checking`
- In the CI, `make install` should be used instead of `make`
- For Alectryon, the order in which the files are considered, doesn't matter, because it compiles a single file and not a whole project. The imports come from `make install`. 

The most interesting part would be: how big will the generated `html` files be. This can actually turn out to be quite large (I would not be surprised by several GBs).

nmvdw:
I managed to compile the whole of UniMath using Alectryon. The size of the documentation is actually as bad as I expected, namely 759 MB, but this is still a lot. In addition, the most straightforward way to compile it (`alectryon --frontend coq *.v */*.v */*/*.v */*/*/*.v */*/*/*/*.v --output-directory alectryon`) has some issues, namely naming and the usage of a single core (it would compile UniMath using a single core, which takes ages). Finally, there are some errors during compilation. Some symbols, like `≃`, are not accepted by the lexer of Alectryon (improvements on Alectryon's side). There also were errors in the files `Algebra/Groups2.v`, `Ktheory/GrothendieckGroup.v`, `SyntheticHomotopyTheory/AffineLine.v`. The exact output is in this file:

[output.txt](https://github.com/user-attachments/files/16959091/output.txt)

So, to incorporate Alectryon to UniMath:
- We should switch to `#[global] Unset Universe Checking`
- The compilation script in  the`MakeFile` should handle the usage of multiple cores by itself. The order of the files does not matter. The compilation only needs a list of all files (preferably with the names of the directories in which they are).

Also, we need to think what to do with the documentation. Should the Alectryon documentation be put on the UniMath website?

nmvdw:
It does not seem to be possible to point to identifiers in the documentation compiled using Alectryon. The html code generated for a single definition would be something like:
```html
<span class="alectryon-sentence">
  <span class="alectryon-input">
    <span class="kn">Definition</span> 
    <span class="nf">appel</span>
    {<span class="nv">A</span> : <span class="kt">Type</span>}
    (<span class="nv">a</span> : A) : A := a
</span>
<span class="alectryon-wsp">
</span>
</span>
```
Nothing in this piece of code allows one to link to this identifier.

arnoudvanderleer:
Closing this, as replaced by #2032

PR #1653, opened 2023-03-12
Small follow-up PR to #1643 : remove more uses of Coq.Init.Logic

Skantz:
Hi,

This is a small follow-up PR addressing some imports of Coq.Init.Logic, in addition to those already removed in #1643. As I understand, only one proof remains to be amended after this.

Thanks  @m-lindgren for submitting #1643. Thanks @benediktahrens for help with this PR (I amended two more proofs compared to the [suggested PR](https://github.com/UniMath/UniMath/pull/1643#issuecomment-1465249011), but can also revert those additional changes if preferable).

Work done:

- UniMath/Algebra/GaussianElimination/Elimination.v : remove import and amend proofs
- UniMath/Algebra/GaussianElimination/Corollaries.v : amend proofs that needed Coq.Init.Logic (this file imported it from Elimination.v)
- UniMath/CategoryTheory/categories/HSET/Slice.v : remove import and amended one proof
- UniMath/SubstitutionSystems/MonadsMultiSorted.v : amend one proof that needed Coq.Init.Logic (previously imported from Slice.v)

 
Work remaining:

- UniMath/CategoryTheory/Chains/OmegaCocontFunctors.v : needs Coq.Init.Logic (at `rewrite h2` in the proof for `map_to_K_commutes`), so I add the import at the start of this file.

PR #1674, opened 2023-03-25
Remove last use of Coq.Init.Logic

Skantz:
Here is another small PR as a follow-up to #1643, addressing the remaining work after #1653 to remove the last use of Coq.Init.Logic. It is a small tweak to one proof in `OmegaCocontFunctors.v`.


m-lindgren (approved):

m-lindgren:
This is great, now we can finally close #827 which has been around since 2017. Thanks a lot!

PR #1676, opened 2023-03-30
Add a package about Algebraic Theories

arnoudvanderleer:
Adds the first definitions and a proof in a package about algebraic theories, based on Hyland's paper [Classical lambda calculus in modern dress](https://arxiv.org/pdf/1211.5762.pdf).


UniMath/AlgebraicTheories/AlgebraicBases.v
```diff
@@ -0,0 +1,22 @@
+Require Import UniMath.Foundations.All.
```
nmvdw, line 1: This file can get some more documentation. You could say what an algebraic base is supposed to represent.

UniMath/AlgebraicTheories/AbstractClones/AbstractClones.v
```diff
@@ -0,0 +1,98 @@
+Require Import UniMath.Foundations.All.
+Require Import UniMath.MoreFoundations.All.
+Require Import UniMath.Combinatorics.StandardFiniteSets.
+
+Require Import UniMath.AlgebraicTheories.AlgebraicBases.
+
+Local Open Scope algebraic_theory.
+
+Definition abstract_clone_data := ∑ (B: algebraic_base),
```
nmvdw, line 9: https://ncatlab.org/nlab/show/clone#definition

An alternative presentation of an abstract clone is as a relative monad for the embedding of finite sets (constructively, one would use Bishop finite sets, I think) into sets. For this definition, the involved categories are univalent, while the finite set skeleton category is strict.

Are you planning to study this presentation as well?

UniMath/AlgebraicTheories/AbstractClones/AbstractClones.v
```diff
@@ -0,0 +1,98 @@
...
+
+Definition make_abstract_clone_data
+  (C : algebraic_base)
+  (pr : ∏ {n}, stn n → C n)
+  : abstract_clone_data.
+Proof.
+  exact (C ,, pr).
+Defined.
+
+Definition algebraic_base_from_abstract_clone (C : abstract_clone_data) : algebraic_base := pr1 C.
```
nmvdw, line 20: You can make this function a coercion in 1 line:

```suggestion
Coercion algebraic_base_from_abstract_clone (C : abstract_clone_data) : algebraic_base := pr1 C.
```

UniMath/AlgebraicTheories/AbstractClones/AbstractClones.v
```diff
@@ -0,0 +1,98 @@
...
+Coercion algebraic_base_from_abstract_clone : abstract_clone_data >-> algebraic_base.
+
+Definition pr {C : abstract_clone_data} {n : nat} := pr2 C n.
```
nmvdw, line 23: Can this function be renamed (for example, to `abstract_clone_pr`)? And can you add the type of this function in the code?

UniMath/AlgebraicTheories/AbstractClones/AbstractClones.v
```diff
@@ -0,0 +1,98 @@
...
+
+Definition reindex
+  {C : abstract_clone_data}
+  {m n : nat}
+  (a : stn m → stn n)
+  : C m → C n
+  := (λ f, f • (λ i, pr (a i))).
+
+(* Define the unitality property of the algebraic theory *)
+Definition comp_project_component (C : abstract_clone_data) : Prop := ∏
```
nmvdw, line 33: `hProp` is not the same as `Prop`: the former is about types of which all inhabitants are equal, while the second is some kind of impredicate universe. In UniMath, we don't use `Prop`, so I think you want to use `hProp` here instead.

UniMath/AlgebraicTheories/AbstractClones/AbstractClones.v
```diff
@@ -0,0 +1,98 @@
...
+  (m n : nat)
+  (i : stn m)
+  (f : stn m → C n),
+    (pr i) • f = f i.
+
+(* Define the compatibility of the projection function with composition *)
+Definition comp_identity_projections (C : abstract_clone_data) : Prop := ∏
```
nmvdw, line 40: `Prop` -> `hProp`

UniMath/AlgebraicTheories/AbstractClones/AbstractClones.v
```diff
@@ -0,0 +1,98 @@
...
+  (n : nat)
+  (f : C n),
+    f • (λ i, pr i) = f.
+
+(* Define the associativity property of the algebraic theory *)
+Definition comp_is_assoc (C : abstract_clone_data) : Prop := ∏
```
nmvdw, line 46: `Prop` -> `hProp`

UniMath/AlgebraicTheories/AbstractClones/AbstractClones.v
```diff
@@ -0,0 +1,98 @@
...
+
+Definition abstract_clone := total2 is_abstract_clone.
+
+Definition make_abstract_clone
+  (T : abstract_clone_data)
+  (H : is_abstract_clone T)
+  : abstract_clone
+  := (T ,, H).
+
+Definition abstract_clone_data_from_abstract_clone : abstract_clone -> abstract_clone_data := pr1.
```
nmvdw, line 81:
```suggestion
Coercion abstract_clone_data_from_abstract_clone : abstract_clone -> abstract_clone_data := pr1.
```

UniMath/AlgebraicTheories/AbstractClones/AbstractClones.v
```diff
@@ -0,0 +1,98 @@
...
+Coercion abstract_clone_data_from_abstract_clone : abstract_clone >-> abstract_clone_data.
+
+Lemma abstract_clone_eq
+  (X Y : abstract_clone)
+  (H1 : pr111 X = pr111 Y)
+  (H2 : transportf _ H1 (pr211 X) = (pr211 Y))
+  (H3 : transportf (λ (T : nat → hSet), (∏ n, stn n → T n)) H1 (pr21 X) = (pr21 Y))
+  : X = Y.
+Proof.
+  (* Is it possible to do this proof without destruct? *)
```
nmvdw, line 91: `subtypePath` and `total2_paths_f` might be useful here

UniMath/AlgebraicTheories/AbstractClones/AbstractCloneCategory.v
```diff
@@ -0,0 +1,98 @@
...
+Require Import UniMath.AlgebraicTheories.AlgebraicBases.
+Require Import UniMath.AlgebraicTheories.AbstractClones.AbstractClones.
+Require Import UniMath.AlgebraicTheories.AbstractClones.AbstractCloneMorphisms.
+
+Definition id_morphism_data
+(C : abstract_clone_data)
+: abstract_clone_morphism_data C C
+:= λ _ x, x.
+
+Lemma id_is_morphism (C : abstract_clone_data) : is_abstract_clone_morphism (id_morphism_data C).
```
nmvdw, line 16: To avoid having to unfold `id_morphism_data`, one can fill in its definition directly into this type. This would make the type a bit uglier, but one does not have to unfold `id_morphism_data`. Both solutions are fine though.

UniMath/AlgebraicTheories/AbstractClones/AbstractCloneCategory.v
```diff
@@ -0,0 +1,98 @@
...
+Qed.
+
+Definition abstract_clone_precategory_data : precategory_data.
+Proof.
+  use make_precategory_data.
+  - use make_precategory_ob_mor.
+    + exact abstract_clone.
+    + exact abstract_clone_morphism.
+  - simpl.
+    exact (λ C, make_abstract_clone_morphism (id_morphism_data C) (id_is_morphism C)).
```
nmvdw, line 62: What about having a definition for `id_morphism` of the type `abstract_clone_morphism`? The same for composition.

UniMath/AlgebraicTheories/AlgebraicTheories/AlgebraicTheories.v
```diff
@@ -0,0 +1,172 @@
...
+  (d : algebraic_theory_data)
+  : algebraic_base
+  := pr1 d.
+Coercion algebraic_base_from_algebraic_theory_data : algebraic_theory_data >-> algebraic_base.
+
+Definition e {T : algebraic_theory_data} : T 1 := pr12 T.
+
+Definition Tmor {T : algebraic_theory_data} {m n} : (stn m → stn n) → T m → T n := pr22 T m n.
+
+Definition pr {T : algebraic_theory_data} {n : nat} (i : stn n) : T n := Tmor (λ (x : stn 1), i) e.
```
nmvdw, line 33: The name `pr` is also used in `AbstractClones`, so if you have both files in scope, it becomes annoying to use (one has to explicitly mention the module).

UniMath/AlgebraicTheories/AlgebraicTheories/AlgebraicTheories.v
```diff
@@ -0,0 +1,172 @@
...
+  exact (T ,, e ,, Tmor).
+Defined.
+
+Definition algebraic_base_from_algebraic_theory_data
+  (d : algebraic_theory_data)
+  : algebraic_base
+  := pr1 d.
+Coercion algebraic_base_from_algebraic_theory_data : algebraic_theory_data >-> algebraic_base.
+
+Definition e {T : algebraic_theory_data} : T 1 := pr12 T.
```
nmvdw, line 29: A more descriptive name would be better.

UniMath/AlgebraicTheories/AlgebraicTheories/AlgebraicTheories.v
```diff
@@ -0,0 +1,172 @@
...
+
+Definition pr {T : algebraic_theory_data} {n : nat} (i : stn n) : T n := Tmor (λ (x : stn 1), i) e.
+
+Definition algebraic_theory_data_to_functor_data
+  (T : algebraic_theory_data)
+  : functor_data finite_set_skeleton_category HSET
+  := make_functor_data (T : finite_set_skeleton_category → HSET) (@Tmor T).
+
+(* Define the associativity property of the algebraic theory *)
+Definition comp_is_assoc (T : algebraic_theory_data) : Prop := ∏
```
nmvdw, line 41: This should be of type `hProp`.

UniMath/AlgebraicTheories/AlgebraicTheories/AlgebraicTheories.v
```diff
@@ -0,0 +1,172 @@
...
+  (l m n : nat)
+  (f_l : T l)
+  (f_m : stn l → T m)
+  (f_n : stn m → T n),
+    (f_l • f_m) • f_n = f_l • (λ t_l, (f_m t_l) • f_n).
+
+(* Define the unitality property of the algebraic theory *)
+Definition comp_is_unital (T : algebraic_theory_data) : Prop := ∏
```
nmvdw, line 49: `Prop` -> `hProp`

UniMath/AlgebraicTheories/AlgebraicTheories/AlgebraicTheories.v
```diff
@@ -0,0 +1,172 @@
...
+  (n : nat)
+  (f : T n),
+    e • (λ _, f) = f.
+
+(* Define the compatibility of the projection function with composition *)
+Definition comp_identity_projections (T : algebraic_theory_data) : Prop := ∏
```
nmvdw, line 55: `Prop` -> `hProp`

UniMath/AlgebraicTheories/AlgebraicTheories/AlgebraicTheories.v
```diff
@@ -0,0 +1,172 @@
...
+  (H2 : transportf _ H1 (pr211 X) = pr211 Y)
+  (H3 : transportf (λ (T : nat → hSet), T 1) H1 (pr121 X) = (pr121 Y))
+  (H4 : transportf
+    (λ (T : nat → hSet), ∏ m n, (stn m → stn n) → T m → T n)
+    H1
+    (pr221 X) = (pr221 Y)
+  )
+  : X = Y.
+Proof.
+  destruct X as [[[Xf Xcomp] [Xe Xmor]] HX].
```
nmvdw, line 118: Here `subtypePath` and `total2_paths_f` might be useful.

UniMath/AlgebraicTheories/AlgebraicTheories/AlgebraicTheoryMorphisms.v
```diff
@@ -0,0 +1,97 @@
...
+
+Definition algebraic_theory_morphism_data_to_nat_trans_data
+  {T T'}
+  (F : algebraic_theory_morphism_data T T')
+  : nat_trans_data
+    (algebraic_theory_data_to_functor_data T)
+    (algebraic_theory_data_to_functor_data T')
+  := F.
+
+Definition preserves_projections {T T'} (F : algebraic_theory_morphism_data T T') : Prop := ∏
```
nmvdw, line 25: `Prop` -> `hProp`

UniMath/AlgebraicTheories/AlgebraicTheories/AlgebraicTheoryMorphisms.v
```diff
@@ -0,0 +1,97 @@
...
+  (n : finite_set_skeleton_category)
+  (i : stn n),
+    (F n (pr i)) = (pr i).
+
+Definition preserves_composition {T T'} (F : algebraic_theory_morphism_data T T') : Prop := ∏
```
nmvdw, line 30: `Prop` -> `hProp`

UniMath/AlgebraicTheories/AlgebraicTheories/AlgebraicTheoryMorphisms.v
```diff
@@ -0,0 +1,97 @@
...
+  (m n : finite_set_skeleton_category)
+  (f_m : T m)
+  (f_n : stn m → T n),
+  (F n (comp f_m f_n)) = (comp (F m f_m) (λ i, F n (f_n i))).
+
+Definition is_algebraic_theory_morphism
```
nmvdw, line 36: `Prop` -> `hProp`

UniMath/AlgebraicTheories/AlgebraicTheories/AlgebraicTheoryMorphisms.v
```diff
@@ -0,0 +1,97 @@
...
+  := total2 (@is_algebraic_theory_morphism T T').
+
+Definition make_algebraic_theory_morphism 
+  {T T' : algebraic_theory} 
+  (F : algebraic_theory_morphism_data T T') 
+  (H : is_algebraic_theory_morphism F)
+  : algebraic_theory_morphism T T'
+  := (F ,, H).
+
+Definition algebraic_theory_morphism_to_algebraic_theory_morphism_data
```
nmvdw, line 74:
```suggestion
Coercion algebraic_theory_morphism_to_algebraic_theory_morphism_data 
```

UniMath/AlgebraicTheories/AbstractCloneAlgebraicTheory.v
```diff
@@ -0,0 +1,152 @@
...
+Qed.
+
+Lemma algebraic_theory_weq_abstract_clone : abstract_clone ≃ algebraic_theory.
+Proof.
+  use (algebraic_theory_from_abstract_clone ,, _).
+  use isweq_iso.
+  - exact abstract_clone_from_algebraic_theory.
+  - exact abstract_clone_id.
+  - exact algebraic_theory_id.
+Qed.
```
nmvdw, line 152: Why is this `Qed` instead of `Defined`? This blocks computation with this equivalence, and I don't see a good reason for that.

UniMath/AlgebraicTheories/examples/EndomorphismTheory.v
```diff
@@ -0,0 +1,115 @@
+Require Import UniMath.Foundations.All.
```
nmvdw, line 1: The directory should be renamed into `Examples` instead of `examples`.

UniMath/AlgebraicTheories/examples/EndomorphismTheory.v
```diff
@@ -0,0 +1,115 @@
...
+Require Import UniMath.AlgebraicTheories.AlgebraicTheories.AlgebraicTheories.
+Require Import UniMath.AlgebraicTheories.AbstractClones.AbstractClones.
+Require Import UniMath.AlgebraicTheories.AbstractCloneAlgebraicTheory.
+
+Local Open Scope cat.
+
+Section EndomorphismAlgebraicTheory.
+
+  Context {C : category}.
+  Variable (finite_products: ∏ (n : nat), Products (stn n) C).
```
nmvdw, line 17: This might be good to add as a definition in the folder `CategoryTheory/limits`, so that it can be used and so that statements about it can be proven.

UniMath/AlgebraicTheories/examples/EndomorphismTheory.v
```diff
@@ -0,0 +1,115 @@
...
+  Variable (X : C).
+
+  (* Helper definitions *)
+  Local Definition power_function (n : nat) : Product (stn n) C _ := finite_products n (λ _, X).
```
nmvdw, line 21: This should also be moved to `CategoryTheory/limits`, so that it can be reused.

nmvdw: Same for: `power_function_set`, `power_function_projection`, `power_function_is_product`, `induced_map`, `induced_map_commutes`, `induced_map_equals`

UniMath/AlgebraicTheories/examples/EndomorphismTheory.v
```diff
@@ -0,0 +1,115 @@
...
+    exact (maponpaths pr1 (map_is_unique (g ,, (λ _, idpath _)))).
+  Defined.
+
+  (* Construct an algebraic theory as an abstract clone *)
+  Definition endomorphism_clone_data : abstract_clone_data.
+  Proof.
+    use make_abstract_clone_data.
+    - use make_algebraic_base.
+      + intro n.
+        use make_hSet.
```
nmvdw, line 75: Use: https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/Core/Categories.v#L167

nmvdw (approved):
The error is:
```
Error: Error: Notation "_ • _" is already defined at level 50 with arguments constr
[292](https://github.com/UniMath/UniMath/actions/runs/4569338472/jobs/8065471792?pr=1676#step:5:298)
  at next level, constr at next level while it is now required to be
[293](https://github.com/UniMath/UniMath/actions/runs/4569338472/jobs/8065471792?pr=1676#step:5:299)
  at level 35 with arguments constr at next level, constr at next level.
```
This notation is `\bu`, which is also used here: https://github.com/UniMath/UniMath/blob/master/UniMath/Bicategories/Core/Bicat.v#L1700
but with different priorities.

This should be made consistent over UniMath.

UniMath/AlgebraicTheories/AbstractClones/AbstractClones.v
nmvdw: That is fine too. Proving that they are a proposition in a separate lemma/definition, would be helpful, because then those statements would be reusable.

UniMath/AlgebraicTheories/examples/EndomorphismTheory.v
arnoudvanderleer: Ah. Is there a reason why the `CategoryTheory` package has a folder `categories` and not `Categories`? I think that is where I got the idea.

UniMath/AlgebraicTheories/examples/EndomorphismTheory.v
arnoudvanderleer: Ooooh, my first external definition!

UniMath/AlgebraicTheories/AlgebraicTheories/AlgebraicTheories.v
```diff
@@ -0,0 +1,172 @@
...
+Proof.
+  apply isapropdirprod.
+  - apply isaprop_is_functor.
+    apply SET.
+  - repeat apply isapropdirprod;
+      repeat (apply impred_isaprop; intro);
+      apply setproperty.
+Qed.
+
+Definition algebraic_theory := total2 is_algebraic_theory.
```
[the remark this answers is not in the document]
arnoudvanderleer: The reason I used `total` is because it does not require one to specify over which space we are summing. I saw that this was used in, for example, the definition of `functor` and `precategory` and I thought that it was a good idea. What are your ideas about this?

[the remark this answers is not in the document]
benediktahrens: The definitions of `functor` and `precategory` were written down before unicode was used systematically (or at all), back in 2012/13 :-)  It would be good to use `\sum` instead.

m-lindgren:
This is just a reflection on the library layout: Since this introduces a package called `AlgebraicTheories` should we consider this package a meta-package for everything called algebraic theories? So in potential future formalizations, should material on for example essentially algebraic theories and generalized algebraic theories be placed in this package?

UniMath/AlgebraicTheories/examples/EndomorphismTheory.v
benediktahrens: That code is old, and a bad example :-) Sorry for that!

UniMath/AlgebraicTheories/AlgebraicTheories/AlgebraicTheories.v
```diff
@@ -0,0 +1,172 @@
...
+  := (T ,, H).
+
+Definition algebraic_theory_data_from_algebraic_theory
+  : algebraic_theory -> algebraic_theory_data
+  := pr1.
+Coercion algebraic_theory_data_from_algebraic_theory : algebraic_theory >-> algebraic_theory_data.
+
+Lemma algebraic_theory_eq
+  (X Y : algebraic_theory)
+  (H1 : pr111 X = pr111 Y)
```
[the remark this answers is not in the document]
arnoudvanderleer: I agree that things become unreadable. The main reason that I did not do these, was that I had no clue how to name them.
For `H1`, we have coercions which allow us to rewrite it as `H1 : (X : nat → hSet) = (Y : nat → hSet)`. Would that suffice?
Would you create accessor functions for all the individual properties of an algebraic theory, and if so, what name would you give, for example, an accessor function with type `∏ T : algebraic_theory_data, comp_is_natural_l T` (the one that will replace `pr22 (pr222 T)`)?

UniMath/AlgebraicTheories/AlgebraicTheories/AlgebraicTheories.v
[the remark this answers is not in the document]
m-lindgren: I think if one enters the `AlgebraicTheories` folder one expects to see material on algebraic theories, not another folder called `AlgebraicTheories`. It doesn't seem to add much in terms of structure, since the subfolder `AlgebraicTheories` only contains two files currently. Granted it will be one more if you also add a file `Algebras.v`, but this is still something one would expect to find when one enters the first folder. Or do you expect this package to grow a lot? (even then - unless the scope of the package changes - I would think it makes sense to keep the material related to algebraic theories in the center)

m-lindgren:
By the way, I noticed you made this pull request from arnoudvanderleer:master. I would recommend for the future that you use topic branches for your development, and keep your master branch always in sync with UniMath:master. Having your own commits in your forks master branch will make it difficult to keep it synchronized with UniMath and will make things complicated for you down the line.

- Github calls it [feature branches](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-branches#working-with-branches)
- The git book calls it [topic branches](https://git-scm.com/book/en/v2/Git-Branching-Branching-Workflows)

UniMath/AlgebraicTheories/AlgebraicTheories/AlgebraicTheories.v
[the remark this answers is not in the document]
arnoudvanderleer: Hm, right. So you would propose moving the AlgebraicTheories material one level up, but keeping the AbstractClones where they are? It goes a bit against my desire for symmetry, but if you think this is better, I will change it.

UniMath/AlgebraicTheories/AlgebraicTheories/AlgebraicTheories.v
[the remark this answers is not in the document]
benediktahrens: It is useful to give a name that let's one guess the type. In your example, `algebraic_theory_comp_is_natural_l` would be a reasonable name.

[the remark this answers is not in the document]
arnoudvanderleer: Ah, right. Thanks.

UniMath/AlgebraicTheories/AlgebraicTheories/AlgebraicTheories.v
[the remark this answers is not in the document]
benediktahrens: I don't have a preference right now. 

In general, it seems difficult to me to plan a suitable directory layout in advance, in particular for a topic as open-ended as "Algebraic Theories". It might make sense to postpone the search for a suitable structure until later.

UniMath/AlgebraicTheories/AbstractClones/AbstractClones.v
arnoudvanderleer: I wasn't specifically planning on studying those, but I might at some point formalize the correspondence.

UniMath/AlgebraicTheories/AlgebraicTheories/AlgebraicTheories.v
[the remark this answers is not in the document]
arnoudvanderleer: Hm... For now, I will follow Michael's request, and move the algebraic theory material one directory up. If it turns out that this causes problems in the future, I will move them down again (or somewhere else).

arnoudvanderleer:
> By the way, I noticed you made this pull request from arnoudvanderleer:master. I would recommend for the future that you use topic branches for your development, and keep your master branch always in sync with UniMath:master. Having your own commits in your forks master branch will make it difficult to keep it synchronized with UniMath and will make things complicated for you down the line.
> 
> * Github calls it [feature branches](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-branches#working-with-branches)
> * The git book calls it [topic branches](https://git-scm.com/book/en/v2/Git-Branching-Branching-Workflows)

I will work in feature branches from now on.

arnoudvanderleer:
Thank you both for your thorough reviews, @nmvdw and @m-lindgren. I believe I have addressed all feedback. Would you take a look again?

PR #1679, opened 2023-04-01
General classes of enriched limits/colimits

nmvdw:
- The notion of conical limit/colimit
- Construction of conical limits from products and equalizers (also the dual statement for colimits)
- The notion of (weighted) limit/colimit
- Instances of weighted limits/colimits


m-lindgren:
Interesting, the -dev builds fail. Not sure it's actually related to this PR, it might be changes upstream. I'll look into it.

m-lindgren (approved):

PR #1680, opened 2023-04-07
Limits and colimits in structure enriched categories

nmvdw:
- Refactor the file on structures (because it became too large), and add some material
- Refactor the file on posets, and add some material
- Characterize limits in categories enriched over structures
- Characterize colimits in categories enriched over structures


nmvdw:
The error:
```
File "./Satellite/code/hit_biadjunction/hit_prealgebra_biadj/left_triangle.v", line 74, characters 6-621:
[456](https://github.com/UniMath/UniMath/actions/runs/4641221332/jobs/8213967783?pr=1680#step:6:462)
  Error: Error: Case analysis on private inductive gquot
```
The relevant line: https://github.com/UniMath/GrpdHITs/blob/master/code/hit_biadjunction/hit_prealgebra_biadj/left_triangle.v#L74

This error surprises me, as no pattern matching is done in that particular line.

nmvdw:
Currently, all checks succeed except for the one in GrpdHITs on Coq dev. However, fixing that one would require a separate pull request at GrpdHITs to make it compatible with the latest version of Coq dev (which might take some time). Until that is fixed, we can ignore this part of the CI.

benediktahrens:
You are saying that the issue in GrpdHITs is not coming from this PR?

nmvdw:
> You are saying that the issue in GrpdHITs is not coming from this PR?

It doesn't. GrpdHITs makes use of higher inductive types, so private inductive types are used. There was a recent change in Coq (1 or 2 days ago) that affected private inductive types to ensure they are really private (with some hacks, people could still access the wrong induction principle). A side-effect of that change is that sometimes, one needs to give Coq more info (more arguments need to be provided), and that seems to be the cause of the error in GrpdHITs. As such, GrpdHITs needs a patch.

UniMath/CategoryTheory/DisplayedCats/Structures/ExamplesOfStructures.v
```diff
@@ -0,0 +1,968 @@
...
+ Structures on sets
+
+ In this file, we look at a particular class of structures on the
+ category of set that is closed under products and the terminal
+ object. Key in this development are displayed categories and the
+ structure identity principle.
+
+ The notion of structure that we consider, consists of:
+ - For every hSet, a set of structures on that set
+ - For every function, a proposition that represents wheter this
```
benediktahrens, line 12:
```suggestion
 - For every function, a proposition that represents whether this
```

benediktahrens (approved):

PR #1682, opened 2023-04-15
Definitions for natural transformations between sections

DenSinH:
For my thesis, I needed natural transformations between sections (such that its components lie over identity in the base category). After implementing it myself and asking about it in the UniMath Zulip, they suggested I make a PR. I figured the best spot to put them is just in the same file as the sections / section functors are defined (just called Constructions).


UniMath/CategoryTheory/DisplayedCats/Constructions.v
```diff
@@ -496,6 +496,93 @@ Section Functor_Lifting.
...
+    {C : category} 
+    {D : disp_cat C} 
+    (F F' : section_disp D) : UU :=
+  ∏ (x : C), F x -->[identity _] F' x.
+
+Definition section_nat_trans_disp_axioms 
+    {C : category}
+    {D : disp_cat C} 
+    {F F': section_disp D}
+    (nt : section_nat_trans_disp_data F F') :=
```
nmvdw, line 512:
```suggestion
    (nt : section_nat_trans_disp_data F F') : UU :=
```
For clarity, I think it is good to give the type of the definition

UniMath/CategoryTheory/DisplayedCats/Constructions.v
```diff
@@ -496,6 +496,93 @@ Section Functor_Lifting.
...
+      transportf _
+      (id_right _ @ !(id_left _)) 
+      (section_disp_on_morphisms F f ;; nt x') =
+    nt x ;; section_disp_on_morphisms F' f.
+
+Definition section_nat_trans_disp 
+    {C : category}
+    {D : disp_cat C} 
+    (F F': section_disp D) : UU :=
+  total2 (@section_nat_trans_disp_axioms C D F F').
```
nmvdw, line 523: Using \sum (∑) is more readable, I think. In this case, it would become something like
```suggestion
 ∑ (nt : section_nat_trans_disp_data), section_nat_trans_disp_axioms nt.
```

UniMath/CategoryTheory/DisplayedCats/Constructions.v
```diff
@@ -496,6 +496,93 @@ Section Functor_Lifting.
...
+
+Definition section_nt_disp_data_from_section_nt_disp
+    {C : category}
+    {D : disp_cat C} 
+    {F F': section_disp D}
+    (nt : section_nat_trans_disp F F')
```
nmvdw, line 529:
```suggestion
    (nt : section_nat_trans_disp F F')
    : section_nat_trans_disp_data F F'
```

UniMath/CategoryTheory/DisplayedCats/Constructions.v
```diff
@@ -496,6 +496,93 @@ Section Functor_Lifting.
...
+    (nt : section_nat_trans_disp F F') :
+  ∏ x : ob C, F x -->[identity _]  F' x := section_nt_disp_data_from_section_nt_disp nt.
+Coercion section_nat_trans_data_from_section_nat_trans_disp_funclass :
+    section_nat_trans_disp >-> Funclass.
+  
+Definition section_nt_disp_axioms_from_section_nt_disp
+    {C : category}
+    {D : disp_cat C} 
+    {F F': section_disp D}
+    (nt : section_nat_trans_disp F F')
```
nmvdw, line 545:
```suggestion
    (nt : section_nat_trans_disp F F')
    : section_nat_trans_disp_axioms nt
```

nmvdw (approved):
Looks good in general (be sure to double check whether the types that I added are actually correct).

benediktahrens (approved):

UniMath/CategoryTheory/DisplayedCats/Constructions.v
DenSinH: Ah sure, I just used this approach because it was also used earlier when defining `section_disp`

m-lindgren:
Lots of trailing whitespaces in the lines added. Would you mind removing them? If you're using Emacs you can view them by using `(setq show-trailing-whitespace t)`. If you're using VSCode there's probably a setting named something similar.

DenSinH:
I don't see any trailing whitespace myself...

benediktahrens:
@DenSinH : Thanks for contributing this material.

I confirm that there are some trailing whitespaces. Unfortunately, github does not show them. If @DenSinH cannot see them, maybe someone else could amend this PR?

DenSinH:
I can also not see the whitespace locally though... I also have a definition for the composition of natural transformations between sections that I can add by the way.

benediktahrens:
> I also have a definition for the composition of natural transformations between sections that I can add by the way.

That would be great!

DenSinH:
Also wrote some more utilities

m-lindgren:
Am I understanding this right: the reason we need an ad-hoc construction of natural transformations between (total spaces of) sections is because sections are defined in an ad-hoc manner, and not as a displayed functor from the identity dispcat? (by which I mean the displayed cat with precisely one object for each object in the base, and one morphism for each morphism in the base)

The naming of some things in this PR confused me for this reason, because nowhere is an actual `disp_nat_trans` defined yet some things have names indicating that they are such things (for example `section_nat_trans_disp_data`).

Would an equivalent way of doing this be to show that each `S : section_disp D` induces a displayed functor from the identity (and over the identity functor) into `D`, and then use the definition of _displayed_ natural transformations as they are already defined in the library?

This is not really a suggested change, just a reflection that maybe we should document some of these design choices for the future (like the definition of `section_disp`).

m-lindgren:
> I don't see any trailing whitespace myself...

@DenSinH I made a PR https://github.com/DenSinH/UniMath/pull/1 into your branch removing the trailing whitespace.

m-lindgren (approved):

PR #1683, opened 2023-04-16
Smash products for pointed structures

nmvdw:
- Define a notion of structure that supports smash products
- Instantiate it to pointed posets and pointed sets
- Prove that it gives rise to a symmetric monoidal category
- Improve documentation at several points


UniMath/CategoryTheory/Monoidal/Examples/SmashProductMonoidal.v
```diff
@@ -0,0 +1,1234 @@
+(*****************************************************************
+
+ Monoidal category from smash products
+
+ We show that every notion of structure that supports a smash
+ product, gives rise to a symmetric monoidal closed category. Note
+ that in the construction, we give two (equivalent) definitions
+ for both of the associators. One of the two definition is
+ convenient for calculation, while the other is better for proving
+ that it is a structure preserving map.
```
benediktahrens, line 10: Which one is which? Would it be useful to add a comment in the code below that picks up this comment here and continues it?

benediktahrens (approved):

UniMath/CategoryTheory/Monoidal/Examples/SmashProductMonoidal.v
nmvdw: I added a comment on it

nmvdw:
I slightly modified the definition of posets with a minimum element: before, the bottom element was in a truncation, but I realized that bottom elements are unique (due to anti-symmetry), so the truncation can be removed. In addition, I renamed some stuff in `ExamplesOfStructures`.

One thing that I noticed, is that the file `ExamplesOfStructures` is getting rather large (currently 1480 lines of code), and based on the current division of files, it will only grow larger. As such, it will be needed to find a good way to split this file up into multiple files.

PR #1687, opened 2023-04-21
Split the proof of associators_equiv to make sure it fits inside 32 bits

arnoudvanderleer:
This is necessary to create a jscoq addon for UniMath (https://github.com/UniMath/UniMath-jsCoq).


benediktahrens:
This is great, thanks!

Would it be useful to go one step further, and prove each of the three laws of `nat_trans_associator_has_naturality_laws` separately?

arnoudvanderleer:
What do you mean by "useful"?
With this definition split, jscoq compiles, so splitting it even further would not contribute to that.
Or did you actually mean to ask "I think it is useful if you split it up even further, to make it more reusable. Could you do that, while you are at it?". If that is what you meant, the answer is: of course ^^

benediktahrens:
I meant the latter :-)

rmatthes:
Please just for understanding: in what dimension did the previous proof not fit into the 32bit requirement of jscoq? How did that tool tell you?

arnoudvanderleer:
@rmatthes In short: I tried compiling the UniMath library using the JSCoq tooling, and it gave an error which usually is associated with a very large proof term. The error was raised on a line with `Defined.`, and the people who know more about this stuff confirmed that there probably was a very large definition that caused the error.

For more details, see the discussion over at https://coq.zulipchat.com/#narrow/stream/256336-jsCoq/topic/How.20to.20create.20a.20version.20of.20jsCoq.20with.20more.20packages.3F/near/350703512

PR #1690, opened 2023-05-02
(Re)formulate all categories in the algebraic theory package as displayed categories and prove that algebras are fibered over theories

arnoudvanderleer:


UniMath/AlgebraicTheories/AlgebraicTheoryMorphisms2.v
```diff
@@ -0,0 +1,70 @@
+Require Import UniMath.Foundations.All.
```
benediktahrens, line 1: What is the purpose of this file? Does it duplicate the contents of another file?

benediktahrens (commented):
Could you insert some comments? For instance, at the top of each file you could give an overview of the contents of the file.

UniMath/AlgebraicTheories/AlgebraicTheoryMorphisms2.v
arnoudvanderleer: It provides a different/simpler way to define an algebraic theory morphism. Its contents are used by `AlgebraicTheoryMorphism` in the section `MakeAlgebraicTheoryMorphisms2`.
I thought putting this whole bunch of constructors (and accessors) in the file `AlgebraicTheoryMorphisms2` would be a good idea, because it shows that it is very closely linked to algebraic theory morphisms. Do you think this naming scheme can be improved upon?

benediktahrens: Thanks a lot for the information. I believe it would be very useful to give this kind of information at the top of the file in a comment.

UniMath/AlgebraicTheories/AlgebraicTheoryAlgebraFibration.v
```diff
@@ -0,0 +1,152 @@
...
+Local Open Scope mor_disp.
+
+Lemma concat_displayed_cartesian_morphisms
+  {C C' : category}
+  (D := disp_cartesian C C')
+  {c c' c'' : C}
+  {F : C⟦c, c'⟧} {F' : C⟦c', c''⟧}
+  {A : D c} {A' : D c'} {A'' : D c''}
+  (G : A -->[F] A') (G' : A' -->[F'] A'')
+  : (G ;; G') = G · G'.
```
arnoudvanderleer, line 30: This feels like it could be an external lemma, but I am not sure whether it is general enough.

benediktahrens:
Merging this now, any issues should be solved in a future PR.

Issue #1693, opened 2023-05-14
GrpdHITs currently not building over Coq dev


peterlefanulumsdaine:
Currently, the CI job “BuildGrpdHITs (Coq dev)” is consistently failing, see e.g. https://github.com/UniMath/UniMath/actions/runs/4971891863/jobs/8896730557?pr=1227

For those who know GrpdHITs better than me — is this something we should be looking to address here in UniMath, or that GrpdHITs should be fixing there?  It’d be good to have a clear path towards fixing the CI failure, somewhere.

nmvdw:
Recently, in Coq dev there was a patch in Coq dev that affected private inductive types (https://github.com/coq/coq/pull/9609). GrpdHITs was affected by it, because it uses private inductive types to implement HITs. The fix is thus completely independent of stuff in UniMath, but it just is caused by this particular change. Coq might also have rewrite rules in the near future ('The Rewster: The Coq Proof Assistant with Rewrite Rules', https://types2023.webs.upv.es/Accepted.html), and that might give a better way of implementing HITs, so I am waiting with fixing it.

I think the best solution would be to temporarily remove the Coq dev check of GrpdHITs from the UniMath CI.

peterlefanulumsdaine:
#1694 gives the short-term fix for this issue, but I suggest keeping the issue open (with slightly changed title), with the longer-term goal of adding GrpdHITs back in once they build again.

nmvdw:
> @nmvdw This is resolved now, right?

Yes

PR #1694, opened 2023-05-14
Temporarily remove GrpdHITs-dev from CI.

m-lindgren:
See #1693.

### DIFF
--- a/.github/workflows/build-unimath.yml
+++ b/.github/workflows/build-unimath.yml
@@ -145,6 +145,9 @@ jobs:
         satellite: [SetHITs, largecatmodules, GrpdHITs, TypeTheory]
         coq-version: [latest, dev]
         ocaml-version: [4.14-flambda]
+        exclude:
+          - satellite: GrpdHITs
+            coq-version: dev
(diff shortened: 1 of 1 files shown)


nmvdw (approved):

peterlefanulumsdaine:
lgtm!

benediktahrens:
Are there any plans to fix GrpdHITs? How can we avoid forgetting GrpdHITs after deactivating it in CI?

m-lindgren:
> How can we avoid forgetting GrpdHITs after deactivating it in CI?

By next release of Coq the GrpdHITs-latest build will fail as well so that'll remind us to either fix it or remove GrpdHITs from the CI completely (or only build GrpdHITs with 8.17).

nmvdw:
There are plans to fix it. However, the precise fix depends on the features in the new version of Coq. Currently, one implements HITs using private inductive types, and a nicer way would be to add rewrite rules. This is not possible at the moment, but some people are actually working on adding that feature (if I heard correctly, they have an open PR). If rewrite rules are added, then the fix might be different compared to if they are not added. In addition, the CI will indeed remind us when we switch to a new version of Coq. 

I would like to keep GrpdHITs actively maintained and in the CI.

PR #1695, opened 2023-05-15
Formalized notion of model of enriched effect calculus

nmvdw:
I used the formalization on enriched categories to formalize Definition 6.5, Proposition 6.9, and Example 6.10 from http://www.itu.dk/people/mogel/papers/eec.pdf.

Note: I created a new folder for this, because I think that in the short run, there will be more stuff added to that folder. In addition, I am hesitant of putting it in CategoryTheory, because I would like to be able to import Bicategories for these files (although that currently is not done, yet). As such, the two options were to either add it to Bicategories or to create a new folder for it. I found the second to be more logical.


nmvdw:
> New bug in sanity-checks. Seems to be triggered by `UniMath/Semantics/.package/files` not existing.

That was caused by me forgetting to commit that file

benediktahrens (approved):

UniMath/CategoryTheory/EnrichedCats/Colimits/CopowerFunctor.v
```diff
@@ -0,0 +1,407 @@
...
+      internal_eval v₁ v₂ · mor_of_copow_ob v₂ r.
+  Proof.
+    pose (maponpaths
+            (λ z, z #⊗ identity _ · internal_eval _ _)
+            (mor_to_copower_commutes
+               EC
+               v₁ r
+               (pr2 (copowC v₁ r))
+               (internal_postcomp _ (mor_of_copow_ob v₂ r)))).
+    cbn in p.
```
m-lindgren, line 214:
```suggestion
               (internal_postcomp _ (mor_of_copow_ob v₂ r))))
      as p.
    cbn in p.
```

m-lindgren (approved):

Issue #1696, opened 2023-05-15
Should we replace `-type-in-type` with `Unset Universe Checking` in some way?


peterlefanulumsdaine:
(This is proposed by @Bigstep22 in #1671 as it’s a prerequisite for adding Alectryon, but seems worth splitting out for separate discussion since it’s conceptually its own thing, and potentially impactful for the whole library.)

For at least a few years now, Coq has had the flag `Universe Checking`; [according to the documentation](https://coq.inria.fr/refman/proof-engine/vernacular-commands.html#coq:flag.Universe-Checking), `Unset Universe Checking` is equivalent to `-type-in-type`.  **So we could replace our use of `-type-in-type` with `Unset Universe Checking` in `Foundations/Init.v`.  Should we?**

I see one significant and one very small upside, and no clear downsides.  Positives:

- having it in file(s) rather than command-line flags seems like it could really help with incrementally removing reliance on it, i.e. making progress on #648, and integrating partial progress from there into master
- less reliance on command-line flags seems good in general, although this doesn’t really give much payoff unless we could eliminate them entirely, which seems unlikely (currently we use `-type-in-type`, `-noinit`, and `-indices-matter`)

Possible downsides:

- I see none — but this possibility has been available for at least a few years now, so was there a reason we didn’t switch over to it sooner?  Or did we just never consider it?

DanGrayson:
That's a good suggestion.  I wasn't aware of the new flag.

benediktahrens:
@peterlefanulumsdaine : thanks a lot for the summary. I was not aware of this option until #1672 . The change Peter is suggesting seems beneficial.

m-lindgren:
Possible downside: Universe checking would be disabled for all packages importing Foundations.Init. If this is the case then this has the complete opposite effect of the intended goal to incrementally remove `-type-in-type`.

Example: SetHITs builds without `-type-in-type` today. I suspect that the above flag would disable universe checking for SetHITs as well, and that would be a shame.

nmvdw:
> Constants relying on "type in type" are printed by [Print Assumptions](https://coq.inria.fr/refman/proof-engine/vernacular-commands.html#coq:cmd.Print-Assumptions).

I think this is a big upside. This way it is easier to detect what precisely relies on type-in-type and what not.

nmvdw:
[...]

PR #1697, opened 2023-05-22
Replace `-type-in-type` with `#[global] Unset Universe Checking` in `UniMath.Foundations.Init`

peterlefanulumsdaine:
This addresses #1696 : it replaces the use of the `-type-in-type` flag (in the Makefile, emacs local variables file, etc) with the command `Export Unset Universe Checking` in `UniMath.Foundations.Init`.

As discussed in detail in #1696 , these should be entirely equivalent in their current effect within UniMath; the two effective differences are:

- The new setting will be more “infectious” downstream — projects importing UniMath will get universe checking unset, unless they specifically re-enable it with `Set Universe Checking` after importing UniMath.
- This makes it easier in future to reduce our reliance on type-in-type, by setting/unsetting universe checking on a file-by-file basis.

Compared to f9175188 in #1671 , I’ve used `Export` rather than `Global` — according to the guidance in the Coq Reference Manual [here](https://coq.inria.fr/refman/language/core/modules.html#controlling-the-scope-of-commands-with-locality-attributes) and [here](https://coq.inria.fr/refman/language/core/basic.html#set-unset-scope-qualifiers), that seems to me like the correct one.


benediktahrens:
@peterlefanulumsdaine thanks a lot for preparing this. There is a warning about deprecated "Export Set" - should we address this before merging?

m-lindgren:
I think we should be very careful about merging this. This will effectively enable `-type-in-type` in **every** project that use UniMath, without them knowing it. It is my opinion that it's not very good practice for a library to export a setting that collapses the entire universe hierarchy for the user, even if UniMath internally relies on it. 

In particular since we have no way of informing projects that rely on UniMath that we are making this change in **their projects**. We do not know every user of UniMath, and any problems related to this will only cause headaches for them, not us unfortunately. (Imagine spending hours trying to figure out why your project suddenly has a flat universe hierarchy, despite you being sure that you don't have `-type-in-type` in your Makefile. Recloning/cleaning the project, reinstalling/upgrading/downgrading Coq, clearing caches, reading Coq documentation, filing Coq issues ...)

I like the idea of taking steps towards removing `type-in-type` but this looks to me like a step in the opposite direction since it makes more code overall compile with universe checking disabled, not less, and the effort saved seems to amount only to the effort required to add `Export Unset Universe Checking` to one file, so I don't really see how this helps.

JasonGross:
> @peterlefanulumsdaine thanks a lot for preparing this. There is a warning about deprecated "Export Set" - should we address this before merging?

This warming is easy to address, btw, just use `#[export] Set` instead of `Export Set`

JasonGross:
Why not add `#[local] Unset Universe Checking` just to the minimal set of files that need it?  (Just run `make` and add the setting to whatever files fail, and do this in a loop)

peterlefanulumsdaine:
@m-lindgren I feel fairly mixed about this.  I agree with your reservations to some extent, but two counter-points:

- (pragmatic) If a downstream importer notices their universe hierarchy has collapsed, I see how the trouble-shooting would take long: stepping through their files and seeing where it first collapses would immediately show that it’s coming from importing UniMath; then googling why UniMath does that should quickly bring up these discussions.
 
- (principled) I think it’s probably (if anything) slightly *better* that type-in-type is propagated downstream by default, since (as I mentioned in #1696 ) working with universe checking but importing un-checked objects seems theoretically pretty unclear.  In particular, it certainly *doesn’t* re-gain consistency — the main thing people might hope it does — so I think it gives a false sense of security, unless a development is very conscious and careful about restricting which objects it imports from us.

These are why in the long-term, I don’t think the change is negative.  In the short-term I fully agree it’s not great to affect downstream users without notification — but I’m not sure how we can notify them, other than trying this out.

So I’d lean to merging this (if others agree) but thinking of it as provisional for a while (at least a month, say), and being willing to roll it back if downstream users report any problems.  As long as we don’t merge #1671 (which is blocked anyway if we don’t merge this), I don’t think anything else is likely to interact with this and make it difficult to roll back.

nmvdw:
> Why not add `#[local] Unset Universe Checking` just to the minimal set of files that need it? (Just run `make` and add the setting to whatever files fail, and do this in a loop)

I really like this solution. I think it addresses Michael’s comments and it properly indicates where type-in-type is used, which is useful for removing the dependency on type-in-type.

JasonGross:
What about something like (assuming no spaces in names):
```bash
debug=1 # change to debug="" to disable waiting on user input
while ! make -k >log 2>&1; do
  files="$(grep -B1 "^Error" log | grep -o '^File "[^"]*' | sed 's/^File "//g')"
  echo "Errors in $files"
  if [ ! -z "$debug" ]; then echo "Press enter to continue, ctrl+c to break..."; read dummyvar; fi
  for f in $files; do
    echo "editing $f"
    if grep -q '#.local. Unset Universe Checking' "$f"; then
      echo "File $f requires manual intervention"
      exit 1
    fi
    cp -f "$f" "$f.tmp"
    { echo "#[local] Unset Universe Checking."; cat "$f.tmp"; } > "$f"
  done
done
```
or as a one-liner:
```bash
debug=1; while ! make -k >log 2>&1; do files="$(grep -B1 "^Error" log | grep -o '^File "[^"]*' | sed 's/^File "//g')"; echo "Errors in $files"; if [ ! -z "$debug" ]; then echo "Press enter to continue, ctrl+c to break..."; read dummyvar; fi; for f in $files; do echo "editing $f"; if grep -q '#.local. Unset Universe Checking' "$f"; then; echo "File $f requires manual intervention"; exit 1; fi; cp -f "$f" "$f.tmp"; ( ( echo "#[local] Unset Universe Checking."; cat "$f.tmp") > "$f"); done; done
```
Once you set `debug=""`, then you should be able to just let this run in the background while you do other tasks

peterlefanulumsdaine:
@JasonGross Thanks very much for the script — it came through just as I was starting to scratch my head over putting together such a script myself.  Works like a charm, no syntax errors.

I’m trying it out in a branch [`localise-type-in-type-per-file`](https://github.com/peterlefanulumsdaine/UniMath/tree/localise-type-in-type-per-file).  Under this naïve approach, it’s needed in a *lot* of files.

peterlefanulumsdaine:
Shall we merge this now (assuming it still passes CI after updating), changing from global `-type-in-type` to a global `Export Universe Checking` in `Foundations.Init`?  In discussion at #1925, @arnoudvanderleer , @benediktahrens and I all support the switch; in #1696, @m-lindgren sounded in favour as well; and as @arnoudvanderleer noted, the one arguable downside mentioned in #1969 no longer applies.

nmvdw:
> Shall we merge this now (assuming it still passes CI after updating), changing from commandline `-type-in-type` to `#[global] Unset Universe Checking` in `Foundations.Init`? In discussion at #1925, @arnoudvanderleer , @benediktahrens and I all support the switch; in #1696, @m-lindgren sounded in favour as well; and as @arnoudvanderleer noted, the one arguable downside mentioned in #1969 no longer applies.

I also am in favor.

benediktahrens:
What will the satellite repositories need to be compatible with this change?

peterlefanulumsdaine:
I don’t think satellite repos will need to do anything in response — certainly the ones in the CI are compiling just fine (except for `Schools` which has been failing already on `main` for a while).  They won’t require `-type-in-type` any more, but having it shouldn’t break their builds.

arnoudvanderleer:
:partying_face:

PR #1703, opened 2023-05-30
Displayed natural transformations

EwenBC:


UniMath/CategoryTheory/DisplayedCats/NaturalTransformations.v
```diff
@@ -759,3 +759,122 @@ Proof.
...
+    apply pathsinv0.
+    etrans. apply transport_b_b.
+    etrans. apply transport_b_b.
+    etrans. apply transport_b_b.
+    etrans. apply transport_b_b.
+    etrans. apply transport_b_b.
+    apply two_arg_paths.
+    * apply uip.
+      apply homset_property.
+    * reflexivity.
```
benediktahrens, line 831: This part should be opacified. It could be put into a separate lemma, the type of which is copied from the goal statement.

benediktahrens (commented):
Lgtm, in general. The construction of the inverse should have the proof part opacified, to make computation more efficient.

PR #1708, opened 2023-06-03
Refactor sub bicategories

nmvdw:
In this PR:
- The notion of subbicategory (`subbicat`) is made more general
- I added two examples displayed bicategories over the bicategory of categories (not necessarily univalent ones). Both of these are about terminal objects: one is for chosen terminal objects, the other is for existing terminal objects. 

The main reason to look at these bicategories is to study how it interacts with the Rezk completion. More specifically, what can we say about the Rezk completion of a category with a terminal object? Does it matter whether the terminal object is chosen or just exists?


UniMath/Bicategories/Limits/Examples/LimitsStructuredCategories.v
```diff
@@ -87,15 +87,15 @@ Definition disp_has_binprod_univ_cat_with_terminal_obj
   : disp_has_binprod disp_bicat_terminal_obj has_binprod_bicat_of_univ_cats.
 Proof.
   use subbicat_disp_binprod.
-  - exact (λ C₁ C₂, terminal_category_binproduct (pr22 C₁) (pr22 C₂)).
+  - exact (λ C₁ C₂, terminal_category_binproduct (pr12 C₁) (pr12 C₂)).
```
benediktahrens, line 90: It seems to me that this file shows what happens when one does not use access functions. Could we add access functions instead of those projections?

UniMath/Bicategories/DisplayedBicats/Examples/Sub1Cell.v
```diff
@@ -167,25 +167,33 @@ End Sub1CellBicategory.
...
+          (P₁ : ∏ (x y : B), P₀ x → P₀ y → x --> y -> UU)
+          (P₁id : ∏ (x : B) (Px : P₀ x), P₁ _ _ Px Px (id₁ x))
+          (P₁comp : ∏ (x y z : B)
+                      (Px : P₀ x)
+                      (Py : P₀ y)
+                      (Pz : P₀ z)
+                      (f : x --> y) (g : y --> z),
+              P₁  _ _ Px Py f
+              → P₁ _ _ Py Pz g
+              → P₁ _ _ Px Pz (f · g)).
```
benediktahrens, line 179: Are the definitions in this file well-behaved when P_0 or P_1 are not pointwise propositions?

nmvdw: If `P_0` and `P_1` are not pointwise propositions, then it will not give a univalent bicategory even if `B` is so. In that sense, it is not well-behaved. In addition, if `P_0` and `P_1` are not pointwise propositions, then the resulting inclusion functor might lose some properties, and then it (arguably) is not a subbicategory.

In the material that I added in `CategoriesWithStructure.v`, there is an example where `P_0` is not a pointwise proposition (`disp_bicat_chosen_terminal_obj`). However, in that file, the bicategory of (not necessarily univalent) categories is used, so the resulting bicategory wouldn't have been univalent anyway. It is more used as a shortcut to obtain the desired definition in a compact way without needing to redefine much.

UniMath/Bicategories/Limits/Examples/LimitsStructuredCategories.v
nmvdw: The nicest way to do so, would be as follows:
```coq
#[nonuniform] Coercion univ_cat_of_univ_cat_with_terminal_obj
         (C : univ_cat_with_terminal_obj)
  : univalent_category
  := pr1 C.

Definition terminal_of_univ_cat_with_terminal_obj
           (C : univ_cat_with_terminal_obj)
  : Terminal C
  := pr12 C.

#[nonuniform] Coercion functor_of_univ_cat_with_terminal_obj_mor
         {C₁ C₂ : univ_cat_with_terminal_obj}
         (F : C₁ --> C₂)
  : C₁ ⟶ C₂
  := pr1 F.

Definition preserves_terminal_univ_cat_with_terminal_obj_mor
           {C₁ C₂ : univ_cat_with_terminal_obj}
           (F : C₁ --> C₂)
  : preserves_terminal F
  := pr22 F.
```
Both coercions don't respect the uniform inheritance condition, so I needed to add `#[nonuniform]` in order to suppress the warning arising from that (they do work though).

Would this solution be fine?

benediktahrens: Yes, I think that's good!

UniMath/Bicategories/DisplayedBicats/Examples/Sub1Cell.v
benediktahrens: I think this explanation is very useful. Would it be a good idea to add it, as a comment, to the definition?

UniMath/Bicategories/Limits/Examples/LimitsStructuredCategories.v
nmvdw: Okay, I tried it further and by adding these coercions to `univ_cat_with_binprod`, 
```coq
#[nonuniform] Coercion univ_cat_of_univ_cat_with_binprod
         (C : univ_cat_with_binprod)
  : univalent_category
  := pr1 C.

Definition binprod_of_univ_cat_with_binprod
           (C : univ_cat_with_binprod)
  : BinProducts C
  := pr12 C.
```
The second definition is not accepted, and the first one gives the following warning:
```coq
univ_cat_of_univ_cat_with_binprod is defined
Warning:
New coercion path [univ_cat_of_univ_cat_with_binprod] : ob >-> univalent_category (reversible) is ambiguous with existing 
[univ_cat_of_univ_cat_with_terminal_obj] : ob >-> univalent_category (reversible).
[ambiguous-paths,typechecker]
univ_cat_of_univ_cat_with_binprod is now a coercion
```
The error for the second definition is:
```coq
Error:
In environment
C : univ_cat_with_binprod
The term "C" has type "ob univ_cat_with_binprod" while it is expected to have type "category".
```
So: it is not a good idea to have a coercion from the objects of a bicategory to something, because this coercion seem as a coercion from `ob` to categories and it does not take the bicategory into account.

Note that this idea was already used in:
https://github.com/UniMath/UniMath/blob/master/UniMath/Bicategories/DaggerCategories/BicatOfDaggerCats.v#L73
Now I think we should delete that coercion.

benediktahrens: I am not fully understanding the matter of coercions, but would like to. Could we discuss this in a meeting sometime? 

Anyway, this is possibly something that is separate from the current PR. I would be happy with a solution without coercions for now. Also, the substitution of projections for access functions could happen in a later PR.

UniMath/Bicategories/DisplayedBicats/Examples/Sub1Cell.v
nmvdw: I added this documentation in the latest commit

UniMath/Bicategories/Limits/Examples/LimitsStructuredCategories.v
nmvdw: > I am not fully understanding the matter of coercions, but would like to. Could we discuss this in a meeting sometime?

Sure.

> Anyway, this is possibly something that is separate from the current PR. I would be happy with a solution without coercions for now. Also, the substitution of projections for access functions could happen in a later PR.

I think that is fine.

PR #1714, opened 2023-06-07
Pushforward for displayed categories renamed as ReindexingForward

EwenBC:
Screenshot_2023-06-06_10-43-44
Given D C and C', creation of a re-indexation of a displayed category D over C as a category D' over C'. Such that for any D'' over C'', H and DG a display functor over F·H, a unique DH exists over H such that DF·DH=DG.


EwenBC:
@rmatthes there was a typo in the last name sorry about that. I've tried to answer to your question through the description but maybe I didn't understood it.

EwenBC:
> You define
> 
> ```coq
> Definition reindexing_f_ob_mor : disp_cat_ob_mor C'.
> Proof.
>   use make_disp_cat_ob_mor.
>   - exact (λ c', ∑c:C,(F c=c' × ob_disp D c)).
>   - intros a' b' x' y' f'.
>     exact (∑f:(↙ x')-->(↙ y'),
>            (transportf_mor (¤ x') (¤ y') (# F f)=f' ×
>             (← x')-->[f](← y'))).
> Defined.
> ```
> 
> The definition `exact (λ c', ∑c:C,(F c=c' × ob_disp D c)).` surprises me. Would an isomorphism be suitable here? Should the equality be truncated?

I'm not sure of what you mean but yes, the "reindexing_f functor" that I use is more or less an isomorphism. (It is a displayed functor so to inverse it the base functor would need to be invertible so it isn't invertible but the total functor is invertible.)
I did try at some point to truncate the equality and I think that it would indeed works but it also works without.

nmvdw:
I meant: what do you think of the following definitions:
- `exact (λ c', ∑c:C,(z_iso (F c) c' × ob_disp D c)).`. Note that the following notion should be possible `exact (λ c', ∑c:C,(F c=c' × D c)).`
- `exact (λ c', ∑c:C,(|| F c=c' || × ob_disp D c)).`

Both of these definitions should also work, but there's a choice to be made which one is the most suitable for your application.

EwenBC:
> I meant: what do you think of the following definitions:
> 
>     * `exact (λ c', ∑c:C,(z_iso (F c) c' × ob_disp D c)).`. Note that the following notion should be possible `exact (λ c', ∑c:C,(F c=c' × D c)).`
> 
>     * `exact (λ c', ∑c:C,(|| F c=c' || × ob_disp D c)).`
> 
> 
> Both of these definitions should also work, but there's a choice to be made which one is the most suitable for your application.

To me, the truncation adds an extra layer of abstraction which can be avoided because whenever the equality is used for morphisms and we can apply uid thanks to the homset property which solves almost all cases.
I haven't thought of using an isomorphism and maybe it would be a good idea but I don't really have any problems with the equality. (I could finish the proofs that I wanted to do.)
So I don't really know how to measure the "suitability" of this choice.

benediktahrens (approved):

m-lindgren:
Looks good overall modulo the remarks made by @nmvdw. 

Two style comments:

1. Very few files in UniMath are currently named using underscores, so `Reindexing_forward.v` would stick out a bit. I suggest renaming it to `ReindexingForward.v`.
2. In UniMath identifiers are usually named using English words concatenated by underscores (see https://github.com/UniMath/UniMath/tree/master/UniMath#readme) , would you mind renaming `reindex_f_*` to something like `reindex_forward_*`?

EwenBC:
I have done these proofs for any categories not just univalent ones so maybe reasoning up to isomorphism would implies a loss of generality.

benediktahrens:
> I meant: what do you think of the following definitions:
> 
>     * `exact (λ c', ∑c:C,(z_iso (F c) c' × ob_disp D c)).`. Note that the following notion should be possible `exact (λ c', ∑c:C,(F c=c' × D c)).`
> 
>     * `exact (λ c', ∑c:C,(|| F c=c' || × ob_disp D c)).`
> 
> 
> Both of these definitions should also work, but there's a choice to be made which one is the most suitable for your application.

I am not sure which of these definitions is the most suitable. For now, I think it is good to go with the one that has been implemented, while keeping the alternatives in mind.

PR #1720, opened 2023-06-22
[WIP] Double categories via 2-sided displayed categories

nmvdw:
This PR contains an alternative definition of double categories, formulated with 2-sided displayed categories. In addition, it contains a function that maps this definition of double categories to the usual one (as formalized by Nima some time ago). It is worthwhile to note that on the data, it is the identity, but that only the laws are formulated a bit differently.


nmvdw:
There are currently 4 places where there is an admit:
1. `predoublecategory_ver_assoc_naturality_double_cat`: in my definition, the naturality of the associator is formulated using the left associator while in the other, it is formulated for the right associator
2. `predoublecategory_ver_unitor_coherence_double_cat`: in this definition, different (but equivalent) formulations for the triangle law are chosen
3. `predoublecategory_ver_assoc_coherence_double_cat`: same point for the pentagon
4. `double_cat_to_doublecategory`: we have different set conditions. In my definition, I require that for every `f, g, h, k`, the type of squares forms a set, whereas in the other a sigma type is required to be a square: `isaset (∑ (g : pr1 x -h-> pr1 y) (k : pr12 x -h-> pr12 y), sqq g (pr22 x) (pr22 y) k)`. 

For (1)-(3): the best way to deal with these points, is by making a consistent choice between the two definitions.
For (4): I think the requirement in `doublecategory` is too strong. The horizontal morphisms don't form a set, so why should this sigma type form a set? I think it should be weakened by asking that for all `g` and `k` this type forms a set.

One more thing that I would like to change: currently, I have to write `apply Categories.homset_property`. This can be avoided by renaming `homset_property` in `UniMath.CategoryTheory.DoubleCategories.Core.DoubleCategories`.

benediktahrens:
@nmvdw : re point 4: what are the stronger assumptions under which having a two-sided inverse becomes a prop?

nmvdw:
> @nmvdw : re point 4: what are the stronger assumptions under which having a two-sided inverse becomes a prop?

The extra assumptions are that the squares form an hSet. Nima's definition where you have a left-inverse and a right-inverse, is a proposition irregardless of whether you require the squares to be a set. However, if you only require one inverse, then either you must require a coherence or that the squares form a set (which is actually required in the end anyway).

benediktahrens:
Right, I think we are not interested in "wild" double categories.

nmvdw:
> Right, I think we are not interested in "wild" double categories.

I agree, which is why I prefer the version with one inverse.

benediktahrens (approved):

nmvdw:
Changes:
- The definition in `DoubleCategories.Core.DoubleCategories` now only has 1 inverse for the associators/unitors
- Small renaming to prevent name clashes

PR #1725, opened 2023-06-26
Reactivate and speed up the commented section in AlgebraicTheoryAlgebraCategory

### LINKED ISSUE #1723: Compilation seems to hang at Qed.
Compilation of `AlgebraicTheories/AlgebraicTheoryAlgebraCategory.v` seems to hang or take an unreasonable amount of time, specifically at the `Qed.` of ` is_univalent_disp_algebraic_theory_algebra_data_full_disp_cat`.

I have only taken a very brief look, but I would not be surprised if bad programming in the definition of `algebraic_theory_algebra_data_full_disp_cat` contributes to the slowness. Specifically, there are several proofs embedded in that construction; these proofs should be made opaque.

@arnoudvanderleer : could you take a look at that?

arnoudvanderleer:
Closes #1723


nmvdw:
My guess is that the `Qed` finishes in less than a second. Is that correct?

arnoudvanderleer:
It takes about 0.3 seconds now

nmvdw (approved):
Nice.

Issue #853, opened 2017-12-19
Tactics in UniMath


siddharthist:
As discussed on #828, we need to come up with a general policy for what kinds of tactics may be included and used in UniMath, and in what situations. 

I'll show my hand early: I generally support tactics, I think if UniMath is to become more widely used then automation is essential. The people who currently work with proof assistants are already used to the extra effort of doing trivial proofs, but mathematicians in general dislike this. 

Moreover, tactics that solve trivial goals seem useless to those who understand how to prove these goals already (see e.g. my proposed tactic `hleveltac` in #828), but more casual contributors might not know the names of these lemmas or where to find them. It lowers the barrier to entry to have them "packaged" in this way. 

I agree with @mortberg that we should be very careful about complex tactics that might generate non-optimal proof terms, and should not use tactics for goals where a specific _choice_ has to be made (such as proofs of contractibility), and that a reasonable compromise is only using heavy-handed tactics in opaque proofs (although the use of `Opaque` and `Qed` is also not agreed upon yet).

mortberg:
I generally prefer type classes or canonical structures to obtain the kind of automation that a tactic like `hleveltac` provides. I have always felt like complicated special purpose tactics is the least elegant way to obtain automation in Coq. I think the two methods I mention are more fundamental and less ad hoc. In some sense they provide ways to let the user "teach" Coq basic mathematical facts and Coq can then use this knowledge to solve subgoals. I would definitely prefer if we added some simple type classes instead of adding more complicated tactics, however I'm not convinced that we need more automation at all...

DanGrayson:
It was sound of Vladimir to insist we not use type classes or canonical structures, in case we move to another proof assistant in the future.

mortberg:
Ah right, with that reasoning we should not rely on Ltac either. I think it is much more likely that future proof assistants have type classes. At the moment all of Agda, Coq Lean and Isabelle have type classes, but only Coq has Ltac...

mortberg:
PS: It seems to me like the future for Ltac is a lot more uncertain than the future for type classes. I know there are at least two prototypes for replacements for Ltac out there (called Mtac and Mtac2), but I know of no proposed replacement for type classes...

DanGrayson:
Ltac is a low-level construct that would be easy to implement in a new proof assistant.  Alternatively, we could replace every Ltac call by the term the tactic finds.

mortberg:
[...]

PR #1727, opened 2023-06-27
Add a tactic for proofs about hlevels

arnoudvanderleer:


UniMath/MoreFoundations/Tactics.v
```diff
@@ -83,3 +83,62 @@ Ltac split19 := split; [| split18].
...
+  end.
+
+(* Try to close the goal in one of the standard ways *)
+Local Ltac finish :=
+  match goal with
+  | [ |- isofhlevel 0 _ ] => apply iscontrunit
+  | [ |- isofhlevel 1 _ ] => apply propproperty
+  | [ |- isofhlevel 2 _ ] => apply setproperty
+  | _ =>
+    apply isofhlevel_HLevel ||
```
nmvdw, line 119: For this tactic, you can match the goal in the following way
```
| [ |- isofhlevel (S ?n) (HLevel n)
```
(something like this)

UniMath/MoreFoundations/Tactics.v
```diff
@@ -83,3 +83,62 @@ Ltac split19 := split; [| split18].
...
+  refine (_ : isofhlevel _ (∑ _, _)) ||
+  refine (_ : isofhlevel _ (_ → _)) ||
+  refine (_ : isofhlevel _ (∏ tmp, _)).
+
+(* Reduce a goal about hlevels to its components *)
+Ltac prove_hlevel' n :=
+  convert_to_hlevel;
+  repeat progress_hlevel;
+  match n with
+  | _ ?n' => unfold_hlevel_expression; prove_hlevel' n'
```
nmvdw, line 136: I don't understand this part of the code. Why do you need to match `n`?

UniMath/MoreFoundations/Tactics.v
```diff
@@ -83,3 +83,62 @@ Ltac split19 := split; [| split18].
...
+  match goal with
+  | [ |- isofhlevel 0 _ ] => apply iscontrunit
+  | [ |- isofhlevel 1 _ ] => apply propproperty
+  | [ |- isofhlevel 2 _ ] => apply setproperty
+  | _ =>
+    apply isofhlevel_HLevel ||
+    apply hlevelntosn; finish
+  end.
+
+Ltac unfold_hlevel_expression :=
```
nmvdw, line 123: Can you add a comment that explains what this tactic does?

UniMath/MoreFoundations/Tactics.v
```diff
@@ -83,3 +83,62 @@ Ltac split19 := split; [| split18].
...
+  repeat progress_hlevel;
+  match n with
+  | _ ?n' => unfold_hlevel_expression; prove_hlevel' n'
+  | _ => try finish
+  end.
+
+Tactic Notation "prove_hlevel" :=
+  prove_hlevel' 0.
+
+Tactic Notation "prove_hlevel" constr(n) :=
```
nmvdw, line 143: Why do you need `constr` around the `n`?

UniMath/MoreFoundations/Tactics.v
```diff
@@ -83,3 +83,62 @@ Ltac split19 := split; [| split18].
...
+  | [ |- isaset _ ] => refine (_ : isofhlevel 2 _)
+  | _ => fail
+  end.
+
+(* Reduce the different possible constructions *)
+Local Ltac progress_hlevel :=
+  match goal with
+  | [ |- isofhlevel ?n        (_ = _)  ] => refine ((_ : isofhlevel (S n) _) _ _)
+  | [ |- isofhlevel 1         (_ ⨿ _)  ] =>
+    refine (isapropcoprod _ _ (_ : isofhlevel _ _) (_ : isofhlevel _ _) _)
```
nmvdw, line 104: This branch generates a goal `P -> Q -> empty`. Could that cause unintended errors?

nmvdw (approved):
This tactic seems quite nice. How was the performance in the cases in which you tried it?

I only have one major comment: the tactic is a bit complicated (there is some conversion, some matching), so I think it is necessary to add some tests to it.

benediktahrens:
- This tactic seems to save very little work, judging from the diff.
- The need for unfolding seems to indicate that the tactic is not yet ready.
- In general, I am hesitant to support automation, since it is a pain to fix in case of failure - in particular in code we are not familiar with.

UniMath/AlgebraicTheories/AlgebraicTheoryMorphisms.v
```diff
@@ -182,8 +181,7 @@ Proof.
   - do 2 (apply funextsec; intro).
     apply H1.
   - apply isaprop_is_nat_trans, homset_property.
-  - apply setproperty.
-  - unfold preserves_composition.
-    prove_hlevel.
-  - exact isapropunit.
+  - prove_hlevel 1.
```
nmvdw, line 184: Why do you need to use `prove_hlevel` multiple times here?

UniMath/MoreFoundations/Tactics.v
arnoudvanderleer: This does not close the goal `isofhlevel 3 hSet` though. Is there a reason why we would want to match the goal to something in this case?

UniMath/MoreFoundations/Tactics.v
arnoudvanderleer: Ah. I thought this was a very nifty trick. The `n` in this case (maybe I should rename it) allows one to specify how many times definitions should be unfolded before being handed back to the user. This to both be able to do one-liners like
```
Lemma isaprop_is_algebraic_theory_algebra
  {T : algebraic_theory_data}
  (A : algebraic_theory_algebra_data T)
  : isaprop (is_algebraic_theory_algebra A).
Proof.
  prove_hlevel 2.
Qed.
```
while simultaneously allowing one to use `isaprop_is_algebraic_theory_algebra` in the cases where we are proving that something (with `is_algebraic_theory_algebra` as a component) is a proposition.

UniMath/MoreFoundations/Tactics.v
arnoudvanderleer: Yes, will do!

UniMath/MoreFoundations/Tactics.v
arnoudvanderleer: I believe it wouldn't, because after `convert_to_hlevel`, anything that cannot be processed further and cannot be finished by the tactic will just be handed back to the user as a goal.
Also, from my experience, it doesn't.
However, I will specifically test this just once more.

UniMath/MoreFoundations/Tactics.v
arnoudvanderleer: This is because, as far as I understand it (but the docs are not very easy to read), one needs to specify what kind of parameter we are expecting. In this case, we just want some term (usually something of type `nat`, but if someone wants to provide a list, that is also not a problem): https://coq.inria.fr/refman/user-extensions/syntax-extensions.html#coq:cmd.Tactic-Notation

nmvdw:
> * This tactic seems to save very little work, judging from the diff.
> * The need for unfolding seems to indicate that the tactic is not yet ready.
> * In general, I am hesitant to support automation, since it is a pain to fix in case of failure - in particular in code we are not familiar with.

My comments to this would be as follows:
- The kind of goals that are solved, are of the form `isaprop`, `isaset`, `iscontr` or `isofhlevel`
- These goals can be solved systematically and always in the same way.
- The added automation would then go through the term systematically and build the expected proof
- It is kinda similar to `is_iso` for bicategories (https://github.com/UniMath/UniMath/blob/master/UniMath/Bicategories/Core/Invertible_2cells.v#L313)

However, it is indeed the case that automation can be difficult to fix in case of failure. For that reason, I think there are four main requirements for this tactic:
1. It must be robust.
2. It must be well-tested in a wide variety of cases.
3. It must be simple.
4. The goals that it solves, should be simple enough so that a proof for them can easily be reproduced.

Especially (1) and (2) are important, I would say.

UniMath/MoreFoundations/Tactics.v
nmvdw: Hmm, good point. Maybe they could have separate clauses?

m-lindgren:
This is not the first time something like this has been proposed, see: https://github.com/UniMath/UniMath/pull/828.

arnoudvanderleer:
> This is not the first time something like this has been proposed, see: #828 and the discussion in #853.

Ah, thanks for mentioning this. I had not seen those. It seems like there is very much not a consensus about this, and not a final decision has been made in the 2017 and 2018 discussions.

Even though I think this tactic would be a good idea, and not a lot can go wrong, I will close this PR for now. If, at some point in the future, a consensus has been reached in favour of this kind of tactic, someone can reopen it.

UniMath/MoreFoundations/Tactics.v
arnoudvanderleer: I checked: it won't cause unintended errors.

PR #1732, opened 2023-07-13
Rewrite some proofs in ReindexindForward.v

EwenBC:


UniMath/CategoryTheory/DisplayedCats/ReindexingForward.v
```diff
@@ -22,23 +21,19 @@ Require Import UniMath.CategoryTheory.Core.Isos.
...
 Local Open Scope cat. 
 
 (*%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%*)
 
-Definition transportf_mor {C:category} {a b a' b':C} 
-(p:a=a') (q:b=b') (x:a-->b) : a'-->b'
-:= transportf (precategory_morphisms a') q (transportf (λ z, z--> b) p x). 
-
-Definition transportf_mor_disp {C C':category} {D':disp_cat C'} {a b a' b':C}
+Definition double_transport_disp {C C':category} {D':disp_cat C'} {a b a' b':C}
```
benediktahrens, line 33: Should this definition go into `DisplayedCats.Core`?

UniMath/CategoryTheory/Core/Univalence.v
```diff
@@ -106,6 +106,44 @@ Definition double_transport {C : precategory_ob_mor} {a a' b b' : ob C}
...
+  : double_transport px py f
+  · double_transport py pz g =
+  double_transport px pz (f · g).
+Proof.
+  unfold double_transport.
+  destruct px, py, pz.
+  reflexivity.
+Qed.
+
+
```
benediktahrens, line 146: Would this material be better placed into `CategoryTheory.Core.Categories`?

EwenBC: double_transport is defined just above in Univalence so I'd have to move it to. I don't know why it was written there so I don't want to take the risk of moving it.

UniMath/CategoryTheory/DisplayedCats/ReindexingForward.v
EwenBC: Done !

PR #1734, opened 2023-07-14
add info on different ways of using UniMath to README

benediktahrens:


nmvdw (approved):

rmatthes (commented):
I checked for Lecture 4 of the UniMath school. As I mentioned in discussion #1729, I wonder how the window size can be divided into the three panels. The defaults are not good for that lecture.

rmatthes (approved):

PR #1735, opened 2023-07-18
The Grothendieck construction as a biequivalence

nmvdw:
This PR contains a proof of the Grothendieck construction as a biequivalence of bicategories (see https://ncatlab.org/nlab/show/Grothendieck+construction#statement_of_the_equivalence).

There currently is 1 hole in the proof. I make use of univalent bicategories, and it still needs to be proven that the bicategory of fibrations over a fixed category `C` is univalent. This is for a later PR.


benediktahrens:
Where do you use that the underlying base category C is univalent?

nmvdw:
> Where do you use that the underlying base category C is univalent?

I think that, in hindsight, that assumption is needed. Since I was looking at pseudofunctors to univalent categories, I required that `C` is univalent as well, because why not. I will fix that and drop that assumption when it is not necessary.

nmvdw:
> Where do you use that the underlying base category C is univalent?

This assumption is now removed, but we do need that the displayed category is univalent, because it is about pseudofunctors to the bicategory of univalent categories.

benediktahrens:
> > Where do you use that the underlying base category C is univalent?
> 
> This assumption is now removed, but we do need that the displayed category is univalent, because it is about pseudofunctors to the bicategory of univalent categories.

Can the construction be modularized in a way that the univalence on either side is put on top of a biequivalence of bicategories? In the style of a displayed biequivalence - but without necessarily implementing this concept?

nmvdw:
> > > Where do you use that the underlying base category C is univalent?
> > 
> > 
> > This assumption is now removed, but we do need that the displayed category is univalent, because it is about pseudofunctors to the bicategory of univalent categories.
> 
> Can the construction be modularized in a way that the univalence on either side is put on top of a biequivalence of bicategories? In the style of a displayed biequivalence - but without necessarily implementing this concept?

The bicategory of univalent categories is not defined as a displayed bicategory over the bicategory of categories, which would be a necessary precondition for that. So, that would not be possible.

However, I don't think this is an issue (examples of displayed categories that are not univalent, are very scarce).

nmvdw:
> > > Where do you use that the underlying base category C is univalent?
> > 
> > 
> > This assumption is now removed, but we do need that the displayed category is univalent, because it is about pseudofunctors to the bicategory of univalent categories.
> 
> Can the construction be modularized in a way that the univalence on either side is put on top of a biequivalence of bicategories? In the style of a displayed biequivalence - but without necessarily implementing this concept?

If we use this approach then:
- Several definitions need to be redefined. For example, the fibrational slice is defined for displayed univalent categories.
- The proof becomes more complicated. I make use of the fact that pointwise adjoint equivalences are adjoint equivalences in the bicategory of pseudofunctors. The proof of this uses univalence, because that allows us to do induction on adjoint equivalences. This simplifies dealing with coherencies, which is a big advantage. Without this, then the inverse of the unit and counit need to be defined by hand, and that would add quite some code (couple of thousand lines).
- For displayed univalent categories, the result would then become more inconvenient to use. That is because we look at the full subbicategory of pseudofunctors such that they land in univalent categories. As a result, there are added `tt` to 1-cells and 2-cells, and objects are not definitionally equal to pseudofunctors into univalent categories.

I don't think there is much benefit to this. Examples of displayed categories that are not univalent are very scarce. As such, I think it is better to keep the assumption that the displayed category is univalent.

PR #1739, opened 2023-07-20
formalizing category of monads as displayed category

rmatthes:
this is a reimplementation that needs adaptation in the use as well

issues:
hssMor_Monad_Mor_eq is admitted in
UniMath/SubstitutionSystems/MonadsFromSubstitutionSystems.v UniMath/SubstitutionSystems/SimplifiedHSS/MonadsFromSubstitutionSystems.v compilation failure for:
UniMath/SubstitutionSystems/MonadsMultiSorted.v
UniMath/CategoryTheory/Monads/KTriplesEquiv.v
UniMath/CategoryTheory/Monads/Derivative.v
UniMath/CategoryTheory/Monads/KleisliCategory.v
UniMath/CategoryTheory/Monads/Kleisli.v
(this may hide many more problems downstream in compilation)

PR #1743, opened 2023-07-23
contributes a displayed definition of the category of comonads

rmatthes:
- this dualizes the most recent version of Monads.v (PR #1739)
- the Eilenberg-Moore construction for the comonads is purely by dualization


UniMath/CategoryTheory/Monads/Comonads.v
```diff
@@ -0,0 +1,480 @@
...
+  Proof.
+    exact (pr1 (pr122 T)).
+  Qed.
+
+  Lemma Comonad_law2 {T : Comonad} : ∏ c : C, δ T c · #T (ε T c)  = identity (T c).
+  Proof.
+    exact (pr2 (pr122 T)).
+  Qed.
+
+  Lemma Comonad_law3 {T : Comonad} : ∏ c : C, δ T c · #T (δ T c) = δ T c · δ T (T c).
```
nmvdw, line 133: For `Comonad_law1`, `Comonad_law2`, `Comonad_law3`, do you think it is better to have `T` implicit or explicit? When I formalized the enriched Kleisli category of a monad, I often had to provide the relevant monad by hand.

nmvdw (approved):

UniMath/CategoryTheory/Monads/Comonads.v
rmatthes: You could also locally change the settings for implicit arguments. I just wanted to stick with the declarations that were used for monads.

nmvdw (approved):

rmatthes:
@nmvdw Did you approve the Eilenberg-Moore dualization as well in such a short time span?

nmvdw:
> @nmvdw Did you approve the Eilenberg-Moore dualization as well in such a short time span?

I recognized a lot of the code, because it is very similar to what I wrote. So, I only checked whether the definitions were okay, and I think they are.

rmatthes:
@nmvdw Okay, so the system was right and you applied your expertise.

PR #1744, opened 2023-07-24
Monoidal adjunctions

nmvdw:
- Define the notion of monoidal adjunction
- An adjunction is monoidal if the left adjoint is a strong monoidal functor


UniMath/CategoryTheory/Monoidal/Adjunctions.v
```diff
@@ -0,0 +1,1025 @@
...
+
+ In this file, we define monoidal adjunctions. We define this notion in the following
+ way: a monoidal adjunction  is adjunction of categories such that both functors are
+ lax monoidal and such that the unit and counit are monoidal transformations. This is
+ given in [monoidal_adjunction].
+
+ However, in practice it is often useful to use the following characterization: an
+ adjunction is monoidal if the left adjoint is a strong monoidal functor. A reference
+ for this result is Lemma 13 in https://hal.science/hal-00154229/document. The relevant
+ statements are [monoidal_adjunction_from_strong] and [is_strong_left_adjoint].
```
benediktahrens, line 13: Since there are many choices for what a "monoidal adjunction" is, it would be good to give a reference, e.g., a (versioned) link to nlab or similar.

benediktahrens (approved):

UniMath/CategoryTheory/Monoidal/Adjunctions.v
nmvdw: I added a reference and more explanation

Issue #1655, opened 2023-03-13
Move a file


nmvdw:
The file https://github.com/UniMath/UniMath/blob/master/UniMath/Bicategories/PseudoFunctors/DisplayedUniversalArrow.v should be moved to the directory `DisplayedBicats`, because it is about displayed gadgets.

PR #1749, opened 2023-08-01
move displayed universal arrow

ReimannJ:
See #1655


benediktahrens:
The files to be moved were deleted. To move them instead, use
```
$ git move <old> <new>
```

PR #1751, opened 2023-08-08
Bicategory of univalent double categories

nmvdw:
This PR is currently work in progress. The content is going to be:
- The bicategory of univalent double categories (defined via a displayed bicategory)
- A proof that this bicategory is univalent
- Refactor the notion of double category (in the directory on 2-sided displayed categories) accordingly, so that it is the same to the obtained notion of double category via the construction of this displayed bicategory.

This required a refactoring in the directory on 2-sided displayed categories to make certain proofs more efficient (regarding memory) and simpler.


benediktahrens:
> The thing that is currently missing from this PR is a file in the directory `DoubleCategories` which gives an interface for the objects, 1-cells, and 2-cells of the bicategory of double categories. This can either be done in this PR or in a separate PR.
> 
> What would be better in this case? @benediktahrens

Whatever suits the developer(s) best!

nmvdw:
> > The thing that is currently missing from this PR is a file in the directory `DoubleCategories` which gives an interface for the objects, 1-cells, and 2-cells of the bicategory of double categories. This can either be done in this PR or in a separate PR.
> > What would be better in this case? @benediktahrens
> 
> Whatever suits the developer(s) best!

Sure. I think it is ready for review now.

UniMath/Bicategories/DoubleCategories/DoubleFunctor/Associator.v
```diff
@@ -0,0 +1,183 @@
...
+Require Import UniMath.CategoryTheory.TwoSidedDisplayedCats.TwoSidedDispCat.
+Require Import UniMath.CategoryTheory.TwoSidedDisplayedCats.Isos.
+Require Import UniMath.CategoryTheory.TwoSidedDisplayedCats.DisplayedFunctor.
+Require Import UniMath.Bicategories.DoubleCategories.DoubleCategoryBasics.
+Require Import UniMath.Bicategories.DoubleCategories.DoubleFunctor.Basics.
+
+Local Open Scope cat.
+
+Unset Kernel Term Sharing.
+(* this is to reduce the memory consumption of this file *)
```
benediktahrens, line 22: Does this work? Do you have the numbers with and without this option?

nmvdw: `Unset Kernel Term Sharing` switches from lazy to strict evaluation. So, it might increase time consumption while reducing memory consumption

With that option:
- The `Qed` takes 33.027 seconds
- At the end of the file, Coq uses 5.92 GB RAM

Without:
- The `Qed`: takes 42.759
- At the end of the file, Coq uses 7.27 GB RAM

So, I think it is very helpful

benediktahrens: Could you add this information to the comment in the source file?

nmvdw: I added this comment

nmvdw: The documentation is here: https://coq.github.io/doc/master/refman/proofs/writing-proofs/equality.html#coq:flag.Kernel-Term-Sharing

UniMath/Bicategories/DoubleCategories/BicatOfDoubleCats.v
```diff
@@ -0,0 +1,1711 @@
+(**********************************************************************************
+
+ The Bicategory of Double Categories
+
+ In this file, we define the bicategory of univalent double categories, and we
+ prove that this bicategory is univalent. To do so, we make heavy use of the
+ machinery of displayed bicategories.
+
+ The idea behind this construction is to split up the notion of double category into
+ several independent parts. More specifically, a univalent double category is a
```
benediktahrens, line 10: "Independent" is not exactly correct, since the later layers depend on the earlier ones.
```suggestion
 several layers. More specifically, a univalent double category is a
```

nmvdw:
> @nmvdw : could you remove the file `UniMath/CategoryTheory/TwoSidedDisplayedCats/.DS_Store` from this PR?

Removed.

UniMath/Bicategories/DoubleCategories/DoubleCategoryBasics.v
```diff
@@ -0,0 +1,917 @@
+(**********************************************************************************
+
+ Basics of double categories
+
+ In this file, we define the basic notions for double categories. There are a
+ couple of ideas behind the definition that we use
```
benediktahrens, line 6:
```suggestion
 couple of ideas behind the definition that we use:
```

benediktahrens (approved):
Looks great!

PR #1752, opened 2023-08-08
Symmetric monoidal comonads

rmatthes:
the definition is there, but it is not yet tested in examples (other than the construction of the Eilenberg-Moore category)


UniMath/CategoryTheory/Monoidal/Displayed/FullMonoidalSubcategory.v
```diff
@@ -0,0 +1,54 @@
+(** constructs the full monoidal subcategory through a displayed monoidal category *)
```
nmvdw, line 1: How does this file compare to
https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/Monoidal/Examples/Fullsub.v

UniMath/CategoryTheory/Monoidal/Functors.v
```diff
@@ -268,6 +278,19 @@ Section MonoidalFunctors.
...
+  Lemma fmonoidal_lax_eq
+             {C D : category}
+             {M : monoidal C}
+             {N : monoidal D}
+             {F : functor C D}
+             (fmd fmd' : fmonoidal_lax M N F) :
+    pr1 fmd = pr1 fmd' -> fmd = fmd'.
+  Proof.
+    intro H.
+    use total2_paths_f; [apply H |].
```
nmvdw, line 290: `subtypePath` might be useful here (the proof remains similar though)

UniMath/CategoryTheory/Monoidal/FunctorCategories.v
```diff
@@ -0,0 +1,264 @@
...
+  is_univalent_disp disp_cat_lax_monoidal_functors.
+Proof.
+  use is_univalent_disp_from_SIP_data.
+  - exact lax_monoidal_functors_Pisset.
+  - intros F Fm Fm'. apply lax_monoidal_functors_Hstandard.
+Defined.
+
+End LaxMonoidalFunctorCategory.
+
+Definition is_univalent_category_lax_monoidal_functors
```
nmvdw, line 99: My experience is as follows: if you want to prove `is_univalent`, then it is often convenient to have `is_univalent` in the assumptions, because then you can use that statement and the required assumptions will be generated. However, if you want a definition for a `univalent_category`, then it is nicer to have `univalent_category` in the assumptions.

nmvdw (approved):
It looks quite good so far.

UniMath/CategoryTheory/Monoidal/Displayed/FullMonoidalSubcategory.v
rmatthes: So bad that I did not look into the examples folder for that.

rmatthes: Since there are now two variants, we can compare: I see no reason why the "predicate" P has to go into hProp. I understand that my two files are a bit too short, so better have all in one. I see several unnecessary import statements in Fullsub.v. And I think it is good to spell out the total categories. @nmvdw If you agree, I can merge my stuff into that file.

nmvdw: > I see no reason why the "predicate" P has to go into hProp. 

Agreed. It can make it a bit more difficult to instantiate this construction.

> I think it is good to spell out the total categories.

Definitely. It would also be good if `total_category (disp_full_sub C P)` would be named.

>If you agree, I can merge my stuff into that file. 

I agree.

rmatthes:
Compilation is rather slow on the last steps in SymmetricMonoidalCoEilenbergMoore.v - I marked those positions.

UniMath/CategoryTheory/Monoidal/Examples/LiftPoset.v
```diff
@@ -330,6 +331,22 @@ Proof.
     induction x ; cbn ; apply idpath.
 Qed.
 
+Definition lift_poset_symmetric_monoidal_comonad :
+  symmetric_monoidal_comonad (pr21 pointed_poset_sym_mon_closed_cat).
```
nmvdw, line 335: A builder for symmetric monoidal comonads might be nice

nmvdw (approved):
Looks good to me.

rmatthes:
Would it be okay to do a "rebase and merge" and not a squash merge?

nmvdw (approved):

PR #1753, opened 2023-08-12
Fiberwise equivalences between fibrations

nmvdw:
This PR contains a rather nice application of the Grothendieck construction: to check whether a cartesian functor between fibrations gives rise to an equivalences between the fibrations, it suffices to check whether it gives rise to an adjoint equivalence on the fibers. This follows directly from the Grothendieck construction.


UniMath/Bicategories/Grothendieck/FiberwiseEquiv.v
```diff
@@ -0,0 +1,118 @@
...
+    exact (HF x).
+Defined.
+
+Definition ff_and_eso_to_adjequiv
+           {C : univalent_category}
+           {P₁ P₂ : fib_slice_bicat C}
+           (F : P₁ --> P₂)
+           (F_full : ∏ (x : C), full (fiber_functor (pr1 F) x))
+           (F_faithful : ∏ (x : C), faithful (fiber_functor (pr1 F) x))
+           (F_eso : ∏ (x : C), Functors.essentially_surjective (fiber_functor (pr1 F) x))
```
benediktahrens, line 105: Why is the qualification `Functor.` necessary here?

UniMath/Bicategories/PseudoFunctors/Biequivalence.v
```diff
@@ -328,3 +330,54 @@ Proof.
...
+(**
+ Biequivalences reflect adjoint equivalences
+ *)
+Section BiequivReflectAdjequiv.
+  Context {B₁ B₂ : bicat}
+          (L : biequivalence B₁ B₂).
+
+  Let R : psfunctor B₂ B₁ := inv_psfunctor_of_is_biequivalence L.
+  Let η := unit_of_is_biequivalence L.
+  Let Hη := is_biequivalence_adjoint_unit L.
```
benediktahrens, line 343: It would be good to give type annotations here.

UniMath/Bicategories/PseudoFunctors/Biequivalence.v
```diff
@@ -328,3 +330,54 @@ Proof.
...
+    - exact f'.
+    - use comp_left_adjoint_equivalence.
+      + use comp_left_adjoint_equivalence.
+        * apply inv_left_adjoint_equivalence.
+        * use (psfunctor_preserves_adjequiv' R).
+          exact Hf.
+      + exact (pointwise_adjequiv _ Hη y).
+    - exact biequiv_reflect_adjequiv_cell.
+    - apply property_from_invertible_2cell.
+  Defined.
```
benediktahrens, line 382: Is the proof-irrelevant stuff used in this definition opacified?

UniMath/Bicategories/Grothendieck/FiberwiseEquiv.v
nmvdw: `essentially_surjective` is used both for functors and pseudofunctors, so there is a name clash.

UniMath/Bicategories/PseudoFunctors/Biequivalence.v
nmvdw: I can interpret this question in 2 ways. If you just look at this function, the only proof irrelevant part comes from the `property_from_invertible_2cell`, so then yes.

I can also look at the total construction. Regarding the invertible 2-cells, this function makes use of:
- `psnaturality_of`: this is a projection, so that is okay
- `left_equivalence_counit_iso`: this is basically a projection, so they are okay
- `comp_of_invertible_2cell`
- `linvunitor_invertible_2cell`
- `rwhisker_of_invertible_2cell`
- `inv_of_invertible_2cell`
- `rassociator_invertible_2cell`
- `lwhisker_of_invertible_2cell`

The last 6: the proof that they are invertible is mostly opaque, but they can be made a bit more opaque (not an issue). These are opaque enough.

Regarding the adjoint equivalences, this function uses:
- `property_from_invertible_2cell`: this is a projection, so that is okay
- `left_adjoint_equivalence_invertible`
- `comp_left_adjoint_equivalence`
- `inv_left_adjoint_equivalence`
- `psfunctor_preserves_adjequiv'`
- `pointwise_adjequiv`

For `left_adjoint_equivalence_invertible`, `comp_left_adjoint_equivalence`, `inv_left_adjoint_equivalence`, and `psfunctor_preserves_adjequiv`: same as for the invertible 2-cell. The `is_iso` part can be made opaque, but that doesn't lead to a very big term.

For `pointwise_adjequiv`: this is the only one that I think would slow down a computation. It is constructed as follows:
```coq
Definition pointwise_adjequiv
           {B₁ B₂ : bicat}
           {F₁ F₂ : psfunctor B₁ B₂}
           (σ : pstrans F₁ F₂)
           (Hf : left_adjoint_equivalence σ)
  : ∏ (X : B₁), left_adjoint_equivalence (σ X).
Proof.
  intro X.
  pose (pr1 (left_adjoint_equivalence_total_disp_weq _ _ Hf)) as t₁.
  pose (pr1 (left_adjoint_equivalence_total_disp_weq _ _ t₁)) as t₂.
  pose (pr1 (left_adjoint_equivalence_total_disp_weq _ _ t₂)) as t₃.
  exact (is_adjequiv_to_all_is_adjequiv _ _ _ t₃ X).
Defined.
```
The `left_adjoint_equivalence_total_disp_weq` is expensive to compute. This can also be witnessed experimentally: if I do
```coq
      pose (pr2 biequiv_reflect_adjequiv_cell) as i.
      cbn -[pointwise_adjequiv] in i.
```
before the `property_from_invertible_2cell`, then the `cbn` is fast, whereas if I did `cbn in i` instead, then it would have been slow (longer than 5 seconds, and then I stop the computation). So, only `pointwise_adjequiv` is not opaque enough.

UniMath/Bicategories/PseudoFunctors/Biequivalence.v
nmvdw: Fixed

UniMath/Bicategories/PseudoFunctors/Biequivalence.v
nmvdw: My judgment on this matter:
- I would keep `biequiv_reflect_adjequiv` the way it is
- While `pointwise_adjequiv` could be made opaque, I would be tended to leave it the way it is. This is because it might be that with some refactoring in other places, the computation wouldn't be as slow.
- It might be good to go through all code of bicategories again. Refactor some parts and improve some parts (regarding memory consumption). I think that `left_adjoint_equivalence_total_disp_weq` might be the expensive part here, because it is a big chain of equivalences, and maybe a direct definition would be more efficient.

For this PR, I would leave it the way it is.

PR #1755, opened 2023-08-15
Comonoids vs cartesian

Kfwullaert:
In this PR, comonoids internal to a (symmetric) monoidal category has been studied.
More precisely:
1. the definitions of comonoids, commutative comonoids and comonoid morphisms; and the categories involved, have been defined;
2. the category of (commutative) comonoids is shown to be a (commutative) monoidal category;
3. some builders for (displayed, symmetric, or cartesian)  monoidal categories have been formalized;
4. a (symmetric monoidal) category of commutative comonoids is shown to be cartesian.

In CategoryTheory/Monoidal/Structure/SymmetricDiagonal.v, two admits are present.
These gaps will be solved, hopefully soon.


UniMath/CategoryTheory/catiso_displayed.v
```diff
@@ -0,0 +1,200 @@
+Require Import UniMath.Foundations.All.
```
nmvdw, line 1: Can this file name be capitalized? In addition, it might fit better in the directory `DisplayedCats`

UniMath/CategoryTheory/catiso_displayed.v
```diff
@@ -0,0 +1,200 @@
...
+Section DisplayedCatIso.
+
+  Context {C : category} (D : disp_cat C).
+
+  Context
+    (uo : ∏ c : C, iscontr (D c)).
+  Context (uf : ∏ (c1 c2 : C) (f : C⟦c1,c2⟧),
+              iscontr (pr1 (uo c1) -->[f] (pr1 (uo c2)))).
+
+  Lemma uf_eq {c1 c2 : C} (f : C⟦c1,c2⟧)
```
nmvdw, line 23: What does `uf` stand for?

UniMath/CategoryTheory/catiso_displayed.v
nmvdw: Also, can you add a comment describing the contents of this file?

UniMath/CategoryTheory/catiso_displayed.v
```diff
@@ -0,0 +1,200 @@
...
+Require Import UniMath.CategoryTheory.DisplayedCats.Total.
+
+Local Open Scope cat.
+
+Section DisplayedCatIso.
+
+  Context {C : category} (D : disp_cat C).
+
+  Context
+    (uo : ∏ c : C, iscontr (D c)).
```
nmvdw, line 19: I feel this assumption can be simplified by using univalence nicely. I would be tended to take the following assumptions:
- `D` is univalent
- above every `c : C`, there is some `cc : D c`
- the type `xx -->[ f ] yy` is contractible
- all morphisms above an isomorphism are isomorphisms (I think that follows from the previous one)

Suppose that we have `cc` and `cc'` above `c`. To show they are equal, we need to construct an isomorphism between them above the identity. This isomorphism exists by the last assumption.

UniMath/CategoryTheory/Monoidal/Displayed/SymmetricMonoidalBuilder.v
```diff
@@ -0,0 +1,279 @@
...
+    - exact (λ _ _ _ _ _ _ yy gg, Tr yy gg).
+    - exact Tl_preserves_id.
+    - intro ; intros ; apply Tr_preserves_id.
+    - exact Tl_preserves_comp.
+    - intro ; intros ; apply Tr_preserves_comp.
+    - intro ; intros ; apply T_equalwhiskers.
+  Defined.
+
+  Context (II : D I_{ M})
+    (lulu : disp_leftunitor_data make_symmetric_monoidal_disp_cat_tensor II)
```
nmvdw, line 193: Can you either name this `ll` or `lu_lu`?

UniMath/CategoryTheory/Monoidal/Comonoids/Comonoids.v
```diff
@@ -0,0 +1,435 @@
...
+
+Local Open Scope cat.
+
+Import MonoidalNotations.
+
+Section ComonoidsAndMorphisms.
+
+  Context {C : category} (M : monoidal C).
+
+  Notation "x ⊗ y" := (x ⊗_{M} y).
```
nmvdw, line 35: If you want to use this notation, then why not use the bundled approach? https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/Monoidal/Categories.v#L1301

For the bundled approach, we have this notation, and the whole infrastructure is built around it.

The advantage of the displayed approach is that it is more flexible, because one can directly express what a monoidal structure over a category is. This is necessary to define a displayed bicategory of monoidal structures In addition, it is also useful for theorems that express when an adjunction is monoidal.

The advantage of the bundled approach is that one can simplify the notation and work directly with it. With both approaches established, one can choose which one is the most fitting in certain context, because switching between them is not too difficult (calling the functions is a bit different, but still fien).

I think that the bundled approach is more appropriate here.

UniMath/CategoryTheory/Monoidal/Comonoids/Comonoids.v
```diff
@@ -0,0 +1,435 @@
...
+  Let ruinv : rightunitorinv_data M (monoidal_unit M) := monoidal_rightunitorinvdata M.
+  Let αinv : associatorinv_data M := monoidal_associatorinvdata M.
+
+  Definition comonoid_data (x : C) : UU
+    := C⟦x, x ⊗ x⟧ × C⟦x, I⟧.
+
+  Definition comonoid_data_comultiplication {x : C} (m : comonoid_data x)
+    : C⟦x, x ⊗ x⟧
+    := pr1 m.
+  Notation "μ_{ m }" := (comonoid_data_comultiplication m).
```
nmvdw, line 54: better to name it \delta instead of \mu

UniMath/CategoryTheory/Monoidal/Comonoids/Comonoids.v
```diff
@@ -0,0 +1,435 @@
...
+
+  Definition comonoid_data_counit {x : C} (m : comonoid_data x)
+    : C⟦x, I⟧
+    := pr2 m.
+  Notation "η_{ m }" := (comonoid_data_counit m).
```
nmvdw, line 59: better to name it epsilon instead of eta

UniMath/CategoryTheory/Monoidal/Comonoids/Comonoids.v
```diff
@@ -0,0 +1,435 @@
...
+        apply (pr2 pg).
+      }
+      apply (pr2 pf).
+  Qed.
+
+End ComonoidsAndMorphisms.
+
+Module ComonoidNotations.
+
+  Notation "μ_{ m }" := (comonoid_data_comultiplication _ m).
```
nmvdw, line 171: Delta and epsilon are better

UniMath/CategoryTheory/Monoidal/Comonoids/Comonoids.v
```diff
@@ -0,0 +1,435 @@
...
+  Notation "η_{ m }" := (comonoid_data_counit _ m).
+
+End ComonoidNotations.
+
+Section ComonoidAux.
+
+  Context {C : category} (M : monoidal C).
+
+  Notation "x ⊗ y" := (x ⊗_{M} y).
```
nmvdw, line 180: Here the bundled approach is more fitting as well.

UniMath/CategoryTheory/Monoidal/Comonoids/Comonoids.v
```diff
@@ -0,0 +1,435 @@
...
+    now rewrite id_right.
+  Qed.
+
+End ComonoidAux.
+
+Section CommutativeComonoids.
+
+  Context {C : category} {M : monoidal C} (S : symmetric M).
+
+  Notation "x ⊗ y" := (x ⊗_{M} y).
```
nmvdw, line 245: Same as above

UniMath/CategoryTheory/Monoidal/Comonoids/Comonoids.v
nmvdw: It is even possible to solve this using a `let`. You could write `Let M' : monoidal_cat := C ,, M.` and use `M` in the remainder. This way calling the function remains the same.

UniMath/CategoryTheory/Monoidal/Comonoids/ComonoidsMonoidalCategory.v
```diff
@@ -0,0 +1,818 @@
...
+Require Import UniMath.CategoryTheory.Monoidal.Comonoids.ComonoidsCategory.
+
+Local Open Scope cat.
+Import MonoidalNotations.
+
+Section SymmetricMonoidalCategoryOfComonoids.
+
+  Context {C : category} {M : monoidal C} (S : symmetric M).
+
+  Notation "x ⊗ y" := (x ⊗_{M} y).
```
nmvdw, line 44: The bundled approach might be better

UniMath/CategoryTheory/Monoidal/Comonoids/ComonoidsMonoidalCategory.v
```diff
@@ -0,0 +1,818 @@
...
+    (xx : comonoid M x) (yy : comonoid M y) (zz : comonoid M z)
+    : is_comonoid_mor M (tensor_of_comonoids xx (tensor_of_comonoids yy zz))
+        (tensor_of_comonoids (tensor_of_comonoids xx yy) zz) αinv^{M}_{x, y, z}.
+  Proof.
+    split.
+    - exact (comonoid_disp_associatorinv_mult xx yy zz).
+    - exact (comonoid_disp_associatorinv_unit xx yy zz).
+  Qed.
+
+  Definition comonoid_disp_symmetric_monoidal
```
nmvdw, line 703: I am thinking of whether it is possible to simplify this construction. Note the following:
- We have a functor `F : C -> C` which sends `X` to `X \otimes X`. If `C` is symmertic, then this functor is strong monoidal.
- We have a functor `G : C -> C` which sends `X` to `I`. This functor is monoidal because `I` is the unit
- We can take the dialgebras for identity and `F`, this gives the comultiplication, and for the identity and `G`, which gives the counit.
- We take a product of these displayed categories, and we have the data of comonoids.
- Afterwards, we take a full subcategory for the laws.

UniMath/CategoryTheory/Monoidal/Comonoids/MonoidalCartesianBuilder.v
```diff
@@ -0,0 +1,354 @@
...
+Require Import UniMath.CategoryTheory.Monoidal.Structure.SymmetricDiagonal.
+
+Require Import UniMath.CategoryTheory.Monoidal.Comonoids.Comonoids.
+Require Import UniMath.CategoryTheory.Monoidal.Comonoids.ComonoidsCategory.
+Require Import UniMath.CategoryTheory.Monoidal.Comonoids.ComonoidsMonoidalCategory.
+
+Local Open Scope cat.
+Import MonoidalNotations.
+
+(* Definition make_section_into_comonoids
```
nmvdw, line 34: Is this one needed?

UniMath/CategoryTheory/Monoidal/Comonoids/MonoidalCartesianBuilder.v
```diff
@@ -0,0 +1,354 @@
...
+    unfold section_disp.
+    use tpair.
+    - exists m.
+      exact mf.
+    - split ; intro ; intros ; apply isaprop_is_comonoid_mor.
+  Defined. *)
+
+Section CartesianBuilder.
+
+  Context {C : category} (M : monoidal C).
```
nmvdw, line 46: The bundled approach might be nicer here. The reason why I prefer the bundled approach, is that it allows to define the relevant notation globally. However, now you need to manage the notation in every section (including levels).

nmvdw (approved):
I think there are several points where the code can be simplified. One construction can be simplified and some notation can be simplified.

UniMath/CategoryTheory/catiso_displayed.v
Kfwullaert: It does not have a specific meaning. I called uo the "section on objects" and because I usually denote morphisms as f, I just changed uo to uf.

UniMath/CategoryTheory/Monoidal/Comonoids/MonoidalCartesianBuilder.v
Kfwullaert: No, that is why it is commented out. Should I remove it?

Kfwullaert:
@nmvdw In the latest commit, I have redefined the category of comonoids in terms of dialgebras (as you suggested).

UniMath/CategoryTheory/Monoidal/Comonoids/ComonoidsMonoidal.v
```diff
@@ -29,17 +29,20 @@ Require Import UniMath.CategoryTheory.Monoidal.Structure.Cartesian.
...
 
 Section SymmetricMonoidalCategoryOfComonoids.
 
-  Context {C : category} {M : monoidal C} (S : symmetric M).
+  Context (V : sym_monoidal_cat).
+  Let C : category := V.
+  Let M : monoidal C := V.
+  Let S : symmetric M := pr2 V.
 
   Notation "x ⊗ y" := (x ⊗_{M} y).
```
nmvdw, line 47: This notation should be redundant now.

UniMath/CategoryTheory/Monoidal/Comonoids/ComonoidsMonoidal.v
```diff
@@ -29,17 +29,20 @@ Require Import UniMath.CategoryTheory.Monoidal.Structure.Cartesian.
...
   Notation "x ⊗l f" := (x ⊗^{M}_{l} f) (at level 31).
```
nmvdw, line 48: What about adding this notation globally?

UniMath/CategoryTheory/Monoidal/Comonoids/ComonoidsMonoidal.v
```diff
@@ -456,19 +459,19 @@ Section SymmetricMonoidalCategoryOfComonoids.
...
+        apply (tensor_sym_mon_braiding).
       }
       rewrite ! assoc'.
       apply comult_before_rearrange_and_swap.
     - unfold is_comonoid_mor_unit.
       cbn.
       rewrite ! assoc.
       apply maponpaths_2.
-      etrans. { apply pathsinv0, (tensor_sym_mon_braiding ((C,,M),,S)). }
+      etrans. { apply pathsinv0, (tensor_sym_mon_braiding V). }
```
nmvdw, line 470: Did `rewrite <- tensor_sym_mon_braiding` work here?

UniMath/CategoryTheory/Monoidal/Comonoids/CommComonoidsCartesian.v
```diff
@@ -35,17 +36,18 @@ Import ComonoidNotations.
 Section CartesianMonoidalCategoryOfCommutativeComonoids.
 
   Context {C : category} {M : monoidal C} (S : symmetric M).
+  Let V : sym_monoidal_cat := (C,,M),,S.
 
   Lemma diagonal_is_comonoid_mor_mult
     {x : C} {m : comonoid M x} (c : is_commutative S m)
```
nmvdw, line 42: `x : V` might be nicer than `x : C` here

UniMath/CategoryTheory/Monoidal/Comonoids/ComonoidsCategoryAsDialg.v
```diff
@@ -0,0 +1,601 @@
...
+
+Local Open Scope cat.
+
+Import MonoidalNotations.
+
+Section DiagFunctor.
+
+  Context (V : sym_monoidal_cat).
+
+  Notation "x ⊗ y" := (monoidal_cat_tensor_pt x y).
```
nmvdw, line 50: This notation is redundant.

UniMath/CategoryTheory/Monoidal/Comonoids/ComonoidsCategoryAsDialg.v
```diff
@@ -0,0 +1,601 @@
...
+  Notation "f ⊗⊗ g" := (monoidal_cat_tensor_mor f g) (at level 31).
```
nmvdw, line 51: This notation is redundant and I think another one was used.

UniMath/CategoryTheory/Monoidal/Comonoids/ComonoidsCategoryAsDialg.v
```diff
@@ -0,0 +1,601 @@
...
+
+
+  Definition V_counit
+    : disp_monoidal (dialgebra_disp_cat (functor_identity V) (constant_functor _ _ I_{V})) V.
+  Proof.
+    use (dialgebra_disp_monoidal (identity_fmonoidal V)
+           (constant_functor_fmonoidal V unit_monoid _ _)).
+    - refine (_ ,, _).
+      apply (monoidal_leftunitorisolaw V).
+    - apply Isos.identity_is_z_iso.
```
nmvdw, line 352: Imports should be managed

nmvdw (approved):

nmvdw:
> @nmvdw In the latest commits, I have redefined the category of comonoids in terms of dialgebras (as you suggested).

Which construction do you prefer?

Kfwullaert:
> > @nmvdw In the latest commits, I have redefined the category of comonoids in terms of dialgebras (as you suggested).
> 
> Which construction do you prefer?

I prefer the dialgebra construction for multiple reasons, such as the possibility to reuse the constructions/results of dialgebras.

rmatthes:
I also used dialgebras for the Eilenberg-Moore construction given a symmetric monoidal comonad (as suggested by @nmvdw ).

UniMath/CategoryTheory/Monoidal/Comonoids/CartesianAsComonoids.v
```diff
@@ -0,0 +1,449 @@
...
+    etrans.
+    2: {
+      apply (bifunctor_distributes_over_id (F := M)) ; apply M.
+    }
+    apply maponpaths.
+    apply (semi_cart_to_unit_eq Ccart).
+  Qed.
+
+  Lemma cartesian_lunitor (x : C)
+    : semi_cart_tensor_pr2 (pr1 Ccart) I_{ M} x = lu^{M}_{x}.
```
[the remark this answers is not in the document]
nmvdw: Semi-cartesian means that the unit is a terminal object. In that case, we can make a first and second projections from the tensor: so `semi_cart_tensor_pr1 : x \otimes y -> x` and `semi_cart_tensor_pr2 : x \otimes y -> y`. These were defined outside of this PR: https://github.com/UniMath/UniMath/blob/56465756fba8e2079389b3e57c13ec082fbf1239/UniMath/CategoryTheory/Monoidal/Structure/Cartesian.v#L61.

To formulate cartesianness, we say that these two functions turn `x \otimes y` into a product.

What name would you prefer for this function?

UniMath/CategoryTheory/Monoidal/Comonoids/CommComonoidsCartesian.v
```diff
@@ -0,0 +1,194 @@
...
+Require Import UniMath.CategoryTheory.Monoidal.Comonoids.MonoidalCartesianBuilder.
+
+Local Open Scope cat.
+Import MonoidalNotations.
+Import ComonoidNotations.
+
+Section CartesianMonoidalCategoryOfCommutativeComonoids.
+
+  Context {C : category} {M : monoidal C} (S : symmetric M).
+  Let V : sym_monoidal_cat := (C,,M),,S.
```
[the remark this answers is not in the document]
nmvdw: There is not. Note: in the parts where `sym_monoidal_cat` is used the most often, the bundled approach is used consistently, so in those parts, the context solely consists of `V : sym_monoidal_cat`.

UniMath/CategoryTheory/Monoidal/Comonoids/ComonoidsMonoidal.v
```diff
@@ -0,0 +1,824 @@
...
+  Let I : C := monoidal_unit M.
+  Let lu : leftunitor_data M (monoidal_unit M) := monoidal_leftunitordata M.
+  Let ru : rightunitor_data M (monoidal_unit M) := monoidal_rightunitordata M.
+  Let α : associator_data M := monoidal_associatordata M.
+  Let luinv : leftunitorinv_data M (monoidal_unit M) := monoidal_leftunitorinvdata M.
+  Let ruinv : rightunitorinv_data M (monoidal_unit M) := monoidal_rightunitorinvdata M.
+  Let αinv : associatorinv_data M := monoidal_associatorinvdata M.
+
+  Notation "σ_{ x , y }" := (monoidal_braiding_data (symmetric_to_braiding S) x y).
+  Notation "μ_{ m }" := (comonoid_data_comultiplication _ m).
```
rmatthes, line 61: The comonoid here still has components called $\mu$ and $\eta$.

UniMath/CategoryTheory/Monoidal/Comonoids/Comonoids.v
```diff
@@ -0,0 +1,435 @@
...
+
+Require Import UniMath.CategoryTheory.Monoidal.WhiskeredBifunctors.
+Require Import UniMath.CategoryTheory.Monoidal.Categories.
+Require Import UniMath.CategoryTheory.Monoidal.Structure.Symmetric.
+Require Import UniMath.CategoryTheory.Monoidal.Structure.SymmetricDiagonal.
+Require Import UniMath.CategoryTheory.Monoidal.Structure.Cartesian.
+Require Import UniMath.CategoryTheory.Monoidal.Displayed.WhiskeredDisplayedBifunctors.
+Require Import UniMath.CategoryTheory.Monoidal.Displayed.Monoidal.
+Require Import UniMath.CategoryTheory.Monoidal.Displayed.TotalMonoidal.
+Require Import UniMath.CategoryTheory.Monoidal.Displayed.Symmetric.
```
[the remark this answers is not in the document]
Kfwullaert: You are completely right. This is something that I'm going to adapt soon.

UniMath/CategoryTheory/DisplayedCats/CatIsoDisplayed.v
```diff
@@ -0,0 +1,286 @@
...
+      }
+      exact p.
+  Qed.
+
+  Lemma weq_hom
+    {c1 : C} {c2 : C}
+    (d1 : D c1)
+    (d2 : D c2)
+    (f : C ⟦ c1, c2 ⟧)
+    : ∃! x : ∑ f0 : C ⟦ c1, c2 ⟧, d1 -->[ f0] d2, pr1 x = f.
```
nmvdw, line 91: Would it be possible to use parentheses here?

UniMath/CategoryTheory/DisplayedCats/CatIsoDisplayed.v
```diff
@@ -0,0 +1,286 @@
...
+      {
+        intro.
+        apply isapropifcontr.
+        apply uf0.
+      }
+      exact q.
+  Qed.
+
+  Lemma weq_ob (c : C)
+    : ∃! x : ∑ x : C, D x, pr1 x = c.
```
nmvdw, line 112: And here

UniMath/CategoryTheory/Monoidal/Comonoids/CartesianAsComonoids.v
```diff
@@ -0,0 +1,449 @@
...
+    etrans.
+    2: {
+      apply (bifunctor_distributes_over_id (F := M)) ; apply M.
+    }
+    apply maponpaths_2.
+    apply (semi_cart_to_unit_eq Ccart).
+  Qed.
+
+  Lemma identity_of_rwhisker_with_unit (x : C)
+    :  monoidal_cat_tensor_mor (V := C,,M) (identity x) (aug I_{ M})
```
nmvdw, line 79: The notation that I developed for monoidal categories can be used here: #⊗. Also, use `x : V` so that you take full advantage of the bundled approach.

UniMath/CategoryTheory/Monoidal/Comonoids/CartesianAsComonoids.v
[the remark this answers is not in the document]
nmvdw: It can be a bit misleading, because if one only has semi-cartesianess, then the tensor is not necessarily a product.

UniMath/CategoryTheory/Monoidal/Comonoids/CartesianAsComonoids.v
```diff
@@ -0,0 +1,449 @@
...
+      }
+      rewrite assoc.
+      refine (_ @ id_left _).
+      apply maponpaths_2.
+      unfold diag.
+      refine ((BinProductPr2Commutes) _ _ _ (is_cartesian_BinProduct Ccart _ _) _ _ _).
+  Qed.
+
+  Lemma diag_is_symmetric (x : C)
+    :  diag x · cartesian_to_braiding_data Ccart x x = diag x.
```
nmvdw, line 232:
```suggestion
    : diag x · cartesian_to_braiding_data Ccart x x = diag x.
```

UniMath/CategoryTheory/Monoidal/Comonoids/CommComonoidsCartesian.v
```diff
@@ -0,0 +1,194 @@
...
+    {x : C} {m : comonoid M x} (c : is_commutative S m)
+    : comonoid_data (sym_monoidal_cat_commutative_comonoids V) (x,,m,,c).
+  Proof.
+    refine (_ ,, _).
+    - refine (pr11 m ,, (_ ,, tt) ,, tt).
+      split ; cbn.
+      + refine (diagonal_is_comonoid_mor_mult c @ _).
+        apply maponpaths.
+        cbn.
+        unfold MonoidalDialgebras.dialgebra_disp_tensor_op.
```
nmvdw, line 97: Manage the imports so that you don't have to qualify.

UniMath/CategoryTheory/Monoidal/Comonoids/CommComonoidsCartesian.v
```diff
@@ -0,0 +1,194 @@
...
+        apply maponpaths_2.
+        apply pathsinv0, id_left.
+      + refine (_ @ ! diagonal_is_comonoid_mor_unit m @ _).
+        * apply id_right.
+        * cbn.
+          apply maponpaths.
+          unfold MonoidalDialgebras.dialgebra_disp_tensor_op.
```
nmvdw, line 104: Manage the imports so you don't have to qualify

UniMath/CategoryTheory/Monoidal/Comonoids/ComonoidsCategoryAsDialg.v
nmvdw: #⊗ is the right one

UniMath/CategoryTheory/Monoidal/Comonoids/ComonoidsCategoryAsDialg.v
```diff
@@ -0,0 +1,601 @@
...
+    use make_functor_data.
+    - exact (λ x, x ⊗ x).
+    - exact (λ _ _ f, f ⊗⊗ f).
+  Defined.
+
+  Lemma diag_is_functor
+    : is_functor diag_functor_data.
+  Proof.
+    split ; intro ; intros.
+    - apply (bifunctor_distributes_over_id (F := V)) ; apply (pr21 V).
```
nmvdw, line 65: With the bundled approach, these laws have slight different names. You can the appropriate one in `Monoidal.Categories.v`, and there's no need to mention the bifunctor directly.

UniMath/CategoryTheory/Monoidal/Comonoids/ComonoidsCategoryAsDialg.v
```diff
@@ -0,0 +1,601 @@
...
+
+  Definition diag_functor_fmonoidal_lax
+    : fmonoidal_lax V V diag_functor.
+  Proof.
+    exists diag_functor_fmonoidal_data.
+    exact diag_functor_fmonoidal_laxlaws.
+  Defined.
+
+  Definition diag_functor_is_strong_fmonoidal
+    :  fmonoidal_stronglaws
```
nmvdw, line 207:
```suggestion
    : fmonoidal_stronglaws
```

UniMath/CategoryTheory/Monoidal/Comonoids/ComonoidsCategoryAsDialg.v
```diff
@@ -0,0 +1,601 @@
...
+    exact diag_functor_is_strong_fmonoidal.
+  Defined.
+
+  Definition diag_functor_is_symmetric
+    : is_symmetric_monoidal_functor (pr2 V) (pr2 V) diag_functor_fmonoidal.
+  Proof.
+    intro ; intro.
+    apply pathsinv0.
+    apply (rearrange_commute_with_swap (pr2 V)).
+  Defined.
```
nmvdw, line 232:
```suggestion
 Qed.
```

UniMath/CategoryTheory/Monoidal/Comonoids/ComonoidsCategoryAsDialg.v
```diff
@@ -0,0 +1,601 @@
...
+
+  Definition constant_functor_fmonoidal
+    : fmonoidal V V (constant_functor _ _ x).
+  Proof.
+    exists constant_functor_fmonoidal_lax.
+    exact constant_functor_fmonoidal_strong.
+  Defined.
+
+  Definition constant_functor_is_symmetric
+               (m_is_comm : (pr12 V) x x · pr11 m = pr11 m)
```
nmvdw, line 295: What is `pr12 V`? The notation should be used. It might be `sym_mon_braiding`

UniMath/CategoryTheory/Monoidal/Comonoids/ComonoidsCategoryAsDialg.v
```diff
@@ -0,0 +1,601 @@
...
+    : is_symmetric_monoidal_functor (pr2 V) (pr2 V) constant_functor_fmonoidal.
```
nmvdw, line 296: Does `V` work instead of `pr2 V`? If not, can you add the appropriate coercion?

UniMath/CategoryTheory/Monoidal/Comonoids/MonoidalCartesianBuilder.v
```diff
@@ -0,0 +1,359 @@
...
+    - split ; intro ; intros ; apply isaprop_is_comonoid_mor.
+  Defined. *)
+
+Section CartesianBuilder.
+
+  Context (V : monoidal_cat).
+  Let C : category := V.
+  Let M : monoidal C := V.
+
+  Notation "x ⊗ y" := (x ⊗_{M} y).
```
nmvdw, line 50: These lines are mostly redundant, because the notation is already available.

UniMath/CategoryTheory/Monoidal/Comonoids/MonoidalCartesianBuilder.v
```diff
@@ -0,0 +1,359 @@
...
+  Notation "x ⊗l f" := (x ⊗^{M}_{l} f) (at level 31).
+  Notation "f ⊗r y" := (f ⊗^{M}_{r} y) (at level 31).
+  Notation "f ⊗⊗ g" := (f ⊗^{M} g) (at level 31).
+
+  Let I : C := monoidal_unit M.
```
nmvdw, line 55: These also are redundant. Use `mon_lunitor`, `mon_lassociator` and so on instead.

UniMath/CategoryTheory/Monoidal/Comonoids/MonoidalCartesianBuilder.v
```diff
@@ -0,0 +1,359 @@
...
+  Let α : associator_data M := monoidal_associatordata M.
+  Let luinv : leftunitorinv_data M (monoidal_unit M) := monoidal_leftunitorinvdata M.
+  Let ruinv : rightunitorinv_data M (monoidal_unit M) := monoidal_rightunitorinvdata M.
+  Let αinv : associatorinv_data M := monoidal_associatorinvdata M.
+
+  Context
+    (m : ∏ x : C, comonoid M x)
+      (mf : ∏ (x y : C) (f : C⟦x,y⟧), is_comonoid_mor M (m x) (m y) f).
+
+  Notation "∇_{ x }" := (pr11 (m x)).
```
nmvdw, line 67: Projections of monoids should be named.

UniMath/CategoryTheory/Monoidal/Comonoids/MonoidalCartesianBuilder.v
```diff
@@ -0,0 +1,359 @@
...
+      refine (_ @ pr12 (m z)) ; cbn.
+      apply maponpaths_2, maponpaths.
+      apply (when_bifunctor_becomes_rightwhiskering M).
+    Qed.
+
+    Context (p : ∇_{x ⊗ y} · ((x ⊗l !_{y}) ⊗⊗ (!_{x} ⊗r y)) · (ru x ⊗⊗ lu y)
+                 = identity (x ⊗ y)).
+    Lemma make_is_binprod_from_comonoids_uniqueness
+      (f : C ⟦ z, x ⊗ y ⟧)
+      (px : f · (identity x ⊗^{ M} !_{y} · ru x) = fx)
```
nmvdw, line 177: In the top of the file, notation was defined, but here it is not used. 

I think all notation regarding tensors/associators/unitor that is defined in any section of this section, should be deleted.

nmvdw (approved):

UniMath/CategoryTheory/Monoidal/Comonoids/CartesianAsComonoids.v
```diff
@@ -0,0 +1,449 @@
...
+    unfold semi_cart_tensor_pr2.
+    refine (assoc _ _ _ @ _).
+    refine (_ @ mon_linvunitor_lunitor (V := C,,M) x).
+    apply maponpaths_2.
+    refine (_ @ cartesian_linvunitor' x).
+    apply maponpaths.
+    apply (when_bifunctor_becomes_rightwhiskering M).
+  Qed.
+
+  (* Lemma diagonal_commutes_with_assoc (x : C)
```
rmatthes, line 259: I see the lemma is the equation of the new hypothesis in the context, but without the leading diag x on both sides. If the lemma is supposed to hold, one should better name the hypothesis with a suffix indicating that this is only an instance of a provable law.

rmatthes: I now think that the lemma is not true. But I have a proof of the hypothesis - a PR to @Kfwullaert 's fork.

Kfwullaert:
I have reorganized the code, however, small syntactic changes might need to happen, as already noticed by e.g., @nmvdw @benediktahrens . In particular, the comments made before have not yet been fully addressed (even though they are probably now outdated. I'll let you known when your comments have been taken into account.

Kfwullaert:
@benediktahrens @rmatthes @nmvdw 
I think I have addressed every comment made by you.

UniMath/CategoryTheory/Monoidal/Comonoids/CommComonoidsCartesian.v
```diff
@@ -0,0 +1,163 @@
...
+    (m : commutative_comonoid V)
+    : comonoid (symmetric_cat_commutative_comonoids V).
+  Proof.
+    exists m.
+    exists (pr2 (commutative_comonoid_to_comonoid_of_comonoids_data m)).
+    exact (commutative_comonoid_to_comonoid_of_comonoids_laws m).
+  Defined.
+
+  Definition comonoid_mor_is_comonoid_mor {x y : commutative_comonoid V} (f : _⟦x,y⟧)
+    :  comonoid_mor_struct (symmetric_cat_commutative_comonoids V)
```
nmvdw, line 128:
```suggestion
    : comonoid_mor_struct (symmetric_cat_commutative_comonoids V)
```

UniMath/CategoryTheory/Monoidal/Comonoids/MonoidalCartesianBuilder.v
```diff
@@ -0,0 +1,350 @@
...
+    intro ; apply terminal_from_aug_id.
+    assumption.
+  Defined.
+
+  Section make_cartesian.
+
+    Context (pI : εI = identity (monoidal_unit V))
+      {x y z : V} (fx : V⟦z, x⟧) (fy : V⟦z, y⟧).
+
+    Let δx := δ_{(x ,, m x) : comonoid V}.
```
nmvdw, line 83: Can you add types for these?

UniMath/CategoryTheory/Monoidal/Comonoids/MonoidalCartesianBuilder.v
```diff
@@ -0,0 +1,350 @@
...
+    Let εz := ε_{(z ,, m z) : comonoid V}.
+
+    Definition make_isbinprod_from_comonoid_existence_mor
+      : V⟦z, x ⊗_{V} y⟧
+      := δz · fx ⊗^{V} fy.
+
+    Let k := make_isbinprod_from_comonoid_existence_mor.
+
+    Lemma make_is_binprod_from_comonoids_existence_mor_1
+      : δz · fx ⊗^{V} fy · (identity x ⊗^{V} εy · ru^{V}_{x}) = fx.
```
nmvdw, line 97: Use `#⊗`

UniMath/CategoryTheory/Monoidal/Comonoids/MonoidalCartesianBuilder.v
```diff
@@ -0,0 +1,350 @@
...
+      rewrite ! assoc.
+      refine (_ @ id_left _).
+      apply maponpaths_2.
+      refine ( _ @ comonoid_to_law_unit_right V _).
+      apply maponpaths_2, maponpaths.
+      apply (when_bifunctor_becomes_leftwhiskering V).
+    Qed.
+
+    Lemma make_is_binprod_from_comonoids_existence_mor_2
+      : δz · fx ⊗^{V} fy · (εx ⊗^{V} identity y · lu^{V}_{y}) = fy.
```
nmvdw, line 130: Use `#⊗` and you can use `mon_lunitor`

UniMath/CategoryTheory/Monoidal/Comonoids/MonoidalCartesianBuilder.v
```diff
@@ -0,0 +1,350 @@
...
+      apply maponpaths_2, maponpaths.
+      apply (when_bifunctor_becomes_rightwhiskering V).
+    Qed.
+
+    Context (p : identity (x ⊗_{V} y) =
+                   δ_{(x ⊗_{V} y ,, m _) : comonoid V}
+                     · ((identity x ⊗^{V} εy) ⊗^{V} (εx ⊗^{V} identity y) · ru^{V}_{x} ⊗^{V} lu^{V}_{y})).
+
+    Lemma make_is_binprod_from_comonoids_uniqueness
+      (f : V⟦z, x ⊗_{V} y⟧)
```
nmvdw, line 167: Use `⊗` and `#⊗`

UniMath/CategoryTheory/Monoidal/Comonoids/MonoidalCartesianBuilder.v
```diff
@@ -0,0 +1,350 @@
...
+      apply maponpaths.
+      apply p.
+    Qed.
+
+  End make_cartesian.
+
+  Lemma monoidal_is_binproduct_from_comonoid
+    (pI : εI = identity (monoidal_unit V))
+    (pT : ∏ x y : V,
+          δ_{(x ⊗_{V} y ,, m _) : comonoid V} · ((identity x ⊗^{V} ε_{(y ,, m y) : comonoid V}) ⊗^{V} (ε_{(x ,, m x) : comonoid V} ⊗^{V} identity y) · ru^{V}_{x} ⊗^{V} lu^{V}_{y})
```
nmvdw, line 207: You can use `⊗` and `#⊗`

UniMath/CategoryTheory/Monoidal/Comonoids/MonoidalCartesianBuilder.v
```diff
@@ -0,0 +1,350 @@
...
+    - intro f.
+      use subtypePath.
+      { intro ; apply isapropdirprod ; apply homset_property. }
+      exact (make_is_binprod_from_comonoids_uniqueness fx fy (! pT x y) (pr1 f) (pr12 f) (pr22 f)).
+  Qed.
+
+  Definition monoidal_is_cartesian_from_comonoid
+    (pI : εI = identity (monoidal_unit V))
+    (pT : ∏ x y : V,
+          δ_{(x ⊗_{V} y ,, m _) : comonoid V} · ((identity x ⊗^{V} ε_{(y ,, m y) : comonoid V}) ⊗^{V} (ε_{(x ,, m x) : comonoid V} ⊗^{V} identity y) · ru^{V}_{x} ⊗^{V} lu^{V}_{y})
```
nmvdw, line 227: You can use `⊗` and `#⊗`

UniMath/CategoryTheory/Monoidal/Comonoids/Symmetric.v
```diff
@@ -0,0 +1,103 @@
...
+      + exact (pr2 V).
+      + apply is_symmetric_monoidal_identity.
+      + apply constant_functor_is_symmetric.
+        refine (sym_mon_braiding_lunitor _ _ @ _).
+        apply pathsinv0.
+        apply unitors_coincide_on_unit.
+  Defined.
+
+  Definition disp_symmetric_comonoids
+    : disp_symmetric (disp_monoidal_comonoids V) (pr2 V).
```
nmvdw, line 73: You should be able to use `V` instead of `pr2 V`

UniMath/CategoryTheory/Monoidal/Comonoids/Tensor.v
```diff
@@ -0,0 +1,746 @@
...
+
+  Context (M : sym_monoidal_cat).
+
+  Definition tensor_of_comonoids_data
+    (mx my : comonoid M)
+    : comonoid_data M.
+  Proof.
+    exists (mx ⊗ my).
+    split.
+    - refine (δ_{mx} ⊗^{M} δ_{my} · _).
```
nmvdw, line 51: Use `#⊗` (also in the remainder of the file)

UniMath/CategoryTheory/Monoidal/Examples/ConstantFunctor.v
```diff
@@ -0,0 +1,93 @@
...
+    : fmonoidal_lax V V (constant_functor _ _ x).
+  Proof.
+    exists constant_functor_fmonoidal_data.
+    exact constant_functor_fmonoidal_laxlaws.
+  Defined.
+
+  Context (comul_iso : is_z_isomorphism (monoid_data_multiplication _ m))
+    (counit_iso : is_z_isomorphism (monoid_data_unit _ m)).
+  Definition constant_functor_fmonoidal_strong
+    :   fmonoidal_stronglaws
```
nmvdw, line 65:
```suggestion
    : fmonoidal_stronglaws
```

UniMath/CategoryTheory/Monoidal/Examples/ConstantFunctor.v
```diff
@@ -0,0 +1,93 @@
...
+  Definition constant_functor_is_symmetric
+               (S : symmetric V)
+               (m_is_comm : pr1 S x x · monoid_data_multiplication _ m
+                            = monoid_data_multiplication _ m)
+    : is_symmetric_monoidal_functor S S constant_functor_fmonoidal.
+  Proof.
+    intro ; intro.
+    rewrite id_right.
+    apply m_is_comm.
+  Defined.
```
nmvdw, line 91: Use `Qed`

UniMath/CategoryTheory/Monoidal/Examples/DiagonalFunctor.v
```diff
@@ -0,0 +1,219 @@
...
+
+Section DiagFunctor.
+
+  Context (V : monoidal_cat).
+
+  Definition diag_functor_data
+    : functor_data V V.
+  Proof.
+    use make_functor_data.
+    - exact (λ x, x ⊗_{V} x).
```
nmvdw, line 32: `⊗` and `#⊗`

UniMath/CategoryTheory/Monoidal/Examples/DiagonalFunctor.v
```diff
@@ -0,0 +1,219 @@
...
+  Definition diag_preserves_tensor_data
+    : preserves_tensordata V V diag.
+  Proof.
+    exact (λ x y, rearrange_prod V x x y y).
+  Defined.
+
+  Definition diag_preserves_unit
+    : preserves_unit V V diag.
+  Proof.
+    exact (luinv^{V}_{I_{V}}).
```
nmvdw, line 68: `mon_linvunitor`

UniMath/CategoryTheory/Monoidal/Examples/DiagonalFunctor.v
```diff
@@ -0,0 +1,219 @@
...
+  Defined.
+
+  Definition diag_functor_fmonoidal_data
```
nmvdw, line 71: It might be good to check out https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/Monoidal/Functors.v#L1788, because this is easier to use in the bundled case.

UniMath/CategoryTheory/Monoidal/Structure/SymmetricDiagonal.v
```diff
@@ -0,0 +1,720 @@
...
+
+Local Open Scope cat.
+
+Import MonoidalNotations.
+
+Section Rearranging.
+
+  Context {C : category} {M : monoidal C} (S : symmetric M).
+
+  Notation "x ⊗ y" := (x ⊗_{M} y).
```
nmvdw, line 28: Use the bundled version here

UniMath/CategoryTheory/Monoidal/Structure/SymmetricDiagonal.v
```diff
@@ -0,0 +1,720 @@
...
+    rewrite ! assoc'.
+    etrans.
+    2: {
+      apply maponpaths.
+      rewrite assoc.
+      (*
+         The following is: sym_mon_hexagon_lassociator ((C,, M),, S).
+         Difference: sym_mon_hexagon_lassociator is written in whiskered form
+       *)
+      apply (pr1 (pr222 S) _ _ _).
```
nmvdw, line 102: Proper accessors are needed here

UniMath/CategoryTheory/DisplayedCats/CatIsoDisplayed.v
```diff
@@ -0,0 +1,286 @@
...
+  Defined.
+
+End DisplayedCatIso.
+
+Section DisplayedCatIsoUnivalence.
+
+  Context {C : category}
+    (D : disp_cat C) (Duniv : is_univalent_disp D).
+
+  Context
```
nmvdw, line 150: To me, this version is much more natural and sensible than the one in the previous section. In addition, univalent displayed categories are more much common than the non-univalent ones, so I think only the version in this section should be used.

UniMath/CategoryTheory/Monoidal/Structure/SymmetricDiagonal.v
```diff
@@ -121,61 +106,61 @@ Section Rearranging.
     apply sym_monoidal_braiding_hexagon1_variant.
   Qed.
 
-  Lemma rearrange_prod_characterization' (x y z w : C)
+  Lemma rearrange_prod_characterization' (x y z w : V)
```
rmatthes, line 109: I do not see a naming scheme here for the many variations on `rearrange_prod_characterization`. How about having not some primes but two natural numbers, the first for the number of primes on `rearrange_prod` to the left, the second for the number of primes on the right?

UniMath/CategoryTheory/Monoidal/Comonoids/CartesianAsComonoids.v
```diff
@@ -0,0 +1,424 @@
...
+    : semi_cart_tensor_pr2 (pr1 Ccart) I_{ M} x = lu^{M}_{x}.
+  Proof.
+    refine (_ @ id_left _).
+    unfold semi_cart_tensor_pr2.
+    apply maponpaths_2.
+    apply identity_of_lwhisker_with_unit.
+  Qed.
+
+  Lemma cartesian_runitor (x : V)
+    : semi_cart_tensor_pr1 (pr1 Ccart) x I_{M} = ru^{M}_{x}.
```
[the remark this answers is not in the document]
Kfwullaert: I actually don't really have a preference, although, mon_runitor is nicer because the monoidal category need not be specified.

UniMath/CategoryTheory/Monoidal/Structure/SymmetricDiagonal.v
```diff
@@ -18,100 +18,85 @@ Import BifunctorNotations.
...
-  Let ru : rightunitor_data M (monoidal_unit M) := monoidal_rightunitordata M.
-  Let α : associator_data M := monoidal_associatordata M.
-  Let luinv : leftunitorinv_data M (monoidal_unit M) := monoidal_leftunitorinvdata M.
-  Let ruinv : rightunitorinv_data M (monoidal_unit M) := monoidal_rightunitorinvdata M.
-  Let αinv : associatorinv_data M := monoidal_associatorinvdata M.
-  Let σ := pr1 S.
-
-  Definition rearrange_prod (x y z w : C)
-    : C ⟦(x ⊗_{ M} y) ⊗_{ M} (z ⊗_{ M} w), (x ⊗_{ M} z) ⊗_{ M} (y ⊗_{ M} w)⟧.
+  Definition rearrange_prod (x y z w : V)
```
rmatthes, line 29: In https://agda.github.io/agda-categories/, such an operation is called swapInner. Is the name taken from somewhere? They also prove that `rearrange_prod` is  but like its own inverse (modulo rearrangement of the indices):  https://agda.github.io/agda-categories/Categories.Category.Monoidal.Interchange.Symmetric.html

UniMath/CategoryTheory/Monoidal/Comonoids/CartesianAsComonoids.v
```diff
@@ -0,0 +1,424 @@
...
+    exact (cartesian_monoidal_has_enough_comonoids_laws x).
+  Defined.
+
+  Definition cartesian_monoidal_has_enough_comm_comonoids
+    : ∏ x : V, is_commutative V (cartesian_monoidal_has_enough_comonoids x).
+  Proof.
+    exact (λ x, diag_is_symmetric x).
+  Qed.
+
+  Lemma cartesian_monoidal_has_unique_comonoids
```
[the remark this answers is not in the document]
Kfwullaert: @nmvdw I would have to show univalence first. Do you agree that I would do this in another PR?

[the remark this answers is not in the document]
nmvdw: It should follow directly from these:
```
dirprod_disp_cat_is_univalent
is_univalent_dialgebra
disp_full_sub_univalent
is_univalent_sigma_disp 
```

UniMath/CategoryTheory/Monoidal/Structure/SymmetricDiagonal.v
Kfwullaert: swapInner is indeed a more appropriate name, it did not cross my mind. 
The name is just something that I came up with.

rmatthes: I only think it does not conform with the UniMath coding style because of the capitalization. But swap_inner should be fine.

rmatthes:
Beware that there is a compilation error during several commits by now.

UniMath/CategoryTheory/Monoidal/Examples/DiagonalFunctor.v
Kfwullaert: Thanks, I was unaware of this! 
I do have to modify/combine certain proofs if I want to use this (because the laws are formulated no longer in whiskered form).

nmvdw: Sure. It is not needed for this PR

UniMath/CategoryTheory/DisplayedCats/CatIsoDisplayed.v
Kfwullaert: @nmvdw This version is now used.

nmvdw: This version makes it simpler, right?

UniMath/CategoryTheory/Monoidal/Examples/Sigma.v
```diff
@@ -0,0 +1,355 @@
+Require Import UniMath.Foundations.All.
```
rmatthes, line 1: Please a short contents overview as header info.

UniMath/CategoryTheory/Monoidal/Displayed/SymmetricMonoidalBuilder.v
```diff
@@ -0,0 +1,279 @@
+Require Import UniMath.Foundations.All.
```
rmatthes, line 1: Please some header info, e.g., that the builder only rests on asymmetric ingredients (Tl is in the context, Tr is then derived, and further properties of Tr come from extra assumptions on Tl, and analogously with the right unitor that is derived from the left unitor). This is all obvious to you now.

UniMath/CategoryTheory/Monoidal/Structure/SymmetricDiagonal.v
```diff
@@ -0,0 +1,715 @@
...
+  Qed.
+
+  Lemma inner_swap_characterization03 (x y z w : V)
+    : inner_swap x y z w = inner_swap''' x y z w.
+  Proof.
+    etrans. { apply inner_swap_characterization02. }
+    apply inner_swap_characterization23.
+  Qed.
+
+  Lemma precompose_inner_swap
```
rmatthes, line 147:
```suggestion
  Lemma naturality_inner_swap
```
This should just be naturality between 4-ary functors (than need not be given).

UniMath/CategoryTheory/Monoidal/Structure/SymmetricDiagonal.v
```diff
@@ -0,0 +1,715 @@
...
+      apply pathsinv0, tensor_sym_mon_braiding.
+    }
+    cbn.
+    rewrite <- (when_bifunctor_becomes_leftwhiskering V).
+    refine (_ @ tensor_comp_mor _ _ _ _).
+    apply maponpaths_2.
+    apply pathsinv0, id_right.
+  Qed.
+
+  Lemma rearrange_along_unit (x y : V)
```
rmatthes, line 227:
```suggestion
  Lemma inner_swap_along_unit (x y : V)
```

rmatthes: There are still six more identifiers afterwards in the file that start with `rearrange` that should become `inner_swap`.

UniMath/CategoryTheory/Monoidal/Structure/SymmetricDiagonal.v
```diff
@@ -0,0 +1,715 @@
...
+      apply maponpaths, mon_lunitor_triangle.
+    }
+    rewrite assoc.
+    refine (! id_left _ @ _).
+    apply maponpaths_2.
+    apply pathsinv0.
+    apply monoidal_associatorisolaw.
+  Qed.
+
+  Lemma rightwhisker_of_lunitor_with_unit (x : V)
```
rmatthes, line 332:
```suggestion
  Lemma leftwhisker_of_lunitor_with_unit (x : V)
```

UniMath/CategoryTheory/Monoidal/Comonoids/Tensor.v
```diff
@@ -0,0 +1,747 @@
...
+
+Section TensorOfCommutativeComonoids.
+
+  Context (V : sym_monoidal_cat).
+
+  Lemma tensor_of_comm_comonoids
+    (mx my : commutative_comonoid V)
+    : is_commutative V (tensor_of_comonoids V mx my).
+  Proof.
+    unfold disp_cat_of_commutative_comonoids.
```
rmatthes, line 715:
```suggestion
    red.
```

UniMath/CategoryTheory/Monoidal/Examples/Sigma.v
```diff
@@ -0,0 +1,365 @@
+(*
+In this file, we construct a displayed monoidal structure on the sigma construction (of displayed categories), given both displayed categories have a displayed monoidal structure.
+For simplicity, we assume that the most upper displayed category is locally propositional (this assumption is satisfied for the instantiations).
```
rmatthes, line 3:
```suggestion
For simplicity, we assume that the upper-most displayed category is locally propositional (this assumption is satisfied for the instantiations).
```

UniMath/CategoryTheory/Monoidal/Comonoids/Tensor.v
benediktahrens: I think tactics such as red and simpl should be avoided if possible, for performance reasons.

nmvdw: Do you have a reason why `red` and `simpl` decrease performance? In my experience, this tends to be very case dependent.

rmatthes: I use `red` mostly to unfold the outermost definition - without having to name it.

benediktahrens: > I use `red` mostly to unfold the outermost definition - without having to name it.

I believe more "precise" tactics are to be preferred; if anything, I would recommend to replace `red` by `unfold`, not the other way round. Here, I would leave the construction as is.

UniMath/CategoryTheory/Monoidal/Comonoids/Symmetric.v
```diff
@@ -0,0 +1,103 @@
...
+  Proof.
+    use (sigma_disp_cat_monoidal_symmetric).
+    - exact disp_symmetric_comonoids.
+    - apply disp_symmetric_fullsub.
+  Defined.
+
+  Definition symmetric_cat_commutative_comonoids
+    : sym_monoidal_cat.
+  Proof.
+    Check disp_monoidal_commutative_comonoids.
```
rmatthes, line 98:
```suggestion
```

rmatthes:
I still have to make a final walk through two files of this PR. Please wait for my approval that should come this afternoon.

UniMath/CategoryTheory/Monoidal/Comonoids/MonoidalCartesianBuilder.v
```diff
@@ -0,0 +1,355 @@
...
+    Let δz := δ_{(z ,, m z) : comonoid V} : V⟦z, z ⊗ z⟧.
+    Let εx := ε_{(x ,, m x) : comonoid V} : V⟦x, monoidal_unit V⟧.
+    Let εy := ε_{(y ,, m y) : comonoid V} : V⟦y, monoidal_unit V⟧.
+    Let εz := ε_{(z ,, m z) : comonoid V} : V⟦z, monoidal_unit V⟧.
+
+    Definition make_isbinprod_from_comonoid_existence_mor
+      : V⟦z, x ⊗ y⟧
+      := δz · fx #⊗ fy.
+
+    Let k := make_isbinprod_from_comonoid_existence_mor.
```
rmatthes, line 95:
```suggestion
```

rmatthes:
I made another shallow PR against @Kfwullaert 's fork.

UniMath/CategoryTheory/Monoidal/Comonoids/CartesianAsComonoids.v
```diff
@@ -0,0 +1,479 @@
+(*
+In this file, we show:
+   if a symmetric monoidal category is cartesian, the forgetful functor from the category of commutative comonoids is an isomorphism of categories.
+
+The converse of this statement remains to be done.
+At this moment, one property remains to be proven. The condition is axiomatized in the context.
```
rmatthes, line 6:
```suggestion
```

rmatthes (approved):
module my small PR to the fork and my suggestions

UniMath/CategoryTheory/Monoidal/Comonoids/CartesianAsComonoids.v
```diff
@@ -0,0 +1,479 @@
...
+
+  Definition cartesian_monoidal_has_enough_comonoids
+    : ∏ x : V, comonoid M.
+  Proof.
+    intro x.
+    refine (x ,, _ ,, _).
+    exact (cartesian_comonoid_laws x).
+  Defined.
+
+  (* Definition cartesian_monoidal_has_enough_commcomonoids
```
nmvdw, line 285: Is this statement needed?

UniMath/CategoryTheory/Monoidal/Comonoids/CartesianAsComonoids.v
```diff
@@ -0,0 +1,479 @@
...
+        apply maponpaths.
+        unfold monoidal_cat_tensor_mor.
+        rewrite (when_bifunctor_becomes_rightwhiskering M).
+        apply maponpaths.
+        apply (semi_cart_to_unit_eq Ccart).
+    - apply (semi_cart_to_unit_eq Ccart).
+  Qed.
+
+  Lemma cartesian_monoidal_has_unique_comm_comonoids
+    : ∏ x : V,  iscontr (disp_cat_of_commutative_comonoids V x).
```
nmvdw, line 334:
```suggestion
    : ∏ x : V, iscontr (disp_cat_of_commutative_comonoids V x).
```

UniMath/CategoryTheory/Monoidal/Comonoids/Monoidal.v
```diff
@@ -0,0 +1,200 @@
...
+Local Open Scope moncat.
+
+Section MonoidalCategoryOfComonoids.
+
+  Context (V : sym_monoidal_cat).
+
+  Let V_comult : disp_monoidal (dialgebra_disp_cat _ _) V
+      := dialgebra_disp_monoidal (identity_fmonoidal V) (diag_functor_fmonoidal V).
+
+  Local Definition V_counit
```
nmvdw, line 57: Why is this `Local Definition` rather than `Let`? You can also use `Proof.` for `Let`.

UniMath/CategoryTheory/Monoidal/Comonoids/MonoidalCartesianBuilder.v
```diff
@@ -0,0 +1,355 @@
...
+    intro ; apply terminal_from_aug_id.
+    assumption.
+  Defined.
+
+  Section make_cartesian.
+
+    Context (pI : εI = identity (monoidal_unit V))
+      {x y z : V} (fx : V⟦z, x⟧) (fy : V⟦z, y⟧).
+
+    Let δx := δ_{(x ,, m x) : comonoid V} : V⟦x, x ⊗ x⟧.
```
nmvdw, line 84: For `Let`: you can add the typing information on the left of the semi-colon. If you do that, then the assumptions are a bit cleaner in the context.

UniMath/CategoryTheory/Monoidal/Comonoids/MonoidalCartesianBuilder.v
nmvdw: This one did not get fixed.

UniMath/CategoryTheory/Monoidal/Displayed/SymmetricMonoidalBuilder.v
```diff
@@ -0,0 +1,286 @@
...
+
+Local Open Scope cat.
+Import MonoidalNotations.
+
+Section Construct_SymmetricMonoidal_On_LocallyProp_DisplayedCategories.
+
+  Context {C : category} {M : monoidal C} (S : symmetric M) (D : disp_cat C).
+
+  Context (T_ob : ∏ x y : C, D x → D y → D (x ⊗_{ M} y)).
+  Definition T {x y : C} (xx : D x) (yy : D y) : D (x ⊗_{ M} y) := T_ob _ _ xx yy.
```
nmvdw, line 40: The names `T`, `Tl` and so on, are not very informative. I think `Let` is more appropriate here, because I don't think these definitions should stay in the name space.

nmvdw (approved):

UniMath/CategoryTheory/Monoidal/Displayed/SymmetricMonoidalBuilder.v
rmatthes: the index V was not even fixed in my PR to @Kfwullaert 's fork

UniMath/CategoryTheory/Monoidal/Displayed/SymmetricMonoidalBuilder.v
```diff
@@ -30,59 +30,63 @@ Require Import UniMath.CategoryTheory.Monoidal.Structure.Cartesian.
...
-  Context (T_ob : ∏ x y : C, D x → D y → D (x ⊗_{ M} y)).
-  Definition T {x y : C} (xx : D x) (yy : D y) : D (x ⊗_{ M} y) := T_ob _ _ xx yy.
+  Context (disp_tensor_ob : ∏ x y : V, D x → D y → D (x ⊗ y)).
+  Let T {x y : V} (xx : D x) (yy : D y) : D (x ⊗ y) := disp_tensor_ob _ _ xx yy.
 
-  Context (T_mor : ∏ (x y1 y2 : C) (g : C ⟦ y1, y2 ⟧) (xx : D x) (yy1 : D y1) (yy2 : D y2),
-              yy1 -->[ g] yy2 → T xx yy1 -->[ x ⊗^{ M}_{l} g] T xx yy2).
+  Context (disp_lwhisker : ∏ (x y1 y2 : V) (g : V⟦ y1, y2 ⟧) (xx : D x) (yy1 : D y1) (yy2 : D y2),
+              yy1 -->[ g] yy2 → T xx yy1 -->[ x ⊗^{V}_{l} g] T xx yy2).
   Definition Tl
```
rmatthes, line 45: why is this a definition of `Tr` given by `Let`?

UniMath/CategoryTheory/Monoidal/Displayed/SymmetricMonoidalBuilder.v
```diff
@@ -30,59 +30,63 @@ Require Import UniMath.CategoryTheory.Monoidal.Structure.Cartesian.
...
+              Tl xx (id_disp yy) =
+                transportb _ (bifunctor_leftid V x y) (id_disp (T xx yy))).
 
-  Context (B_inv : ∏ (x y : C) (xx : D x) (yy : D y), B x y xx yy ;; B y x yy xx = transportb _ (pr1 (pr122 S x y)) (id_disp _)).
+  Context (B_inv : ∏ (x y : V) (xx : D x) (yy : D y),
+              B xx yy ;; B yy xx = transportb _ (sym_mon_braiding_inv V x y) (id_disp _)).
 
-  Lemma Tr_preserves_id
-    {x y : C} (xx : D x) (yy : D y)
+  Lemma disp_rwhisker_preserves_id
```
rmatthes, line 84: Is there no risk that this unspecific identifier once clashes elsewhere? May rather a local lemma?

UniMath/CategoryTheory/Monoidal/Displayed/SymmetricMonoidalBuilder.v
```diff
@@ -108,23 +112,24 @@ Section Construct_SymmetricMonoidal_On_LocallyProp_DisplayedCategories.
...
+    : ∏ (x y1 y2 y3 : V) (g1 : V⟦ y1, y2 ⟧) (g2 : V⟦ y2, y3 ⟧) (xx : D x)
+        (yy1 : D y1) (yy2 : D y2) (yy3 : D y3)
+        (gg1 : yy1 -->[ g1] yy2) (gg2 : yy2 -->[ g2] yy3),
+      Tl xx (gg1 ;; gg2)
+      = transportb _ (bifunctor_leftcomp V x y1 y2 y3 g1 g2)
           (Tl _ gg1 ;; Tl _ gg2)).
 
-  Lemma Tr_preserves_comp
-    {x1 x2 x3 y : C} {f1 : C ⟦ x1, x2 ⟧} {f2 : C ⟦ x2, x3 ⟧} {xx1 : D x1}
+  Lemma disp_rwhisker_preserves_comp
```
rmatthes, line 123: idem

rmatthes (approved):

UniMath/CategoryTheory/Monoidal/Comonoids/MonoidalCartesianBuilder.v
rmatthes: Still there - I overlooked it in my "review".

rmatthes:
GitHub tells me that "this branch cannot be rebased due to conflicts". I now have the choice between a merge commit and a squash. I would prefer the former. What do @benediktahrens and @nmvdw think about this question?

nmvdw:
> GitHub tells me that "this branch cannot be rebased due to conflicts". I now have the choice between a merge commit and a squash. I would prefer the former. What do @benediktahrens and @nmvdw think about this question?

I have no preference whether a squash or a merge commit is used.

benediktahrens:
In general i prefer squash. When there are commits from different people, merge is also fine.

UniMath/CategoryTheory/Monoidal/Displayed/SymmetricMonoidalBuilder.v
```diff
@@ -30,59 +30,63 @@ Require Import UniMath.CategoryTheory.Monoidal.Structure.Cartesian.
...
 
-  Lemma Tr_preserves_id
-    {x y : C} (xx : D x) (yy : D y)
+  Lemma disp_rwhisker_preserves_id
+    {x y : V} (xx : D x) (yy : D y)
     : Tr yy (id_disp xx) =
-        transportb (mor_disp (T_ob x y xx yy) (T_ob x y xx yy)) (bifunctor_rightid M y x)
-          (id_disp (T_ob x y xx yy)).
+        transportb _ (bifunctor_rightid V y x) (id_disp (T xx yy)).
   Proof.
```
nmvdw, line 88: Would `apply homset_property` solve this directly (without any setting)?

UniMath/CategoryTheory/Monoidal/Displayed/SymmetricMonoidalBuilder.v
nmvdw: One could choose a `Lemma`, `Local Lemma` and `Let`. Differences:
- `Lemma`: enters the name space by default. Can cause clashes
- `Local Lemma`: enter the name space, but as a qualified version with the section name. Does not cause clashes
- `Let`: does not enter the name space. However, even if you make it opaque, it is transparent in the end-result.

For this specific identifier, a `Let` would not make sense (opaqueness). The name is very generic, so it is likely that it can cause clashes. In addition, it is something that is not needed outside of this section (I think). As such, I think a `Local Lemma` would be the most suitable.

UniMath/CategoryTheory/Monoidal/Displayed/SymmetricMonoidalBuilder.v
Kfwullaert: I'm not sure what you mean/refer to with this comment. 
The proof itself can not be solved with only applying the homset property.

rmatthes:
I'll do a merge commit as soon as the current checks have completed successfully.

PR #1758, opened 2023-08-17
Linear-non-linear models

nmvdw:
This PR is still work in progress. Currently, it contains:
- Definition of linear-non-linear model
- The bang modality
- Definition of linear category
- Example of a linear category
- Prewhiskering and postwhiskering of monoidal transformations.
- Lafont categories


UniMath/CategoryTheory/Monoidal/Adjunctions.v
```diff
@@ -1026,3 +1034,205 @@ Section MonoidalAdjunctionFromStrong.
...
+         {S₂ : symmetric M₂}
+         {A : adjunction C₁ C₂}
+         (HA : sym_monoidal_adjunction S₁ S₂ A)
+  : is_sym_monoidal_adjunction S₁ S₂ HA
+  := pr2 HA.
+
+(**
+ 7. Symmetric monoidal adjunctions from strong functors
+ *)
+Coercion sym_monoidal_cat_to_symmetric
```
nmvdw, line 1115: This will be deleted once #1755 is merged.

nmvdw:
> > * Prewhiskering and postwhiskering of monoidal transformations. @Kfwullaert you constructed the bicategory of monoidal categories, and this should contain these operations. Do you have some code that I can fill in there?
> 
> I have identified the code lines and I'll try to do the operation.

Thanks!

UniMath/Semantics/LinearLogic/LinearCategory.v
```diff
@@ -0,0 +1,270 @@
+(****************************************************************************
+
+ Linear categories
+
+ In this file, we define linaer categories.
```
rmatthes, line 5:
```suggestion
 In this file, we define linear categories.
```

UniMath/Semantics/LinearLogic/LinearNonLinear.v
```diff
@@ -0,0 +1,120 @@
...
+           (HA : sym_monoidal_adjunction 𝕄 𝕃 A)
+  : linear_non_linear_model
+  := 𝕃 ,, 𝕄 ,, A ,, HM ,, HA.
+
+Definition make_linear_non_linear_from_strong
+           (𝕃 : sym_mon_closed_cat)
+           (𝕄 : sym_monoidal_cat)
+           (A : adjunction 𝕄 𝕃)
+           (HM : is_cartesian 𝕄)
+           (L := left_adjoint A)
```
rmatthes, line 97:
```suggestion
```

rmatthes:
I have no clear picture as how to present the definitions such as of linear category - if it is better to favor instances of abstract concepts or to write the details so that the correspondence with presentations in books and articles can be checked without running Coq. @nmvdw has a very rich experience in those matters. If the definition is abstract, the accessors should show the exact formulas as return types. If the definition is concrete, there should be builders that take instances of abstract concepts as arguments.

nmvdw:
This PR now contains Lafont categories, and it is ready for review

nmvdw:
> I have no clear picture as how to present the definitions such as of linear category - if it is better to favor instances of abstract concepts or to write the details so that the correspondence with presentations in books and articles can be checked without running Coq. @nmvdw has a very rich experience in those matters. If the definition is abstract, the accessors should show the exact formulas as return types. If the definition is concrete, there should be builders that take instances of abstract concepts as arguments.

I chose an unfolded way, because I think it is more convenient to use. My reasons for that are as follows:
- There is a clear separation between data and properties. As such, it is simple to make the data transparent and the properties opaque.
- One does not need to write a large amount of definitions to define a linear category.
- One is not need to manage the precise typing of the natural transformations.

UniMath/Semantics/LinearLogic/LafontCategory.v
```diff
@@ -0,0 +1,88 @@
+(****************************************************************************
+
+ Lafont Categories
+
+ In this file, we define Lafont categories and we show that every lafont
```
rmatthes, line 5:
```suggestion
 In this file, we define Lafont categories and we show that every Lafont
```

UniMath/Semantics/LinearLogic/LiftingModel.v
```diff
@@ -0,0 +1,108 @@
...
+Local Open Scope cat.
+Local Open Scope moncat.
+
+Definition lifting_linear_category_data
+  : linear_category_data.
+Proof.
+  use make_linear_category_data.
+  - exact pointed_poset_sym_mon_closed_cat.
+  - exact lift_poset_symmetric_monoidal_comonad.
+  - exact lift_poset_comult.
```
rmatthes, line 37: To my mind, it is very implicit that this term `lift_poset_comult` is of the right type. In the file where it is defined, it should be identified as being the comultiplication of a suitable comonoid. And the proofs below of the laws should refer to what is done there already (naturality is already part of the definition of this term).

nmvdw: Do you prefer the updated version?

rmatthes: Definitely. Even if there is the small twist with the composition with identity for unit (because of the dialgebra "nature"). But I still think that the comonoidal perspective should also be developed upstream and more of those results imported when checking the laws.

nmvdw: I added an accessor that gives a commutative comonoid on `! x` for every object `x` and that the comultiplication gives a comonoid morphism.

nmvdw: I also added accessors for the natural transformations. For the coalgebra morphism part: I will leave that the way it is now

UniMath/CategoryTheory/Monoidal/Comonoids/Category.v
```diff
@@ -246,6 +246,70 @@ Section CategoryOfCommutativeComonoids.
...
+  Definition make_commutative_comonoid
+    : commutative_comonoid V.
+  Proof.
+    simple refine (x ,, ((δ ,, ε) ,, _) ,, _) ; cbn.
+    - exact make_commutative_comonoid_laws.
+    - exact comm.
+  Defined.
+End MakeCommutativeComonoid.
+
+Arguments make_commutative_comonoid_data /.
```
rmatthes, line 311: Can you please remind me what `/` does with zero declared arguments? There are very few occurrences of this command in UniMath.

UniMath/Semantics/LinearLogic/LiftingModel.v
rmatthes: > To my mind, it is very implicit that this term `lift_poset_comult` is of the right type. In the file where it is defined, it should be identified as being the comultiplication of a suitable comonoid. And the proofs below of the laws should refer to what is done there already (naturality is already part of the definition of this term).

I think my message was a bit unclear. I did not think about the precise cascade of dependencies between the source files, but I thought it would be good to develop the announcement of having a comonoidal structure in https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/Monoidal/Examples/LiftPoset.v#L346 in the very same file (currently, there are only nat. trafos that are called comultiplication and counit, but it is not clear where they live in). The proofs are in the present PR, which is why I discuss this within the present PR.

UniMath/CategoryTheory/Monoidal/Comonoids/Category.v
nmvdw: Sometimes `cbn` does not automatically unfold definitions. A small example:
```coq
Definition idfun { A : UU } (x : A) : A := x.

Definition test
           {A : UU}
           (x : A)
  : UU.
Proof.
  pose (idfun x) as y.
  cbn in y.
```
After `cbn in y`, the `idfun` in that assumption is not unfolded. However, in the following example, the unfolding is done:
```coq
Definition idfun { A : UU } (x : A) : A := x.

Arguments idfun /.

Definition test
           {A : UU}
           (x : A)
  : UU.
Proof.
  pose (idfun x) as y.
  cbn in y.
```

I added it out of precaution, but I checked again, and I think that in this case, it is not necessary.

UniMath/Semantics/LinearLogic/LiftingModel.v
nmvdw: I added the proof that it forms a comonoid

rmatthes: I have made a PR to @nmvdw 's fork to show what we can import from that file.

UniMath/CategoryTheory/Monoidal/Adjunctions.v
rmatthes: It is still there.

nmvdw: Thanks! I forgot

benediktahrens (approved):

rmatthes (approved):

PR #1759, opened 2023-08-18
avoids duplication of a big proof

rmatthes:
in passing, displays more information: the right adjoint is the same 

also a comment on the choice of name "left closed"


benediktahrens:
Looks great, thanks!
Remark: the lemma `adj_unique_up_to_nat_z_iso` should rather be called `adj_closed_under_iso` or similar. The statement that one would expect under the name `adj_unique_up_to_nat_z_iso` is the other implication: if F1 and F2 are both adjoints to G, then they are isomorphic.

rmatthes:
I changed the names in the core part as well.

nmvdw (approved):

PR #1761, opened 2023-08-22
The relational model of linear logic

nmvdw:
This PR contains a formalization of the relational model of linear logic. It makes use of Lafont categories.


UniMath/Semantics/LinearLogic/RelationalModel.v
```diff
@@ -0,0 +1,732 @@
...
+ 1. In `SET`, we have free commutative monoids. In UniMath, we can construct
+    these as quotients, but they can also be constructed with higher inductive
+    types.
+ 2. `REL` is equivalent to the Kleisli category of the powerset monad on `SET`.
+ 3. We can lift monoids in `SET` to monoids in `REL`.
+ 4. `REL` is self-dual.
+
+ The main idea behind the third point is that there is a distributive law
+ between the power set monad and the free commutative monoid monad. The proof
+ in this file is a combines these 4 facts.
```
benediktahrens, line 25:
```suggestion
 in this file combines these 4 facts.
```

UniMath/Semantics/LinearLogic/RelationalModel.v
```diff
@@ -0,0 +1,732 @@
...
+ over `X` in `REL` gives a monoid in `SET` whose carrier is the power set of `X`.
+
+ Third we show that monoid homomorphisms in `SET` give rise to comonoid
+ homomorphisms in `REL` ([cofree_comonoid_REL_map]). The universal property of
+ the free commutative monoid gives a monoid homomorphism, but we want a comonoid
+ homomorphism.
+
+ Fourth we show the converse: comonoid morphisms in `REL` give rise to monoid
+ homomorphisms in `SET` ([comonoid_mor_REL_to_monoid_mor]). This is so that we
+ can instantiate the uniqueness of the universal property.
```
benediktahrens, line 45: I would like to understand this better, but it is difficult to read the code, due to section variables etc.
Is there a way to write this down in terms of functors between categories of (co)monoids in some categories?

nmvdw: One would combine the following facts:
- `Mon(C)` and `Comon(C^op)` are equivalent (where we use that a symmetric monoidal structure on `C` gives a symmetric monoidal structure on `C^op`)
- If `C` and `C'` are equivalent, then `Comon(C)` and `Comon(C')` are equivalent
- `Rel` and `Rel^op` are equivalent
- We have a functor `Mon(Set) -> Mon(Rel)` that sends a monoid `X` to `P(X)`. If we think of `Mon` as the Eilenberg-Moore category of the free monoid monad `M`, then what we need to do, is given `c : M(X) -> X`, make `M(P(X)) -> P(X)`. To do so, we need `M(P(X)) -> P(M(X))`, and this is where a distributive law would be used.

This gives:
```
Mon(Set) -> Mon(Rel) ≃ Comon(Rel^op) ≃ Comon(Rel)
```

The explanation in Galal's thesis is good (Section 1.1.4 in https://github.com/zgalal/manuscript/raw/main/main.pdf).

benediktahrens: Can we copy this explanation into the comment? I think it complements nicely the exisiting comment.

nmvdw: I added this comment to the documentation

rmatthes:
I might be a bit slow on this review because of the heat wave, but I have begun studying the PR.

nmvdw:
> I might be a bit slow on this review because of the heat wave, but I have begun studying the PR.

No problem!

rmatthes (approved):
Hard to follow the details, but seems plausible with the extensive header info. I still have a minor PR against @nmvdw 's fork.

nmvdw:
> Hard to follow the details, but seems plausible with the extensive header info. I still have a minor PR against @nmvdw 's fork.

I merged it.

In my experience, proofs involving the category of relation tend to be quite technical and complicated. This makes it difficult to read, I think.

nmvdw:
> > > Hard to follow the details, but seems plausible with the extensive header info. I still have a minor PR against @nmvdw 's fork.
> > 
> > 
> > I merged it.
> > In my experience, proofs involving the category of relation tend to be quite technical and complicated. This makes it difficult to read, I think.
> 
> If we were to work in relations for a longer time, we ought to encapsulate the reasoning principles in nicer syntax. The Isabelle system is great for such optimizations. It could be a nice challenge for Ltac2. Dealing with the squashed existence of the middle terms is not even the worst thing here. For me, it is the use of equations between morphisms in REL that has to be done so extensionally and then turned from equalities into implications.

I agree. Another interesting approach would be to define `REL` as the Kleisli category of the powerset monad.

rmatthes:
> > > > Hard to follow the details, but seems plausible with the extensive header info. I still have a minor PR against @nmvdw 's fork.
> > > 
> > > 
> > > I merged it.
> > > In my experience, proofs involving the category of relation tend to be quite technical and complicated. This makes it difficult to read, I think.
> > 
> > 
> > If we were to work in relations for a longer time, we ought to encapsulate the reasoning principles in nicer syntax. The Isabelle system is great for such optimizations. It could be a nice challenge for Ltac2. Dealing with the squashed existence of the middle terms is not even the worst thing here. For me, it is the use of equations between morphisms in REL that has to be done so extensionally and then turned from equalities into implications.
> 
> I agree. Another interesting approach would be to define `REL` as the Kleisli category of the powerset monad.

That would be a solution within category theory, not needing computer science.

UniMath/CategoryTheory/Monoidal/Comonoids/Category.v
```diff
@@ -246,6 +246,56 @@ Section CategoryOfCommutativeComonoids.
...
+
+  Definition make_commutative_comonoid_mor
+    : C₁ --> C₂.
+  Proof.
+    refine (f ,, ((_ ,, _) ,, tt) ,, tt).
+    - exact fδ.
+    - cbn.
+      rewrite id_right.
+      exact fε.
+  Defined.
```
m-lindgren, line 266: Since this is not opaque would it be better to use `apply` instead of `rewrite` here, or does it not matter in this instance?

m-lindgren (approved):

UniMath/CategoryTheory/Monoidal/Comonoids/Category.v
nmvdw: I actually forgot to make it opaque.

PR #1766, opened 2023-08-24
two properties of Scott-continuous maps

co-dan:
+ Scott continuous maps are continuous in the Scott topology
+ Scott continuous maps reflect apartness


nmvdw (approved):

co-dan:
I have updated the .package file.
I am not sure of the benefits of creating another file just for these two lemmas.

nmvdw (approved):

PR #1768, opened 2023-08-25
Linear to linear non-linear

Kfwullaert:
Every linear category is part of a linear non-linear model.
The associated cartesian (monoidal) category is the eilenberg-moore category. 
This construction has been formalized in this PR.
Some laws in the definition of a linear category were missing, these are now added. 
Consequently, a construction in LiftingModel has now become incomplete (@nmvdw ).

### DIFF
--- a/UniMath/Semantics/LinearLogic/LinearCategory.v
+++ b/UniMath/Semantics/LinearLogic/LinearCategory.v
@@ -20,6 +20,7 @@ Require Import UniMath.CategoryTheory.Monoidal.Categories.
 Require Import UniMath.CategoryTheory.Monoidal.Functors.
 Require Import UniMath.CategoryTheory.Monoidal.FunctorCategories.
 Require Import UniMath.CategoryTheory.Monoidal.Structure.Symmetric.
+Require Import UniMath.CategoryTheory.Monoidal.Structure.SymmetricDiagonal.
 Require Import UniMath.CategoryTheory.Monoidal.Structure.Closed.
 Require Import UniMath.CategoryTheory.Monoidal.Comonoids.Category.
 
@@ -152,7 +153,39 @@ Definition linear_category_laws
       linear_category_comult 𝕃 x
       · sym_mon_braiding 𝕃 _ _
       =
-      linear_category_comult 𝕃 x).
+        linear_category_comult 𝕃 x)
+     ×
+     (* counit preserves tensor *)
+     (∏ x y : 𝕃, mon_functor_tensor (linear_category_bang_functor 𝕃) x y
+                           · linear_category_counit 𝕃 (x ⊗ y)
+                         = linear_category_counit 𝕃 x #⊗ linear_category_counit 𝕃 y
+                             · mon_lunitor (monoidal_unit 𝕃))
+     ×
+     (* counit preserves unit *)
+     (mon_functor_unit (linear_category_bang_functor 𝕃)
+                · linear_category_counit 𝕃 I_{𝕃}
+              = identity I_{𝕃})
+     ×
+     (* comult preserves unit *)
+     (mon_functor_unit (linear_category_bang_functor 𝕃)
+                · linear_category_comult 𝕃 I_{𝕃}
+              = mon_linvunitor I_{𝕃}
+                  · mon_functor_unit (linear_category_bang_functor 𝕃)
+                  #⊗ mon_functor_unit (linear_category_bang_functor 𝕃))
+     ×
+     (* comult preserves tensor *)
+     (∏ x y : 𝕃,
+         mon_functor_tensor (linear_category_bang_functor 𝕃) x y
+           · linear_category_comult 𝕃 (x ⊗ y)
+         = (linear_category_comult 𝕃 x) #⊗ (linear_category_comult 𝕃 y)
+             · (inner_swap 𝕃
+                  (linear_category_bang 𝕃 x)
+                  (linear_category_bang 𝕃 x)
+                  (linear_category_bang 𝕃 y)
+                  (linear_category_bang 𝕃 y)
+                  · mon_functor_tensor (linear_category_bang_functor 𝕃) x y
+                  #⊗ mon_functor_tensor (linear_category_bang_functor 𝕃) x y)).
+
 
 Definition linear_category
   : UU
@@ -268,8 +301,53 @@ Section AccessorsLaws.
       =
       linear_category_comult 𝕃 x.
   Proof.
-    exact (pr222 (pr222 (pr222 𝕃)) x).
+    exact (pr1 (pr222 (pr222 (pr222 𝕃))) x).
   Qed.
+
+  Proposition linear_category_counit_preserves_tensor
+    (x y : 𝕃)
+    : mon_functor_tensor (linear_category_bang_functor 𝕃) x y
+        · linear_category_counit 𝕃 (x ⊗ y)
+      = linear_category_counit 𝕃 x #⊗ linear_category_counit 𝕃 y
+          · mon_lunitor (monoidal_unit 𝕃).
+  Proof.
+    exact (pr12 (pr222 (pr222 (pr222 𝕃))) x y).
+  Qed.
+
+  Proposition linear_category_counit_preserves_unit
+    : mon_functor_unit (linear_category_bang_functor 𝕃)
+        · linear_category_counit 𝕃 I_{𝕃}
+      = identity I_{𝕃}.
+  Proof.
+    exact (pr122 (pr222 (pr222 (pr222 𝕃)))).
+  Qed.
+
+  Proposition linear_category_comult_preserves_unit
+    : mon_functor_unit (linear_category_bang_functor 𝕃)
+        · linear_category_comult 𝕃 I_{𝕃}
+      = mon_linvunitor I_{𝕃}
+          · mon_functor_unit (linear_category_bang_functor 𝕃)
+          #⊗ mon_functor_unit (linear_category_bang_functor 𝕃).
+  Proof.
+    exact (pr1 (pr222 (pr222 (pr222 (pr222 𝕃))))).
+  Qed.
+
+  Proposition linear_category_comult_preserves_tensor
+    (x y : 𝕃)
+    : mon_functor_tensor (linear_category_bang_functor 𝕃) x y
+           · linear_category_comult 𝕃 (x ⊗ y)
+         = (linear_category_comult 𝕃 x) #⊗ (linear_category_comult 𝕃 y)
+             · (inner_swap 𝕃
+                  (linear_category_bang 𝕃 x)
+                  (linear_category_bang 𝕃 x)
+                  (linear_category_bang 𝕃 y)
+                  (linear_category_bang 𝕃 y)
+                  · mon_functor_tensor (linear_category_bang_functor 𝕃) x y
+                  #⊗ mon_functor_tensor (linear_category_bang_functor 𝕃) x y).
+  Proof.
+    exact (pr2 (pr222 (pr222 (pr222 (pr222 𝕃)))) x y).
+  Qed.
+
 End AccessorsLaws.
 
 (**
--- a/UniMath/Semantics/LinearLogic/LinearCategoryEilenbergMooreAdjunction.v
+++ b/UniMath/Semantics/LinearLogic/LinearCategoryEilenbergMooreAdjunction.v
@@ -1,6 +1,5 @@
 (*
-In this file, we show how any linear category induces a linear non linear model.
-This boils down to proving how the eilenberg moore category of the (symmetric monoidal) comonad, given by a linear category, is cartesian monoidal.
+In this file, the cofree adjunction between a linear category and its eilenberg-moore category is constructed.
 *)
 
 Require Import UniMath.MoreFoundations.All.
--- a/UniMath/Semantics/LinearLogic/LinearCategoryEilenbergMooreAdjunction.v
+++ b/UniMath/Semantics/LinearLogic/LinearCategoryEilenbergMooreAdjunction.v
@@ -42,7 +42,7 @@ Section CofreeAdjunction.
   Context (L : linear_category).
 
   Definition eilenberg_moore_cofree
-    :  pr1 L ⟶ co_eilenberg_moore_cat (linear_category_bang L).
+    : L ⟶ co_eilenberg_moore_cat (linear_category_bang L).
   Proof.
     use functor_to_co_eilenberg_moore_cat.
     - apply (linear_category_bang L).
@@ -61,20 +61,8 @@ Section CofreeAdjunction.
           exact (Comonad_law3 (T := linear_category_bang L) x)).
   Defined.
 
-  (* Lemma test_ob (x : L)
-    : pr11 (eilenberg_moore_cofree x) = linear_category_bang L x.
-  Proof.
-    apply idpath.
-  Qed.
-
-  Lemma test_mor (x : L)
-    : pr21 (eilenberg_moore_cofree x) =  δ (linear_category_bang L) x.
-  Proof.
-    apply id_left.
-  Qed. *)
-
   Local Definition eilenberg_moore_forget
-    : full_subcat (dialgebra (functor_identity (pr1 L)) (linear_category_bang L))
+    : full_subcat (dialgebra (functor_identity L) (linear_category_bang L))
         mon_cat_co_eilenberg_moore_extra_condition ⟶ L.
     (* : co_eilenberg_moore_cat (linear_category_bang L) ⟶ L. *)
   Proof.
@@ -83,7 +71,7 @@ Section CofreeAdjunction.
 
   Local Definition eilenberg_moore_adj_unit
     : functor_identity
-        (full_subcat (dialgebra (functor_identity (pr1 L)) (linear_category_bang L))
+        (full_subcat (dialgebra (functor_identity L) (linear_category_bang L))
            mon_cat_co_eilenberg_moore_extra_condition) ⟹
         eilenberg_moore_forget ∙ eilenberg_moore_cofree.
   Proof.
@@ -121,7 +109,7 @@ Section CofreeAdjunction.
 
   Definition eilenberg_moore_cmd_adj
     : adjunction
-    (full_subcat (dialgebra (functor_identity (pr1 L)) (linear_category_bang L))
+    (full_subcat (dialgebra (functor_identity L) (linear_category_bang L))
        mon_cat_co_eilenberg_moore_extra_condition) L.
   Proof.
     use make_adjunction.
--- a/UniMath/Semantics/LinearLogic/LinearCategoryEilenbergMooreAdjunction.v
+++ b/UniMath/Semantics/LinearLogic/LinearCategoryEilenbergMooreAdjunction.v
@@ -1,0 +1,136 @@
+(*
+In this file, we show how any linear category induces a linear non linear model.
+This boils down to proving how the eilenberg moore category of the (symmetric monoidal) comonad, given by a linear category, is cartesian monoidal.
+*)
+
+Require Import UniMath.MoreFoundations.All.
+Require Import UniMath.CategoryTheory.Core.Prelude.
+Require Import UniMath.CategoryTheory.Adjunctions.Core.
+Require Import UniMath.CategoryTheory.whiskering.
+Require Import UniMath.CategoryTheory.Monads.Comonads.
+Require Import UniMath.CategoryTheory.Monoidal.WhiskeredBifunctors.
+Require Import UniMath.CategoryTheory.Monoidal.Categories.
+Require Import UniMath.CategoryTheory.Monoidal.Functors.
+Require Import UniMath.CategoryTheory.Monoidal.Adjunctions.
+Require Import UniMath.CategoryTheory.Monoidal.FunctorCategories.
+Require Import UniMath.CategoryTheory.Monoidal.Structure.Cartesian.
+Require Import UniMath.CategoryTheory.Monoidal.Structure.Symmetric.
+Require Import UniMath.CategoryTheory.Monoidal.Structure.Closed.
+Require Import UniMath.CategoryTheory.Monoidal.Structure.SymmetricDiagonal.
+
+Require Import UniMath.CategoryTheory.DisplayedCats.Core.
+Require Import UniMath.CategoryTheory.DisplayedCats.Total.
+Require Import UniMath.CategoryTheory.DisplayedCats.Constructions.
+Require Import UniMath.CategoryTheory.categories.Dialgebras.
+
+Require Import UniMath.CategoryTheory.Monoidal.Displayed.WhiskeredDisplayedBifunctors.
+Require Import UniMath.CategoryTheory.Monoidal.Displayed.Monoidal.
+
+Require Import UniMath.CategoryTheory.categories.CoEilenbergMoore.
+Require Import UniMath.CategoryTheory.Monoidal.Examples.MonoidalDialgebras.
+Require Import UniMath.CategoryTheory.Monoidal.Examples.SymmetricMonoidalCoEilenbergMoore.
+
+Require Import UniMath.Semantics.LinearLogic.LinearCategory.
+
+Import MonoidalNotations.
+
+Local Open Scope cat.
+Local Open Scope moncat.
+
+Section CofreeAdjunction.
+
+  Context (L : linear_category).
+
+  Definition eilenberg_moore_cofree
+    :  pr1 L ⟶ co_eilenberg_moore_cat (linear_category_bang L).
+  Proof.
+    use functor_to_co_eilenberg_moore_cat.
+    - apply (linear_category_bang L).
+    - use nat_trans_comp.
+      2: { apply nat_trans_functor_id_left. }
+      exact (δ (linear_category_bang L)).
+    - abstract (
+          intro x;
+          refine (_ @ Comonad_law1 (T := linear_category_bang L) x);
+          refine (assoc' _ _ _ @ _);
+          apply id_left).
+    - abstract (
+          intro x;
+          cbn;
+          rewrite id_left;
+          exact (Comonad_law3 (T := linear_category_bang L) x)).
+  Defined.
+
+  (* Lemma test_ob (x : L)
+    : pr11 (eilenberg_moore_cofree x) = linear_category_bang L x.
+  Proof.
+    apply idpath.
+  Qed.
+
+  Lemma test_mor (x : L)
+    : pr21 (eilenberg_moore_cofree x) =  δ (linear_category_bang L) x.
+  Proof.
+    apply id_left.
+  Qed. *)
+
+  Local Definition eilenberg_moore_forget
+    : full_subcat (dialgebra (functor_identity (pr1 L)) (linear_category_bang L))
+        mon_cat_co_eilenberg_moore_extra_condition ⟶ L.
+    (* : co_eilenberg_moore_cat (linear_category_bang L) ⟶ L. *)
+  Proof.
+    exact (functor_composite (pr1_category _) (pr1_category _)).
+  Defined.
+
+  Local Definition eilenberg_moore_adj_unit
+    : functor_identity
+        (full_subcat (dialgebra (functor_identity (pr1 L)) (linear_category_bang L))
+           mon_cat_co_eilenberg_moore_extra_condition) ⟹
+        eilenberg_moore_forget ∙ eilenberg_moore_cofree.
+  Proof.
+    use make_nat_trans.
+    - intro x.
+      use make_mor_co_eilenberg_moore.
+      + exact (pr21 x).
+      + abstract (
+            refine (! pr22 x @ _);
+            apply maponpaths,
+              pathsinv0,
+              id_left).
+    - abstract (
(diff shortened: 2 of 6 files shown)


m-lindgren:
Will the `Admitted` material be added to this PR?

UniMath/Semantics/LinearLogic/LinearCategory.v
```diff
@@ -152,7 +153,39 @@ Definition linear_category_laws
       linear_category_comult 𝕃 x
       · sym_mon_braiding 𝕃 _ _
       =
-      linear_category_comult 𝕃 x).
+        linear_category_comult 𝕃 x)
+     ×
+     (* counit preserves tensor *)
```
rmatthes, line 158: I am sorry for this shallow observation, but most of the time, the comultiplication is dealt with before the counit. Also the tensor comes before the unit in the def. of monoidal cats. (And, generally, the unitor rules came before the associator rules as a warm-up in the proofs.)
So, here, the order should be:
- comult preserves tensor
- comult preserves unit
- counit preserves tensor
- counit preserves unit

However, the laws 5 and 6 of a linear category do not follow this pattern.

nmvdw: Agreed (with the request to fix this after merging https://github.com/Kfwullaert/UniMath/pull/37).

UniMath/Semantics/LinearLogic/LinearCategoryEilenbergMooreAdjunction.v
```diff
@@ -0,0 +1,123 @@
...
+          apply id_left).
+    - abstract (
+          intro x;
+          cbn;
+          rewrite id_left;
+          exact (Comonad_law3 (T := linear_category_bang L) x)).
+  Defined.
+
+  Local Definition eilenberg_moore_forget
+    : full_subcat (dialgebra (functor_identity L) (linear_category_bang L))
```
nmvdw, line 64: Can you use an identifier here? Why do you go via the full subcat here?

UniMath/Semantics/LinearLogic/LinearToLinearNonLinear.v
```diff
@@ -0,0 +1,811 @@
...
+Require Import UniMath.Semantics.LinearLogic.LinearNonLinear.
+
+Import MonoidalNotations.
+
+Local Open Scope cat.
+Local Open Scope moncat.
+
+Import ComonoidNotations.
+
+Section ConstructionOfComonoidsInLinearCategory.
```
nmvdw, line 48: This statement is not specific to linear categories, right? It might be good to change `L : linear_category` to `V : sym_monoidal_cat` (and possibly move it to another file).

UniMath/Semantics/LinearLogic/LinearToLinearNonLinear.v
```diff
@@ -0,0 +1,811 @@
...
+    - apply id_right.
+  Qed.
+
+End ConstructionOfComonoidsInLinearCategory.
+
+Section LiftingPropertyCoalgebraMorSection.
+
+  Lemma postcomp_with_section_reflect_coalg_mor
+    (L : linear_category)
+    (xx aa bb : @sym_monoidal_cat_co_eilenberg_moore _ _ _ (linear_category_bang L))
```
nmvdw, line 277: Why do you need the `@` here?

UniMath/Semantics/LinearLogic/LinearToLinearNonLinear.v
```diff
@@ -0,0 +1,811 @@
...
+    (i : _⟦aa,bb⟧)
```
nmvdw, line 278:
```suggestion
    (i : aa --> bb)
```

nmvdw: There is no reason to use the `⟦` notation if you don't specify the category.

UniMath/Semantics/LinearLogic/LinearToLinearNonLinear.v
```diff
@@ -0,0 +1,811 @@
...
+      apply maponpaths.
+      refine (_ @ functor_id (linear_category_bang L) _).
+      apply maponpaths.
+      exact ir_id.
+  Qed.
+
+  Definition lifting_is_coalg_mor
+    {L : linear_category}
+    {xx aa bb : (@sym_monoidal_cat_co_eilenberg_moore _ _ _ (linear_category_bang L))}
+    {g : _⟦xx,bb⟧} {i : _⟦aa,bb⟧} {f : L⟦pr11 xx, pr11 aa⟧}
```
nmvdw, line 314:
```suggestion
    {g : xx --> bb} {i : aa --> bb} {f : L⟦pr11 xx, pr11 aa⟧}
```

UniMath/Semantics/LinearLogic/LinearToLinearNonLinear.v
```diff
@@ -0,0 +1,811 @@
...
+    }
+    rewrite assoc'.
+    rewrite <- tensor_comp_mor.
+    refine (_ @ id_right _).
+    apply maponpaths.
+    rewrite <- tensor_id_id.
+    use two_arg_paths ; apply Comonad_law1.
+  Qed.
+
+  Local Lemma yank (* any suggestions for the name are welcome *)
```
nmvdw, line 366: What about `transport_comonoid_from_free_lem`?

UniMath/Semantics/LinearLogic/LinearToLinearNonLinear.v
```diff
@@ -0,0 +1,811 @@
...
+
+  Let bang : sym_monoidal_cmd L
+      := linear_category_bang L.
+
+  Context (xx : EM).
+  Let x : L := pr11 xx.
+  Let hx : L⟦x, bang x⟧ := pr21 xx.
+
+  Lemma comonoid_in_eilenberg_moore_from_coalg_counit_alg_mor
+    :  hx · #bang ε_{(x,, transport_comonoid_from_free L xx) : comonoid L}
```
nmvdw, line 503: It might be good to make an identifier for `transport_comonoid_from_free` where the resulting type is a comonoid

UniMath/Semantics/LinearLogic/LinearToLinearNonLinear.v
```diff
@@ -0,0 +1,811 @@
...
+End ConstructionOfComonoidsInEilenbergMoore.
+
+Section EilenbergMooreCartesian.
+
+  Context (L : linear_category).
+
+  (* naturality of the comultiplication and counit *)
+  Lemma comonoid_mor_in_eilenberg_moore
+    {x y : (@sym_monoidal_cat_co_eilenberg_moore _ _ _ (linear_category_bang L))}
+    (f : _⟦x, y⟧)
```
nmvdw, line 618:
```suggestion
    (f : x --> y)
```

nmvdw (approved):

UniMath/Semantics/LinearLogic/LinearToLinearNonLinear.v
Kfwullaert: sym_monoidal_cat_co_eilenberg_moore only has implicit arguments

nmvdw: It might be better to make one of its arguments explicit.

UniMath/Semantics/LinearLogic/LinearToLinearNonLinear.v
Kfwullaert: You are absolutely right, thanks for noticing this!

UniMath/Semantics/LinearLogic/LinearCategoryEilenbergMooreAdjunction.v
Kfwullaert: If I would have used "co_eilenberg_moore_cat (linear_category_bang L) ⟶ L. ",
Coq would not recognize this as a total category, i.e., I could not just use pr1_category.

rmatthes:
Commit https://github.com/UniMath/UniMath/pull/1768/commits/f7a562a984c43c2897bd46039429d4ea6f56b91f seems to run indefinitely.

Kfwullaert:
> Commit [f7a562a](https://github.com/UniMath/UniMath/commit/f7a562a984c43c2897bd46039429d4ea6f56b91f) seems to run indefinitely.

My machine hangs when trying to compile UniMath/Semantics/LinearLogic/LiftingModel.v
The admits that I first added, and solved by @nmvdw , a cbn is used, for me, it hangs there.

nmvdw:
The cause was that the order of the laws was changed, but not my code. A fix is in https://github.com/Kfwullaert/UniMath/pull/38

UniMath/Semantics/LinearLogic/LinearCategory.v
rmatthes: The order of the laws 5 and 6 should also be inverted (with propagation).

UniMath/Semantics/LinearLogic/LinearCategoryEilenbergMooreAdjunction.v
```diff
@@ -0,0 +1,123 @@
+(*
+In this file, the cofree adjunction between a linear category and its eilenberg-moore category is constructed.
```
rmatthes, line 2:
```suggestion
In this file, the cofree adjunction between a linear category and its Eilenberg-Moore category is constructed.
```

UniMath/Semantics/LinearLogic/LinearToLinearNonLinear.v
nmvdw: Now the arguments are explicit, you can remove the `@` (multiple occurrences in this file)

UniMath/Semantics/LinearLogic/LinearToLinearNonLinear.v
```diff
@@ -0,0 +1,591 @@
...
+      · linear_category_comult L (pr11 x ⊗ pr11 y)
+      · ε (linear_category_bang L) (pr11 x ⊗ pr11 y) #⊗ ε (linear_category_bang L) (pr11 x ⊗ pr11 y)
+    = rightwhiskering_on_morphisms (pr1 L) (pr11 y) _ _
+        (pr21 x · linear_category_comult L (pr11 x) · ε (linear_category_bang L) (pr11 x) #⊗ ε (linear_category_bang L) (pr11 x))
+        · leftwhiskering_on_morphisms (pr1 L) (pr11 x ⊗ pr11 x) _ _
+        (pr21 y · linear_category_comult L (pr11 y)
+           · ε (linear_category_bang L) (pr11 y) #⊗ ε (linear_category_bang L) (pr11 y))
+  · pr11 (inner_swap (sym_monoidal_cat_co_eilenberg_moore (linear_category_bang L)) x x y y).
+  Proof.
+    Opaque SymmetricDiagonal.inner_swap. (* I do not include the file SymmetricDiagonal because this is the only place I use it *)
```
nmvdw, line 477: I don't think it is a problem to include a file even if you use it in only 1 place.

nmvdw (approved):

rmatthes:
Now that the development has been done, it would be nice to use the blackboard L as identifier for a linear category, as is done in the later part of `LinearCategory.v`. Emacs tells me to type \BbbL to enter that letter.

I think it would be worth it since the letter L has such a strong association with left adjoints.

UniMath/Semantics/LinearLogic/LinearToLinearNonLinear.v
```diff
@@ -0,0 +1,591 @@
+(*
+In this file, we show how any linear category induces a linear non linear model.
```
rmatthes, line 2:
```suggestion
In this file, we show how any linear category induces a linear/non-linear model.
```

UniMath/Semantics/LinearLogic/LinearToLinearNonLinear.v
```diff
@@ -0,0 +1,591 @@
...
+This boils down to proving how the eilenberg moore category of the (symmetric monoidal) comonad, given by a linear category, is cartesian monoidal.
```
rmatthes, line 3:
```suggestion
This boils down to proving that the Eilenberg-Moore category of the (symmetric monoidal) comonad, given by a linear category, is cartesian monoidal.
```
Or is it really a question of choosing the binary products?

Kfwullaert: @rmatthes No, you do not choose it because the (tensor) product is shown to be cartesian. Thus, your suggestion is indeed a valid observation.

UniMath/Semantics/LinearLogic/LinearToLinearNonLinear.v
rmatthes: @nmvdw What is your take on the name? You often wrote "linear-non-linear". Benton had an "and" between "linear and "non-linear" also in the title of his paper you mention there.

Kfwullaert:
> Now that the development has been done, it would be nice to use the blackboard L as identifier for a linear category, as is done in the later part of `LinearCategory.v`. Emacs tells me to type \BbbL to enter that letter.
> 
> I think it would be worth it since the letter L has such a strong association with left adjoints.

I'll make that change.

UniMath/CategoryTheory/Monoidal/Comonoids/TransportComonoidAlongRetraction.v
```diff
@@ -0,0 +1,239 @@
+Require Import UniMath.Foundations.All.
+Require Import UniMath.MoreFoundations.All.
+
+Require Import UniMath.CategoryTheory.Core.Categories.
+Require Import UniMath.CategoryTheory.Monoidal.Categories.
+
+Require Import UniMath.CategoryTheory.DisplayedCats.Core.
+
+Require Import UniMath.CategoryTheory.Monoidal.Comonoids.Category.
```
benediktahrens, line 9: - It would be great if this file had a header describing its contents.
- Is there a more systematic way to "order" the import statements?

UniMath/Semantics/LinearLogic/LinearCategory.v
```diff
@@ -152,7 +153,39 @@ Definition linear_category_laws
...
+     (∏ x y : 𝕃, mon_functor_tensor (linear_category_bang_functor 𝕃) x y
+                           · linear_category_counit 𝕃 (x ⊗ y)
+                         = linear_category_counit 𝕃 x #⊗ linear_category_counit 𝕃 y
+                             · mon_lunitor (monoidal_unit 𝕃))
+     ×
+     (* counit preserves unit *)
+     (mon_functor_unit (linear_category_bang_functor 𝕃)
+                · linear_category_counit 𝕃 I_{𝕃}
+              = identity I_{𝕃}).
+
```
benediktahrens, line 188: Is there a standard notation for the operations, such that using this notation one could express the laws precisely, at least in a comment?

UniMath/Semantics/LinearLogic/LinearCategoryEilenbergMooreAdjunction.v
```diff
@@ -0,0 +1,123 @@
+(*
+In this file, the cofree adjunction between a linear category and its Eilenberg-Moore category is constructed.
```
benediktahrens, line 2: Can you give a reference for this construction?

UniMath/Semantics/LinearLogic/LinearCategoryEilenbergMooreAdjunction.v
```diff
@@ -0,0 +1,123 @@
...
+Require Import UniMath.CategoryTheory.categories.Dialgebras.
+
+Require Import UniMath.CategoryTheory.Monoidal.Displayed.WhiskeredDisplayedBifunctors.
+Require Import UniMath.CategoryTheory.Monoidal.Displayed.Monoidal.
+
+Require Import UniMath.CategoryTheory.categories.CoEilenbergMoore.
+Require Import UniMath.CategoryTheory.Monoidal.Examples.MonoidalDialgebras.
+Require Import UniMath.CategoryTheory.Monoidal.Examples.SymmetricMonoidalCoEilenbergMoore.
+
+Require Import UniMath.Semantics.LinearLogic.LinearCategory.
```
benediktahrens, line 32: Would it be useful to order the import statements somehow?

Kfwullaert: I must admit that I consider this ordered.
The first list consists of the category theory files (including monoidal categories).
The second list consists of the general displayed categories constructions.
The third list consists of the displayed monoidal stuff.
The fourth is about the Eilenberge-Moore.

UniMath/Semantics/LinearLogic/LinearToLinearNonLinear.v
nmvdw: Paul-André Mellies wrote linear-non-linear adjunction in https://www.irif.fr/~mellies/mpri/mpri-ens/biblio/categorical-semantics-of-linear-logic.pdf, and that is the one I followed. The nLab follows the same spelling: https://ncatlab.org/nlab/show/linear-non-linear+logic. In the paper by Benton (https://link.springer.com/chapter/10.1007/BFb0022251), linear/non-linear is used.

I don't have a strong preference, but consistency is good.

nmvdw:
> Now that the development has been done, it would be nice to use the blackboard L as identifier for a linear category, as is done in the later part of `LinearCategory.v`. Emacs tells me to type \BbbL to enter that letter.
> 
> I think it would be worth it since the letter L has such a strong association with left adjoints.

I use `\bL` for blackboard L

UniMath/Semantics/LinearLogic/LinearCategory.v
nmvdw: Do you mean notation for `linear_category_counit`?

benediktahrens: > Do you mean notation for `linear_category_counit`?

I mean notations for the operations mentioned in the comments, like "counit preserves tensor", "comult preserves unit", etc. It would be great to instead write an equation in mathematical notation. It would make it easier (for me) to understand the code.

UniMath/CategoryTheory/Monoidal/Comonoids/TransportComonoidAlongRetraction.v
```diff
@@ -0,0 +1,242 @@
+(*
+Consider a morphism i : B --> a in a monoidal category.
+Given a comonoid structure on B, we show how the comonoid structure can be transported to a, provided i is part of a retraction pair (and a compatibility condition).
+Furthermore, i becomes a comonoid homomorphism.
```
benediktahrens, line 4: Is there a reason why one letter ("B") is capital, one is small ("a")? The question might sound silly, but it really makes me wonder if there is a significance in this choice.

UniMath/Semantics/LinearLogic/LinearCategory.v
rmatthes: I would like to recall my concerns in this matter: https://github.com/UniMath/UniMath/issues/1758#issuecomment-1684942340_

benediktahrens: I don't understand that comment. Can you give an example?

rmatthes: `linear_category_comult_coalgebra_mor` is not used in the file `LinearCategory.v`, but it could be used to construct a coalgebra morphism, which would make it clear that the law indeed corresponds to its name. This is then not a formula of ordinary mathematics but a high-level representation.

benediktahrens (approved):

UniMath/Semantics/LinearLogic/LinearToLinearNonLinear.v
```diff
@@ -0,0 +1,591 @@
...
+
+  Context (𝕃 : linear_category).
+  Let bang : sym_monoidal_cmd 𝕃
+      := linear_category_bang 𝕃.
+
+  Context (xx : sym_monoidal_cat_co_eilenberg_moore (linear_category_bang 𝕃)).
+  Let x : 𝕃 := pr11 xx.
+  Let hx : 𝕃⟦x, bang x⟧ := pr21 xx.
+
+  Let comonoid_on_bang_x := linear_category_cocommutative_comonoid 𝕃 x.
```
rmatthes, line 121: Why is this called cocommutative? The problem comes from upstream.

nmvdw: The laws of a comonoid can be called "co-unitality", "co-associativity" and co-commutativity (see i.e. https://ncatlab.org/nlab/show/cocommutative+coalgebra). I used that convention in the code I wrote, but I think Kobe used co-unitality, co-associativity, and commutativity.

rmatthes: I see. Thank you. Not optimal to have both conventions.

UniMath/Semantics/LinearLogic/LinearToLinearNonLinear.v
```diff
@@ -0,0 +1,591 @@
...
+  Lemma linear_category_comult_factors_through_comult_bang
+    :  δ (linear_category_bang 𝕃) x
+         · (linear_category_comult 𝕃 (bang x)
+              · ε bang (bang x) #⊗ ε bang (bang x))
+       = linear_category_comult 𝕃 x.
+  Proof.
+    rewrite assoc.
+    etrans. {
+      apply maponpaths_2.
+      apply linear_category_counit_comonoid_mor_comult.
```
rmatthes, line 132: This seems to be the only use of this axiom of a linear category. I am puzzled that the companion axiom `linear_category_counit_comonoid_mor_counit` is never referenced. Is this maybe hidden through some projections?

rmatthes: I have an answer to this that might interest @nmvdw : if I replace the axiom `linear_category_counit_comonoid_mor_counit` by a trivial statement (`0=0`), I can easily restore compilation. So, this axiom is not needed yet. Would it anyway follow from the others?

Kfwullaert: @rmatthes You are indeed right that I've only used two (out of 4) axioms that I've added in the definition of a linear category (if I recall correctly), and are used not much.
However, I don't see why they should follow from the other axioms. 
Anyway, I don't consider this is as something _strange_. The added axioms tell you that you have a monoidal structure on the natural transformations (each consisting of two axioms), so why not ask for both?

nmvdw: Interesting. It might then indeed be the case that these axioms are not needed. It is worthwhile to check the literature.

It might be the case that somehow they follow from the other axioms, because every LNL model gives rise to a linear category.

UniMath/CategoryTheory/Monoidal/Comonoids/TransportComonoidAlongRetraction.v
Kfwullaert: I've used the same notation as Paul-André Melliès used in his paper "Categorical semantics of linear logic".
B is a comonoid, while a is only an object.

UniMath/Semantics/LinearLogic/LinearToLinearNonLinear.v
rmatthes: So, if I understand right, two of @Kfwullaert 's axioms and one of @nmvdw ' axioms are not used *yet*. This might warrant having those in an extra bunch that becomes optional (even before knowing if they are derivable).

UniMath/Semantics/LinearLogic/LinearCategory.v
rmatthes: I have made this thought concrete in https://github.com/Kfwullaert/UniMath/pull/40 . Thus, none of the original 9 axioms put by @nmvdw stays unexplained mathematically (they enter into identifiable structures in the same file). For the 4 axioms added by @Kfwullaert, I do not feel competent yet.

UniMath/Semantics/LinearLogic/LinearToLinearNonLinear.v
Kfwullaert: @nmvdw The latter is indeed something to check. I do want to point out that this is not the first time such a phenomena occurs, indeed: If you want to prove the characterization of a (not necessarily symmetric) monoidal category being cartesian, you ask for every object to have a comonoid structure. However, one does not actually need the (co)associativity law of the comonoid.

UniMath/Semantics/LinearLogic/LinearCategory.v
Kfwullaert: @rmatthes Niels has actually provided some of the mathematical structures at the end of the file; such as the (co)commutative comonoid, the natural transformations.

nmvdw: I think Ralph also wants to have accessors for the **monoidal** natural transformations

Kfwullaert: I consider this something to be done in another PR, because the diagonal functor has to be shown to be monoidal (here is where all the diagrams regarding the inner_swap come into play).

UniMath/Semantics/LinearLogic/LinearToLinearNonLinear.v
rmatthes: I think it would be valuable to record that. But I see no easy way to make this in Coq code.

UniMath/Semantics/LinearLogic/LinearCategory.v
rmatthes: Okay, Niels provided structures using 7 of the 9 laws - some of them not used afterwards, but that is fine. I spelt out the coalgebra homomorphisms to illustrate the last two laws. It is already a compromise since it builds up the coalgebras by hand instead of referring to more structure here. For 3 of the 4 new rules, it does not look so much different to me. Could they be reassembled into structure? And for `linear_category_comult_preserves_tensor`, could this not be done in the file `LinearToLinearNonLinear.v`? I mean as a motivation for the mysterious lemma `aux` (I do not ask for presenting that lemma in a structure-revealing format).

rmatthes: For the two last laws, I have a new PR: https://github.com/Kfwullaert/UniMath/pull/41
It appears to me that their names were not to the point.

rmatthes: Now even the last two laws: https://github.com/Kfwullaert/UniMath/pull/42

UniMath/Semantics/LinearLogic/LinearCategoryEilenbergMooreAdjunction.v
rmatthes: This is now properly addressed in https://github.com/Kfwullaert/UniMath/pull/39

rmatthes (approved):
modulo my two pull requests to the fork

UniMath/CategoryTheory/Monoidal/Comonoids/TransportComonoidAlongRetraction.v
```diff
@@ -0,0 +1,242 @@
+(*
```
nmvdw, line 1:
```suggestion
(**
```
Coqdoc uses (** while normal comments use (*, so it is better to use (**

UniMath/Semantics/LinearLogic/LinearCategoryEilenbergMooreAdjunction.v
```diff
@@ -0,0 +1,123 @@
+(*
```
nmvdw, line 1:
```suggestion
(**
```

UniMath/Semantics/LinearLogic/LinearToLinearNonLinear.v
```diff
@@ -0,0 +1,591 @@
+(*
```
nmvdw, line 1:
```suggestion
(**
```

UniMath/Semantics/LinearLogic/LinearToLinearNonLinear.v
```diff
@@ -0,0 +1,591 @@
...
+      apply maponpaths.
+      exact (! linear_category_counit_nat (pr11 f)).
+  Qed.
+
+  (* Unfortunately, I'm unable to clean up this lemma.
+     If I do this, I will have to do some manipulation in
+     definition linear_category_eilenberg_moore_cartesian.
+     The purpose of this lemma is to avoid having to prove this property in that definition.
+   *)
+  Local Lemma aux
```
nmvdw, line 460: Maybe call it `linear_category_eilenberg_moore_cartesian_lem`, so that it is clear what it is related to.

nmvdw (approved):
I also think it is ready for merging (except for a couple of minor comments).

rmatthes:
Good, so let's merge as soon as CI gives us the green light.

rmatthes:
This should be a rebased merge, not a squash merge.

rmatthes:
I'll prepare a new PR to the fork that will remove the problems coming from issue #1770.

rmatthes:
The announced PR is https://github.com/Kfwullaert/UniMath/pull/43.

PR #1773, opened 2023-08-28
Solve issue1770

rmatthes:
at least so that PR #1768 can be merged

This PR is only for displaying a possible solution. If the remaining problems are solved, I would file a PR against @Kfwullaert 's fork, so that it becomes part of PR #1768.


rmatthes:
is_univalent_full_subcat -> is_univalent_full_sub_category will be required for satellite SetHITs

rmatthes:
I now have found a proof that is accepted by Coq in still acceptable time. I'll make a PR against @Kfwullaert 's fork, and after it has been merged, @nmvdw could improve the situation in a subsequent PR.

PR #1777, opened 2023-09-02
Monoidal functor lifting - resurrection of work in December 2022 by Kobe Wullaert

rmatthes:
This is only the minimal adaptation so that it is compatible with the current master branch.

My "original" work consists in identifying several superfluous import statements in the UniMath library concerning the more specific monoidal sections (that Kobe's work generalizes).

The use case (for which I had asked if Kobe could do this work) is the rather important simplification of the construction of inserters in the bicategory of monoidal categories, as seen in https://github.com/UniMath/UniMath/compare/master...rmatthes:monoidalfunctorlifting?expand=1#diff-e5483c470a83be919400406e128e649e555570269098297ca2920948406e1de8


UniMath/CategoryTheory/Monoidal/Displayed/WhiskeredDisplayedBifunctors.v
```diff
@@ -350,7 +350,7 @@ Section DisplayedBifunctor.
 
   Definition is_disp_bifunctor
              {A B C : category}
-             {F : bifunctor A B C}
+             (F : bifunctor A B C)
```
nmvdw, line 353: Why do you make this argument explicit instead of implicit?

UniMath/CategoryTheory/Monoidal/Displayed/WhiskeredDisplayedBifunctors.v
```diff
@@ -442,7 +442,7 @@ Section DisplayedBifunctor.
              {DC : disp_cat C}
              (DF : disp_bifunctor F DA DB DC)
     : disp_bifunctor_rightcompax DF
-    := pr1 (pr2 (pr2 (pr2 (is_disp_bifunctordata_from_disp_bifunctor DF)))).
+    := pr1 (pr2 (pr2 (pr2 (is_disp_bifunctor_from_disp_bifunctor DF)))).
```
nmvdw, line 445: The `pr122` notation is useful here

UniMath/CategoryTheory/Monoidal/MonoidalFunctorLifting.v
```diff
@@ -0,0 +1,388 @@
...
+  Import DisplayedBifunctorNotations.
+  Import DisplayedMonoidalNotations.
+
+  Context {C' C : category} {F : functor C' C} {D : disp_cat C}.
+  Local Notation TD := (total_category D).
+  Context {M' : monoidal C'}
+          {M : monoidal C}
+          (Fm : fmonoidal_lax M' M F)
+          (DM : disp_monoidal D M).
+  Local Notation TM := (total_monoidal DM).
```
nmvdw, line 34: A `Let` might be nicer here instead of `Local Notation`

UniMath/CategoryTheory/Monoidal/Displayed/WhiskeredDisplayedBifunctors.v
rmatthes: This is because I had to specify the implicit argument F in six instances. This came from my other changes concerning more distinction between data and the full structure.

benediktahrens (approved):

PR #1779, opened 2023-09-07
builder when displayed cat is locally propositional

rmatthes:
first example: monoidal full subcategory
this does not avoid code but is conceptually clearer

also builders for displayed binary products and displayed terminal objects in the locally propositional case

### DIFF
--- a/UniMath/Bicategories/MonoidalCategories/ActionBasedStrongFunctorsWhiskeredMonoidal.v
+++ b/UniMath/Bicategories/MonoidalCategories/ActionBasedStrongFunctorsWhiskeredMonoidal.v
@@ -547,9 +547,11 @@ Section Main.
       change (montrafotargetbicat_disp w) in π.
       exact (montrafotargetbicat_tensor_comp_aux v w v' w f (identity w) η π η' π Hyp (id_disp π)).
     Qed.
+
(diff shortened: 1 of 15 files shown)


rmatthes:
I think that the new builder in `Projections` should already appear in `Core` since it will be useful in `Constructions`. Many more simplifications of examples seem possible.

nmvdw (approved):
Nice.

nmvdw:
> I think that the new builder in `Projections` should already appear in `Core` since it will be useful in `Constructions`. Many more simplifications of examples seem possible.

My opinion on this matter is stronger: the whole directory on displayed categories needs cleaning and refactoring. The current structure makes it hard to find stuff and difficult to find where stuff should be put. I would thus suggest to keep this refactoring maybe outside of this PR.

benediktahrens:
```
--- checking the ordering prescribed by the files UniMath/*/.packages/files ---
error: *** UniMath/CategoryTheory/coslicecat.vo should not require UniMath/CategoryTheory/DisplayedCats/Projection.vo
1 dependency order errors in package listings
```

rmatthes:
Okay, no refactoring. But I moved locally propositional to `Core.v`. This also solves the problem Benedikt spotted on compilation ordering. I'll do the other examples of monoidal categories.

nmvdw (approved):

PR #1780, opened 2023-09-07
Characterization of adjoint equivalences/invertible 2cells of double categories

nmvdw:
This PR strengthens a result from sufficient conditions to a characterization.


benediktahrens (approved):
Awesome!

PR #1781, opened 2023-09-09
exploit groupoidal disp. cat. in disp. mon. cat.

rmatthes:
showcase: mon. cat. of dialgebras
also a forgotten use of local propositionality there

This is work in the continuation of PR #1779.

### DIFF
--- a/UniMath/CategoryTheory/DisplayedCats/Projection.v
+++ b/UniMath/CategoryTheory/DisplayedCats/Projection.v
@@ -122,7 +122,7 @@ Defined.
  *)
 Definition groupoidal_disp_cat
            {C : category}
-           (D : disp_cat C)
+           (D : disp_cat_data C)
   : UU
   := ∏ (x y : C)
        (f : x --> y)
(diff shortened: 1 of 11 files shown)


rmatthes:
I see an issue: how to automatically get the iso laws for the unitors and the associator when the displayed monoidal data was constructed with the builder that exploits the disp. cat. being groupoidal?

nmvdw:
> I see an issue: how to automatically get the iso laws for the unitors and the associator when the displayed monoidal data was constructed with the builder that exploits the disp. cat. being groupoidal?

This might depend on the order of how you define stuff, right?
If you first define the displayed category without the monoidal data, then this should be fine?

rmatthes:
@nmvdw : the dilemma should be very visible with the definition in my commit https://github.com/UniMath/UniMath/pull/1781/commits/bcae0d8d2d0df806a62c5b02d1ae58ee871693f3

nmvdw:
So, we are looking at a constructor where we assume that all displayed morphisms are invertible, but we don't necessarily assume that all displayed morphisms are equal. As such, in this case, one cannot automatically deduce the laws for the displayed monoidal structure.

However, a constructor that combines both assumption would be very nice to have.

rmatthes:
But I think this would also be easy by hand: get the disp_monoidal_data from being groupoidal and then use the builder for local propositionality.

benediktahrens:
Should this PR be marked WIP?

rmatthes:
Agreed. The PR looked complete, but more instances of the usefulness showed up and more are likely to show up very soon.

benediktahrens (approved):

PR #1782, opened 2023-09-18
Unfolded Definition of DoubleCats

nimarasekh:


nmvdw (approved):
Can you delete the file in `CategoryTheory/DoubleCats/Core`? It is being moved to a different location

PR #1784, opened 2023-09-19
Equivalence between the two definitions of double categories

nmvdw:
We have an unfolded one, and it is equivalent to the one construct via displayed bicategories. The proof is a matter of reorganizing the data.


nmvdw:
Note: some `Qed` are changed to `Defined`. This does not matter much for memory consumption, because the involved functions are projections. It does simplify the proof though.

benediktahrens (approved):

PR #1795, opened 2023-10-12
Enrichment for the category of presheaves

nmvdw:
In this PR, an enrichment for the category of functors from $C$ to $\mathcal{V}$ is defined for arbitrary categories $C$. Note that this is not a special case of the enriched functor category, because we do not assume that $C$ is enriched over $\mathcal{V}$.


UniMath/CategoryTheory/EnrichedCats/Examples/Presheaves.v
```diff
@@ -0,0 +1,1253 @@
...
+ In this file, we define an enrichment for the category of presheaves.
+
+ In the file `FunctorCategory.v`, we showed that the category of
+ enriched functors between two enriched categories forms an enriched
+ category as well. However, this does not capture all examples that
+ one might want to have, such as the category of presheaves. If we were
+ to use the category of enriched functors, then we only get presheaves
+ on categories `C` enriched over `V`. As such, we do not naturally get
+ all the examples that one might want to consider, such as presheaves
+ over partial orders.
```
benediktahrens, line 14:
```suggestion
 In the file `FunctorCategory.v`, we showed that the category of
 V-enriched functors between two V-enriched categories forms a V-enriched
 category as well. However, this does not capture all examples that
 one might want to have, such as the category of V-valued presheaves. If we were
 to use the category of V-enriched functors, then we only get V-valued presheaves
 on categories `C` enriched over `V`. As such, we do not naturally get
 all the examples that one might want to consider, such as presheaves
 over partial orders.
```

benediktahrens (approved):

PR #1797, opened 2023-10-20
Universal property of the Rezk completion for enriched categories

nmvdw:
This PR builds forth on https://github.com/UniMath/UniMath/pull/1796.

It contains a proof of the universal property of the Rezk completion of enriched categories. The main idea behind the proof is the same as for ordinary categories. The only difference is that the enrichments must be constructed as well, and those can be dealt with using the same methods.


UniMath/CategoryTheory/EnrichedCats/Examples/Precomposition.v
```diff
@@ -0,0 +1,68 @@
+(********************************************************************************
+
+ Precomposition
+
+ In this file, we define the precomposition functor for enriched categories.
+ More specifically, given an enriched functor from `C₁` to `C₂`, we get a functor
+ from the category `E₂ ⟶ E₃` of enriched functors to the category `E₁ ⟶ E₃`
+ of enriched functors. Note that we do not require these two categories to be
+ enriched. We are only interested in them as categories.
```
benediktahrens, line 9: The word "require" is usually used for assumptions. Here, the enrichedness of the functor categories is not an assumption, but a consequence, isn't it?
```suggestion
 of enriched functors. Note that we do not use that these two categories are
 enriched. We are only interested in them as categories.
```

benediktahrens (approved):

UniMath/CategoryTheory/EnrichedCats/Examples/Precomposition.v
nmvdw: To guarantee that the functor categories are enriched, we need to make assumptions over `V` (complete and symmetric monoidal closed). So, if we want to use that the functor categories are enriched, we need to make additional assumptions. In the file, we only assume `V` to be a monoidal category without any assumption on whether it is complete or symmetric monoidal closed.

Do you think that 'require' or 'use' is a better fit here?

benediktahrens: Thanks for the explanations. I think require is fine, then - I retract my suggestion.

PR #1799, opened 2023-10-23
Kleisli objects in the bicategory of enriched categories

nmvdw:
This PR contains:
- A formalization of Kleisli objects in the bicategory of univalent enriched categories. From this, we obtain the usual theorems about enriched Kleisli categories.
- The displayed bicategory of enrichments is locally groupoidal
- A refactoring that simplifies the change of base pseudofunctor (remove unnecessary identifiers)

The most interesting thing to notice is the following. By using the usual construction for Kleisli categories, one can obtain Kleisli objects without any assumption on `V` (this is done in Street's formal theory of monads). However, for univalent categories this does not work: the underlying category of the usual Kleisli category is not univalent. Instead, we can construct the Kleisli category as a full subcategory of the Eilenberg-Moore category. This works fine, but in this case, we need to add assumptions on `V` (namely that it has equalizers). A possible way to prevent this assumption on `V` is by constructing a Rezk completion for enriched categories that preserves the universe level and for that, one has to assume the existence of a Rezk completion that preserves the universe level. Another possible approach would be to construct the enrichment for the univalent Kleisli category directly instead of using the enrichment of the Eilenberg-Moore category.


UniMath/CategoryTheory/EnrichedCats/Examples/UnivalentKleisliMapping.v
```diff
@@ -0,0 +1,453 @@
+(****************************************************************************************
+
+ The universal mapping property of the enriched univalent Kleisli category
+
+ In this file, we establish the universal mapping property of the enriched univalent
+ Kleisli category. More precisely, we establish that mapping properties that show that
+ the bicategory of enriched categories has Kleisli objects.
```
benediktahrens, line 7:
```suggestion
 Kleisli category. More precisely, we establish the mapping properties that show that
 the bicategory of enriched categories has Kleisli objects.
```

benediktahrens (approved):

PR #1805, opened 2023-11-27
The univalent category of strict double categories

nmvdw:
This PR contains a construction of the univalent category of strict double categories. There also are some minor modifications/renamings in some files.


UniMath/CategoryTheory/TwoSidedDisplayedCats/Strictness.v
```diff
@@ -0,0 +1,227 @@
+(*******************************************************************************
+
+ Strict 2-sided displayed categories
+
+ A strict 2-sided displayed category is a 2-sided displayed category for which
+ the objects form a set. Generally speaking, this notion is rather restricting.
```
benediktahrens, line 6: "the objects form a set" - does that refer to the displayed objects or the base objects or both?

UniMath/Bicategories/DoubleCategories/StrictDoubleCatBasics.v
```diff
@@ -0,0 +1,756 @@
+(**********************************************************************************
+
+ Basics for strict double categories
+
+ In this file, we lay down the basics notions to define strict double categories.
```
benediktahrens, line 5:
```suggestion
 In this file, we lay down the basic notions to define strict double categories.
```

benediktahrens (approved):

UniMath/CategoryTheory/TwoSidedDisplayedCats/Strictness.v
nmvdw: In the definition that I formalized in Coq, only the displayed objects are required to form a set. However, morally, we want both the displayed objects and the base objects to form a set, because if that would be the case, then the total category is guaranteed to be a strict category. One loses some flexibility by only defining strict 2-sided displayed categories over strict categories, and that's the reason why there requirement that the base objects form a set is not there.

Note that for univalence, the situation is analogous. Displayed univalence is formulated without mentioning univalence of the base, and these two requirements are only combined to prove that the total category is univalent.

Do you have suggestions to make the documentation clearer?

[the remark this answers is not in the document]
nmvdw: Sure

PR #1820, opened 2024-01-03
More theory on double bicategories

nmvdw:
This PR contains:
- Horizontal dual of a Verity double bicategory
- Conjoints in Verity double bicategories
- Companion pairs and conjoints in pseudo double categories, and examples of these


nmvdw:
Note: on my laptop the sanity checks succeed.

PR #1823, opened 2024-01-15
Derive the types of objects from their categories, instead of defining them separately

arnoudvanderleer:
Additionally, shorten some property names (and make them consistent), and transition from using functors to using (better behaved) indexed sets.


nmvdw:
What did you think of defining the types as objects of the category instead?

nmvdw:
> I am not sure whether unification has gotten any faster, but at least I have not encountered any slow unification inside my proofs (only at a couple of Qeds of RepresentationTheorem).

I don't think that either version would perform significantly better, because their definitions are basically the same.

DanGrayson:
Is there a good reason to do this?

arnoudvanderleer:
> Is there a good reason to do this?

I think there is a good reason to do this. This PR does an overhaul of the categories in my package:
 * Previously, the category `algebra_cat` and the type `algebra` had no relation, apart from the fact that the type `ob algebra_cat` was equal to `algebra`. To make this possible, one had to keep track of the precise structure of `ob algebra_cat` and make sure that `algebra` machted this. Now, `algebra` is defined as `ob algebra_cat`, which expresses their relation more strongly, and avoids duplication, which is more sustainable. The same goes for the morphisms, and for the categories `algebraic_theory_cat`, `lambda_theory_cat` and `presheaf_cat`.
  * Previously, the categories were built on top of a functor category `FinSet --> SET`, because Hyland constructs them that way. However, most of such a functor `F` is reduntant. The only thing that we want is the sequence of sets `fun n => F(n)`. The fact that our objects were functors, meant that coq could not infer automatically that `F n` was a set, which lead to undesirable definitions like
```coq
    Definition some_def
      (F : algebraic_theory)
      (f : F n : hSet)
      : UU
      := comp f pr = f.
```
After this change, one can leave out the `: hSet`.
  * Previously, one had to use some special constructor to build the objects by just specifying the sequence of sets (indexed set) and some additional data. The constructor then inferred the functor structure. However, we almost never actually used the fact that it was a functor. The transition from the functor to the indexed set means that don't need these additional constructors (`make_algebraic_theory'`, `make_presheaf'` etc.) and can remove the files that supply some parts of these constructors `AlgebraicTheories2.v`, `AlgebraicTheoryMorphisms2.v`.
  * This PR shortens a lot of unreadable and unwritable names to more consistent and workable names. For example, `algebraic_theory_comp_projects_component` becomes `pr_comp`, since it describes the interplay between the composition operator with the projections. On the other hand, `algebraic_theory_comp_identity_projections` becomes `comp_pr`, because this too describes the interplay, but then with `pr` as the second argument, instead of the first. These names are easier to work with, and actually more descriptive in some way.

DanGrayson:
@arnoudvanderleer -- thanks for the explanation.
@benediktahrens -- it looks good to me.

benediktahrens (approved):

arnoudvanderleer:
@benediktahrens @DanGrayson @nmvdw 
Are we still waiting for a review or can this be merged? (If yes, which of the two?)

PR #1825, opened 2024-01-16
Hack/metacoq

jmikedupont2:
First version of unimath compiled on top of metacoq

The changes are to the following notations
,, → ,,m
× → ×u
<-  → <--

Also there was another change tpair takes only 2 args not 3

I did only up to Combinatorics for our purposes that should be enough


benediktahrens:
Dear @jmikedupont2 , we are not interested in merging these changes into UniMath. If you would like to have a copy of UniMath with these changes, please maintain a fork yourself.

jmikedupont2:
Understood. I wanted to point out that Unimath is not compatible with MetaCoq because of the conflicts in symbology. Do you have another means of extraction and reflection in unimath?

DanGrayson:
What is "symbology"?

nmvdw:
> Understood. I wanted to point out that Unimath is not compatible with MetaCoq because of the conflicts in symbology. Do you have another means of extraction and reflection in unimath?

Are the only clashes caused by the usage of symbols or are there other causes to the incompatibility?

rmatthes:
Can nothing been done here with interpretation scopes ?

PR #1826, opened 2024-01-22
Fix the sanity-checks "argument list too long" bug

arnoudvanderleer:
... by moving a large list of files, which was used twice, to a bash variable.


DanGrayson:
I'll take a lpok...

DanGrayson:
Looks good to me.  Also all the checks pass, which is a positive sign.

PR #1827, opened 2024-01-23
Include file missing in package declaration

m-lindgren:


m-lindgren:
I'll merge this immediately without waiting for the rest of the CI to complete (it shouldn't fail with this change anyway, since it's using Dune and not make), so that we don't block Coq-CI.

benediktahrens:
@m-lindgren : thanks for taking care of this!

PR #1834, opened 2024-01-31
WIP Feature/ltac2

jmikedupont2:


benediktahrens:
Dear @jmikedupont2 , I don't think that code is meant for merging, but rather to illustrate a behaviour that you consider to be a bug. I will thus close this PR.

PR #1646, opened 2023-03-10
Refactor monoidal categories (by Niels van der Weide)

m-lindgren:
This is a duplication of the material in PR #1644, but squashed into one commit for better preservation of git history. See #1644 for details.

PR #1838, opened 2024-02-06
Make file and directory names in the CategoryTheory package more consistent

arnoudvanderleer:


nmvdw (approved):
Nice.

Did you use a script for this? It might be needed for one of the other repositories.

arnoudvanderleer:
Unfortunately, it involved a lot of manual work. When I asked around in the coq zulip for a tool that could automatically update references, I got no response, other than someone saying "You could probably use the .glob files for that".

nmvdw:
My plan would be to do a global find and replace like this:
```
Require Import UniMath.CategoryTheory.categories. -> Require Import UniMath.CategoryTheory.Categories.
Require Import UniMath.CategoryTheory.limits.zero. -> Require Import UniMath.CategoryTheory.Limits.Zero.
Require Import UniMath.CategoryTheory.limits.binproducts. -> Require Import UniMath.CategoryTheory.limits.Binproducts.
...
```
There are bash commands for this (and I always forget those).

What about something like this?

arnoudvanderleer:
I did something like that (but then separately for each line) in the find-replace of VScode, but that still needs a manual update of `.package/files`, as well as [some qualified names](https://github.com/UniMath/UniMath/pull/1838/commits/d150e8773f3e80fb47866ba17204779bef936120).

nmvdw:
> I did something like that (but then separately for each line) in the find-replace of VScode, but that still needs a manual update of `.package/files`, as well as [some qualified names](https://github.com/UniMath/UniMath/pull/1838/commits/d150e8773f3e80fb47866ba17204779bef936120).

I would expect those to occur less in the other repositories, so maybe that problem will be less.

arnoudvanderleer:
Can we merge this now, or do we need to wait till more checks pass?

nmvdw:
> Can we merge this now, or do we need to wait till more checks pass?

I think @benediktahrens can make the decision for merging

rmatthes:
This is a risky endeavour. Git does not have a proper renaming function, so the history of all files in CategoryTheory.categories can potentially be lost for Git. But in GitHub, studying the changes in the 445 files, there seem to be no lost connections, other than from Categories/HSET/All.v to  categories/HSET/All.v.
In the past (PR #1644, PR #1646), we were faced with more difficult problems, and there, it was vital to do a squash merge - but this is anyway the default now.

@arnoudvanderleer Did you try to visit the Git history (e.g. through git blame) of some of the renamed files? Are you ready to propagate the changes to the satellite repos?

arnoudvanderleer:
@rmatthes I will take a look at the git history of some files.
Regarding the satellite repositories, I have actually never worked with those. Is there a quick primer on the satellite repositories? (I.e., why are they here, how actively are they maintained, what is their intended purpose/scope?)

rmatthes:
@arnoudvanderleer The four satellites that fall within the continuous integration of this GitHub repository are definitely maintained, even if this often only means propagating upstream changes. The satellites are less integrated parts of UniMath, just like the "contributions" that Coq had already a long time ago. Satellites do not need to conform to the coding style of UniMath as consistently as the UniMath library itself (HITs use more inductive types). With dune, the satellites can be very comfortably compiled together with the UniMath library. It suffices to clone the Git repos of the satellites as subdirectories of the root directory of the main repo.  The command `\time -v dune build -j 6` (the one I use on Ubuntu) then takes care of the satellites as well. Of course, the code updates have to be done separately for all the satellites.

benediktahrens:
@arnoudvanderleer : thanks a lot for filing this PR - this fixes a long-standing weirdness in the library. I won't have time to adapt the satellites any time soon. But that should not prevent us from merging this PR soon.

arnoudvanderleer:
@rmatthes Thanks, that explains a lot. Is this documented somewhere, or is it folklore? (Would it be an idea to add this to the wiki?)
I will try to adapt the satellites today then :-)

arnoudvanderleer:
I added 4 PR's in the satellite repositories

rmatthes:
> @rmatthes Thanks, that explains a lot. Is this documented somewhere, or is it folklore? (Would it be an idea to add this to the wiki?) I will try to adapt the satellites today then :-)

This is not relevant for the typical PR to UniMath that only adds new material. But it would be a good thing to have this in the documentation.

PR #1839, opened 2024-02-07
replaces v3 by v4 in all actions (this should do the migration from N…

rmatthes:
…ode 16 to Node 20)

### DIFF
--- a/.github/workflows/build-unimath.yml
+++ b/.github/workflows/build-unimath.yml
@@ -27,7 +27,7 @@ jobs:
     name: Sanity Checks
     runs-on: ubuntu-22.04
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Install build dependencies
         run: |
           sudo apt-get update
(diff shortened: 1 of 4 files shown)


rmatthes:
I did not dig deeper into the .yml file. I only applied the changes to one of the three files for the moment. I am surprised that the macOS build does not profit from the cache while the other checks are as successful or unsuccessful as before, seen in max. 2 minutes.

benediktahrens:
@rmatthes : thanks a lot for preparing this! For my education, could you point me to a resource explaining what you briefly mention about "nodes"? Relatedly, is there a way to receive updates about new versions of the scripts?

rmatthes:
@benediktahrens I just followed warnings and fell on https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/ . This is not an organized way to keep track of version updates.

benediktahrens (approved):
It looks fine to me, but I am not competent. It would be great to have @m-lindgren 's feedback.

rmatthes:
Changing the language versions triggers a full rebuild. The satellites are not touched by this since their CI has not yet been updated (see the pending PR in the repos).

m-lindgren (approved):
Looks good to me!

PR #1844, opened 2024-02-12
more consequent use of the sum of a constant functor and the domain-s…

rmatthes:
…pecific functor

This is for the construction of the variable inclusion and the other constructors.

SubstitutionSystems.v: Id_H generalized to Const_plus_H, Id_H is then an instance
                       the constructors eta and tau are now for the more general situation
LiftingInitial_alt.v only instantiates these notions from SubstitutionSystems.v LiftingInitial.v: idem
MonadsFromSubstitutionSystems.v: def. of Id_H deleted STLC.v now refers to SubstitutionSystems.v for Id_H and uses eta for var_map STLC_alt.v uses eta for var_map
MLTT79.v now refers to SubstitutionSystems.v for Id_H SimplifiedHSS gets the same changes

MultiSortedMonadConstruction_(coind_)actegorical.v now refers to SubstitutionSystems.v for Id_H

Lam.v exactly uses eta and tau (the latter for the pseudo-constructor Lam_App_Abs) the development now gets help with type information and there is a certain reorganization of the proof of bracket_for_LamE_algebra_on_Lam_unique SimplifiedHSS/Lam.v is similar but does not have the statement bracket_for_LamE_algebra_on_Lam_unique


rmatthes:
Given that there are many other developments and that I paused mine, it would be better to merge this unfinished work.

UniMath/SubstitutionSystems/MultiSorted_alt.v
```diff
@@ -139,14 +141,50 @@ use tpair.
...
+    intro i.
+    refine (_,, uni).
+    set (nbbindersallargs := BindingSigMap s i).
+    exact (map translateArity  nbbindersallargs).
+  Defined.
+
+  Definition MultiSortedSigFromBindingSig : MultiSortedSig
+    := I ,, arFromBindingSig.
+
+End MultiSortedSigFromBindingSig.
```
benediktahrens, line 178: What does the code in this section do?

rmatthes: It is meant to make a multisorted binding signature out of a binding signature by using a single sort. I cannot specify what it does so easily.

benediktahrens (approved):

PR #1845, opened 2024-02-15
Strengthen foundation of retracts and equivalences, ...

arnoudvanderleer:
leading up to some results about the Karoubi envelope


UniMath/CategoryTheory/Core/Categories.v
```diff
@@ -309,26 +309,118 @@ Defined.
...
 
-  Definition is_retraction {A B : ob C} (m : A --> B) (r : B --> A) :=
-    m · r = identity A.
-
-  Lemma isaprop_is_retraction {A B : ob C} (m : A --> B) (r : B --> A) :
-    has_homsets C -> isaprop (is_retraction m r).
+  Definition is_retraction
+    {A B : C}
+    (s : A --> B)
+    (r : B --> A)
```
nmvdw, line 315:
```suggestion
    (r : B --> A)
    : UU
```

UniMath/CategoryTheory/Core/Categories.v
```diff
@@ -309,26 +309,118 @@ Defined.
...
+    := ∑ (s : A --> B)
+        (r : B --> A),
+        is_retraction s r.
+
+  Section Accessors.
+
+    Context {A B : C}.
+    Context (H : retraction A B).
+
+    Definition retraction_section
```
nmvdw, line 340: Do you want a coercion here?

UniMath/CategoryTheory/Core/Categories.v
```diff
@@ -309,26 +309,118 @@ Defined.
...
 
 End SectionsAndRetractions.
+
+Section Idempotents.
+
+  Context {C : precategory}.
+
+  Definition is_idempotent
+    {c : C}
+    (e : c --> c)
```
nmvdw, line 373:
```suggestion
    (e : c --> c)
    : UU
```

UniMath/CategoryTheory/Core/Categories.v
```diff
@@ -309,26 +309,118 @@ Defined.
...
+    := e · e = e.
+
+  Definition idempotent
+    (c : C)
```
nmvdw, line 377:
```suggestion
    (c : C)
    : UU
```

UniMath/CategoryTheory/Core/Categories.v
```diff
@@ -309,26 +309,118 @@ Defined.
 Section SectionsAndRetractions.
   Context {C : precategory}.
 
-  Definition is_retraction {A B : ob C} (m : A --> B) (r : B --> A) :=
-    m · r = identity A.
-
-  Lemma isaprop_is_retraction {A B : ob C} (m : A --> B) (r : B --> A) :
-    has_homsets C -> isaprop (is_retraction m r).
+  Definition is_retraction
```
nmvdw, line 312: It might be good to put the stuff on retractions and idempotents in a separate file. The advantage would be:
- The `Categories.v` remains more focussed
- It is easier to navigate and find the relevant material (instead of having to look in every file)

UniMath/CategoryTheory/Equivalences/Core.v
```diff
@@ -456,38 +602,29 @@ Defined.
...
 
-Section from_fully_faithful_and_ess_surj_to_equivalence.
+Section FullyFaithfulAndSplitEssentiallySurjectiveToEquivalence.
 
 Variables A B : category.
-Hypothesis HA : is_univalent A.
 Variable F : functor A B.
 Hypothesis HF : fully_faithful F.
-Hypothesis HS : essentially_surjective F.
+Hypothesis HS : split_essentially_surjective F.
```
nmvdw, line 610: Nice. This is a good improvement

UniMath/CategoryTheory/Limits/Graphs/Coequalizers.v
```diff
@@ -383,3 +383,38 @@ Proof.
      intro e.
      induction e; apply idpath.
 Defined.
+
+(* For a section-retraction pair (g, f) between b and a, b is the coequalizer of id_a and f · g *)
+Lemma retract_is_coequalizer
```
nmvdw, line 388: There are two files on coequalizers: one uses graphs while the other is direct. I tend to use the file Limits/Coequalizers.v, while you seem to use a different one.

Could this inconsistency lead to problems?

UniMath/AlgebraicTheories/FundamentalTheorem/SurjectivePrecomposition.v
```diff
@@ -0,0 +1,98 @@
...
+Section PrecompositionEssentiallySurjective.
+
+  Context {C C' : category}.
+  Context (F : C ⟶ C').
+  Context (H : fully_faithful F).
+
+  Context (D : category).
+  Context (HD : Colims D).
+
+  Let Fweq := weq_from_fully_faithful H.
```
nmvdw, line 35: What is the type of `Fweq`?

UniMath/AlgebraicTheories/FundamentalTheorem/SurjectivePrecomposition.v
```diff
@@ -0,0 +1,98 @@
...
+          refine (!invmap_eq _ _ _ (!_));
+          refine (functor_comp _ _ _ @ _);
+          refine (maponpaths _ (homotweqinvweq (Fweq _ _) _) @ _);
+          exact (!pr2 e @ id_right _)
+        ).
+    Defined.
+
+    Lemma pre_comp_z_iso_is_inverse
+      : is_inverse_in_precat (C := D)
+        ((unit_from_right_adjoint (is_right_adjoint_precomposition HD F) P : _ ⟹ _) c)
```
nmvdw, line 61: It might be good to introduce the unit as a variable in this section. Something like
```
Let \eta : _ ==> _ := ...
```

UniMath/AlgebraicTheories/FundamentalTheorem/CommonUtilities/KaroubiEnvelope.v
```diff
@@ -0,0 +1,449 @@
...
+  Context (C : category).
+
+(** * 1. The Karoubi envelope and its embedding of C *)
+
+  Definition karoubi_envelope_data
+    : precategory_data.
+  Proof.
+    use make_precategory_data.
+    - use make_precategory_ob_mor.
+      + exact (∑ (c : C), idempotent c).
```
nmvdw, line 73: Why not use a displayed category for this construction?

UniMath/AlgebraicTheories/FundamentalTheorem/CommonUtilities/KaroubiEnvelope.v
```diff
@@ -0,0 +1,449 @@
...
+        karoubi_envelope_is_precategory)
+      karoubi_envelope_has_homsets.
+
+  Definition karoubi_envelope_inclusion
+    : C ⟶ karoubi_envelope.
+  Proof.
+    use make_functor.
+    - use make_functor_data.
+      + intro c.
+        repeat use tpair.
```
nmvdw, line 144: It might be good to add a builder for the morphisms

nmvdw (approved):

UniMath/CategoryTheory/Core/Categories.v
arnoudvanderleer: That is a good idea. I added it here because Categories already contained some stuff about retractions.
Shall I put it in `UniMath.CategoryTheory.Retracts`?

UniMath/CategoryTheory/Core/Categories.v
arnoudvanderleer: I maybe could add a coercion that gives the retraction `r: B --> A` indeed. Do you think that is a good idea?

UniMath/CategoryTheory/Limits/Graphs/Coequalizers.v
arnoudvanderleer: Oh, right. For some definition of problems: I don't think it will pose unsurmountable problems, but I think it can cause some major inconvenience. It feels like it is related to #395 .
I will change this definition over to `Limits/Coequalizers`.
Would it maybe be an idea to standardize the practice of proving things about `Limits/...` instead of `Limits/Graphs/...`, and to only use the latter in the proofs and not the statements of theorems?

UniMath/CategoryTheory/Core/Categories.v
nmvdw: I think that would be fine. Does @benediktahrens agree with this reorganization?

UniMath/CategoryTheory/Core/Categories.v
nmvdw: That's the one you will use the most often, right? I think then it would be good to have.

UniMath/CategoryTheory/Limits/Graphs/Coequalizers.v
nmvdw: Standardization might be a bit too much for this PR, and it will also require further discussion. I was mainly wondering about the motivation of your choice.

UniMath/AlgebraicTheories/FundamentalTheorem/CommonUtilities/KaroubiEnvelope.v
arnoudvanderleer: Good question. I tried it for a bit, but then realized that the identity morphisms don't lie over identity morphisms.

The objects in this category are idempotents, and the identity on the idempotent `f: c --> c` is `f`, not `identity c`.

nmvdw: I see. Thanks

UniMath/CategoryTheory/Limits/Graphs/Coequalizers.v
arnoudvanderleer: Oh, yes, not in this PR. But to prevent having a bunch of theorems around one definition and a bunch around the other, it may be a good idea to, at some point, standardize use of one of them, and only use the other as a stepping stone for some results

UniMath/CategoryTheory/Core/Categories.v
arnoudvanderleer: (Or maybe IdempotentsAndRetractions or something like that?)

arnoudvanderleer: Or also split those two up into two different files?

nmvdw: It might be ok to have them in the same file for now (otherwise, you might want to manage export files).

PR #1853, opened 2024-02-27
Pseudofunctors for comprehension categories

nmvdw:
This PR contains the pseudofunctors that (will) constitute a biequivalence between categories with finite limits and comprehension categories that are democratic and have unit/product/equalizer/sigma types. In addition, it contains the results from category theory that are necessary to construct these pseudofunctors.


UniMath/CategoryTheory/DisplayedCats/Codomain.v
```diff
@@ -305,52 +304,6 @@ Proof.
...
-  split.
-  - intros x yf.
-    use eq_cod_mor.
-    rewrite transportb_cod_disp.
-    cbn.
-    rewrite functor_id.
-    apply idpath.
-  - intros x₁ x₂ y₃ yf₁ yf₂ yf₃ f₁ f₂ gp₁ gp₂.
-    use eq_cod_mor.
-    rewrite transportb_cod_disp.
```
benediktahrens, line 338: Where did this code go?

UniMath/CategoryTheory/Limits/PreservationProperties.v
```diff
@@ -0,0 +1,676 @@
...
+ `C₁ ≃ D₁` and `C₂ ≃ D₂` such that the resulting square commutes.
+
+ We also compare preservation of pullbacks and preservation of products and equalizers.
+
+ Contents
+ 1. Preservation and natural isomorphisms
+ 2. Preservation and equivalences in the arrow bicategory
+ 3. Preservation of binary products and equalizers from pullbacks
+ 4. Preservation of pullbacks via binary products and equalizers
+
```
benediktahrens, line 23: This could be a nice application of univalence, if you assumed that the categories involved are univalent. Aren't you?

benediktahrens (approved):

UniMath/CategoryTheory/DisplayedCats/Codomain.v
nmvdw: I moved it to another file (`FunctorCod`).

UniMath/CategoryTheory/Limits/PreservationProperties.v
nmvdw: I do. However, I proved these statements generally, so that they would be reusable in other cases as well and because for these statements, it is not very complicated to do so. However, for more difficult stuff (such as preservation of exponentials), I would have used univalence.

PR #1856, opened 2024-02-29
restore compilation with Coq 8.20 dev version

rmatthes:
This has the same objective as PR #1848 and deals with a future problem of code that was merged in the meantime.


benediktahrens:
@rmatthes : thanks a lot for preparing this!

PR #1862, opened 2024-03-12
Extending the biequivalence by Clairambault and Dybjer to local properties

nmvdw:
This PR contains an extension of the biequivalence by Clairambault and Dybjer to include local properties of categories. Currently, the example of this is given by lextensive categories, so we get a biequivalence between the bicategory of univalent lextensive categories and democratic full comprehension categories with ∑, extensional identity, empty types, and sum types.


UniMath/CategoryTheory/DisplayedCats/Codomain/IteratedSlice.v
```diff
@@ -0,0 +1,326 @@
...
+
+ Iterated slice categories
+
+ The slice construction can be iterated. More specifically, if we have a category `C`,
+ an object `y : C`, and a morphism `f : x --> y`, then we can consider the slice
+ category `C/y/f`. This slice category is equivalent to a simpler one, namely `C/x`.
+ In this file, we construct this equivalence. In addition, if `C` has pullbacks, then
+ we characterize the fiber functor of the iterated slice category via the fiber functor
+ of the ordinary slice category. This allows us to conclude that preservation of colimits
+ for the pullback functor of the iterated slice category follows from presrvation of
```
benediktahrens, line 11:
```suggestion
 for the pullback functor of the iterated slice category follows from preservation of
```

benediktahrens (commented):
Should "property" be renamed to "structure"?

nmvdw:
> Should "property" be renamed to "structure"?

It is good to have some discussion about this. The way I think about this notion, is as predicate `P_0 : univalent_category -> hProp` together with some stuff. I implemented it via a type family `P_0 : category -> UU` (which is only an hProp for univalent categories), because that it is a bit more convenient to use (it doesn't require one to carry the proofs of univalence around). For this reason, I called `P_0` a predicate on categories (this part of the documentation needs to be clarified)

Because I looked at the formalization from this perspective, I was a bit in doubt whether I should use 'property' or 'structure'. In the definition, I am describing `P_0 : univalent_category -> hProp`, and that suggests a property on univalent categories. However, in the examples I look at initial objects and binary coproducts, and those are structure (in fact, they are 'property-like structure). 

Do you think it is better to consistently talk about 'local structure' (this would also mean renaming files and definitions)?

UniMath/Bicategories/DisplayedBicats/Examples/DisplayedCatToBicat.v
```diff
@@ -22,24 +22,6 @@ Require Import UniMath.Bicategories.DisplayedBicats.DispUnivalence.
 Local Open Scope cat.
 Local Open Scope mor_disp_scope.
 
-Definition is_chaotic
```
nmvdw, line 25: This was actually redundant: as it is the same as `disp_2cells_iscontr`

benediktahrens:
> > Should "property" be renamed to "structure"?
> 
> It is good to have some discussion about this. The way I think about this notion, is as predicate `P_0 : univalent_category -> hProp` together with some stuff. I implemented it via a type family `P_0 : category -> UU` (which is only an hProp for univalent categories), because that it is a bit more convenient to use (it doesn't require one to carry the proofs of univalence around). For this reason, I called `P_0` a predicate on categories (this part of the documentation needs to be clarified)
> 
> Because I looked at the formalization from this perspective, I was a bit in doubt whether I should use 'property' or 'structure'. In the definition, I am describing `P_0 : univalent_category -> hProp`, and that suggests a property on univalent categories. However, in the examples I look at initial objects and binary coproducts, and those are structure (in fact, they are 'property-like structure).
> 
> Do you think it is better to consistently talk about 'local structure' (this would also mean renaming files and definitions)?

Thanks for these explanations. When you specialize to set-categories, the predicates are not hprops, right? In any case, I am not sure that this question deserves a lot of rewrite work.

nmvdw:
> Thanks for these explanations. When you specialize to set-categories, the predicates are not hprops, right? In any case, I am not sure that this question deserves a lot of rewrite work.

They indeed are not. I only assume that they are `hProps` for univalent categories, because that's the best that one can expect without excluding key examples (arising from initial objects or coproducts).

I would then suggest that in the documentation, I add:
- that the notion of local property is designed around univalent categories, so we think of it as a map `univalent_category -> hProp`
- why the implementation slightly differs from this
- the key examples are property-like structure (but if one would not assume univalence, then these would be given as structure)

What do you think of that?

nmvdw:
The latest commit should address each of the comments.

PR #1869, opened 2024-03-25
the STLC example in the actegorical setting, with Church numerals

rmatthes:
simply-typed lambda calculus represented through a multi-sorted binding signature, with inductive and with coinductive interpretation, over base category HSET (this continues the rudimentary example presentation in PR #1844 in addressing and establishing naturality) and over an abstract category

construction of the typed Church numerals, including the Church numeral for infinity in the coinductive interpretation  (with its characteristic fixed-point equation)


benediktahrens (approved):

rmatthes:
@benediktahrens Did you also approve the overall idea of my latest commit with the "watchdog" file? I wonder if I should expand that file by the other constructions that are scattered around with the functor categories built on top of `sortToC`.

benediktahrens:
> @benediktahrens Did you also approve the overall idea of my latest commit with the "watchdog" file? I wonder if I should expand that file by the other constructions that are scattered around with the functor categories built on top of `sortToC`.

Do you mean [UniMath/SubstitutionSystems/SortIndexing.v](https://github.com/UniMath/UniMath/pull/1869/files#diff-4b4fa1812d1e00c40f25dee18b3028ff93cc3bf9bef389b50a1045924461d295) ? That file seems useful, a sort of interface. We should probably have more of those.

rmatthes:
@benediktahrens Thank you. I'll thus add a bit more to it - even if the compile time goes for all those checks.

The PR is thus likely to evolve a bit in mathematically irrelevant ways.

rmatthes:
I declare the PR as draft since my progress with `sortToC` was so good that I should be able to remove the multiple identical definitions based on `sortToSet` as well.

UniMath/SubstitutionSystems/STLC_actegorical.v
```diff
@@ -153,14 +195,49 @@ Section IndAndCoind.
...
+    apply idpath.
+  Qed.
+
+  Lemma app_source_gen_mor_pr2 (s t : sort) (ξ ξ' : sortToSet) (f : sortToSet ⟦ ξ, ξ' ⟧)
+    (u : sort) (arg : pr1 (pr1 (pr1 (app_source_gen s t) ξ) u)) :
+    pr2 (pr1 (# (pr1 (app_source_gen s t)) f) u arg) =
+      pr1 (# (pr1 (functor_compose STLC_gen (projSortToSet s ∙ hat_functorSet t))) f) u (pr2 arg).
+  Proof.
+    apply idpath.
+  Qed.
```
benediktahrens, line 212: Are these ever needed? Should they be named, or are they sanity checks that could be written using `Goal`?

UniMath/SubstitutionSystems/STLC_actegorical_abstractcat.v
```diff
@@ -0,0 +1,1037 @@
...
+
+Syntax of the simply-typed lambda calculus as a multisorted signature based on the actegorical development,
+with an abstract base category [C] instead of [HSET] (the objects of [C] serve to represent collections of terms of a given sort in a given context, but also collections of names of variables of a given sort).
+
+The development is "point-free" in the sense that no assumption of well-pointedness of [C] is made (that would allow to replay the point-wise reasoning that is possible in [HSET], as it is done in STLC_actegorical.v).
+
+Thanks to that actegorical development, the inductive and the coinductive calculus are exposed in parallel.
+The Church numerals are developed independently from the choice for inductive or coinductive syntax.
+
+There is also the construction (by primitive corecursion) of the Church numeral for infinity in the coinductive calculus, with a proof that it satisfied a proper recursive equation.
```
benediktahrens, line 11:
```suggestion
There is also the construction (by primitive corecursion) of the Church numeral for infinity in the coinductive calculus, with a proof that it satisfies a proper recursive equation.
```

UniMath/SubstitutionSystems/STLC_actegorical_abstractcat.v
```diff
@@ -0,0 +1,1037 @@
...
+                            (projSortToC (s ⇒ t) ∙ hat_functorC t))) f) u).
+    - exact (pr1 (# (pr1 (functor_compose STLC_gen
+                            (projSortToC s ∙ hat_functorC t))) f) u).
+  Defined.
+
+  Lemma app_source_gen_mor_eq (s t : sort) {ξ ξ' : sortToC} (f : sortToC ⟦ ξ, ξ' ⟧) (u : sort)
+    : app_source_gen_mor_eq_statement s t f u.
+  Proof.
+    apply idpath.
+  Qed.
```
benediktahrens, line 256: Is this ever needed?

benediktahrens (approved):
lgtm. 

There are several results proved by reflexivity - are those ever needed, or are they sanity checks. If the latter, it would be good to signal this via the use of `Goal`.

UniMath/SubstitutionSystems/STLC_actegorical.v
rmatthes: They are used, and the first one is definitely needed since it allows to develop a goal that would become unreadable through simplification. I kept the second one to indicate the analogy in the overall proof structure.

UniMath/SubstitutionSystems/STLC_actegorical_abstractcat.v
rmatthes: Yes, it is applied in a situation where the right-hand side of the goal is a meta-variable. Applying `idpath` instead would not give any progress. Interestingly, the proof would go through, but even the execution of `Qed.` would be slowed down by a factor of ten (and other steps before would be slowed down massively as well).

PR #1870, opened 2024-03-28
Clairambault&Dybjer's biequivalence for regular/exact categories and pretoposes

nmvdw:
This PR contains:
- Formalization of regular and of exact categories
- A proof that being regular and being exact are local properties
- The product of displayed biequivalences

From the second point of these, we can extend the biequivalence by Clairambault and Dybjer to regular/exact categories. As a consequence, we can also extend it to pretoposes.


UniMath/Bicategories/DisplayedBicats/ProductDispBiequiv.v
```diff
@@ -0,0 +1,290 @@
+(*******************************************************************************************
+
+ The product of displayed biequivalences
+
+ If we have two displayed bicategories `D₁` and `D₃` over a bicategory `B`, then we can take
+ their product `D₁ × D₃`, which is a displayed bicategory over `B`. If `D₁` is biequivalent
+ to `D₂` and `D₃` is biequivalent to `D₄`, then the products `D₁ × D₃` and `D₂ × D₃` are
+ biequivalent as well. We prove that in this file under the assumption that all displayed
+ 2-cells in the involved displayed bicategories are equal and invertible.
```
benediktahrens, line 9: How much work do you save with these assumptions? Is it worth it?

UniMath/CategoryTheory/RegularAndExact/RegularEpi.v
```diff
@@ -0,0 +1,251 @@
+(********************************************************************************************
+
+ Regular epimorphisms
+
+ There are several variations on the notion of epimorphism. One of them, is the notion of
```
benediktahrens, line 5:
```suggestion
 There are several variations on the notion of epimorphism. One of them is the notion of
```

UniMath/CategoryTheory/DisplayedCats/Codomain/CodRegular.v
```diff
@@ -0,0 +1,767 @@
+(********************************************************************************************
+
+ The slice category is regular and exact
+
+ One key theorem about regular categories, is that they are closed under slicing.
```
benediktahrens, line 5:
```suggestion
 One key theorem about regular categories is that they are closed under slicing.
```

benediktahrens (approved):

UniMath/Bicategories/DisplayedBicats/ProductDispBiequiv.v
nmvdw: It allows me to skip the coherences, and proving that cells are invertible. The main work behind constructing pseudofunctors/transformations/modifications always lies in proving the coherences. In this case, each coherence would be around 4~5 steps (using a transport lemma, coordinate wise equality, and for both coordinates using the corresponding law). It thus does save some work, and most examples of displayed bicategories satisfy these assumptions..

benediktahrens: If you are happy, I'm happy.

PR #1871, opened 2024-04-02
avoids one kind of warning when compiling UniMath

rmatthes:
only one occurrence, but right in the beginning, which might be irritating


nmvdw (approved):

rmatthes:
I also checked that the command still protects against unstructured code.

PR #1880, opened 2024-04-30
Enriched profunctors

nmvdw:
This PR develops the basics of enriched profunctors. The main idea is to use a whiskered presentation of profunctors, which is similar to how monoidal categories and bicategories are developed. The reason for this choice, is that the ultimate definition of profunctor becomes simpler and more usable. For that purpose, whiskered and curried enriched bifunctors are developed, and they are shown to be equivalent to the usual definition. 

Other stuff in this PR are transformations between enriched profunctors (again in a whiskered style), and some of the main examples of enriched profunctors and transformations between them.


nmvdw:
The final commits contain a small refactoring and squares of profunctors.

UniMath/CategoryTheory/EnrichedCats/Profunctors/ProfunctorSquares.v
```diff
@@ -0,0 +1,1445 @@
+(******************************************************************************************
+
+ Squares between enriched profunctors
+
+ Squares of profunctors are one of the ingredients needed to define the double (bi)category
+ of (univalent) enriched categories, functors, and profunctors, and in this file, we define
+ squares of enriched profunctors. We define squares using transformations betweene enriched
```
benediktahrens, line 7:
```suggestion
 squares of enriched profunctors. We define squares using transformations between enriched
```

benediktahrens (approved):

PR #1884, opened 2024-05-06
Improve notation in BinaryOperations, Groups and Monoids, split Groups and Monoids

arnoudvanderleer:
This PR changes to the following unicode notations:
  * `∑ _, _` instead of `total2 _ _`
  * `_ = _` instead of `paths _ _`
  * `_ @ _` instead of `pathscomp0 _ _`
  * `!_` instead of `pathsinv0 _ _`
  * `_ × _` instead of `dirprod _ _`
  * `_ ,, _` instead of `make_dirprod _ _` or `tpair _ _ _`
  * `_ → _` instead of `_ -> _`
  * `∃ _, _` instead of `hexists _ _`
  * `λ _, _` instead of `fun _ => _`
  * `1` or `0` instead of `unel _ _` wherever possible
  * `_ * _` or `_ + _` instead of `op _ _ _` wherever possible
  * `x^-1` or `-x` instead of `grinv _ _` wherever possible
 
Additionally, it splits `Groups.v` into `Groups2.v` and `AbelianGroups.v` (keeping `Groups.v` as a file with two `Require Export` statements, to not break dependencies), and the same for `Monoids.v`.


UniMath/Algebra/Monoids2.v
```diff
@@ -0,0 +1,601 @@
+(** ** Standard Algebraic Structures *)
```
nmvdw, line 1: I think a more descriptive name would be better than `Monoids2.v`

UniMath/Algebra/Groups2.v
```diff
@@ -0,0 +1,744 @@
+(** ** Contents
```
nmvdw, line 1: I think a more descriptive name would be better than `Groups2.v`

nmvdw (approved):
I think this is an improvement, but the new files should get more descriptive names.

UniMath/Algebra/Monoids2.v
arnoudvanderleer: I indend to rename this to `Monoids` in the future, but that would be a "breaking change".

rmatthes:
Just one (very late) comment: readability becomes much better with these changes, but for "->", already Company-Coq provides a nice rendering: it shows a "prettified" version when working with ProofGeneral.

PR #1886, opened 2024-05-22
The double bicategory of enriched profunctors

nmvdw:
This PR contains:
- The Verity double bicategory of enriched profunctors
- The pseudo double category of profunctors enriched over a quantale

For the second point, I also developed some theory of posetal categories.


benediktahrens (approved):
LGTM

PR #1887, opened 2024-05-26
Quantale valued relation and coalgebras

nmvdw:
This PR adds two more examples of double categories:
- relations valued in a quantale
- coalgebras for a lax double functor


benediktahrens (approved):
LGTM

PR #1888, opened 2024-05-30
Basic theory of lax extensions

nmvdw:
This PR contains some basic theory of lax extension. 
- Definition of lax extensions
- Equivalence of functors with a lax extension and lax double functors

PR #1889, opened 2024-06-01
Some more theory of double categories

nmvdw:
This PR also contains #1888 . The additional content is:
- flat double categories and examples
- double terminal objects
- graphs and collages of enriched profunctors


benediktahrens (approved):
lgtm

PR #1890, opened 2024-06-04
Original representation theorem

arnoudvanderleer:
* Partially add Scott's original (1980) proof of his representation theorem;
* Define a tactic (in `Combinators`) to help generate `refine` statements for proving equality between lambda terms;
* To make this work, re-add some definitions (like `subst_subst_ax`) to the object files (as opposed to the category files), because the old approach broke the tactic (generating terms like `pr121 T f g` instead of `subst f g`);
* Using these two, rewrite parts of the "algebra to theory" file, deprecating `LambdaTerms.v`;
* Rename `pr` to `var` and `comp` to `subst`, to be more consistent with the lambda calculus view.


rmatthes:
@arnoudvanderleer I checked out the branch related to your PR, and compiled all of AlgebraicTheories with Dune. `make UniMath/CONTENTS.md` does not do anything, and `make check-for-changes-to-CONTENTS.md` is happy. So, this really seems to be a CI problem.

arnoudvanderleer:
@rmatthes Ah, thanks. That reduces the "works on my machine" factor a bit :-)

UniMath/dune
```diff
@@ -4,6 +4,7 @@
 (coq.theory
  (name UniMath)
  (package coq-unimath)
+ (theories Ltac2)
```
benediktahrens, line 7: @arnoudvanderleer : can you explain what the effect of this line is?

UniMath/AlgebraicTheories/CategoryOfRetracts.v
```diff
@@ -0,0 +1,1174 @@
...
+        refine '(_ @ !var_subst _ _ _).
+        exact (!extend_tuple_inl _ _ _).
+      Qed.
+
+      Context (B C : R_ob).
+
+      Lemma exponential_idempotent
+        : exponential_term B C ∘ exponential_term B C = exponential_term B C.
+      Proof.
+        refine '(curry'_compose _ Lβ _ _ @ _).
```
benediktahrens, line 557: What does the backtick do?

benediktahrens (approved):
lgtm

UniMath/dune
arnoudvanderleer: To be honest, I am not 100% sure of the ramifications. I added it, because it allows me to write tactics using the (slightly more modern / advanced / better-behaved) Ltac2 tactic language. I believe it only instructs dune to allow using these tactics (but again, I'm not 100% sure whether this is a complete description). Would you like me to verify this at the coq zulip?

benediktahrens: That would be great.
Background: it could be beneficial, for the long-term maintainability, to keep dependencies to a minimum, and not rely on experimental features. I don't know if any of this applies to Ltac2. Is it "external" in any sense? Is it experimental?

UniMath/AlgebraicTheories/CategoryOfRetracts.v
arnoudvanderleer: It has something to do with Ltac2 syntax. When Ltac2 is imported, one needs to add an apostrophe before the parentheses of a `refine`. I believe this is a shorthand for
`refine (fun () => (curry'_compose _ Lβ _ _ @ _))`
IIRC, the design of Ltac2 changes the types of some tactics to be nullary or any-ary functions instead of values, because they act upon the proof state or something lik that.

arnoudvanderleer: Ah, I believe I found the [reference](https://coq.inria.fr/doc/v8.17/refman/changes.html#:~:text=Ability%20to%20put%20any%20pattern%20in%20binders%2C%20prefixed%20by%20quote%2C%20e.g.%20%22fun%20%27(a%2Cb)%20%3D%3E%20...%22%2C%20%22%CE%BB%20%27(a%2C(b%2Cc))%2C%20...%22%2C%20%22Definition%20foo%20%27(x%2Cy)%20%3A%3D%20...%22.%20It%20expands%20into%20a%20%22let%20%27pattern%20%3A%3D%20...%22) again.

UniMath/dune
arnoudvanderleer: Can you define "external" and "experimental"?
In this case, Ltac2 is only needed for some tactics that generate proofs. Once the proofs are generated, the tactics and the dependency on Ltac2 can in principle be removed. (And one needs to change `refine '(...)` to `refine (...)` in a couple of places).

nmvdw: > Once the proofs are generated, the tactics and the dependency on Ltac2 can in principle be removed.

What do you mean?

nmvdw: I have some questions:
- How local is the dependence on `Ltac2`? If file `A` uses `Ltac2` and file `B` imports `A`, does `B` have to import `Ltac2` as well to use the tactics in `A`?
- What is the exact location where you use `Ltac2`? I only briefly looked at the files, and I don't see where it is used. Based on the documentation, I expect that the only difference between `Ltac1` and `Ltac2` lies in the kind of tactics that you can define?

benediktahrens: > Can you define "external" and "experimental"? 

External: does Ltac2 come with Coq, or is it separate?
Experimental: have the Coq developers promised to take care of Ltac2, and will they consider backwards compatibility when developing it further?

rmatthes: @nmvdw The file AlgebraicTheories/Combinators.v has plenty of explicit references to Ltac2.

arnoudvanderleer: @nmvdw 
> What do you mean?
I mean that the tactic generates a series of `refine` statements that can replace the tactic invocation. That means that the resulting proofs don't have to depend on the tactic anymore, and so the tactic can, in principle, be erased  when the proofs are finished (although for future proofs, it would be nice to still have the tactic around).

arnoudvanderleer: > I have some questions:
> 
> * How local is the dependence on `Ltac2`? If file `A` uses `Ltac2` and file `B` imports `A`, does `B` have to import `Ltac2` as well to use the tactics in `A`?
> * What is the exact location where you use `Ltac2`? I only briefly looked at the files, and I don't see where it is used. Based on the documentation, I expect that the only difference between `Ltac1` and `Ltac2` lies in the kind of tactics that you can define?

* In my experience, if file `A` defines a tactic using `Ltac2`, and file `B` wants to use those tactics, it also needs to import `Ltac2` indeed. If `Ltac2` is not imported, I get a `The reference ... was not found in the current environment`.
* It is used to define the tactics `propagate_subst` and `generate_refine` in `Combinators.v`, and those tactics have subsequently been used in `CategoryOfRetracts.v` and `OriginalRepresentationTheorem.v`.

UniMath/AlgebraicTheories/CategoryOfRetracts.v
arnoudvanderleer: This was the place where I originally found it:
https://github.com/coq/coq/issues/13806

UniMath/dune
arnoudvanderleer: And about the difference between Ltac and Ltac2: I would guess that they are both turing complete, so the only limit to their capacities would be what kind of interfaces have been written for them to act on the proof states. However, maybe the following experience can shed some light on the matter:
 * In the past, I have dabbled with some Ltac tactics, and eventually got some of them to work. In some sense, working with Ltac was easy. However, it was also sometimes easy to make mistakes, because (IIRC) Ltac does not really do much typechecking, and does not have a very advanced typesystem.
 * I wanted to design a modular tactic. A tactic to reduce lambda terms, that would allow one to add clauses later on, as more and more terms were defined. When I asked on the coq zulip how that could be done, people told me that I either could do that with a hint database (I believe that is Ltac) (but that only supports rewrites), or with mutable tactics, using Ltac2.
 * Ltac2 took a bit of getting used to. The language seems to take a bit more code for the same result. However, I also got the feeling that it was harder to make mistakes here. The typesystem is stronger, which helps you catch mistakes earlier. Also, the typesystem makes more sense, because it makes side effects more explicit. A lot of things that were just "tactics" or "values" before, now are functions.

arnoudvanderleer: @benediktahrens [About experimental and external:](https://coq.zulipchat.com/#narrow/stream/237977-Coq-users/topic/How.20is.20Ltac2.20supported.20by.20humans.20and.20systems.3F/near/442870517)
> ltac2 is packaged with Coq (currently the plugin part in coq-core and the rest in coq-stdlib as far as opam is concerned)
> dune wants a flag because it wants dependencies to be explicit
> we will maintain ltac2, and we try to keep backwards compatibility (about the same as the rest of coq I think)
> (source: I am a coq developer and one of the main ltac2 maintainers)

arnoudvanderleer: Does this answer your questions, @benediktahrens @nmvdw ?

UniMath/AlgebraicTheories/CategoryOfRetracts.v
benediktahrens: Answered, thanks a lot.

UniMath/dune
rmatthes: @arnoudvanderleer I have an extra question: https://coq.inria.fr/doc/V8.19.0/refman/proof-engine/ltac2.html#ltac2 clearly says Ltac2 is experimental. But Coq 8.20 will come soon, and it may be considered less experimental there. Do you know of such "upgrades"?

rmatthes:
I identified the problem: making `UniMath/CONTENTS.md` does not notice deletion of files, but this PR deletes two files. Remedy: just delete `UniMath/CONTENTS.md` and regenerate it. This will kick out the two lines associated to those files.

UniMath/dune
arnoudvanderleer: @rmatthes Ah, good that you mention it. I had read over the "experimental" part in the documentation. I am not aware of such upgrades.
However, my code only lightly depends on Ltac2. If things break down, we can remove the dependence entirely without changing any of my proofs (apart from the `refine '(...)` syntax). The reason that I would still like to include the tactics with their dependence in this PR is that if they still work, they make reasoning with the lambda calculus a lot easier.

rmatthes (approved):
A full review would take as much time as for a conference paper. This looks good, does not seem to harm anything else, and the proofs are now in one place instead of referring to an auxiliary file for the tedious stuff. The proposed reasoning style might be used in other places of the library as well.

arnoudvanderleer:
@rmatthes @benediktahrens Can we merge this, or would you like an additional review from, e.g., Niels?

benediktahrens:
@arnoudvanderleer : thanks for the reminder!

PR #1891, opened 2024-06-11
The theory of extensions

arnoudvanderleer:
Formalize the theory algebras `T_n`, algebra pullbacks `F* A` and the theory of extensions `T_A` as a functor from `T`-algebras to algebraic theories. Show that the category of `T_A`-algebras is equivalent to the coslice category under `A`, and show that any algebraic theory morphism `F : S -> T` factors through `S_{F* T_0}`.


arnoudvanderleer:
@benediktahrens Can you maybe review this, or assign a reviewer?

arnoudvanderleer:
Suddenly, I am getting errors along the lines of
```
Error: Error: Universe inconsistency. Cannot enforce UU.u0 <= PartA.cast.u1 because
  PartA.cast.u1 <= PartA.cast.u0 < UU.u0.
```
However, I fail to see how these have anything to do with my changes. Is there an explanation?

benediktahrens:
The error seems to occur on the first file that is not from UniMath/UniMath. Are we passing the correct command line arguments to Coq?

@m-lindgren : do you have an idea?

UniMath/CategoryTheory/coslicecat.v
```diff
@@ -152,3 +153,73 @@ Defined.
 Definition coslice_cat_total : category := total_category coslice_cat_disp.
 
 End coslice_cat_displayed.
+
+(** ** Isos in slice categories *)
+
+Section Isos.
+
+  Local Notation "x / C" := (coslice_cat C x).
```
rmatthes, line 161:
```suggestion
  Local Notation "x / C" := (coslice_cat_total C x).
```
I would suggest to work with the "modern" construction through displayed categories. This file will then still compile but not the rest of this PR.

arnoudvanderleer: What are the pros and cons of using the displayed approach?

rmatthes: The negative side is that one builds on a concept that not everybody assimilated. Otherwise, I only see advantages of a modular definition. Lemmas such as is_precategory_coslice_precat_data and has_homsets_coslice_precat are superfluous, and one can later also use that the coslices come from a displayed category.

rmatthes: It would be good to apply this move to the displayed coslices also in Free.v.

arnoudvanderleer: So would it be a good idea to:
 - Start using the displayed version everywhere?
 - Add a comment to coslicecats that `coslice_cat_total` is the preferred way to go?
 - Make `coslice_cat` a local definition inside a section, to further discourage (but not quite disable it altogether) its usage?

rmatthes: @arnoudvanderleer You are right that it would be good to make thinks clear for everybody in the future. In your case, the change was minimal. And we do not yet exploit the displayed nature of the coslices.

rmatthes: I tried out if the displayed approach to coslices can be beneficial to FreeTheory.v, and I think so now.

nmvdw:
> The error seems to occur on the first file that is not from UniMath/UniMath. Are we passing the correct command line arguments to Coq?
> 
> @m-lindgren : do you have an idea?

We should note the following:
- The error only happens in the checks using Coq dev and not with Coq latest
- The repositories that got affected are SetHITs and TypeTheory, and neither of them use type-in-type

The problem is thus not related to this PR, but the universe checking probably improved in Coq dev, and thus more errors are caught. It is caused by importing foundations.

UniMath/CategoryTheory/coslicecat.v
arnoudvanderleer: @rmatthes I already updated FreeTheory without any problems. However, I wonder in what way the displayed approach to coslices can be beneficial to this file.

rmatthes: @arnoudvanderleer I demonstrated this in a PR to your branch.

arnoudvanderleer: Ah, I am sorry, I overlooked that one.

UniMath/AlgebraicTheories/Examples/ExtensionsTheory.v
```diff
@@ -0,0 +1,649 @@
...
+
+      End Ob.
+
+      Section Mor.
+
+        Context {B B' : algebra T}.
+        Context {F : algebra_morphism A B}.
+        Context {F' : algebra_morphism A B'}.
+        Context {G : algebra_morphism B B'}.
+        Context (HG : F · G = F').
```
rmatthes, line 379: What is the trick behind having this extra equation instead of using the LHS in place of F'? Regardless, the proof using HG below could be by induction on HG - so the proof itself is then as if this extra equation was not there.

arnoudvanderleer: This is part of the definition of a functor from the coslice category `A / algebra_cat T` to the algebra category of the theory of extensions `T_A`. This particular section is about lifting a coslice morphism `G: (B, F) -> (B', F')` (which, by definition, commutes with `F` and `F'`) to a `T_A`-algebra morphism between `B` and `B'`.

rmatthes: Sorry for my uninspired question. But I still see that proof possibility by induction on the supposed equality.

UniMath/CategoryTheory/coslicecat.v
rmatthes: Yesterday, I have made another PR in a similar spirit.

UniMath/AlgebraicTheories/Examples/ExtensionsTheory.v
arnoudvanderleer: I altered the proof using induction now. Although I believe that the gains from this change are minimal: we only can replace a `refine (_ @ maponpaths (fun (x : algebra_morphism _ _) => x a) HG).` by `induction HG.`, so the line count stays the same.

rmatthes: @arnoudvanderleer I did not pretend that this is really easier. But it was curious that the proof does not need flexibility in the supposed proof of the equation.

UniMath/AlgebraicTheories/Examples/ExtensionsTheory.v
```diff
@@ -0,0 +1,627 @@
+(**************************************************************************************************
+
+  The theory of extensions of an algebra
+
+  Given an algebraic theory T and a T-algebra A, we can construct the "theory of extensions" of A
+  as the theory T_A with (T_A)_n = A + T_n as a coproduct of T-algebras. This is functorial in A.
+  The category of algebras of this theory is equivalent to the coslice category (A / T-alg). For
+  example, if T encodes the theory of rings and A is a T-algebra (equivalent to a ring R), then T_A
```
rmatthes, line 8: is this example somewhere in code? would it be easy to do? (sorry for this late comment)

arnoudvanderleer: No problem.

The example is not somewhere in the code. The current extent of the code is:
 - For any category C, with an forgetful-free (with free functor `F`) adjunction to SET, we can define the "free object theory" as `T_n = F({1, ..., n})`. Then any object `A` of `C` can be given a `T`-algebra structure, and this is functorial in `A`. This is in `FreeObjectTheory.v`.
 - For monoids, this functor from `C` to `T`-algebras is an equivalence of categories (actually, this special case is developed independently from `FreeObjectTheory`). This is in `FreeMonoidTheory.v`.
 - The proof that the category of `T_A`-algebras is equivalent to the coslice category under `A` is in `ExtensionsTheory.v`.

Intuitively, the proof in `FreeMonoidTheory` is very easy. The unit and the monoid operation are given by the actions of appropriate elements of the algebraic theory, and the monoid laws are proven like
```
1 * a
= action (x1 * x2) (action 1 (), a)
= action (x1 * x2) (action 1 (a), action x1 (a))
= action (1 * x1) (a)
= action x1 (a)
= a
```
However, this involves a lot of rewriting using the algebraic theory laws, which is harder than one would like. We can do exactly the same construction for rings. This would just be more of the same tedious rewriting.

And then I'm also not sure whether we even have a definition of ring algebras in the library. Although we can just say that a ring algebra is an object in the coslice category.

rmatthes: Thanks a lot. Is there something of your explanations you would like to put somewhere in the comments in addition?

arnoudvanderleer: @rmatthes I think most of this is in my thesis, in the comments or in the code. But do you think there is anything worthwile of adding to those comments?

rmatthes: I had overlooked the FreeObjectTheory.v and FreeMonoidTheory.v files. I now see that the new element in your above comment was mostly the reasoning example and the critical comments on the way the formalization has to do it.

rmatthes (approved):
This development provides code of very good quality.

PR #1893, opened 2024-06-11
Reasoning with foldr1_map

rmatthes:
The idea is to have a lemma as interface instead of going into the details of the encoding of lists.

This is shown to be useful for non-trivial situations in package SubstitutionSystems.

### DIFF
--- a/UniMath/Combinatorics/Lists.v
+++ b/UniMath/Combinatorics/Lists.v
@@ -253,9 +253,7 @@ Defined.
 Lemma foldr1_foldr1_map {A B : UU} (f : B -> B -> B) (b : B) (h : A -> B) (xs : list A) :
   foldr1_map f b h xs = foldr1 f b (map h xs).
 Proof.
-  revert xs.
   set (P := fun xs b' => b' = foldr1 f b (map h xs)).
-  intro xs.
   apply (foldr1_map_ind f b h P).
   - apply idpath.
   - intro a. apply idpath.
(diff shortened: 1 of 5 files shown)


rmatthes:
The problem in satellite largecatmodules is serious. If I cannot fix that, I would have to rewind the latest commit.

rmatthes:
I have made a tiny pull request for largecatmodules that solves the issue on my machine.

UniMath/Combinatorics/Lists.v
```diff
@@ -219,25 +219,52 @@ Proof.
   apply idpath.
 Qed.
 
-Lemma foldr1_foldr1_map {A B : UU} (f : B -> B -> B) (b : B) (h : A -> B) (xs : list A) :
-  foldr1_map f b h xs = foldr1 f b (map h xs).
+Definition foldr1_map_ind {A B : UU} (f : B -> B -> B) (b : B) (h : A -> B) (P : list A -> B -> UU)
+  (P0 : P nil b)
+  (P1 : ∏ a, P (cons a nil) (h a))
+  (P2 : ∏ a1 xs a2 b, P (cons a1 xs) b -> P (cons a2 (cons a1 xs)) (f (h a2) b))
+  (xs : list A) : P xs (foldr1_map f b h xs).
```
benediktahrens, line 226: @rmatthes : is there an intuition you could share with others (and possibly future you?) that might help understand this induction principle?

benediktahrens (approved):
This looks very useful! Could you add a comment helping others to understand and use the induction principle?

UniMath/Combinatorics/Lists.v
rmatthes: a good point - I have now tried to do this in a comment right before the definition

PR #1895, opened 2024-06-19
 Show that under some circumstances, two precomposition functors that together are an adjoint equivalence, are also themselves adjoint equivalences

arnoudvanderleer:


benediktahrens (approved):
This code looks good. Should it be in package `CategoryTheory` instead of `AlgebraicTheories`?

arnoudvanderleer:
> This code looks good. Should it be in package `CategoryTheory` instead of `AlgebraicTheories`?

Thanks. And yes, I suppose so. Just in the root directory, or maybe in Equivalences?

benediktahrens:
> > This code looks good. Should it be in package `CategoryTheory` instead of `AlgebraicTheories`?
> 
> Thanks. And yes, I suppose so. Just in the root directory, or maybe in Equivalences?

@nmvdw ?

nmvdw:
> > This code looks good. Should it be in package `CategoryTheory` instead of `AlgebraicTheories`?
> 
> Thanks. And yes, I suppose so. Just in the root directory, or maybe in Equivalences?

In Equivalences it would fit quite well, right? Throwing everything in the root directory makes the root directory messy

PR #1899, opened 2024-06-28
Formalize the construction of a one-object category from a monoid

arnoudvanderleer:


arnoudvanderleer:
In the process, I removed a lemma from setcategories that did not depend on the category being a setcategory, and was very similar to `path_functor_mor_alt`. However, it seems that this causes `GrothendieckConstruction/IsPullback.v` to hang at line 373.

nmvdw:
> However, it seems that this causes `GrothendieckConstruction/IsPullback.v` to hang at line 373.

That might have been the reason for the lemma

nmvdw:
> > That might have been the reason for the lemma
> 
> Do you have a hypothesis why switching from a lemma about setcategories to a lemma about categories causes a proof to hang for a very long time?

Actually, I just looked at it more closely. I don't think the lemma should be removed. While they are similar, they work at different sides of the path. If we have `F = G`, one expresses `#F f` via `G` and the other in the other way around. If one uses the other one, then one has to add an inverse. This probably breaks the proof in the hanging file, because there some rewritiing is used where the concrete path is used. 

Hence, I suggest to not delete the lemma.

rmatthes:
@arnoudvanderleer What is the purpose of the new file Test.v at the root of the UniMath library? I guess you did not want to commit it here.

rmatthes:
The title of the PR hints to this being the one-dimensional version of the relation between one-object bicategories and monoidal categories. See https://github.com/UniMath/UniMath/blob/master/UniMath/Bicategories/Core/Examples/BicategoryFromMonoidal.v, https://github.com/UniMath/UniMath/blob/master/UniMath/Bicategories/Core/Examples/BicategoryFromWhiskeredMonoidal.v, https://github.com/UniMath/UniMath/blob/master/UniMath/Bicategories/MonoidalCategories/MonoidalFromBicategory.v, and https://github.com/UniMath/UniMath/blob/master/UniMath/Bicategories/MonoidalCategories/WhiskeredMonoidalFromBicategory.v.

arnoudvanderleer:
> > However, it seems that this causes `GrothendieckConstruction/IsPullback.v` to hang at line 373.
> 
> That might have been the reason for the lemma

In hindsight, it seems that this Qed is also very slow on the master branch. When I tested it (n = 1), it seemed that my code even speeds up this part.

arnoudvanderleer:
> Hence, I suggest to not delete the lemma.

I moved it to `Functors.v`, because I fits better in there

arnoudvanderleer:
> @arnoudvanderleer What is the purpose of the new file Test.v at the root of the UniMath library? I guess you did not want to commit it here.

Ah, my bad. This was just an attempt to make Duuk's code work, and it is unrelated to this PR indeed. I removed it now.

arnoudvanderleer:
> The title of the PR hints to this being the one-dimensional version of the relation between one-object bicategories and monoidal categories. See https://github.com/UniMath/UniMath/blob/master/UniMath/Bicategories/Core/Examples/BicategoryFromMonoidal.v, https://github.com/UniMath/UniMath/blob/master/UniMath/Bicategories/Core/Examples/BicategoryFromWhiskeredMonoidal.v, https://github.com/UniMath/UniMath/blob/master/UniMath/Bicategories/MonoidalCategories/MonoidalFromBicategory.v, and https://github.com/UniMath/UniMath/blob/master/UniMath/Bicategories/MonoidalCategories/WhiskeredMonoidalFromBicategory.v.

Ah, I didn't know that that existed, but indeed. It is the one-dimensional version of this. Did you mention this because you think this relation should be documented somewhere?

rmatthes:
I think you do more for this special case. This could therefore warrant a comment with link.

UniMath/AlgebraicTheories/FundamentalTheorem/CommonUtilities/MonoidToCategory.v
```diff
@@ -0,0 +1,313 @@
...
+
+  If we have a monoid, we can construct a one-object category from it. This is the one-dimensional
+  version of what happens in `Bicategories/MonoidalCategories/MonoidalFromBicategory.v`. In fact,
+  this gives a fully faithful functor from monoids to setcategories. Presheaves on such a one-object
+  category are equivalent to sets with a right monoid action.
+
+  Contents
+  1. The functor from monoids to categories [monoid_to_category]
+  1.2. Fully faithful [monoid_to_category_fully_faithful]
+  2. The equivalence between presheaves and monoid actions [monoid_presheaf_action_equivalence]
```
benediktahrens, line 13: How about moving this file into the package `CategoryTheory`? It might be easier to find there.

benediktahrens (approved):
LGTM (modulo moving a file?)

UniMath/AlgebraicTheories/FundamentalTheorem/CommonUtilities/MonoidToCategory.v
arnoudvanderleer: That is a good idea. Will do! Just in the main directory, or maybe `Categories`?

arnoudvanderleer: @nmvdw

PR #1900, opened 2024-06-30
Forest june

JeanGalte:
state of development at the date of the end of my internship with mr Matthes : 
-Development of the SubsitutionSystems/UntypedForest file : add church integers as an example of the calculus, sigma-monoids point of view, until defining initial and final algebras for the calculus, strength usual definitions (following  SubsitutionSystems/STLC_actegorical). 
-Development of Forests file :  add church integers as an example of the calculus, begin of the study of "detyping" process


rmatthes (approved):
I also checked compilation with Coq 8.20 (rc1). Even the satellites TypeTheory and SetHITs compile when activation option `type-in-type`.

PR #1903, opened 2024-07-11
Formalize the original proof of Dana Scott's 1980 representation theorem

### LINKED ISSUE #1910: Weaken assumptions for the accessors of `exponentials`
The accessors in `exponentials` currently assume that a category has `Exponentials`, whereas most of the accessors only need that some object `is_exponentiable`. It would be good to weaken the assumptions (and adapt the places where the accessors are used to this change)

arnoudvanderleer:
Fixes #1910 on the go.


UniMath/AlgebraicTheories/RepresentationTheorem.v
```diff
@@ -584,7 +561,7 @@ Section RepresentationTheorem.
         refine (maponpaths (λ x, pr1 x _ _ • _) (ProductPrCommutes _ _ _ (pow _) _ _ _) @ _).
         apply var_subst.
       - apply idpath.
-    Qed.
+    Time Qed.
```
nmvdw, line 564:
```suggestion
    Qed.
```

How long does this `Qed` take?

nmvdw (commented):
Can you explain:
- Why did you add `Exponentials'`?
- What is the advantage of using `Exponentials'` over `Exponentials`?
- There are possibilities of inconsistenties. For instance, if one defines toposes, then one uses `Exponentials`. An alternative approach would be to use a proper interface. This way the internal representation would stay the same and consistent, while one would still have all desired operations. What is the advantage of this approach over adapting the interface?
- When should one use `Exponentials` instead of `Exponentials'`?

It would also be good if these comments are documented in the code. If not, then on the long run it could create two competing developments of exponentials in the code, which would then not be consistent with each other.

benediktahrens:
I agree with the mild inconveniences of the current definition of exponentials. However, I think they are very minor, and they don't justify the introduction of another definition, or a refactoring of the code.

nmvdw:
> I agree with the mild inconveniences of the current definition of exponentials. However, I think they are very minor, and they don't justify the introduction of another definition, or a refactoring of the code.

A lot of the inconveniences can be mitigated by developing and consistently using a proper interface.

UniMath/AlgebraicTheories/RepresentationTheorem.v
arnoudvanderleer: About 17 seconds.

arnoudvanderleer:
> Can you explain:
> 
> * Why did you add `Exponentials'`?

I can see the disadvantages of using a definition like `Exponentials'`, so for now, I will try to remove it again. The main reason why I added it is the following:

The current definition seems "wrong", because it leads us to curry a function `f : B, A -> C` as `fun b a, f (a, b)`. When trying to show directly that a category has exponential objects, this gives all kinds of quirky definitions. For example:
 * The standard lambda calculus combinator `curry f := abs (abs (app f (x1, x2)))` cannot be used to define exponential objects, instead we need to use `curry' := abs (abs (app f (x2, x1)))` (the same goes for the `uncurry'` operators).
 * At some point, I need an equivalence between `C[U^n, U^U]` and `C[U^(n + 1), U]`. Using the current definition of `Exponentials`, this means that `U^(n + 1) = U x U^n`, so `U^n = U x (U x (... x (U x U)...))`, instead of the usual left associative `U^n = ((...(U x U) x ...) x U) x U`. This leads to definitions like
```coq
  Definition sn_power_projection
    (i : stn (S n))
    : C ⟦ sn_power_object, c ⟧.
  Proof.
    induction (invmap stnweq i) as [i' | i'].
    - exact (
        BinProductPr2 _ _ ·
        ProductPr _ _ _ i'
      ).
    - apply BinProductPr1.
  Defined.
```
Note that `BinProductPr2` occurs in the first branch, and `BinProductPr1` in the second. This is not a major problem and we can work with this, but for me it means that something is conceptually wrong with defining products in this way.

The choice of definition for `U^n` led to more problems down the line. I had a lot of proofs already working with the "wrong" definition, but then in `OriginalRepresentationTheorem` I decided to define n-tuples via pairs as `(((x0, x1), ...), xn)`, which to me seems like a natural idea. However, I realized that I could not relate such n-tuples to properties of the object `U^n`, because the parentheses are the wrong way. That is when I decided that at least `EndomorphismTheory` should work with the "right" adjunction.

This gives have two options: either `EndomorphismTheory` takes the "wrong" adjunction as input, and then converts it using
```coq
  Definition is_exponentiable'_to_is_exponentiable
    (a : C)
    (HF : is_left_adjoint (constprod_functor2 a))
    : is_exponentiable a
    := is_left_adjoint_closed_under_iso _ _ (z_iso_inv (flip_z_iso a)) HF.
```
Since I want to define my objects in `CategoryOfRetracts` and `RepresentationTheorem` using the right adjunction, because it is conceptually better, I have to use
```coq
  Definition is_exponentiable'_to_is_exponentiable
    (a : C)
    (HF : is_left_adjoint (constprod_functor2 a))
    : is_exponentiable a
    := is_left_adjoint_closed_under_iso _ _ (z_iso_inv (flip_z_iso a)) HF.
```
to convert them to the wrong adjunction. This has the disadvantage of adding precomposition with an isomorphism and its inverse to every instance of `φ_adj` and `φ_adj_inv`. If I want to prove properties of objects that are defined using `φ_adj`, I now have to deal with this. I am currently trying to make my proofs work with this approach, and dealing with this currently makes proofs a lot slower for some reason (17 seconds vs 0.1 seconds for one proof) and adds a couple of long, complicated lines:
```coq
      refine (maponpaths
        (λ (x : presheaf_morphism (PO (S n)) _), pr1 x m l)
        (φ_adj_inv_under_iso _ _ _ (flip_z_iso (bin_products_presheaf_cat L) _) _ _ _ _) @ _).
      refine (maponpaths
        (λ x, pr1 ((_ : presheaf_morphism (PO (S n)) _) · x) m l)
        (φ_adj_inv_under_iso _ _ _ _ _ _ _ _) @ _).
      refine (maponpaths (λ (x : presheaf_morphism _ _), pr1 x m _) (assoc _ _ _) @ _).
      refine (maponpaths
        (λ x, pr1 (x · _ : presheaf_morphism (PO (S n)) _) m l)
        (z_iso_after_z_iso_inv (nat_z_iso_pointwise_z_iso
          (BinProduct_of_functors_commutes _ _
            (bin_products_presheaf_cat L)
            (constant_functor _ _ (theory_presheaf L))
            (functor_identity _)
          )
          (PO n))) @ _).
      refine (maponpaths (λ (x : presheaf_morphism (PO (S n)) _), pr1 x m l) (id_left _) @ _).
```

The easy alternative was adding another definition, `exponentials'`, which captures the right kind of adjunction, and meant that I did not have to deal with converting back and forth (and the consequences of that for my definitions).

nmvdw:
>  This has the disadvantage of adding precomposition with an isomorphism and its inverse to every instance of φ_adj and φ_adj_inv. If I want to prove properties of objects that are defined using φ_adj, I now have to deal with this.

If I understand correctly, the inefficiency arises from how the exponentials concretely are defined? By unfolding the definition, you get a more complicated definition?

arnoudvanderleer:
> If I understand correctly, the inefficiency arises from how the exponentials concretely are defined? By unfolding the definition, you get a more complicated definition?

Yes, some definitions use the equivalence of morphisms. For example (here called `hom_weq`):
```coq
  Definition endomorphism_lambda_theory_data
    : lambda_theory_data.
  Proof.
    use make_lambda_theory_data.
    - exact endomorphism_theory.
    - intros n f.
      exact (invmap (hom_weq n) (f · app)).
    - intros n f.
      exact (hom_weq n f · abs).
  Defined.
```
And if we convert between `exponentials'` and `exponentials` and then back again,
`hom_weq f` becomes something like `flip_z_iso x · (z_iso_inv (flip_z_iso x) · hom_weq f)`, and to show that the latter actually equals the former, we need a couple of unpleasant lines of code in every proof dealing with these definitions.

UniMath/CategoryTheory/exponentials.v
```diff
@@ -40,665 +41,633 @@ Require Import UniMath.CategoryTheory.Core.Univalence.
...
+          rewrite id_left.
+          unfold BinProduct_of_functors_ob, constant_functor, functor_identity.
+          cbn.
+          (* is this supposed to be analogous? *)
+          admit.
+      - intro d. show_id_type. (* F-images in source and target of the morphisms *)
+        cbn.
+        admit.
+    *)
+    Abort.
```
benediktahrens, line 617: @arnoudvanderleer : what is the status of this unfinished construction? Are you planning to finish it?

arnoudvanderleer: I am not planning to finish it. Ralph added it a long time ago, I believe, and I merely moved the whole section down, and properly indented its contents.

benediktahrens (approved):

nmvdw:
Is this PR ready to be merged?

benediktahrens:
> Is this PR ready to be merged?

Yes, I think so!

PR #1908, opened 2024-07-17
adds 8.20 to the Coq versions for building UniMath

rmatthes:


rmatthes:
The newly added check with "Coq 8.20-alpha" in fact is based on a version that identifies itself as 
`The Coq Proof Assistant, version 8.20+rc1`.

benediktahrens:
@rmatthes : thanks for looking into this.

Why do you suggest to "no longer [use] Coq dev for satellite CI"?

rmatthes:
> @rmatthes : thanks for looking into this.
> 
> Why do you suggest to "no longer [use] Coq dev for satellite CI"?

This was part of a debugging effort that was solved in https://github.com/UniMath/UniMath/pull/1908/commits/419915862e423f0b0e5495dba6a5bd59bda816ed

rmatthes:
We learn from this first proper CI run that the satellites SetHITs and TypeTheory do not even compile with the upcoming Coq 8.20 (checked through the first release candidate). Of course, the known remedy is to use the flag for type-in-type. The question should be allowed why this flag is now needed. Probably, one cannot import modules compiled with that option without compiling with that option, too.

benediktahrens (approved):

PR #1912, opened 2024-08-05
Add basic definitions and results regarding pre-bilattices

giorgio93p:
Pre-bilattices are structures comprising two lattices (over the same set).

This PR first defines pre-bilattices, bounded prebilattices, and the displayed category of pre-bilattices.

Moreover, it defines two special cases: interlaced and distributive pre-bilattices, proving that distributive pre-bilattices are also interlaced pre-bilattices.

Also, it defines the construction of product pre-bilattices and proves that product pre-bilattices are interlaced and that every interlaced pre-bilattice is isomoprhic to a product pre-bilattice (representation theorem).

The pre-bilattice FOUR is included as an example.


UniMath/OrderTheory/Lattice/Distributive.v
```diff
@@ -59,3 +66,44 @@ Section Bounded.
   Qed.
 
 End Bounded.
+
+Section Properties.
```
nmvdw, line 70: Each lemma in this section can be made opaque by ending them with `Qed` rather than `Defined`.

UniMath/OrderTheory/Lattice/Lattice.v
```diff
@@ -69,6 +73,18 @@ Definition make_lattice {X : hSet} {min max : binop X} : islatticeop min max →
 Definition Lmin {X : hSet} (lat : lattice X) : binop X := pr1 lat.
 Definition Lmax {X : hSet} (lat : lattice X) : binop X := pr1 (pr2 lat).
 
+Lemma isaset_lattice (X : hSet) : isaset (lattice X) .
+Proof.
+Check (isasetbinoponhSet X).
```
nmvdw, line 78:
```suggestion
Check (isasetbinoponhSet X).
```
This `Check` can be removed.

UniMath/OrderTheory/Lattice/Lattice.v
```diff
@@ -1576,6 +1592,10 @@ Defined.
 
 Close Scope multmonoid.
 
+Section dual_lattice.
```
nmvdw, line 1595: It is not really necessary to use a section here, because there is only 1 definition in it. I would suggest to remove `Section dual_lattice.` and `End dual_lattice.`.

UniMath/OrderTheory/Prebilattice/Prebilattice.v
```diff
@@ -0,0 +1,155 @@
...
+    exact(isaset_dirprod (isaset_lattice X) (isaset_lattice X)).
+  Defined.
+
+  Definition prebilattice_transportf {X1 X2 : hSet} (b1 : prebilattice X1) (b2 : prebilattice X2) (p : X1 = X2)
+             (m : meet (transportf prebilattice p b1) ~ meet b2)
+             (j : join (transportf prebilattice p b1) ~ join b2)
+             (c : consensus (transportf prebilattice p b1) ~ consensus b2)
+             (g : gullibility (transportf prebilattice p b1) ~ gullibility b2)
+    : transportf prebilattice p b1 = b2.
+  Proof.
```
nmvdw, line 48: Would it be useful in this proof to use `induction p`? Would that simplify anything? In addition, this definition canb e made opaque by using `Qed` instead of `Defined`.

UniMath/OrderTheory/Prebilattice/Prebilattice.v
```diff
@@ -0,0 +1,155 @@
...
+         --- apply isaprop_islatticeop.
+  Defined.
+
+  Definition prebilattice_eq {X : hSet} (b1 : prebilattice X) (b2 : prebilattice X) (m : meet b1 ~ meet b2) (j : join b1 ~ join b2) (c : consensus b1 ~ consensus b2) (g : gullibility b1 ~ gullibility b2): b1 = b2.
+  Proof.
+    use (@prebilattice_transportf X X b1 b2 (idpath X)); now rewrite idpath_transportf.
+  Defined.
+End Def.
+
+Section Properties .
```
nmvdw, line 72:
```suggestion
Section Properties.
```

UniMath/OrderTheory/Prebilattice/PrebilatticeDisplayedCat.v
```diff
@@ -0,0 +1,106 @@
...
+do 3 (try use make_dirprod);
+rewrite <- (homotweqinvweq f x), <- (homotweqinvweq f y);
+[
+  rewrite (pr1 (p (invmap f x) (invmap f y))) |
+  rewrite (pr12 (p (invmap f x) (invmap f y))) |
+  rewrite (pr122 (p (invmap f x) (invmap f y))) |
+  rewrite (pr222 (p (invmap f x) (invmap f y)))
+];
+now rewrite homotweqinvweq,  homotweqinvweq, homotinvweqweq.
+Defined.
```
nmvdw, line 38:
```suggestion
Qed.
```

UniMath/OrderTheory/Prebilattice/PrebilatticeDisplayedCat.v
```diff
@@ -0,0 +1,106 @@
+(** Displayed category of pre-bilattices *)
+(** Georgios V. Pitsiladis, Aug. 2024 - *)
+Require Import UniMath.OrderTheory.Prebilattice.Prebilattice.
+Require Import UniMath.CategoryTheory.Core.Categories.
```
nmvdw, line 4:
```suggestion
Require Import UniMath.CategoryTheory.Core.Prelude.
```
This way, you also have `is_iso` and `is_z_isomorphism` in scope

UniMath/OrderTheory/Prebilattice/PrebilatticeDisplayedCat.v
```diff
@@ -0,0 +1,106 @@
...
+- use isaset_prebilattice.
+- intros X b1 b2 p1 p2.
+  use prebilattice_eq; intro x; use weqfunextsec; intro y.
+  -- exact (pr1 (p2 x y)).
+  -- exact (pr12 (p2 x y)).
+  -- exact (pr122 (p2 x y)).
+  -- exact (pr222 (p2 x y)).
+Defined.
+
+Definition prebilattice_iso {X1 : hSet} {X2 : hSet} (b1 : prebilattice X1) (b2 : prebilattice X2) (f : weq X1 X2) (p : respects_prebilattice_structure b1 b2 f) : @Isos.iso prebilattice_disp_category (X1,,b1) (X2,,b2).
```
nmvdw, line 73: It tends be nicer to user `z_iso` instead of `iso`. They are defined slightly differently, and in the proof you also go via `z_iso`.

UniMath/OrderTheory/Prebilattice/Interlaced.v
```diff
@@ -0,0 +1,245 @@
...
+
+Section equalities.
+  Definition interlaced_prebilattice_eq {X : hSet} (b1 : interlaced_prebilattice X) (b2 : interlaced_prebilattice X) (m : meet b1 ~ meet b2) (j : join b1 ~ join b2) (c : consensus b1 ~ consensus b2) (g : gullibility b1 ~ gullibility b2): b1 = b2.
+  Proof.
+    use total2_paths_f.
+    - use (prebilattice_eq b1 b2 m j c g).
+    - apply isaprop_is_interlaced.
+  Defined.
+
+  Definition interlaced_prebilattice_transportf {X1 X2 : hSet} (b1 : interlaced_prebilattice X1) (b2 : interlaced_prebilattice X2) (p : X1 = X2)
```
nmvdw, line 56: Would it be possible to simplify this proof by using `induction p`?

nmvdw (approved):
Looks good! Thank you for this contribution. I only have small comments.

UniMath/OrderTheory/Lattice/Distributive.v
giorgio93p: Indeed. Isn't this contrary to UniMath style, however?

nmvdw: `Qed` is permitted by the style. Before the style guide said that `Qed` was not permitted, but that changed in 2021 (see https://github.com/UniMath/UniMath/pull/1428).

UniMath/OrderTheory/Prebilattice/Prebilattice.v
giorgio93p: Observe that `p` is not directly used in the text of the proof of `prebilattice_transportf`. I am not super familiar with HoTT and UniMath, but I interpret this fact as an indication that `induction p` would not simplify much.

As far as I can tell, it is mostly a matter of style:

- either `prebilattice_transportf` is proved first and then `prebilattice_eq` follows as a simple corollary (as is the case now)
- or `prebilattice_eq` is proved first (exactly the same proof as the current one for `prebilattice_transportf` would work; after a little experimentation I could not simplify it further) and then `prebilattice_transportf` follows as a simple corollary using path induction: `Proof. induction p. use prebilattice_eq; assumption. Qed.`

UniMath/OrderTheory/Prebilattice/Interlaced.v
giorgio93p: `induction p` is already used

UniMath/OrderTheory/Prebilattice/Prebilattice.v
nmvdw: Yes, it might be a matter of style. By doing `induction p`, I guess one can avoid the lemma `prebilattice_transportf`.

nmvdw:
I will update the branch (i.e., merge with the latest version of UniMath). If no new errors arise, then it can be merged.

PR #1914, opened 2024-08-17
Add schools to the CI config

arnoudvanderleer:


arnoudvanderleer:
Note that as long as Schools#52, which adds a dune config to the Schools repo, is not merged, this CI step does not really build anything.

PR #1918, opened 2024-08-31
Make two definitions of sets in HSET/Core into aliases ...

### LINKED ISSUE #1092: duplication
... between
```
Definition unitset : hSet := hSetpair unit isasetunit.
```
and
```
Definition unitHSET : HSET.
Proof.
  exists unit.
  abstract (apply isasetunit).
Defined.
```
The second definition should use the first one.

arnoudvanderleer:
... for the definitions of which they are duplicates. Resolves #1092.

### DIFF
--- a/UniMath/CategoryTheory/Categories/HSET/Core.v
+++ b/UniMath/CategoryTheory/Categories/HSET/Core.v
@@ -75,20 +75,14 @@ Notation "'SET'" := hset_category : cat.
 Definition emptyHSET : HSET.
 Proof.
   exists empty.
-  abstract (apply isasetempty).
+  exact isasetempty.
 Defined.
 
-Definition unitHSET : HSET.
-Proof.
-  exists unit.
-  abstract (apply isasetunit).
-Defined.
+Definition unitHSET : HSET
+  := unitset.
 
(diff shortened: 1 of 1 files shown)


benediktahrens (approved):

nmvdw:
It seems like this PR causes some of the satellite repositories to break (i.e., largecatmodules). Error:
```
 bp := bcpS : BinCoproducts SET
  The term "ff" has type
   "∏ (x x' : unit) (x0 x'0 : x = x'), iscontr (x0 = x'0)"
  while it is expected to have type "SET ⟦ bp unitHSET X, ?x' ⟧"
  (cannot unify "pr1hSet (bp unitHSET X)" and "unit").
  ```

nmvdw:
Note: this change might thus also affect SetHITs, because this repository also uses stuff about the category `HSET`.

arnoudvanderleer:
So apparently, this change caused some weird problems: All three of the errors seem to be caused by a `set (ff := fun x => _).` suddenly being something else than expected. This is easily solved by specifying the type of the function that we are looking for: `set (ff :=  fun (x : unit ⨿ _) => _).`

PR #1926, opened 2024-09-05
Clean up rezk_completion, show that the Rezk completion is unique

arnoudvanderleer:


arnoudvanderleer:
@benediktahrens and @nmvdw I have introduced code that uses bicategories into an ancient file. I would like your opinion on this.

Currently, the (bicategory) imports for this new section are at the bottom of the file. If you think that introducing bicategories is okay, I can also merge these imports into the rest of the import statements at the top.

nmvdw:
> @benediktahrens and @nmvdw I have introduced code that uses bicategories into an ancient file. I would like your opinion on this.
> 
> Currently, the (bicategory) imports for this new section are at the bottom of the file. If you think that introducing bicategories is okay, I can also merge these imports into the rest of the import statements at the top.

Generally, we avoid dependencies on bicategories due to compile-time and memory consumption reasons. I think it is good to adhere to this, and theorems that depend on bicategories can be moved there.

nmvdw:
> @nmvdw Oh, I didn't know that. Is that documented somewhere?

I do not know.

> Where do you propose I move this code?

What about `Bicategories/RezkCompletions`?

arnoudvanderleer:
> I do not know.

[I added it](https://github.com/UniMath/UniMath/wiki/About-UniMath)

> What about `Bicategories/RezkCompletions`?

All right. Will do.

arnoudvanderleer:
Oh, apparently I did break something when moving stuff around. I will fix it soon.
I still find it a bit weird that we are reasoning with equality on functors here. That feels "wrong" somehow.

UniMath/Bicategories/RezkCompletions/Uniqueness.v
```diff
@@ -0,0 +1,132 @@
...
+Proof.
+  induction F as [D F].
+  induction F' as [D' F'].
+  cbn in H1.
+  rewrite <- (idtoiso_2_0_isotoid_2_0 univalent_cat_is_univalent_2_0 H1) in H2.
+  induction (isotoid_2_0 univalent_cat_is_univalent_2_0 H1).
+  apply maponpaths.
+  refine (_ @ H2).
+  exact (!functor_identity_right _ _ _).
+Defined.
```
nmvdw, line 53: Do you really want this one to be `Defined` instead of `Qed`? The way in which this is defined, will most likely not be very usable.

UniMath/Bicategories/RezkCompletions/Uniqueness.v
```diff
@@ -0,0 +1,132 @@
...
+    refine (initial_functor_endo_is_identity F' _ _).
+    refine (!functor_assoc _ _ _ _ _ _ _ @ _).
+    refine (maponpaths (λ x, x ∙ _) (initial_functor_arrow_commutes _ _) @ _).
+    apply initial_functor_arrow_commutes.
+  Qed.
+
+  Let η := idtoiso (C := [_, _]) G'_after_G : z_iso (C := [_, _]) (functor_identity D) (G ∙ G').
+  Let ɛ := idtoiso (C := [_, _]) G_after_G' : z_iso (C := [_, _]) (G' ∙ G) (functor_identity D').
+
+  Definition initial_functor_unique
```
nmvdw, line 87: Same here: do you really want this to be defined? In addition, the adjoint equivalence that you construct, is something that one can compute with. However, it is made opaque.

nmvdw (approved):
Looks good, but I have some comments regarding opaqueness.

If I see correctly, there now are files `RezkCompletion.v` and `rezk_completion.v`. This is a bit confusing, because their names are essentially the same. Would a name change be possible? In addition, a directory called `RezkCompletion` might be useful, but maybe not in scope of this PR.

arnoudvanderleer:
> If I see correctly, there now are files `RezkCompletion.v` and `rezk_completion.v`.
You see correctly. However, these files both already existed, although `RezkCompletion` previously only defined `RezkCat`.

In this PR, I tried to maintain the existing contents of the files (for "backwards compatibility", although I am not sure whether that is necessary), as well as have a clearer view of what is specific to this construction of the Rezk completion, and what is true in general for Rezk completions. (Turns out: almost everything works for Rezk completions in general).

> This is a bit confusing, because their names are essentially the same. Would a name change be possible?

I agree. Would it be an idea to either
 * Merge the current `rezk_completion` into `RezkCompletion`;
 * Rename `rezk_completion` to `RezkCompletionConstruction`?

nmvdw:
> > If I see correctly, there now are files `RezkCompletion.v` and `rezk_completion.v`.
> > You see correctly. However, these files both already existed, although `RezkCompletion` previously only defined `RezkCat`.
> 
> In this PR, I tried to maintain the existing contents of the files (for "backwards compatibility", although I am not sure whether that is necessary), as well as have a clearer view of what is specific to this construction of the Rezk completion, and what is true in general for Rezk completions. (Turns out: almost everything works for Rezk completions in general).
> 
> > This is a bit confusing, because their names are essentially the same. Would a name change be possible?
> 
> I agree. Would it be an idea to either
> 
> * Merge the current `rezk_completion` into `RezkCompletion`;
> * Rename `rezk_completion` to `RezkCompletionConstruction`?

If it would be in a directory called `RezkCompletion`, then the name `Construction` would be sufficient.

What do you think, @benediktahrens ?

UniMath/Bicategories/RezkCompletions/Uniqueness.v
arnoudvanderleer: I made it opaque because in this case, we only need G. But if you think it is better to remove the `abstract` (I agree that at least it is easier for viewing), then I will do that.

nmvdw: What about splitting it up? Making a separate definition for the adjoint equivalence that is transparent?

nmvdw: If the proof of the equality will only be `isotoid` applied to the constructed adjoint equivalence, then it would be fine to make this one transparent. If it involves any non-trivial rewriting, then opaque would be the way to go.

nmvdw (approved):

PR #1928, opened 2024-09-07
Algebras for pointed endofunctors and Kelly's transfinite construction

blinkybool:
This is my first contribution to UniMath so feedback is appreciated.

It consists contains the construction of a free algebra for a well-pointed endofunctor and some relationships between well-pointed endofunctors and reflective subcategories.

I did my best to reuse existing definitions, however I haven't used PointedFunctors.v, and instead just have my own `pointed_endofunctor` definition, since I had issues with `F : category_Ptd C` not coercing to a `functor C C`. Using `ptd_obj C` works, but this naming is obscure in my opinion - I would not guess that it's a functor.

I followed `UniMath/CategoryTheory/Monads/MonadAlgebras.v` to implement the definition of the category of algebras for a pointed endofunctor. Everything is much the same, except we have only 1 algebra law, not 2. I'm not sure if I should also be capitalising like `MonadAlg`? Also worth noting is I've called them `pointed_algebras`, which may imply the algebra X has a point (1 --> X) rather than the endofunctor F, but I don't there's a better alternative.

The relationship to reflective subcategories hasn't been made concrete. I use the term `reflective` in comments and the final theorem, but I'm essentially just using a fully-faithful right adjoint in the code.

### DIFF
--- a/UniMath/CategoryTheory/PointedFunctorAlgebras.v
+++ b/UniMath/CategoryTheory/PointedFunctorAlgebras.v
@@ -486,7 +486,7 @@ Lemma free_ptd_alg_mor_restricts {A B : C} (f : A --> B) (i : nat)
     = # (F ^ i) f · colimIn (CC B) i.
 Proof.
   apply colimArrowCommutes.
-Defined.
+Qed.
 
 Definition free_ptd_alg_mor_is_mor {A B : C} (f : A --> B)
   : is_ptd_alg_mor F _ _ (free_ptd_alg_mor f).
@@ -538,10 +538,8 @@ Definition free_ptd_alg : C ⟶ ptd_alg_category F
 
 Definition unit_FF_adjunction : functor_identity C ⟹ free_ptd_alg ∙ forget_ptd_alg F.
 Proof.
-  use make_nat_trans; red; simpl.
-  - exact (λ A, colimIn (CC A) 0).
-  - intros A B f; simpl.
-    apply pathsinv0, (free_ptd_alg_mor_restricts f 0).
+  apply make_nat_trans with (λ A, colimIn (CC A) 0).
+  abstract (intros A B f; apply pathsinv0, (free_ptd_alg_mor_restricts f 0)).
 Defined.
 
 Definition counit_FF_cocone_arrows (X : ptd_alg F) : ∏ i : nat, C ⟦ (F ^ i) X, X ⟧.
@@ -551,7 +549,7 @@ Proof.
   - exact (# F IH · ptd_alg_map F X). (* Necessary to make ɛ an morphism of algebras *)
 Defined.
 
-Definition counit_FF_forms_cocone (X : ptd_alg F)
+Lemma counit_FF_forms_cocone (X : ptd_alg F)
   : forms_cocone (iter_chain_at X) (counit_FF_cocone_arrows X).
 Proof.
   red; intros i _ []; simpl.
@@ -562,25 +560,20 @@ Proof.
   - simpl. rewrite functor_id, id_right. apply ptd_alg_law.
   - simpl. rewrite <- functor_comp, assoc.
     apply (maponpaths (λ x, # F x · _) IHi).
-Defined.
+Qed.
 
 (* The counit has a component at the pointed algebra X which is a morphism of algebras
   F^ω X --> X. This is the underlying morphism in C.
   (Be aware of the implicit coercion of X to it's underlying object in C)
 *)
-Definition counit_FF_map (X : ptd_alg F) : C ⟦ free_ptd_alg_ob X, X ⟧.
-Proof.
-  simpl.
-  apply colimArrow. use make_cocone.
-  + apply counit_FF_cocone_arrows.
-  + apply counit_FF_forms_cocone.
-Defined.
+Definition counit_FF_map (X : ptd_alg F) : C ⟦ free_ptd_alg_ob X, X ⟧
+  := colimArrow _ _ (make_cocone _ (counit_FF_forms_cocone X)).
 
 Definition counit_FF_map_restricts (X : ptd_alg F) (i : nat)
   : colimIn (CC X) i · counit_FF_map X = counit_FF_cocone_arrows X i.
 Proof.
   apply colimArrowCommutes.
-Defined.
+Qed.
 
 Definition counit_FF_mor_is_mor (X : ptd_alg F) :
   is_ptd_alg_mor F (free_ptd_alg X) X (counit_FF_map X).
@@ -629,26 +622,31 @@ Definition counit_FF_adjunction
   : (forget_ptd_alg F ∙ free_ptd_alg) ⟹ (functor_identity (ptd_alg_category F))
   := make_nat_trans _ _ _ counit_FF_adjunction_is_nat_trans.
 
-(* The adjunction witnessing that the free pointed-algebra functor is indeed free *)
-Definition free_forgetful_are_adjoints : are_adjoints free_ptd_alg (forget_ptd_alg F).
+(** This proves free construction is correctly named, it is left adjoint
+  to the forgetful functor. *)
+Theorem free_forgetful_form_adjunction
+  : form_adjunction free_ptd_alg
+      (forget_ptd_alg F)
+      unit_FF_adjunction
+      counit_FF_adjunction.
 Proof.
-  use make_are_adjoints.
-  - exact unit_FF_adjunction.
-  - exact counit_FF_adjunction.
-  - split; red.
-    + intro A. apply ptd_alg_mor_eq, pathsinv0; simpl.
-      apply colim_endo_is_identity. intros i.
-      rewrite assoc.
-      rw_left_comp (free_ptd_alg_mor_restricts (colimIn (CC A) 0) i).
-      rewrite <- assoc.
-      rw_right_comp (counit_FF_map_restricts (free_ptd_alg_ob A) i).
-      set (ɛ := counit_FF_cocone_arrows).
-      induction i; simpl. { apply id_right. }
-      rewrite assoc, <- functor_comp.
-      apply (transportb (λ x, # F x · _ = _) IHi).
-      apply shift_iter_map_restricts.
-    + intro X. simpl. apply (counit_FF_map_restricts X 0).
-Defined.
+  split; red.
+  + intro A. apply ptd_alg_mor_eq, pathsinv0; simpl.
+    apply colim_endo_is_identity. intros i.
+    rewrite assoc.
+    rw_left_comp (free_ptd_alg_mor_restricts (colimIn (CC A) 0) i).
+    rewrite <- assoc.
+    rw_right_comp (counit_FF_map_restricts (free_ptd_alg_ob A) i).
+    set (ɛ := counit_FF_cocone_arrows).
+    induction i; simpl. { apply id_right. }
+    rewrite assoc, <- functor_comp.
+    apply (transportb (λ x, # F x · _ = _) IHi).
+    apply shift_iter_map_restricts.
+  + intro X. simpl. apply (counit_FF_map_restricts X 0).
+Qed.
+
+Definition free_forgetful_are_adjoints : are_adjoints free_ptd_alg (forget_ptd_alg F)
+  := make_are_adjoints _ _ _ _ free_forgetful_form_adjunction.
 
 (* TODO: Involve fully-faithfulness of forget_ptd_alg F to say something
 about reflective subcategories *)
@@ -691,6 +689,6 @@ Proof.
   }
   apply (post_comp_with_z_iso_is_inj p).
   exact (t2 @ ! t1).
-Defined.
+Qed.
 
 (* TODO: prove the category of pointed algebras for (L∙R,,η) "is" D *)
--- a/UniMath/CategoryTheory/PointedFunctorAlgebras.v
+++ b/UniMath/CategoryTheory/PointedFunctorAlgebras.v
@@ -1,24 +1,41 @@
-(** ***************************************************************
+(**************************************************************************************************
 
-Billy Snikkers
-started September 2024
+  PointedFunctorAlgebras
 
-Contents :
+  An algebra for a pointed endofunctor (F, σ : 1 ⟹ F) is an object X with a structure map F(X) ⟶ X
+  which restricts to id on σ_X : X ⟶ F(X).
 
-  - Definitions of pointed endofunctors, pointed algebras (algebras
-    for a pointed endofunctor), well-pointed endofunctors
+  This file contains the transfinite construction (due to Kelly) of the free-algebra for a
+  *well-pointed*, ω-cocontinuous endofunctor, and the free-forgetful adjunction. There are also
+  connections to reflective subcategories (idempotent monads).
 
-  - The transfinite construction of the free-algebra for a well-pointed
-    endofunctor (due to Kelly), and the free-forgetful adjunction.
+  References:
+  * `nLab: transfinite construction of free algebras https://ncatlab.org/nlab/show/transfinite+construction+of+free+algebras
+  * `Max Kelly, A unified treatment of transfinite constructions for free algebras, free monoids, colimits, associated sheaves, and so on https://www.cambridge.org/core/journals/bulletin-of-the-australian-mathematical-society/article/unified-treatment-of-transfinite-constructions-for-free-algebras-free-monoids-colimits-associated-sheaves-and-so-on/FE2E25E4959E4D8B4DE721718E7F55EE
+  
+  Contents
+  1. Definitions
+  1.1. Pointed endofunctors
+  1.2. Well-pointedness
+  1.3. Pointed endofunctor algebras (algebras for a pointed endofunctor)
+  1.4. The category of pointed endofunctor algebras [ptd_alg_category] [forget_ptd_alg]
+  2. Properties of algebras for well-pointed endofunctors
+  2.1. Structure maps are iso-inverses [well_pointed_point_is_z_iso_at_algebra] 
+  2.2. Any morphism is a morphism of algebras [well_pointed_mor_is_algebra_mor] 
+  2.3. The forgetful functor is fully-faithful [well_pointed_forgetul_fully_faithful] 
+  3. The transfinite construction
+  3.1. The directed colimit A --> F(A) --> F^2(A) --> ... F^ω(A)
+  3.2. The shift map F(F^ω A) --> F^ω A [shift_iter_map]
+  3.3. The shift map forms a pointed algebra [shift_iter_map_forms_ptd_alg]
+  3.4. The free functor C ⟶ ptd_alg_category F [free_ptd_alg]
+  3.5. The free-forgetful adjunction [free_forgetful_form_adjunction]
+  4. Reflective subcategories give well-pointed endofunctors
+    [well_pointed_of_fully_faithful_right_adjoint]
 
-  - Connnection between well-pointed endofunctors and reflective
-    subcategories (idempotent monads).
+  started September 2024
+  Billy Snikkers
 
-References:
-
-* `nLab: transfinite construction of free algebras <https://ncatlab.org/nlab/show/transfinite+construction+of+free+algebras>`_
-* `Max Kelly, A unified treatment of transfinite constructions for free algebras, free monoids, colimits, associated sheaves, and so on <https://www.cambridge.org/core/journals/bulletin-of-the-australian-mathematical-society/article/unified-treatment-of-transfinite-constructions-for-free-algebras-free-monoids-colimits-associated-sheaves-and-so-on/FE2E25E4959E4D8B4DE721718E7F55EE>`_
-******************************************************************)
+ **************************************************************************************************)
 
 Require Import UniMath.Foundations.PartD.
 Require Import UniMath.CategoryTheory.Core.Categories.
@@ -44,7 +61,11 @@ Tactic Notation "rw_left_comp" constr(e) :=
 Tactic Notation "rw_right_comp" constr(e) :=
   apply (transportb (λ x, _ · x = _) ltac:(apply e)).
 
-(* A pointed endofunctor F comes with a natural transformation 1 ⟹ F *)
+(** 1. Definitions *)
+
+(** 1.1. Pointed endofunctors *)
+
+(** A pointed endofunctor F comes with a natural transformation 1 ⟹ F *)
 Definition pointed_endofunctor (C : category) : UU := ∑ F : C ⟶ C, functor_identity C ⟹ F.
 
 #[reversible=no] Coercion pointed_endofunctor_endofunctor
@@ -54,24 +75,24 @@ Definition make_pointed_endofunctor {C : category}
     (F : C ⟶ C) (σ : functor_identity C ⟹ F)
   : pointed_endofunctor C := (F,,σ).
 
-(* The *point* of a pointed endofunctor *)
+(** The *point* of a pointed endofunctor *)
 Definition point {C : category} (F : pointed_endofunctor C) : functor_identity C ⟹ F := pr2 F.
 
-(* Easier to use than nat_trans_ax since no identity functor is mentioned *)
+(** Easier to use than nat_trans_ax since no identity functor is mentioned *)
 Definition point_naturality {C : category} (F : pointed_endofunctor C) {x y : C} (f : x --> y)
   : (f · point F y = point F x · # F f).
 Proof.
   apply (nat_trans_ax (point F)).
 Qed.
 
-(*
+(** 1.2. Well-pointedness
   A well-pointed functor (F,,σ) satisfies Fσ = σF, i.e. for every c : C, the two
   ways of constructing a morphism F c --> F (F c) coincide.
 *)
 Definition well_pointed {C : category} (F : pointed_endofunctor C) : UU
   := ∏ A : C, point F (F A) =  #F (point F A).
 
-(**
+(** 1.3. Pointed endofunctor algebras (algebras for a pointed endofunctor)
   Endofunctors, pointed-endofunctors, monads all have associated categories
   of algebras, consisting of a structure map obeying some subset of the
   laws for the algebras of a monad. These algebra-categories must be disambiguated,
@@ -93,14 +114,15 @@ Definition make_ptd_alg_data
   : ∏ X : C, (F X --> X) → ptd_alg_data
   := tpair (λ X, F X --> X).
 
+(** This coercion is convenient for this file. Be aware of it. *)
 #[reversible=no] Coercion ptd_alg_carrier (X : ptd_alg_data) : C := pr1 X.
 
 Definition ptd_alg_map (X : ptd_alg_data) : F X --> X := pr2 X.
 
 Definition is_ptd_alg (X : ptd_alg_data) : UU := (σ X) · (ptd_alg_map X) = identity X.
 
-(* A pointed algebra for a pointed endofunctor is an algebra (a : F X --> X)
-  for the endofunctor which is a retraction of the point map (σ_X · a = id_X)
+(** A pointed algebra for (F,σ) is a pair (a : F X --> X)
+  which is a retraction of the point map, i.e., σ_X · a = id_X
 *)
 Definition ptd_alg : UU := ∑ X : ptd_alg_data, is_ptd_alg X.
 
@@ -112,7 +134,7 @@ Definition make_ptd_alg (X : C) (s : F X --> X) (r : σ X · s = identity X)
 
 Definition ptd_alg_law (X : ptd_alg) : σ X · (ptd_alg_map X) = identity X := pr2 X.
 
-(** Data for the category of algebras for the pointed endofunctor F, following FunctorAlgebras.v *)
+(** 1.4. The category of pointed endofunctor algebras [ptd_alg_category] [forget_ptd_alg] *)
 
 Section ptd_alg_precategory_data.
 
@@ -129,7 +151,7 @@ Definition make_ptd_alg_mor {X Y : ptd_alg} (f : X --> Y) (p : is_ptd_alg_mor X 
   : X --> Y := pr1 f.
 
 Definition ptd_alg_mor_commutes {X Y : ptd_alg} (f : ptd_alg_mor X Y)
-  : ptd_alg_map X · f = #F f · ptd_alg_map Y  := pr2 f.
+  : ptd_alg_map X · f = #F f · ptd_alg_map Y := pr2 f.
 
 Definition ptd_alg_mor_id (X : ptd_alg) : ptd_alg_mor X X.
 Proof.
@@ -199,7 +221,7 @@ Qed.
 Definition ptd_alg_category : category
   := make_category ptd_alg_precat has_homsets_pointed_alegbra.
 
-(* The functor which forgets the pointed algebra stucture on the underlying object *)
+(** The functor which forgets the pointed algebra stucture on the underlying object *)
 Definition forget_ptd_alg : ptd_alg_category ⟶ C.
 Proof.
   use make_functor.
@@ -211,17 +233,17 @@ Defined.
 
 End PointedAlgebraCategory.
 
-(*
-The notion of algebra for a pointed endofunctor F involves adding
-*structure* to a carrier object X : C, i.e. a map F X --> X which is a retraction
-of the point at X, σ_X : X --> F X.
-Here we discover that when F is well-pointed, such a map is also a section of
-σ_X, and so this structure degenerates to a *property* of X, i.e. is σ_X iso?
+(** 2. Properties of algebras for well-pointed endofunctors
 
-In fact, this makes the category of pointed algebras for F into a reflective
-subcategory of C. Moreover, any reflective subcategory induces an idempotent
-monad, which is in particular, a well-pointed endofunctor, and the algebras
-for this well-pointed endofunctor recover the reflective subcategory.
+The notion of algebra for a pointed endofunctor F involves adding *structure* to a carrier object
+X : C, i.e. a map F X --> X which is a retraction of the point at X, σ_X : X --> F X.
+Here we discover that when F is well-pointed, such a map is also a section of σ_X, and so this
+structure degenerates to a *property* of X, i.e. is σ_X iso?
+
+In fact, this makes the category of pointed algebras for F into a reflective subcategory of C.
+We will see later, that any reflective subcategory induces an idempotent monad, which is in
+particular, a well-pointed endofunctor, and the algebras for this well-pointed endofunctor recover
+the reflective subcategory.
 *)
 
 Section AlgebrasWellPointed.
@@ -232,26 +254,25 @@ Context (H : well_pointed F).
 
 Let σ := point F.
 
-(*
-If X is a pointed algebra for F (well-pointed), then σ_X : X --> F X is iso.
+(** 2.1. Structure maps are iso-inverses [well_pointed_point_is_z_iso_at_algebra]
+If X is a pointed algebra for (F, σ) (well-pointed), then σ_X : X --> F X is iso.
 *)
 Definition well_pointed_point_is_z_iso_at_algebra (X : ptd_alg F) : is_z_isomorphism (σ X).
 Proof.
-  use make_is_z_isomorphism.
-  - exact (ptd_alg_map F X).
-  - split.
-    (* σ_X ⋅ s = id_X is the axiom for pointed algebras *)
-    + apply ptd_alg_law.
-    (* s · σ_X = id_{FX} follows from naturality + well-pointedness + law for pointed algebras *)
-    + rewrite point_naturality.
-      rewrite H.
-      refine (! functor_comp F _ _ @ _).
-      rewrite <- functor_id.
-      apply maponpaths.
-      apply ptd_alg_law.
+  apply make_is_z_isomorphism with (ptd_alg_map F X).
+  split.
+  (* σ_X ⋅ s = id_X is the axiom for pointed algebras *)
+  - apply ptd_alg_law.
+  (* s · σ_X = id_{FX} follows from naturality + well-pointedness + law for pointed algebras *)
+  - rewrite point_naturality.
+    rewrite H.
+    refine (! functor_comp F _ _ @ _).
+    rewrite <- functor_id.
+    apply maponpaths.
+    apply ptd_alg_law.
 Qed.
 
-(* There is at most one pointed algebra structure on a given X : C. *)
+(** There is at most one pointed algebra structure on a given X : C. *)
 Corollary well_pointed_has_ptd_alg_isaprop (A : C)
   : isaprop (∑ a : F A --> A, (σ A) · a = identity A).
 Proof.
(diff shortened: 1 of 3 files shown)


nmvdw:
Error:
```
--- checking every proof file is listed in one of the packages ---
error: *** file UniMath/CategoryTheory/PointedFunctorAlgebras.v not listed in appropriate file UniMath/*/.package/files
```

To solve this, you need to list the file `PointedFunctorAlgebras.v` in `CategoryTheory/.package/files`

UniMath/CategoryTheory/PointedFunctorAlgebras.v
```diff
@@ -0,0 +1,673 @@
...
+  ways of constructing a morphism F c --> F (F c) coincide.
+*)
+Definition well_pointed {C : category} (F : pointed_endofunctor C) : UU
+  := ∏ A : C, point F (F A) =  #F (point F A).
+
+(**
+  Endofunctors, pointed-endofunctors, monads all have associated categories
+  of algebras, consisting of a structure map obeying some subset of the
+  laws for the algebras of a monad. These algebra-categories must be disambiguated,
+  since a monad is a pointed endofunctor is an endofunctor. The terminology
```
nmvdw, line 78: > These algebra-categories must be disambiguated, since a monad is a pointed endofunctor is an endofunctor.

I don't follow this sentence.

UniMath/CategoryTheory/PointedFunctorAlgebras.v
```diff
@@ -0,0 +1,673 @@
...
+(* A pointed algebra for a pointed endofunctor is an algebra (a : F X --> X)
+  for the endofunctor which is a retraction of the point map (σ_X · a = id_X)
+*)
+Definition ptd_alg : UU := ∑ X : ptd_alg_data, is_ptd_alg X.
+
+#[reversible=no] Coercion ptd_alg_data_from_ptd_alg (X : ptd_alg) : ptd_alg_data := pr1 X.
+
+Definition make_ptd_alg (X : C) (s : F X --> X) (r : σ X · s = identity X)
+  : ptd_alg
+  := (tpair (λ X, F X --> X) X s,, r).
```
nmvdw, line 107: Instead of `tpair`, you can also use `,,`, which, I think, is more readable. Would it be possible to write `X ,, s ,, r` here?

UniMath/CategoryTheory/PointedFunctorAlgebras.v
```diff
@@ -0,0 +1,673 @@
...
+
+#[reversible=no] Coercion mor_from_ptd_alg_mor {X Y : ptd_alg} (f : ptd_alg_mor X Y)
+  : X --> Y := pr1 f.
+
+Definition ptd_alg_mor_commutes {X Y : ptd_alg} (f : ptd_alg_mor X Y)
+  : ptd_alg_map X · f = #F f · ptd_alg_map Y  := pr2 f.
+
+Definition ptd_alg_mor_id (X : ptd_alg) : ptd_alg_mor X X.
+Proof.
+  exists (identity X).
```
nmvdw, line 129: What about either `use make_ptd_alg` or
```suggestion
  refine (make_ptd_alg (identity X) _).
```

UniMath/CategoryTheory/PointedFunctorAlgebras.v
```diff
@@ -0,0 +1,673 @@
...
+  abstract (unfold is_ptd_alg_mor;
+            rewrite functor_id, id_right, id_left;
+            apply idpath).
+Defined.
+
+Definition ptd_alg_mor_comp (X Y Z : ptd_alg) (f : ptd_alg_mor X Y) (g : ptd_alg_mor Y Z)
+  : ptd_alg_mor X Z.
+Proof.
+  exists (f · g).
```
nmvdw, line 138: Same here. It is better to use `make_ptd_alg`.

blinkybool:
Okay, I could keep squashing these issues but could you confirm first that this should indeed live in a new file and the file name is good? Out of `PointedFunctorAlgebras`, `PointedEndofunctorAlgebras`, `PointedAlgebras`, I picked `PointedFunctorAlgebras` to make it like `FunctorAlgebras`.

UniMath/CategoryTheory/PointedFunctorAlgebras.v
```diff
@@ -0,0 +1,673 @@
...
+            rewrite assoc;
+            rewrite ptd_alg_mor_commutes;
+            rewrite <- assoc;
+            rewrite ptd_alg_mor_commutes;
+            rewrite functor_comp, assoc;
+            apply idpath).
+Defined.
+
+Definition precategory_ptd_alg_ob_mor : precategory_ob_mor
+  := (ptd_alg,, ptd_alg_mor).
```
nmvdw, line 149: You can use `make_precategory_ob_mor` here

UniMath/CategoryTheory/PointedFunctorAlgebras.v
```diff
@@ -0,0 +1,673 @@
...
+
+Definition precategory_ptd_alg_data : precategory_data
+  := (precategory_ptd_alg_ob_mor,, ptd_alg_mor_id,, ptd_alg_mor_comp).
```
nmvdw, line 152: You can use `make_precategory_data` here.

UniMath/CategoryTheory/PointedFunctorAlgebras.v
```diff
@@ -0,0 +1,673 @@
...
+Defined.
+
+Lemma is_precategory_precategory_ptd_alg_data : is_precategory (precategory_ptd_alg_data F).
+Proof.
+  apply make_is_precategory; intros; apply ptd_alg_mor_eq.
+  - apply id_left.
+  - apply id_right.
+  - apply assoc.
+  - apply assoc'.
+Defined.
```
nmvdw, line 178:
```suggestion
Qed.
```
It is a bit nicer to use `Qed` here, because you won't use these. In addition, you can use `make_is_precategory_one_assoc` so that you only need to prove one associativity law.

UniMath/CategoryTheory/PointedFunctorAlgebras.v
```diff
@@ -0,0 +1,673 @@
...
+
+Lemma has_homsets_pointed_alegbra : has_homsets ptd_alg_precat.
+Proof.
+  intros X Y.
+  apply (isofhleveltotal2 2).
+  - apply homset_property.
+  - intros f.
+    apply isasetaprop.
+    apply homset_property.
+Defined.
```
nmvdw, line 190:
```suggestion
Qed.
```

UniMath/CategoryTheory/PointedFunctorAlgebras.v
```diff
@@ -0,0 +1,673 @@
...
+
+Definition ptd_alg_category : category := (ptd_alg_precat,, has_homsets_pointed_alegbra).
```
nmvdw, line 192: You can use `make_category` here.

UniMath/CategoryTheory/PointedFunctorAlgebras.v
```diff
@@ -0,0 +1,673 @@
...
+
+(* The functor which forgets the pointed algebra stucture on the underlying object *)
+Definition forget_ptd_alg : ptd_alg_category ⟶ C.
+Proof.
+  use make_functor.
+  - use make_functor_data.
+    + simpl. exact (ptd_alg_carrier F).
+    + simpl. intros X Y. exact (mor_from_ptd_alg_mor F).
+  - split; red; intros; apply idpath.
```
nmvdw, line 201:
```suggestion
  - abstract (split; red; intros; apply idpath).
```

UniMath/CategoryTheory/PointedFunctorAlgebras.v
```diff
@@ -0,0 +1,673 @@
...
+Definition well_pointed_point_z_iso_at_algebra (X : ptd_alg F) : z_iso X (F X).
+Proof.
+  use make_z_iso.
+  - exact (σ X).
+  - exact (ptd_alg_map F X).
+  - split.
+    (* σ_X ⋅ s = id_X is the axiom for pointed algebras *)
+    + apply ptd_alg_law.
+    (* s · σ_X = id_{FX} follows from naturality + well-pointedness + law for pointed algebras *)
+    + rewrite point_naturality.
```
nmvdw, line 239: This part can be put in an `abstract`.

UniMath/CategoryTheory/PointedFunctorAlgebras.v
```diff
@@ -0,0 +1,673 @@
...
+Proof.
+  set (point_X_iso := well_pointed_point_z_iso_at_algebra X).
+  apply (pre_comp_with_z_iso_is_inj point_X_iso); simpl.
+  rewrite assoc, assoc.
+  rw_left_comp (ptd_alg_law F X). apply pathsinv0.
+  rw_left_comp (! point_naturality F f).
+  rewrite <- assoc.
+  rw_right_comp (ptd_alg_law F Y).
+  rewrite id_left, id_right; apply idpath.
+Defined.
```
nmvdw, line 270:
```suggestion
Qed.
```

UniMath/CategoryTheory/PointedFunctorAlgebras.v
```diff
@@ -0,0 +1,673 @@
...
+  - intros [a ar].
+    exists (make_ptd_alg F A a ar).
+    apply idpath.
+  - intros [X h].
+    induction h.
+    exists (ptd_alg_map F X).
+    apply ptd_alg_law.
+  - intros [a p]; apply idpath.
+  - intros [X h]; induction h; apply idpath.
+Defined.
```
nmvdw, line 301:
```suggestion
Qed.
```

UniMath/CategoryTheory/PointedFunctorAlgebras.v
```diff
@@ -0,0 +1,673 @@
...
+Context (F_well_pointed : well_pointed F). 
+
+(* Since F is well-pointed, the chain morphisms are all points, i.e. F^i σ = σ F^i *)
+Lemma iter_chain_mor_is_point (i : nat) (A : C) : iter_chain_mor i A = σ ((F ^ i) A).
+Proof.
+  apply pathsinv0.
+  induction i.
+  - exact (idpath _).
+  - exact (F_well_pointed ((F ^ i) A) @ maponpaths (#F) IHi).
+Defined.
```
nmvdw, line 431:
```suggestion
Qed.
```

UniMath/CategoryTheory/PointedFunctorAlgebras.v
```diff
@@ -0,0 +1,673 @@
...
+  rewrite assoc.
+  (* α_A^i · σ_{F^ω A} · s = σ_{F^i A} · F (α_A^i) · s *)
+  rewrite point_naturality.
+  rewrite <- assoc.
+  (* ... = σ_{F^i A} · α_A^(1+i) *)
+  rewrite shift_iter_map_restricts'.
+  simpl. rewrite <- iter_chain_mor_is_point. (* well-pointedness used here! *)
+  (* ... = α_A^i *)
+  apply (colimInCommutes (CC A) i (1+i) (idpath _)).
+Defined.
```
nmvdw, line 454:
```suggestion
Qed.
```

UniMath/CategoryTheory/PointedFunctorAlgebras.v
```diff
@@ -0,0 +1,673 @@
...
+(*
+  The free functor action on morphisms, given by the canonical map between the colimits
+  NOTE: this is the underlying morphism in C, not in the category of algebras
+*)
+Definition free_ptd_alg_mor {A B : C} (f : A --> B)
+  : C ⟦ free_ptd_alg_ob A, free_ptd_alg_ob B ⟧.
+Proof.
+  apply colimOfArrows with (λ i, # (F ^ i) f).
+  intros i _ []; simpl.
+  induction i.
```
nmvdw, line 471: You can put both cases under induction `i` under an `abstract`.

UniMath/CategoryTheory/PointedFunctorAlgebras.v
```diff
@@ -0,0 +1,673 @@
...
+  apply colimArrowCommutes.
+Defined.
+
+Definition free_ptd_alg_data : functor_data C (ptd_alg_category F).
+Proof.
+  exists free_ptd_alg_ob.
+  intros A B f.
+  exists (free_ptd_alg_mor f).
+
+  apply (colim_mor_eq (F_CC A)); simpl.
```
nmvdw, line 489: The part below can be made opaque (either via an abstract or a separate lemma that is made `Qed`)

UniMath/CategoryTheory/PointedFunctorAlgebras.v
```diff
@@ -0,0 +1,673 @@
...
+  simpl.
+  rewrite functor_comp, <- assoc.
+  apply cancel_precomposition.
+  apply pathsinv0, shift_iter_map_restricts.
+Defined.
+
+Definition free_ptd_alg : C ⟶ ptd_alg_category F.
+Proof.
+  apply (make_functor free_ptd_alg_data).
+  split.
```
nmvdw, line 506: I would recommend making a separate lemma for the functor laws

UniMath/CategoryTheory/PointedFunctorAlgebras.v
```diff
@@ -0,0 +1,673 @@
...
+End TransfiniteConstruction.
+
+(*
+We've constructed the reflective subcategory given by the pointed algebras
+of a well-pointed endofunctor.
+Conversely, we can take *any* reflective subcategory of C and extract a
+well-pointed endfunctor on C (in fact, an idempotent monad), whose
+pointed algebras form the "same" reflective subcategory.
+*)
+Theorem well_pointed_of_reflective_subcat
```
nmvdw, line 645: It is also possible to add argument `(L := left_functor adj)` and `(R := right_functor adj)` to this theorem.  So like this:
```coq
Theorem well_pointed_of_reflective_subcat
               {C D : category} (adj : adjunction C D)
               (L := left_functor adj)
```

nmvdw (approved):
This is a nice contribution and the code looks quite good! I only have minor comments. 

Generally, it would be good to make more opaque (using `Qed` or abstract), such a proofs that two morphisms `f` and `g` are equal. Such proofs are never computed, and by making them opaque, Coq won't unfold them. This saves memory and could decrease compile time.

In addition, most of the comments are written using `(*`. However, I think those might not be printed using Coqdoc, so it would be better to use `(**`)

nmvdw:
> Okay, I could keep squashing these issues but could you confirm first that this should indeed live in a new file and the file name is good? Out of `PointedFunctorAlgebras`, `PointedEndofunctorAlgebras`, `PointedAlgebras`, I picked `PointedFunctorAlgebras` to make it like `FunctorAlgebras`.

Yes. I think it is a good idea to put it in a new file, and the file name is fitting.

nmvdw:
Error:
```
--- checking that CONTENTS.md is up-to-date ---
test -z "`git diff UniMath/CONTENTS.md`"
make: *** [Makefile:445: check-for-changes-to-CONTENTS.md] Error 1
```
Try deleting `CONTENTS.md` and then using `make`. After this, you can commit and push the file `CONTENTS.md`

UniMath/CategoryTheory/PointedFunctorAlgebras.v
blinkybool: It's saying "an X is a Y is a Z", which is just a compact way of saying "an X is a Y and a Y is a Z"

UniMath/CategoryTheory/PointedFunctorAlgebras.v
```diff
@@ -0,0 +1,673 @@
...
+  apply colim_mor_eq; intros i. do 2 rewrite assoc.
+  rw_left_comp (free_ptd_alg_mor_restricts f i).
+  rewrite <- assoc.
+  rw_right_comp (counit_FF_map_restricts Y i).
+  rewrite (counit_FF_map_restricts X i).
+  set (ɛ := counit_FF_cocone_arrows).
+
+  induction i; simpl. { rewrite id_left. apply id_right. }
+  rewrite assoc. rewrite <- functor_comp.
+  apply (transportb (λ x, # F x · _ = _) IHi).
```
arnoudvanderleer, line 599: Here again an instance of creative usage of `transportb`

UniMath/CategoryTheory/PointedFunctorAlgebras.v
```diff
@@ -0,0 +1,673 @@
...
+      set (ɛ := counit_FF_cocone_arrows).
+      induction i; simpl. { apply id_right. }
+      rewrite assoc, <- functor_comp.
+      apply (transportb (λ x, # F x · _ = _) IHi).
+      apply shift_iter_map_restricts.
+    + intro X. simpl. apply (counit_FF_map_restricts X 0).
+Defined.
+
+Corollary are_adjoints_FF : are_adjoints free_ptd_alg (forget_ptd_alg F).
+Proof. exact free_forgetful_adjunction. Defined.
```
arnoudvanderleer, line 629: What is the role of this corollary? It seems to merely repeat the lemma above, right?

arnoudvanderleer (commented):
Overall, really nice! Especially for a first PR. However,
 - As Niels pointed out, [there is a lot of things that can be made opaque](https://github.com/UniMath/UniMath/wiki/On-opaqueness)
 - In one or two places, you can use notation for things.
 - It may be nice to use constructors in a couple of places, because that makes the code more resilient against changes in the specific definitions of types.
 - I am curious about your decisions in one or two places.

UniMath/CategoryTheory/PointedFunctorAlgebras.v
blinkybool: With (X,,s,,r) I get this
```
The term "X,, s,, r" has type "∑ y, ?P y" while it is expected to have type
 "ptd_alg" (cannot unify "ob C" and "ptd_alg_data").
```
and ((X,,s),,r) gives me
```
The term "(X,, s),, r" has type "∑ y, ?P y" while it is expected to have type
 "ptd_alg" (cannot unify "∑ y, C ⟦ F X, y ⟧" and "ptd_alg_data").
```
which is why I used tpair. Is there a fix?

UniMath/CategoryTheory/PointedFunctorAlgebras.v
blinkybool: I think you mean make_ptd_alg_mor? (which I'd need to add).
Should I update CategoryTheory/Monads/MonadAlgebras the same way? This code is copy-pasted from there with modifications.

blinkybool: Same is true for the make_precategory_ob_mor and make_precategory_data comment

UniMath/CategoryTheory/PointedFunctorAlgebras.v
blinkybool: I've changed this to prove `is_z_isomorphism (σ X)` with a Qed, because this is a key lemma and it'd be a shame to not be able to step through it.

UniMath/CategoryTheory/PointedFunctorAlgebras.v
nmvdw: Hmm, I didn't expect that. Maybe `tpair` is best, because it fails at some inference.

UniMath/CategoryTheory/PointedFunctorAlgebras.v
nmvdw: > I think you mean make_ptd_alg_mor? (which I'd need to add). Should I update CategoryTheory/Monads/MonadAlgebras the same way? This code is copy-pasted from there with modifications.

No need. That is not part of this PR.

UniMath/CategoryTheory/PointedFunctorAlgebras.v
```diff
@@ -0,0 +1,673 @@
...
+Proof.
+  apply (nat_trans_ax (point F)).
+Defined.
+
+(*
+  A well-pointed functor (F,,σ) satisfies Fσ = σF, i.e. for every c : C, the two
+  ways of constructing a morphism F c --> F (F c) coincide.
+*)
+Definition well_pointed {C : category} (F : pointed_endofunctor C) : UU
+  := ∏ A : C, point F (F A) =  #F (point F A).
```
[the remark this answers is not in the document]
blinkybool: I just defined it to be easily applicable for how I wanted to use it. Well-pointedness is subsumed by idempotent monads (that's essentially the overall content of this file), which is a monad for which T\eta = \etaT (but there are several equivalent definitions https://ncatlab.org/nlab/show/idempotent+monad). I don't think idempotent monads are defined in UniMath.

If you think the whisker definition is better I'm happy to change it.

UniMath/CategoryTheory/PointedFunctorAlgebras.v
```diff
@@ -0,0 +1,673 @@
...
+
+Definition ptd_alg_mor_commutes {X Y : ptd_alg} (f : ptd_alg_mor X Y)
+  : ptd_alg_map X · f = #F f · ptd_alg_map Y  := pr2 f.
+
+Definition ptd_alg_mor_id (X : ptd_alg) : ptd_alg_mor X X.
+Proof.
+  exists (identity X).
+  abstract (unfold is_ptd_alg_mor;
+            rewrite functor_id, id_right, id_left;
+            apply idpath).
```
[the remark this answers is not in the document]
blinkybool: Other UniMath files with `now` compile fine for me, but whenever I try add it to my own code, it either complains about the comma in the rewrite, or it will say "The reference now was not found in the current environment."
Could you provide a correct usage I could try??

UniMath/CategoryTheory/PointedFunctorAlgebras.v
arnoudvanderleer: Ah, yes, I also ran into this whilst trying whether something would work. `ptd_alg_data` cannot be constructed as a pair, so would it be an idea to make a constructor for that?

UniMath/CategoryTheory/PointedFunctorAlgebras.v
arnoudvanderleer: However, if you want, you can do this in a new PR :innocent:
You'll find that a lot of the code of Unimath is not up to standard anymore. By making sure new code is good, by updating old code when we are working on it and by sometimes doing a "housekeeping sweep", we are trying to improve this.

UniMath/CategoryTheory/PointedFunctorAlgebras.v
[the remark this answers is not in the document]
arnoudvanderleer: Hm... I can imagine that the whisker version is slightly less applicable in elementary proofs, although I can also imagine that it may be useful when reasoning about abstract functors and natural transformations.
I think I like the whisker version slightly better (because that ties into stuff that I know best), but if one is not familiar with whiskering, it can also be a barrier.
In the end, it is your call. I merely tried to point out that this equality comes from an equality of natural transformations (but you knew that already). If you decide to keep it this way, maybe add a comment that mentions the whiskering?

UniMath/CategoryTheory/PointedFunctorAlgebras.v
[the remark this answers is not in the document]
arnoudvanderleer: Oooh, I see what is going wrong.
`now` is in `UniMath.MoreFoundations.Tactics.` If you add an import for that to your file, you can either do
```coq
  abstract (unfold is_ptd_alg_mor; now rewrite functor_id, id_right, id_left).
```
or
```coq
  abstract now (unfold is_ptd_alg_mor; rewrite functor_id, id_right, id_left).
```
I am surprised that so much of your code compiles with just `Foundations.PartD`. In code I am working on, I always automatically add imports for `Foundations.All` and `MoreFoundations.All`, because stuff I use is usually scattered throughout those files.

UniMath/CategoryTheory/PointedFunctorAlgebras.v
```diff
@@ -0,0 +1,673 @@
...
+  exists (free_ptd_alg_mor f).
+
+  apply (colim_mor_eq (F_CC A)); simpl.
+  intros i.
+  rewrite assoc, assoc.
+  rw_left_comp shift_iter_map_restricts.
+  refine (free_ptd_alg_mor_restricts f (S i) @ _).
+  change (colimIn (F_CC A) i) with (# F (colimIn (CC A) i)).
+  rewrite <- functor_comp.
+  apply (transportb (λ x, _ = # F x · _) (free_ptd_alg_mor_restricts f i)).
```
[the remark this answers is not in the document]
blinkybool: It's annoying to use transport and would prefer using rewrite if it wasn't so broken in UniMath (can it be fixed?). I guess you're suggesting to trade the = for an @ and maponpaths. Maybe this is an established pattern but isn't this obviously more complicated to understand (and an extra path induction)? Instead of just rewriting a subterm of an equality, you're turning one equality into another and then concatenating.

UniMath/CategoryTheory/PointedFunctorAlgebras.v
blinkybool: A refactoring remnant from when the previous lemma was an `adjunction`. Removing it.

UniMath/CategoryTheory/PointedFunctorAlgebras.v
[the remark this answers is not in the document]
nmvdw: `rewrite` works the same in UniMath as in normal Coq, and I don't think there is a way to improve it. 

The usage of `maponpaths` is clearer than `transport`, I think. For instance, one can go for something like:
```
etrans.
{
   apply maponpaths.
   apply p.
}
```
This way you can use `maponpaths` to find the right location in the goal and then you apply the path.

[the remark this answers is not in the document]
arnoudvanderleer: I think in Unimath it is much more common to use `@` for concatenating paths, and `maponpaths` for making them more specific, than `transportf`, which is used for transporting between types in general.

blinkybool:
@arnoudvanderleer 
> * In one or two places, you can use notation for things.

Could you point out where? I'm not that great at the Notation command.

UniMath/CategoryTheory/PointedFunctorAlgebras.v
```diff
@@ -0,0 +1,673 @@
...
+    endofunctor (due to Kelly), and the free-forgetful adjunction.
+
+  - Connnection between well-pointed endofunctors and reflective
+    subcategories (idempotent monads).
+
+References:
+
+* `nLab: transfinite construction of free algebras <https://ncatlab.org/nlab/show/transfinite+construction+of+free+algebras>`_
+* `Max Kelly, A unified treatment of transfinite constructions for free algebras, free monoids, colimits, associated sheaves, and so on <https://www.cambridge.org/core/journals/bulletin-of-the-australian-mathematical-society/article/unified-treatment-of-transfinite-constructions-for-free-algebras-free-monoids-colimits-associated-sheaves-and-so-on/FE2E25E4959E4D8B4DE721718E7F55EE>`_
+******************************************************************)
```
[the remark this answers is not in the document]
blinkybool: Will come back to fix my comments another time.

arnoudvanderleer:
> Could you point out where? I'm not that great at the Notation command.

Oh, no, I meant the things I already pointed out: using existing notation for things. Mostly the `,,` instead of `tpair`.

UniMath/CategoryTheory/PointedFunctorAlgebras.v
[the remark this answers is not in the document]
blinkybool: > rewrite works the same in UniMath as in normal Coq, and I don't think there is a way to improve it.

What I meant is that rewrite frequently fails for me, even when I make everything fully explicit. Even when I unhide all the coercions, there are many times when it will say "Found no subterm matching <subterm> in the current goal", when that subterm is obviously there. That's why I made the tactic notations `rw_left_comp` and `rw_right_comp`.

For an example, see `well_pointed_mor_is_algebra_mor` and replace `rw_left_comp (ptd_alg_law F X).` with
```
unfold σ.
Set Printing Coercions.
rewrite (ptd_alg_law F X). (* fails *)
```

Anecdotally, there are times when a rewrite will fail if I put a simpl above it, even when the simpl changes nothing (not even with `Set Printing Coercions.`)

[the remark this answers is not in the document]
arnoudvanderleer: I share the experience of coq not recognizing that equal things are equal.
In the example you mention, it seems to be caused by implicit arguments of the coercions :sweat_smile: 
The left hand side of `ptd_alg_law` is
```coq
@compose (precategory_data_from_precategory (category_to_precategory C))
    (@ptd_alg_carrier C F (@ptd_alg_data_from_ptd_alg C F X))
    (@functor_on_objects
       (precategory_ob_mor_from_precategory_data
          (precategory_data_from_precategory (category_to_precategory C)))
       (precategory_ob_mor_from_precategory_data
          (precategory_data_from_precategory (category_to_precategory C)))
       (functor_data_from_functor (precategory_data_from_precategory (category_to_precategory C))
          (precategory_data_from_precategory (category_to_precategory C))
          (@pointed_endofunctor_endofunctor C F))
       (@ptd_alg_carrier C F (@ptd_alg_data_from_ptd_alg C F X)))
    (@ptd_alg_carrier C F (@ptd_alg_data_from_ptd_alg C F X))
    (@nat_trans_data_from_nat_trans_funclass
       (precategory_data_from_precategory (category_to_precategory C))
       (precategory_data_from_precategory (category_to_precategory C))
       (functor_identity_data (precategory_data_from_precategory (category_to_precategory C)))
       (functor_data_from_functor (precategory_data_from_precategory (category_to_precategory C))
          (precategory_data_from_precategory (category_to_precategory C))
          (@pointed_endofunctor_endofunctor C F))
       (@point C F) (@ptd_alg_carrier C F (@ptd_alg_data_from_ptd_alg C F X)))
    (@ptd_alg_map C F (@ptd_alg_data_from_ptd_alg C F X))
```
whereas the goal is
```coq
(@compose (precategory_data_from_precategory (category_to_precategory C))
         (@functor_on_objects
            (precategory_ob_mor_from_precategory_data
               (precategory_data_from_precategory (category_to_precategory C)))
            (precategory_ob_mor_from_precategory_data
               (precategory_data_from_precategory (category_to_precategory C)))
            (functor_data_from_functor
               (precategory_data_from_precategory (category_to_precategory C))
               (precategory_data_from_precategory (category_to_precategory C))
               (functor_identity (precategory_data_from_precategory (category_to_precategory C))))
            (@ptd_alg_carrier C F (@ptd_alg_data_from_ptd_alg C F X)))
         (@functor_on_objects
            (precategory_ob_mor_from_precategory_data
               (precategory_data_from_precategory (category_to_precategory C)))
            (precategory_ob_mor_from_precategory_data
               (precategory_data_from_precategory (category_to_precategory C)))
            (functor_data_from_functor
               (precategory_data_from_precategory (category_to_precategory C))
               (precategory_data_from_precategory (category_to_precategory C))
               (@pointed_endofunctor_endofunctor C F))
            (@ptd_alg_carrier C F (@ptd_alg_data_from_ptd_alg C F X)))
         (@ptd_alg_carrier C F (@ptd_alg_data_from_ptd_alg C F X))
         (@nat_trans_data_from_nat_trans_funclass
            (precategory_data_from_precategory (category_to_precategory C))
            (precategory_data_from_precategory (category_to_precategory C))
            (functor_data_from_functor
               (precategory_data_from_precategory (category_to_precategory C))
               (precategory_data_from_precategory (category_to_precategory C))
               (functor_identity (precategory_data_from_precategory (category_to_precategory C))))
            (functor_data_from_functor
               (precategory_data_from_precategory (category_to_precategory C))
               (precategory_data_from_precategory (category_to_precategory C))
               (@pointed_endofunctor_endofunctor C F))
            (@point C F) (@ptd_alg_carrier C F (@ptd_alg_data_from_ptd_alg C F X)))
         (@ptd_alg_map C F (@ptd_alg_data_from_ptd_alg C F X)))
```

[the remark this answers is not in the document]
blinkybool: What did you do to print this? I only know about `Set Printing Coercions` and `Unset Printing Notations`.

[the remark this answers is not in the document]
arnoudvanderleer: There is also `Set Printing Implicit`. But in this case, I think we need all three of them, and my version of coq won't remember all three at the same time, so I used `Set Printing All` (which I only found out about this morning).
And then manually replaced a couple of parts with `...` for clarity.

UniMath/CategoryTheory/PointedFunctorAlgebras.v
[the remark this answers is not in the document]
blinkybool: Is it appropriate to have my name at the top? I see it in some files but not others.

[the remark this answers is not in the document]
arnoudvanderleer: This varies throughout the library. Personally, I don't do it, because I think the library should be a collaborative effort, where over time more and more people work on the same files. On the other hand, many people do add their name, which makes sense from an intellectual property point of view, I think.
So it's up to you.

UniMath/CategoryTheory/PointedFunctorAlgebras.v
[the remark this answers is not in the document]
blinkybool: I changed to the whisker version.

UniMath/CategoryTheory/PointedFunctorAlgebras.v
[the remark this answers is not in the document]
blinkybool: I made use of `now` and `easy`, but I have to remark that `now` is an odd name. I don't see at all how it suggests what it does.

UniMath/CategoryTheory/PointedFunctorAlgebras.v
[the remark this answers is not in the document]
blinkybool: This really takes the "assistant" out of "proof assistant".

Is this a problem outside of UniMath as well, or is it a result of design decisions (like implict args and coercions) that I've made or that exist in the UniMath library? Before now I assumed it was

blinkybool:
I've finished what I wanted to change for now, but I have some leftover concerns.

- Does it matter that I'm not reusing pointed functors from PointedFunctors.v?
- Are names like `counit_FF_adjunction` (and others from that section) okay? I don't have a good understanding of namespacing and how/when to use `Local`, so I'm not sure if such a generally applicable name could cause problems.
- Should I merge recent changes from master, or will someone else do that when they merge the PR?
- If I were to add more to this file (along the same topic), do I make a completely new branch and PR or do I make another PR on the same branch?

arnoudvanderleer:
> * Does it matter that I'm not reusing pointed functors from PointedFunctors.v?

For now, it is fine that you don't. Can you explain a bit why you are not reusing it? (Depending on that, maybe it would be interesting to experiment with reusing PointedFunctors after this branch has been merged)

> * Should I merge recent changes from master, or will someone else do that when they merge the PR?

Please merge them now, so the CI can run and we can merge :-)

> * If I were to add more to this file (along the same topic), do I make a completely new branch and PR or do I make another PR on the same branch?

It's up to you, but probably a new branch (or delete this branch and create a new one with the same name, both locally and on github). This PR is probably going to be squash-merged, so git won't recognize that the commits on this branch are the same as the new merge commit on master. If you try to merge master (with the squashed merge commit) into this branch again, everything you did on this branch is going to give a merge conflict.
So start with a fresh branch (the name is up to you) from master.

Or do a rebase onto master (so tell git "forget commits a ... q on this branch. From now on, the parent of commit r is the squash commit on master"). I did this a couple of times when my PR had not been merged yet, but I needed it for a new branch. I just branched from my PR branch, and when the PR was merged, rebased onto master.

blinkybool:
> For now, it is fine that you don't. Can you explain a bit why you are not reusing it? (Depending on that, maybe it would be interesting to experiment with reusing PointedFunctors after this branch has been merged)

I mentioned it in the first message of this PR. I thought I should just show how I wanted the definition, and was expecting to be told to update PointedFunctors.v with my definition, or something else. I didn't like how my file looked with `category_Ptd C` or `ptd_obj C` in place of `pointed_endofunctor C` and `ptd_pt C F` in place of `point F`. I think that other file was written for a context where the entire category of pointed endofunctors is the focus, rather than one fixed pointed endofunctor.

UniMath/CategoryTheory/PointedFunctorAlgebras.v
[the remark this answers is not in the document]
arnoudvanderleer: Oh, I agree.

arnoudvanderleer (approved):
Looking good.

arnoudvanderleer:
> * Are names like `counit_FF_adjunction` (and others from that section) okay? I don't have a good understanding of namespacing and how/when to use `Local`, so I'm not sure if such a generally applicable name could cause problems.

Hm... I think in general it is hard to come up with good names (and "good names" is a bit of a subjective idea). Also, I am no expert on namespacing in coq. Personally, I'd prefix things in the section that you mention with the name of the adjunction/construction, so something like `free_ptd_alg_counit` and `free_ptd_alg_unit` (or rather `free_pointed_algebra_counit` and `free_pointed_algebra_unit`. I have a slight dislike for abbreviations, because when chained they have a tendency to become very cryptic). Although there are probably good arguments for and against doing that.

arnoudvanderleer:
> I mentioned it in the first message of this PR. I thought I should just show how I wanted the definition, and was expecting to be told to update PointedFunctors.v with my definition, or something else. I didn't like how my file looked with `category_Ptd C` or `ptd_obj C` in place of `pointed_endofunctor C` and `ptd_pt C F` in place of `point F`. I think that other file was written for a context where the entire category of pointed endofunctors is the focus, rather than one fixed pointed endofunctor.

Ah, I can see what you mean. I just realized that your definitions exactly match those in `PointedFunctors`. In that case, I would propose to merge this PR as it is, although we incur some technical debt with that, and to solve this duplication in a subsequent PR. In that PR, we will let your file reuse as much as possible from PointedFunctors, and potentially renaming a couple of the definitions in `PointedFunctors`.

blinkybool:
I've changed the names in the adjunction section accordingly. Is there some way for me to merge it now or is that for someone with write access to do? Do I need to squash the commits first?

nmvdw:
> I've changed the names in the adjunction section accordingly. Is there some way for me to merge it now or is that for someone with write access to do? Do I need to squash the commits first?

Only those with write access can merge this PR. I didn't have time to pay attention to this PR in the previous days. Is it ready to be merged?

blinkybool:
Yes

nmvdw:
> Yes

Sure. I will merge it today.

UniMath/CategoryTheory/PointedFunctorAlgebras.v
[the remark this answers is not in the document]
nmvdw: The only thing that could affect the functionality of rewrite, is that we use iterated sigma types more compared to other libraries. However, I am not sure whether this actually has that effect, and I did no experiments to verify this.

blinkybool:
@nmvdw whoops, looks like I accidentally committed the UniMath.wiki repo at the top level of this repo before you merged.

nmvdw:
> @nmvdw whoops, looks like I accidentally committed the UniMath.wiki repo at the top level of this repo before you merged.

Can you make a new PR where you remove that file?

PR #1930, opened 2024-09-09
Move RezkCompletion and rezk_completion to the folder RezkCompletions

### LINKED ISSUE #1927: `RezkCompletion.v` and `rezk_completion.v`
Because of historical reasons, there are now two files, with names that only differ in capitalization. Currently, `rezk_completion` contains the specific construction via the Yoneda embedding, whereas `RezkCompletion` is about Rezk completions (fully faithful and essentially surjective functors to a univalent category).

How about we put them together in the folder `CategoryTheory/RezkCompletion`, and rename `rezk_completion` to `Construction`?

What do you think, @benediktahrens ?

_Originally posted by @nmvdw in https://github.com/UniMath/UniMath/issues/1926#issuecomment-2333572193_

arnoudvanderleer:
I used the plural `RezkCompletions` for the folder and file, because I think the subpackage is about objects of this form in general. Of course, the completion is unique, but here we do not know that yet.
Resolves #1927


nmvdw (approved):

benediktahrens (approved):

benediktahrens:
I think using the singular would be slightly better, but I am happy for this PR to be merged nevertheless.

arnoudvanderleer:
If we also merge UniMath/TypeTheory#248, this should work again.

arnoudvanderleer:
@benediktahrens Can we merge this then?

Issue #1896, opened 2024-06-22
CI in trouble with Coq 8.21+alpha


rmatthes:
As of writing, Coq latest in our CI runs is still 8.19.1, but Coq dev jumped from 8.20 to 8.21+alpha.
The problems occur in the satellites TypeTheory and SetHITs.
This became manifest for PR #1891, see the discussion there.

arnoudvanderleer:
The problem you are stating is not entirely clear to me. Can you please clarify whether
 - The version of coq in our CI is "wrong" or "outdated";
 - You'd like to add another version of coq to our CI, because the current versions are not sufficient in some sense;
 - You'd like someone to locally test UniMath on a coq version that is not in CI;
 - You'd like someone to fix the satellites to make sure that they work with a version of coq that is not in CI;
 - A combination of the above?

rmatthes:
A rather quick fix was implemented in PR #1908 and improved in PR #1909. The PR that were merged afterwards triggered 17 CI runs each. The crucial element is that 8.20 is no longer the development version but not the latest version either. In this "limbo state", it was important to run CI with the release candidate, but preferably without hard-coding its name. Currently, 8.20 is hard-coded but at least not the release candidate ordinal. I consider the "limbo state" the one that has the highest risks of problems that should be reported to the Coq developers. For the satellites that do not pass the CI, the question was if the Coq developers wanted that behavior. I have not pursued this question.

arnoudvanderleer:
Closing this, as it seems resolved. Feel free to reopen if you disagree.

PR #1935, opened 2024-09-11
Remove the CI run for 8.20, since coq latest is now at 8.20

### LINKED ISSUE #1934: Adapt the CI
Coq 8.20 is out: https://coq.discourse.group/t/coq-8-20-0-is-out/2425, so the latest version of Coq is 8.20 now. This means that the CI is testing for Coq 8.20 twice, which is unnecessary.

arnoudvanderleer:
Resolves #1934

### DIFF
--- a/.github/workflows/build-unimath.yml
+++ b/.github/workflows/build-unimath.yml
@@ -46,7 +46,7 @@ jobs:
       matrix:
         os: [ubuntu-22.04]
         # https://github.com/coq-community/docker-coq/wiki#ocaml-versions-policy
-        coq-version: ['latest', '8.20', 'dev']
+        coq-version: ['latest', 'dev']
         # (the following is obsolete:) coq-version: [8.16] or [latest, 8.16] (when 8.17 is released)
         ocaml-version: [default]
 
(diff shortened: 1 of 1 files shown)


nmvdw (approved):

rmatthes:
Which version of Coq is now being used for latest - is it really 8.20 or still its release candidate? But in both cases, the current CI setting will have a duplicate entry.

nmvdw:
> Which version of Coq is now being used for latest - is it really 8.20 or still its release candidate? But in both cases, the current CI setting will have a duplicate entry.

Coq 8.20 is the latest version now. It came out today.

arnoudvanderleer:
@rmatthes Note that every CI run prints the versions that it runs:
```
Print versions
  + (script @ line 9) $ opam --version
  2.2.1
  + (script @ line 10) $ opam exec -- dune --version
  3.13.1
  + (script @ line 11) $ opam exec -- coqc --version
  The Coq Proof Assistant, version 8.20.0
  compiled with OCaml 4.13.1
  + (script @ line 12) $ endGroup
```

rmatthes:
Thank you. Yesterday, I heard that the images for CI were not yet available. But they seem to be now.

arnoudvanderleer:
It was useful to have 8.20 explicitly because coq dev was at 8.21 while latest was still at 8.19 (see #1896 )
It is of course conceivable that this will not be the last time that there is a gap between dev and latest, but for now we are fine.

PR #1936, opened 2024-09-13
Algebraic geometry - up to the definition of schemes

### LINKED ISSUE #1316: algebraic geometry

arnoudvanderleer:
Resolves #1316, insofar as that has not yet been done in #1385.


arnoudvanderleer:
So it turns out (as we may have suspected already) that the category of is univalent ^^

arnoudvanderleer:
Note that this just adds the basic definition of a scheme. I expect the package needs a lot of work before we can actually start proving anything real.

UniMath/AlgebraicGeometry/CategoryOfOpens.v
```diff
@@ -0,0 +1,105 @@
+(**************************************************************************************************
+
+  The category of open subsets
```
nmvdw, line 3: Does this functor give rise to an indexed category on `Top`? So to a pseudo functor `Top -> Cat`?

See: https://github.com/UniMath/UniMath/tree/master/UniMath/CategoryTheory/IndexedCategories

UniMath/AlgebraicGeometry/Topology.v
```diff
@@ -1,70 +0,0 @@
-(* Additional definitions and facts from topology. Maybe this should go in UniMath.Topology. *)
```
nmvdw, line 1: What happened to this file?

UniMath/AlgebraicGeometry/PresheafedSpaces.v
```diff
@@ -0,0 +1,585 @@
...
+  2. The embedding of the functor category into a fiber
+    [functor_to_fiber_presheaf_functor] [functor_to_fiber_presheaf_functor_fully_faithful]
+  3. Univalence [is_univalent_presheafed_space_cat]
+  4. Types and accessors [presheafed_space] [presheafed_space_morphism]
+  5. Restriction of a presheafed space to an open subset [presheafed_space_restriction]
+
+ **************************************************************************************************)
+Require Import UniMath.Foundations.All.
+Require Import UniMath.MoreFoundations.All.
+Require Import UniMath.Bicategories.Core.Bicat.
```
nmvdw, line 28: If we add a dependency on `Bicategories`, we should notify the Coq developers to remove `AlgebraicGeometry` from the CI (due to memory consumption).

UniMath/AlgebraicGeometry/PresheafedSpaces.v
```diff
@@ -0,0 +1,585 @@
...
+      exact (functor_opp (continuous_to_functor f) ∙ Xdata ⟹ Ydata).
+  - use tpair.
+    + intros X Xdata.
+      refine (nat_trans_comp _ _ _ _ (lunitor (C := bicat_of_cats) Xdata)).
+      refine (post_whisker _ _).
+      refine (nat_trans_comp _ _ _ (op_nt (inv_from_z_iso (continuous_to_functor_identity _))) _).
+      apply (inv_from_z_iso (C := [_^op, _^op])).
+      apply (z_iso_from_nat_z_iso (homset_property _)).
+      exact (op_triangle_nat_z_iso _).
+    + intros X Y Z f g Xdata Ydata Zdata fdata gdata.
```
nmvdw, line 67: There is a slight disadvantage to this definition: the unitors and associators add unnecessary identities to the pointwise morphism of the natural transformation. So, defining this transformation by hand might make it simpler to use.

UniMath/AlgebraicGeometry/PresheafedSpaces.v
```diff
@@ -0,0 +1,585 @@
...
+      refine (nat_trans_comp _ _ _ _ (rassociator (C := bicat_of_cats) (a := (open_category Z)^op) (b := (open_category Y)^op) _ _ _)
+      ).
+      refine (post_whisker _ Xdata).
+      refine (nat_trans_comp _ _ _ (op_nt (inv_from_z_iso (continuous_to_functor_compose _ _))) _).
+      apply (inv_from_z_iso (C := [_^op, _^op])).
+      apply (z_iso_from_nat_z_iso (homset_property _)).
+      exact (functor_comp_op_nat_z_iso _ _).
+Defined.
+
+Local Lemma aux1
```
nmvdw, line 79: The name can be improved. In addition, this is a transport lemma and those often are reusable.

UniMath/AlgebraicGeometry/PresheafedSpaces.v
```diff
@@ -0,0 +1,585 @@
...
+  - intro P.
+    apply (nat_trans_eq (homset_property _)).
+    intro x.
+    apply id_right.
+  - intros P Q R f g.
+    refine (_ @ !aux1 _ _ _ _).
+    apply (nat_trans_eq (homset_property _)).
+    intro x.
+    cbn.
+    refine (maponpaths (λ x, x · _) (id_right _) @ _).
```
nmvdw, line 232: Does `rewrite id_right` not work in this proof? Especially something like `rewrite !id_right.`

UniMath/AlgebraicGeometry/PresheafedSpaces.v
```diff
@@ -0,0 +1,585 @@
...
+  := pr1 f.
+
+Definition presheafed_space_morphism_to_nat_trans
+  {D : category}
+  {X Y : presheafed_space D}
+  (f : presheafed_space_morphism X Y)
+  : functor_opp (continuous_to_functor f) ∙ presheafed_space_presheaf X ⟹ presheafed_space_presheaf Y
+  := pr2 f.
+
+(** TODO : Move this section to a separate file, or not? *)
```
nmvdw, line 369: `TODO` have a tendency of remaining TODOs by being forgotten. So, it would be good to find a suitable place for general stuff about subtypes. Perhaps in `MoreFoundations`?

UniMath/AlgebraicGeometry/RingedSpaces.v
```diff
@@ -0,0 +1,192 @@
...
+Require Import UniMath.CategoryTheory.Core.Functors.
+Require Import UniMath.CategoryTheory.Core.Univalence.
+Require Import UniMath.CategoryTheory.DisplayedCats.Constructions.
+Require Import UniMath.Topology.Topology.
+
+Local Open Scope cat.
+Local Open Scope subtype.
+Local Open Scope open.
+
+(* TODO: Show that the restriction of a ringed space is again a ringed space *)
```
nmvdw, line 33: Is this TODO a necessary part that is currently missing? Or is it more a 'nice thing to have'?

UniMath/AlgebraicGeometry/RingedSpaces.v
```diff
@@ -0,0 +1,192 @@
...
+  Context (X : presheafed_space commring_category).
+
+  Let F (U : Open (T := X)) : commring := presheafed_space_presheaf X U.
+
+  Definition restriction {U V : Open} (H : U ⊆ V) : ringfun (F V) (F U) := #(presheafed_space_presheaf X) H.
+
+  Definition restrict (A : hsubtype Open) (f : F (⋃ A)) (U : A) : F (pr1 U) :=
+    restriction (contained_in_union_open U) f.
+
+  Definition has_locality
```
nmvdw, line 48: Is `has_locality` an hProp?

UniMath/AlgebraicGeometry/RingedSpaces.v
```diff
@@ -0,0 +1,192 @@
...
+          intro U.
+          exact (pr2 f U @ !pr2 f' U).
+      + intro t.
+        refine (pr1 (isaset_carrier_subset _ (λ _, make_hProp _ _) f f' t _)).
+        apply impred_isaprop.
+        intro.
+        apply setproperty.
+  Qed.
+
+  Definition has_sheaf : UU
```
nmvdw, line 106: The name `has_sheaf` is confusing. What about `isasheaf`?

UniMath/AlgebraicGeometry/Schemes.v
```diff
@@ -0,0 +1,100 @@
...
+Require Import UniMath.CategoryTheory.Core.Univalence.
+Require Import UniMath.CategoryTheory.DisplayedCats.Constructions.
+Require Import UniMath.Topology.Topology.
+
+Local Open Scope cat.
+
+(* TODO: Show that a scheme is not only a ringed space, but also a *locally* ringed space *)
+(* TODO: Show that a presheafed space is a ringed space if it is locally isomorphic to an affine scheme *)
+
+Section Schemes.
```
nmvdw, line 32: Is there a reason for this section? There are no arguments being shared between the identifiers?

UniMath/AlgebraicGeometry/SheavesOfRings.v
```diff
@@ -1,70 +0,0 @@
-Require Import UniMath.Algebra.RigsAndRings.
```
nmvdw, line 1: Just for clarity, the contents of this file has been incorporated into `RingedSpaces.v`?

UniMath/Topology/Topology.v
```diff
@@ -5,101 +5,117 @@
 Require Import UniMath.Foundations.All.
```
nmvdw, line 5: The diffs of this file are a bit confusing. Can you give a short description on what happened in this file?

nmvdw (approved):
Looks good. I have some minor comments.

UniMath/AlgebraicGeometry/CategoryOfOpens.v
arnoudvanderleer: I think so! I will formalize this in the coming week.

UniMath/AlgebraicGeometry/Topology.v
arnoudvanderleer: I upstreamed it.

UniMath/AlgebraicGeometry/PresheafedSpaces.v
arnoudvanderleer: Hm... I believe in this case I just need the unitors and associators of functors. Does this indeed introduce such a large memory consumption, and would there be a way to avoid using bicategories?

UniMath/AlgebraicGeometry/PresheafedSpaces.v
arnoudvanderleer: Well, that would be a solution to the question I just asked ^^
Why do they introduce these identities?
And where should these transformations go? FunctorCategory or NaturalTransformations?

UniMath/AlgebraicGeometry/PresheafedSpaces.v
arnoudvanderleer: You are right

UniMath/AlgebraicGeometry/PresheafedSpaces.v
arnoudvanderleer: In this case, I get `not found in table: core.eq.type`. But in general, I try to use refine instead of rewrite, because it gives smaller terms and faster proofs (and gives more clarity about the location of the rewrite)

UniMath/AlgebraicGeometry/PresheafedSpaces.v
arnoudvanderleer: This was somewhat of a request for comments of the reviewer :-)
Shall I add it to a new file in MoreFoundations, or add it to `MoreFoundations/Subtypes`?

UniMath/AlgebraicGeometry/RingedSpaces.v
arnoudvanderleer: Hm, in this case it is a nice thing to have. I'll remove it. Although it really is one of the first steps, because it is hard to do scheme theory without sheaf theory, this part is about the restriction of a sheaf again being a sheaf.

UniMath/AlgebraicGeometry/RingedSpaces.v
arnoudvanderleer: Yes. Would you want me to define it as an hProp or add a lemma directly afterwards? The proof of this is currently contained in `isaprop_has_sheaf`

UniMath/AlgebraicGeometry/RingedSpaces.v
arnoudvanderleer: Hm... But this is like saying `is_unit` on `monoid_data`: the data here is a presheafed space, and this property is about the presheaf part.
In this case, actually I think it would be best to first develop the theory of presheaves and sheaves a bit, before combining them into (pre)sheafed spaces. A couple of properties (for example, maps of stalks) need you to think about sheafs on the same space, before working with sheafed spaces in general.

UniMath/AlgebraicGeometry/Schemes.v
arnoudvanderleer: The reason was that there was an assumption of the axiom of choice, but then I realized that I just needed isomorphism with `Spec_presheafed_space` instead of `Spec_ringed_space`, and the former does not yet need the axiom of choice. I will remove the section.

UniMath/AlgebraicGeometry/SheavesOfRings.v
arnoudvanderleer: Yes

UniMath/Topology/Topology.v
arnoudvanderleer: Oh, right, sorry, this branch started as a hobby project, and apparently I did some cleaning of Topology.
If you hide whitespace, it already becomes clearer. The main changes are
 - I replaced `X -> hProp` by `hsubtype X`
 - I replaced `(∏ A, P A → O A)` by `(P ⊆ O)`
 - I added the lemmas `subtype_equal_isOpen` and `continuous_iff_reflects_open` (and some stuff related to that last one)
 - I upstreamed `AlgebraicGeometry/Topology` as the section `OpenOperations`

UniMath/AlgebraicGeometry/PresheafedSpaces.v
nmvdw: > Does this indeed introduce such a large memory consumption

The particular files that you import: no

However, it is good to be strict with this policy, because otherwise it would be difficult to enforce. "Import bicategories as long as the total memory consumption is below whatever"

UniMath/AlgebraicGeometry/PresheafedSpaces.v
nmvdw: The associators and unitors of functors are defined to be identity natural transformations. They literally are the identity natural transformation.

UniMath/AlgebraicGeometry/PresheafedSpaces.v
arnoudvanderleer: That is fair

UniMath/AlgebraicGeometry/PresheafedSpaces.v
nmvdw: > because it gives smaller terms and faster proofs

While this is true, this difference tends to be negligible in most cases (this piece of code is probably one of them).

> and gives more clarity about the location of the rewrite

That is true, but it forces the user to do more.

UniMath/AlgebraicGeometry/PresheafedSpaces.v
arnoudvanderleer: Ah, right.

UniMath/AlgebraicGeometry/RingedSpaces.v
nmvdw: It would be good to at least have `isaprop_has_locality`

UniMath/AlgebraicGeometry/RingedSpaces.v
nmvdw: Sheaf is always used to indicate a presheaf that is a sheaf. if you thus say 'has sheaf', then I would understand it 'this data together with a sheaf' and not as 'this data is a sheaf'

UniMath/AlgebraicGeometry/PresheafedSpaces.v
nmvdw: I think it fits in `MoreFoundations/Subtypes`, but I would also like to know @benediktahrens opinion

UniMath/AlgebraicGeometry/RingedSpaces.v
arnoudvanderleer: That is fair. But this data is not a sheaf, that is what I am trying to say. Part of this data turns out to form a sheaf, but the data is a space, together with a (pre)sheaf

PR #1940, opened 2024-09-21
Removes the #[reversible=no] annotations of all coercions,

### LINKED ISSUE #1854: Reversible coercions
Due to a bug in Coq, some coercions were automatically reversible even though they should not have been according to the specification (https://github.com/UniMath/UniMath/pull/1850). For this reason, some coercions are reversible according to Coq even though that is not intended. These coercions should be made non reversible.

arnoudvanderleer:
because in coq 8.20, coercions are no longer reversible by default.
Resolves #1854


nmvdw:
There is a merge conflict (RezkCompletion.v).

nmvdw (approved):
looks good to me

Issue #1931, opened 2024-09-09
Wiki vs README files


arnoudvanderleer:
On GitHub, there are two common ways for documentation. We can document things in `.md` files, like `README.md` or `USAGE.md`, or we can add documentation on the wiki. UniMath has done both for a long time. In many directories, there is a `README.md`, there are `INSTALL_blah.md` and `USAGE.md` files, we have a `CONTENTS.md` file, and there has been a wiki with a couple of pages for a long time, which has recently seen some updates. #1929 highlighted the confusion that this scattering of documentation files (partially mea culpa) causes.

Now, I would like to establish a policy on which instructions we want to go where. Personally, I would propose the following:
 - The GitHub repository needs a front page. This page will contain a short welcoming message, a short description of what UniMath is, and some links (as a list, or embedded in some text) to more detailed documentation, in order of relevance for a user who lands there ("getting help" should be higher up, IMHO. But that is outside the scope of this discussion).
 - Some files need to be hard to change (like a constitution). I have mainly one in mind, containing the (almost) non-negotiable rules of UniMath: We do not allow inductive types, nor do we allow new axioms. Do not make mathematically significant changes to Foundations. And maybe a handful of other rules.
 - It can be argued that semi-detailed documentation about individual packages (the package `README` files) should be close to the source code that it documents. At the same time, I think it would be good to have a page about the packages of UniMath with a short introduction about what each package is and isn't about.

Anything not in these categories is better put on the wiki, because that is easier to maintain.

What do you think of this, @benediktahrens @rmatthes @nmvdw?

nmvdw:
I think that the most important information should be in `.md` files, and that the wiki should only be supplementary information. If I look for something, then I check `.md` files, and I think that in other libraries, `.md` files contain the most relevant information. In addition, if you clone UniMath, you have access to the `.md` files, while the wiki can only be accessed via Github.

arnoudvanderleer:
Can you give a couple of examples of "most important" and "supplementary"?

nmvdw:
The following are subsets:

Most important:
- Installation instructions
- Style guide
- Links to Zulip
- Citation and link to Zenodo

Supplementary material:
- Exact formulation of the resizing rules
- Papers using UniMath

The most important is what a contributor would need to contribute to UniMath. Supplementary is everything that is interesting but not strictly necessary to contribute.

benediktahrens:
[...]

Issue #1929, opened 2024-09-08
Consolidate contribution information


blinkybool:
I think the contribution information should be more consolidated and require fewer clicks to be able to "see" what relevant info exists.

When making my first contribution I overlooked the link to this in the UniMath/README (it appears in the middle of a long list of dotpoints). I had clicked the link in the "Contributing to UniMath" heading of the main page, and assumed everything relevant would appear there. I was not aware there was a wiki in use for UniMath (this could probably be advertised better on the frontpage).

It appears there are two style guides, one is a section in UniMath/README and one is in the wiki, and generally useful info for contribution seems to be spread between. I think it'd be better if this was more consolidated. Since the wiki is more recent I would extract (or at least repeat) the contribution relevant stuff from UniMath/README into the wiki somewhere, and the front page "Contributing to UniMath" section should link to a page in the wiki where the following topics are either on the page or are linked to (the point is visibility).
- Basic PR workflow instructions and policy contributions
- Adding a file to a package
- Adding a new package
- Style Guide (with all info from both locations)
- Resources for new users (direct links to On Opaqueness and My proof is slow)

Even while writing this I forgot where some of these pages were. I had to go to the UniMath/README click the link to the wiki page for the style guide (I see now you can click "Show 6 more pages" on the wiki page to reveal it).

These are just suggestions but I think a good guiding principle is that top-level pages should be widely branching and shallower link hierarchies are preferable.

arnoudvanderleer:
Hi! I very much share this experience. UniMath is maintained by a small group of people with busy lives, and documentation is one of the things that can use a lot of improvement.
I started writing a wiki because I thought we needed more documentation, but of course, that means that, in the current situation, the documentation is scattered.

I think you make a good suggestion. Do you think you could set up a first version of this page (or modification of existing pages) on the wiki? Your experience with being a new contributor is the freshest.

In the near future, we need to have a discussion about which content should be in the wiki and what should be in .md files in the repository, but it is easy to at least prepare the content on the wiki, and then shuffle them around depending on what we decide.

blinkybool:
All completely understandable. Thanks for starting the wiki!
Of course I have a busy life too, but in the name of preventing hypocrisy I'll do my best to help out.

arnoudvanderleer:
> UniMath is maintained by a small group of people with busy lives

Note that this may look like an excuse or a reason, but it is more of an observation.

[...]

Issue #1860, opened 2024-03-08
Coqdoc generation documentation and reduction


DenSinH:
Hello, yesterday I was messing around trying to get the standard coqdoc HTML documentation to generate after contributing the ModelCategories package, and I was wondering what your procedure was, or if it would be valuable to describe this in more detail in one of the READMEs.

The reason for this is the following: I am on Windows, but I can use WSL to do everything you should be able to on Linux. However, installing coq through apt gives me an old version (8.15) for which UniMath does not compile, and coqdoc does not work either. Installing coq through the coq-prover snap package gives me version 8.18, for which UniMath does compile, but coqdoc is not included in that package. I decided to just do it on Windows, where the binary installer gave me a new enough version of Coq (8.17) for UniMath to compile, and _also_ have coqdoc. However, in this case the `make doc` Makefile target produced a command with ALL the UniMath files, which was too long for Windows' CreateProcess... My final resort was to do the following:
- run `make clean` to remove any build files
- run `make ModelCategories`, as to only compile the necessary files for the ModelCategories package.
- Replace the following line in the Makefile:
   ```
   VFILES := $(COQMF_VFILES)
   ```
   with 
   ```
   COMPILEDVO := $(shell find . -type f -name "*.vo")
   VFILES := $(COMPILEDVO:.vo=.v)
   ```
   as to only use the `.v` files necessary to compile the ModelCategories package.
- run `make doc`, now producing a command short enough to be ran by `CreateProcess`

All in all, it was a tedious process, and I feel like it could be improved, or there must be a better way to do this (either through WSL or on Windows itself). Feel free to delete this issue if I missed something somewhere, or if you feel it is not needed, after all, the whole problem would not exist if coqdoc shipped with the Linux snap package...

benediktahrens:
You could try
```
$ make PACKAGES="Foundations MoreFoundations CategoryTheory ModelCategories" html
```
Comments:
1. Target `html` makes a vanilla coqdoc documentation, target `doc` postprocesses the html to make proofs "foldable". Folding requires javascript.
2. You might need more packages in the list above, it does not automatically "complete" with all the needed material.

Issue #1587, opened 2022-10-28
Integrate binary installation instructions for UniMath


benediktahrens:
For the School on Univalent Mathematics at Cortona in July 2022, @amato-gianluca and @maggesi (I believe) wrote very good installation instructions for precompiled Coq and UniMath.

These instructions are available at https://docs.google.com/document/d/1KWSugPK-sJ67pL-P4EtoKZ6HMzewmzewLNkR_arXauY/edit#heading=h.omfmjtwseiix

Would it be a good idea to integrate those instructions into the UniMath documentation? If so, how?

DanGrayson:
We could just include their file as an alternative.  I don't think we
should tell everyone to install Visual Studio.

I attach the exported text file.

> Message ID: ***@***.***>
>

﻿UniMath Installation instructions
Install UniMath on Mac
Install Coq/UniMath and VScode on Mac.


These instructions are for installing and configuring the necessary software for the UniMath School 2022.
Step 1: Install software
Install VScode (Visual Studio Code) the VScoq extension for VScode and Coq Platform following the instructions from the respective web pages:
https://code.visualstudio.com/
https://github.com/coq-community/vscoq
https://github.com/ojsheikh/unicode-latex
https://github.com/coq/platform (TAKE THE VERSION WITH Coq 8.15)
Step 2: Configure and test the software
In Visual Studio Code go to menu
File -> Open
then use the dialog box to create a new directory for your Coq files, for example “Cortona2022”.


Create a new file (Command-N), copypaste the following lines in it


Require Export UniMath.Foundations.All.


Lemma myfirstlemma : 2 + 2 = 4.
Proof.
    apply idpath.
Defined.


Save it (Command-S) with name Test.v (any name ending in .v will work).


Press Shift-Command-P and start typing “open workspace” then select
“Preferences: Open Workspace Settings (JSON)”














Copypaste the following lines


{
    "coqtop.binPath": "/Applications/Coq-Platform~8.15~2022.04.app/Contents/Resources/bin",
    "coqtop.args": [
        "-noinit",
        "-indices-matter",
        "-type-in-type",
        "-w",
        "-notation-overridden",
        "-Q",
        "/Applications/Coq-Platform~8.15~2022.04.app/Contents/Resources/lib/coq/user-contrib/UniMath",
        "UniMath"
    ],
    "coq.format.indentAfterBullet": "indent",
    "coq.format.unindentOnCloseProof": true,
[...]

PR #1941, opened 2024-10-01
Move all documentation to the `/documentation` folder and do a first organization of the files

arnoudvanderleer:
Adding this as a draft, to notify y'all that I am working on this. If you have any input, let me know.
The intention is that this will bring the documentation up to date and make it somewhat coherent, addresing issues like #1860, #1931, #1929, #1587, #1360, #1400,


benediktahrens:
This is a great effort, thanks for taking care of this.
Moving files and modifying them in one go makes things harder to track (at least for me). Would it be a good idea to merge this PR soon and then do further updates to the documentation?

arnoudvanderleer:
Hm, we could do that, although currently almost every link in the documentation is broken. Is this still your preference?

benediktahrens:
Ah, right, maybe not! But any deeper change could be postponed, such as updating installation instructions.

arnoudvanderleer:
All right, fair

arnoudvanderleer:
@benediktahrens Would it be an idea to merge this, and then continue working on the contents of the files in a new PR?

PR #1944, opened 2024-10-18
Expand the material on Grothendieck topoi

arnoudvanderleer:
This pull request
 - Splits the file into four parts for its main definitions;
 - Adds more API to the definitions;
 - Gives two additional equivalent definitions for the sheaf property;
 - Adds a couple of examples.


nmvdw:
This PR is work in progress, right? If so, could you mark it as a draft to make that clear?

arnoudvanderleer:
No, it is supposed to be finished.
Apparently, I forgot to "copy" some of the material from the other branch, so it is still throwing errors. But once I copy the correct files, it should be finished, and the material that is here is in it's (currently) finished state

arnoudvanderleer:
Oh, I realize now that something breaks. Will fix

UniMath/CategoryTheory/GrothendieckToposes/Sheaves.v
```diff
@@ -0,0 +1,504 @@
...
+        := pr1 fWg.
+
+      Definition sheaf_property_equalizer_codomain_index_object
+        (fWg : sheaf_property_equalizer_codomain_index)
+        : C
+        := pr12 fWg.
+
+      Definition sheaf_property_equalizer_codomain_index_morphism
+        (fWg : sheaf_property_equalizer_codomain_index)
+        : C⟦sheaf_property_equalizer_codomain_index_object fWg, sieve_selected_morphism_domain (sheaf_property_equalizer_codomain_index_selected_morphism fWg)⟧
```
nmvdw, line 132: This line also is a bit long. Maybe it might be better to use the `-->` notation and write domain, codomain and `-->` on separate lines.

UniMath/CategoryTheory/GrothendieckToposes/Sheaves.v
```diff
@@ -0,0 +1,504 @@
...
+        {
+          intro.
+          apply impred_isaset.
+          intro.
+          apply setproperty.
+        }
+        apply funextsec.
+        intro f.
+        refine (!eqtohomot (nat_trans_ax (pr11 (H _)) _ _ _) _ @ _).
+        refine (_ @ eqtohomot (nat_trans_eq_pointwise (pr21 (H (make_nat_trans _ _ _ (is_sheaf_to_is_sheaf'_subproof c S w)))) _) (sieve_selected_morphism_preimage f)).
```
nmvdw, line 328: If I guess correctly, `is_sheaf_to_is_sheaf'_subproof` arises from an `abstract` (which is why the `_subproof` is there). However, it is not a good idea to ever refer to the subproofs generated by an abstract, since the names could be changed in a version. It would be better to split out this subproof.

UniMath/CategoryTheory/GrothendieckToposes/Sieves.v
```diff
@@ -0,0 +1,129 @@
+(**************************************************************************************************
+
+  Sieves
+
+  A sieve on X is a collection of morphisms to X, that is closed under precomposition. It is
+  formalized as a subobject of the yoneda embedding of X. In other words, it consists of
```
nmvdw, line 6:
```suggestion
  formalized as a subobject of the Yoneda embedding of X. In other words, it consists of
```

UniMath/CategoryTheory/GrothendieckToposes/Sieves.v
```diff
@@ -0,0 +1,129 @@
...
+  a presheaf F, together with a monomorphism F ⟹ yoneda X. Note that here, a monomorphism is a
+  pointwise injective natural transformation. The `selected` morphisms from Y to X are then the
+  images under this transformation of the elements of F Y.
+  Now, given a morphism h: Y → X, we can use the Yoneda embedding よ(h) of h to pull back a sieve
```
nmvdw, line 10:
```suggestion
  Now, given a morphism h: Y → X, we can use the Yoneda embedding よ(h) of h to pullback a sieve
```

UniMath/CategoryTheory/GrothendieckToposes/Toposes.v
```diff
@@ -0,0 +1,51 @@
+(**************************************************************************************************
+
+  Grothendick toposes
```
nmvdw, line 3:
```suggestion
  Grothendieck toposes
```

UniMath/CategoryTheory/GrothendieckToposes/Toposes.v
```diff
@@ -0,0 +1,51 @@
...
+
+  A Grothendieck topos is a category that is equivalent to the category of sheaves on some site, see
+  Mac Lane and Moerdijk, page 127.
```
nmvdw, line 6: What about adding a section on references to the documentation?

UniMath/CategoryTheory/GrothendieckToposes/Toposes.v
```diff
@@ -0,0 +1,51 @@
...
+Require Import UniMath.CategoryTheory.Core.Functors.
+Require Import UniMath.CategoryTheory.Equivalences.Core.
+Require Import UniMath.CategoryTheory.Subcategory.Core.
+
+Require Import UniMath.CategoryTheory.GrothendieckToposes.Sheaves.
+Require Import UniMath.CategoryTheory.GrothendieckToposes.Sites.
+
+(** * 1. Grothendieck toposes *)
+
+Definition Grothendieck_topos : UU :=
```
nmvdw, line 29: I am not sure whether I agree with this definition.

The notion of Grothendieck topos should be a proposition. One way to characterize Grothendieck toposes, is via Giraud's theorem (https://ncatlab.org/nlab/show/Giraud%27s+theorem). Although there are some technical complicated regarding local presentability (this quantifies over some cardinal number, which comes from the cardinality of the site), this characterization gives us a proposition. For this reason, I would say that the site and the adjoint equivalence in this definition should be under a truncation.

UniMath/CategoryTheory/Monics.v
```diff
@@ -8,13 +8,17 @@
 Require Import UniMath.Foundations.PartD.
 Require Import UniMath.Foundations.Propositions.
 Require Import UniMath.Foundations.Sets.
+Require Import UniMath.MoreFoundations.PartA.
```
nmvdw, line 11: What about `Require Import UniMath.MoreFoundations.All` and `Require Import UniMath.CategoryTheory.Core.Prelude`?

UniMath/CategoryTheory/Monics.v
```diff
@@ -189,6 +205,75 @@ Section monics_subcategory.
...
+      ).
+    - abstract (
+        intro f;
+        apply z_iso_eq;
+        apply Monic_eq;
+        reflexivity
+      ).
+  Defined.
+
+  Definition total2_contr
```
nmvdw, line 234: This should be in another file.

nmvdw (commented):
Looks good! Before this PR can be merged, some comments should be addressed and the CI should give green lights.

UniMath/CategoryTheory/GrothendieckToposes/Toposes.v
nmvdw: Another thing: if `P` is a property on categories (should be valued in `hProp`) such that `P` holds for every sheaf topos, then `P` holds for every **univalent** category that is a Grothendieck topos (where the site and equivalence are truncated). I think this is the key statement about Grothendieck toposes and it is something that would be not too difficult to prove. I think it would be good to add this statement.

We can strengthen this to all type families on categories if we assume that the site is chosen.

UniMath/CategoryTheory/GrothendieckToposes/Examples/DiscreteTopology.v
```diff
@@ -0,0 +1,167 @@
...
+
+(** * 1. The discrete site *)
+
+Section DiscreteTopology.
+
+  Context (C : category).
+
+  Definition discrete_topology_selection
+    (c : C)
+    : hsubtype (Sieves.sieve c)
```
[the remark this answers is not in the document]
arnoudvanderleer: Because `Presheaf` also define sieves. (I am not entirely sure why. It seems to be an old attempt to say something about topoi)

UniMath/CategoryTheory/GrothendieckToposes/Examples/EmptySieve.v
```diff
@@ -0,0 +1,58 @@
...
+  Proof.
+    apply is_nat_trans_monic_from_pointwise_monics.
+    intro a.
+    apply (invmap (MonosAreInjective_HSET _)).
+    intro e.
+    induction e.
+  Qed.
+
+  Definition empty_sieve
+    : Sieves.sieve X
```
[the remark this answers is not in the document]
arnoudvanderleer: Because of `Presheaf`

UniMath/CategoryTheory/GrothendieckToposes/Sheaves.v
arnoudvanderleer: Oh, good catch. I intended to fix this later, but forgot.

UniMath/CategoryTheory/GrothendieckToposes/Sieves.v
arnoudvanderleer: Imho, pullback is only a noun, arising from the verb "to pull back".

UniMath/CategoryTheory/GrothendieckToposes/Toposes.v
arnoudvanderleer: Can you elaborate a bit more on this?

UniMath/CategoryTheory/GrothendieckToposes/Toposes.v
arnoudvanderleer: So you're suggesting taking the propositional truncation of the site, instead of the site itself? Sounds reasonable.

Also, the proof about P holding for every sheaf topos, sounds like using univalence of the bicategory `UnivalentCat` (which is shown in the package `Bicategories` and should therefore not be used in `CategoryTheory`), or is there another way?

UniMath/CategoryTheory/Monics.v
arnoudvanderleer: Fair enough. Somewhere in `MoreFoundations`? (Not sure which one though)

UniMath/CategoryTheory/GrothendieckToposes/Examples/DiscreteTopology.v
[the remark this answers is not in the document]
nmvdw: Oh sure. To define the subobject classifier of presheaves (which is mostly done in UniMath), one needs sieves since the subobject classifier is given by the collection of all sieves. Is the definition of sieves in presheaves usable?

UniMath/CategoryTheory/GrothendieckToposes/Toposes.v
nmvdw: See this: https://github.com/UniMath/UniMath/blob/master/UniMath/Bicategories/ComprehensionCat/DFLCompCat.v#L83

UniMath/CategoryTheory/GrothendieckToposes/Toposes.v
nmvdw: > So you're suggesting taking the propositional truncation of the site, instead of the site itself? Sounds reasonable.

There should be a definition `is_Grothendieck_topos` which is truncated, and a definition `equiv_to_sheaf_cat` which is not truncated (the name could be improved).

> Also, the proof about P holding for every sheaf topos, sounds like using univalence of the bicategory UnivalentCat (which is shown in the package Bicategories and should therefore not be used in CategoryTheory), or is there another way?

You only need that every adjoint equivalence gives an equality. This is a combination of
- https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/CategoryEquality.v#L251 (for univalent categories, it follows directly)
- https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/catiso.v#L563

arnoudvanderleer: How about `Grothendieck_topos_structure` for the latter?

arnoudvanderleer: Oooh, there is stuff in `CategoryTheory` about this

nmvdw: > How about `Grothendieck_topos_structure` for the latter?

Agreed.

UniMath/CategoryTheory/Monics.v
nmvdw: What about this file: https://github.com/UniMath/UniMath/blob/master/UniMath/MoreFoundations/MoreEquivalences.v

nmvdw (approved):

UniMath/CategoryTheory/Monics.v
arnoudvanderleer: In this case, I added a minimal import because the existing imports were kind of minimal. Shall I update imports in files that I encounter to just start with `UniMath.Foundations.All` and `UniMath.MoreFoundations.All`?

nmvdw: I think minimizing imports from `Foundations` and `MoreFoundations` is not very useful. In addition, it can make extending some code more annoying, because imports might need to be added if one wants to use `maponpaths_2`.

UniMath/CategoryTheory/GrothendieckToposes/Examples/DiscreteTopology.v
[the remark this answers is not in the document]
arnoudvanderleer: It is pretty straightforward, but not general enough for arbitrary categories:
```
Definition sieve_def (c : C) : UU.
Proof.
use total2.
- apply (hsubtype (∑ (x : C),C⟦x,c⟧)).
- intros S.
  apply (∏ (s1 : ∑ (x : C),C⟦x,c⟧), S s1 →
         ∏ y (f : C⟦y,pr1 s1⟧), S (y,, f · pr2 s1)).
Defined.
```

UniMath/CategoryTheory/GrothendieckToposes/Sheaves.v
```diff
@@ -0,0 +1,504 @@
...
+  3. Sheaves
+  3.1. The category of sheaves [sheaf_cat]
+  3.2. Sheaves [sheaf]
+  3.3. Sheaf morphisms [sheaf_morphism]
+
+ **************************************************************************************************)
+Require Import UniMath.Foundations.All.
+Require Import UniMath.MoreFoundations.Notations.
+
+Require Import UniMath.CategoryTheory.Categories.HSET.Core.
```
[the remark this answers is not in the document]
arnoudvanderleer: Hm, that feels a bit too general (because there is a lot of material there that we don't use)

nmvdw (approved):

Issue #1958, opened 2024-12-09
Adding Coq8.19 to CI


benediktahrens:
Could we add Coq 8.19 to CI, so that we are alerted when UniMath becomes incompatible with 8.19?

nmvdw:
That's a good idea.

benediktahrens:
Done, closing.

PR #1960, opened 2024-12-09
Add versions to CI

benediktahrens:
relates to #1958


rmatthes:
Is Ubuntu 24.04LTS still not proposed?

PR #1966, opened 2024-12-20
Deduplicate `dirprodeq` and `total2_paths2`

### LINKED ISSUE #1964: Deduplicate `dirprodeq`
It is a duplicate of `dirprod_paths`, both in `Foundations/PartA`

arnoudvanderleer:
Resolves #1964

### DIFF
--- a/UniMath/Foundations/PartA.v
+++ b/UniMath/Foundations/PartA.v
@@ -431,7 +431,12 @@ Proof.
   intros. induction ex. induction ey. apply idpath.
 Defined.
 
-Lemma dirprodeq (A B : UU) (ab ab' : A × B) :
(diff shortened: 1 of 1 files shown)


peterlefanulumsdaine:
I don’t quite support this PR as stands — the deduplication is certainly good, but `total2_paths2` currently sits clearly in a well-organised sequence of related lemmas (after `total2_paths` and variants, and followed by its own variants like `total2_paths_f`), and this PR moves it away from that group, which I think is undesirable.

Generally I support small cleanup like this in Foundations, but I think it needs to be done with consideration for the general organisation of the files, not just looking at individual lemmas.  I’ll open a new issue about this as a big-picture question!

peterlefanulumsdaine:
Closing this as per my previous comment

PR #1995, opened 2025-02-20
Adapt to rocq#19987

Tragicus:
https://github.com/coq/coq/pull/19987 refolds terms before using them to instantiate evars. There is one instance where we need to unfold by hand.


benediktahrens:
@Tragicus : thanks a lot for preparing this!

benediktahrens (approved):

PR #2002, opened 2025-03-11
Show that in a univalent category, being split idempotent is a mere proposition

arnoudvanderleer:


UniMath/CategoryTheory/Retracts.v
```diff
@@ -172,6 +173,74 @@ Section Idempotents.
...
+Lemma isaprop_is_split_idempotent
+  {C : category}
+  {x : C}
+  (f : C⟦x, x⟧)
+  (C_is_univ : is_univalent C)
+  : isaprop (is_split_idempotent f).
+Proof.
+  apply invproofirrelevance.
+  intros X X'.
+  induction X as [X HX], HX as [r HX].
```
KobeWullaert, line 211:
```suggestion
  induction X as [X [r HX]].
```
You can nest inductions. I don't think there is a preference about the use hereof, so this suggestion could be ignored.

KobeWullaert (approved):

UniMath/CategoryTheory/Retracts.v
arnoudvanderleer: "Do not use destruct, match, case, square brackets with intros, or nested square brackets with induction. (The goal is to prevent generation of proof terms using match.)" - [see the style guide](https://github.com/UniMath/UniMath/blob/master/documentation/contributing/Style-guide.md)

PR #2010, opened 2025-03-24
Adapt to coq/coq#20387 (Ltac2 compiled with -noinit)

SkySkimmer:
Ltac2 `now` (used eg in AlgebraicTheories.combinators) relies on Corelib.Init.Ltac.easy_forward_decl so we set it when we define unimath's `easy` in MoreFoundations.Tactics.


SkySkimmer:
Changed to be hopefully backwards compatible

nmvdw (approved):

PR #2012, opened 2025-03-26
Add the adjunction example of the initial and terminal functors, 

arnoudvanderleer:
slightly clean up the examples file


UniMath/CategoryTheory/Adjunctions/Examples.v
```diff
@@ -1,162 +1,163 @@
...
-
-*)
+(**************************************************************************************************
+
+  Examples of Adjunctions
+
+  The binary delta functor C → C × C is a left (right) adjoint to the binary (co)product functor
+  C × C → C, and the general delta functor C → C^I is a left (right) adjoint to the general
+  (co)product functor C^I → C.
+  The swapping functor [C, [D, E]] → [D, [C, E]] is a left adjoint to itself.
```
nmvdw, line 8: This one would be an adjoint equivalent, right?

UniMath/CategoryTheory/Adjunctions/Examples.v
```diff
@@ -1,162 +1,163 @@
...
+  1.3. The adjunction between the binary coproduct and delta functor
+    [is_left_adjoint_bincoproduct_functor]
+  1.4. The adjunction between the general coproduct and delta functor
+    [is_left_adjoint_coproduct_functor]
+  2. Swapping of arguments in functor categories [are_adjoint_functor_cat_swap]
+  3. Adjunctions are closed under natural isomorphism [is_left_adjoint_nat_z_iso]
+  4. The adjunction between the constant initial and terminal functors
+    [initial_terminal_are_adjoints]
+
+  Written by: Anders Mörtberg, 2016
```
nmvdw, line 27: Might be good to say that it got refactored

nmvdw (approved):
Looks good to me.

UniMath/CategoryTheory/Adjunctions/Examples.v
arnoudvanderleer: Yes, but it was only formalized as an adjunction, apparently

UniMath/CategoryTheory/Adjunctions/Examples.v
arnoudvanderleer: Or shall I just remove this line? We have version control and git blame, after all.

nmvdw: @benediktahrens

benediktahrens: Let's switch it to "Started by:", then we don't remove Anders' name and won't have to update it in the future?

arnoudvanderleer: Sure. Although I still feel like we don't need to keep track of who touched a file (or thought first of collecting a few examples), because we have version control.

nmvdw:
Are all the comments addressed now?

arnoudvanderleer:
I believe so?

PR #2013, opened 2025-03-26
Experiment to see why CI fails with coqc-not-found

benediktahrens:


SkySkimmer:
The docker-rocq image for 9.0 does not include the compat coqc executable. You need to install them with eg `opam install coq-core` (should take only a few seconds).

arnoudvanderleer:
@SkySkimmer Has it been renamed to rocq? If so, would you recommend updating all names from coq to rocq instead?

SkySkimmer:
I would not until you drop compatibility for pre 9.0 versions.

PR #2020, opened 2025-04-08
Preservation properties of weak equivalences

KobeWullaert:
In this PR, it is proven that weak equivalences preserve initial objects, binary coproducts, and natural number objects.


arnoudvanderleer:
Do they also preserve (co)limits in general (And can the preservation of (co)products) be deduced from that?

nmvdw (approved):
I would suggest to change NNO to parameterized NNO. For the remainder, it looks good.

KobeWullaert:
> Do they also preserve (co)limits in general (And can the preservation of (co)products be deduced from that)?

They preserve finite limits, this is already formalized. Actually, this was a copy paste from those files, modulo the appropriate changes.
And I expect that the preservation properties can indeed be induced from the preservation of finite limits. Nonetheless, in this case, one does need the appropriate duality results/equivalences; not only for the (co)limits, but also for the preservation properties.

KobeWullaert:
> I would suggest to change NNO to parameterized NNO. For the remainder, it looks good.

Do I leave this for a future PR?

nmvdw:
> > I would suggest to change NNO to parameterized NNO. For the remainder, it looks good.
> 
> Do I leave this for a future PR?

Sure

UniMath/CategoryTheory/WeakEquivalences/Preservation/NNO.v
```diff
@@ -0,0 +1,188 @@
+(**
+   In this file, we show that an arbitrary weak equivalence F : C -> D preserves a natural numbers object.
```
benediktahrens, line 2: "Arbitrary" is a weird way of phrasing things - mathematicians just say "any".
```suggestion
   In this file, we show that any weak equivalence F : C -> D preserves a natural numbers object.
```

UniMath/CategoryTheory/WeakEquivalences/Preservation/Bincoproducts.v
```diff
@@ -0,0 +1,145 @@
+(**
+   In this file, we show that an arbitrary weak equivalence F : C -> D preserves binary coproducts.
```
benediktahrens, line 2:
```suggestion
   In this file, we show that any weak equivalence F : C -> D preserves binary coproducts.
```

UniMath/CategoryTheory/Arithmetic/NNO.v
```diff
@@ -11,6 +11,7 @@ Require Import UniMath.MoreFoundations.All.
 
 Require Import UniMath.CategoryTheory.Core.Prelude.
 Require Import UniMath.CategoryTheory.Limits.Terminal.
+Require Import UniMath.CategoryTheory.Limits.Preservation.
```
benediktahrens, line 14: How about updating the comment at the beginning of the file to describe the new content?

PR #2024, opened 2025-04-10
Univalence of the Bicategory of Display Map Categories

FCsacsa:


UniMath/Bicategories/DispBicatOfDisplayMapCat.v
```diff
@@ -0,0 +1,445 @@
...
+  intros C₁ C₂ F D₁ D₂ HF HF'; cbn in * |-.
+  use isweqcontrcontr; cbn.
+  - apply isPredicate_is_display_map_class_functor.
+  - unfold disp_invertible_2cell, id2_invertible_2cell, is_invertible_2cell_id₂, make_invertible_2cell, make_is_invertible_2cell, is_disp_invertible_2cell, id2_left, nat_trans_id. cbn.
+    rewrite ? transportb_const; simpl.
+    apply iscontr_prod; [exact iscontrunit | apply iscontr_prod; [exact iscontrunit | apply iscontr_prod]].
+    all: apply isapropunit.
+Qed.
+
+Proposition disp_univalent_2_0_disp_bicat_display_map_cat :
```
nmvdw, line 121: Do you have an idea on how to prove this?

FCsacsa: I had an idea, but that might not work out, will probably try to split up the equivalence into 'steps'.

nmvdw: Yes. It is often convenient to turn it into a chain of equivalences: one characterizes the identity of objects and one characterizes the adjoint equivalences (and possibly more).

UniMath/Bicategories/DispBicatOfDisplayMapCat.v
```diff
@@ -118,16 +118,110 @@ Proof.
...
+  {C : bicat_of_univ_cats} (D D' : disp_bicat_display_map_cat C)
+  : is_display_map_class_functor D D' (functor_identity (univalent_category_to_category C))
+    × is_display_map_class_functor D' D (functor_identity (univalent_category_to_category C))
+  -> disp_adjoint_equivalence (idtoiso_2_0 C C (idpath C)) D D'.
+Proof.
+  intros [HF HG].
+  exists HF. use tpair.
+  - exists HG. cbn. exact (tt ,, tt).
+  - split; use tpair.
+    (* TODO: investigate speed *)
```
nmvdw, line 131: What are the observations here?

Also, some more `abstract` would be nice here

FCsacsa: I don't fully get what you mean by 'observations', could you maybe clarify what you mean?

I will go over all the proofs I added in this PR and abstract away non-necessary details, just wanted to get it working first.

nmvdw: The comment says 'investigate speed'.  Is this piece of code slow? Which lines/commands?

FCsacsa: Oh, I see, this is meant for the `rewrite`s in the following sub-goals, some of them took over a second to finish, so I wanted to look into if I could maybe write these proofs differently.

nmvdw (approved):
If I see it correctly, then there are now two constructions of the bicategory of display map categories, right? One via displayed bicategories and one without?

UniMath/CategoryTheory/DisplayedCats/DisplayMapCat.v
```diff
@@ -399,11 +441,144 @@ Qed.
...
+Definition display_map_class_equiv_to_display_map_class_adjoint
+  {C : category}
+  (D D' : display_map_class C)
+  : pr1 D = pr1 D'
+  ->
+    (is_display_map_class_functor D D' (functor_identity C))
+      ×
+    (is_display_map_class_functor D' D (functor_identity C)).
+Proof.
+Set Printing Coercions.
```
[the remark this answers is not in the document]
FCsacsa: I will look into it, as is, I needed them I think because of the rewrites, but a `simpl` or `cbn` or `unfold` should also work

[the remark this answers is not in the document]
nmvdw: Try without it, `Printing` should only make it visible for the user, but not for the tactics (I guess)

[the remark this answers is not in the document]
FCsacsa: It does indeed work!

UniMath/Bicategories/DispBicatOfDisplayMapCat.v
```diff
@@ -0,0 +1,582 @@
...
+
+(** *** Pseudofunctor Data *)
+Definition bicat_terminal_display_map_cat_to_bicat_full_comp_cat
+  : bicat_terminal_display_map_cat → bicat_full_comp_cat.
+Proof.
+  intros D. use make_full_comp_cat.
+  - use make_comp_cat.
+    + use make_cat_with_terminal_cleaving.
+      * use make_cat_with_terminal_disp_cat.
+        -- exact (pr1 D). (* NOTE: I do not know why the coercion [terminal_display_map_cat_to_base_cat] does not work here. *)
```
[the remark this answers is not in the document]
FCsacsa: The type is `D : ob bicat_terminal_display_map_cat` and the goal is `univalent_category`.
But I do have the type
```coq
Definition terminal_display_map_cat : UU := bicat_terminal_display_map_cat.
```
with the coercion:
```coq
Coercion terminal_display_map_cat_to_univ_base (D : terminal_display_map_cat) : univalent_category := pr1 D.
```
and adding a coercion between `bicat_terminal_display_map_cat >-> terminal_display_map_cat` also does not help.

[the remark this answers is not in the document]
nmvdw: In order for the coercion to work, the type of `D` needs to be `terminal_display_map_cat` rather than `ob bicat_terminal_display_map_cat`. Instead of `intros D`, you might want to do something like `refine (λ (D : terminal_display_map_cat), _)`.

FCsacsa:
> If I see it correctly, then there are now two constructions of the bicategory of display map categories, right? One via displayed bicategories and one without?

Yes, that is correct.

nmvdw:
> > If I see it correctly, then there are now two constructions of the bicategory of display map categories, right? One via displayed bicategories and one without?
> 
> Yes, that is correct.

Is there a reason for keeping in both? Which one should be used?

nmvdw:
I suggest the following:
- Keep in both versions
- Mention that the one using displayed bicategories is the 'standard'
- Mention the differences between the two versions in the comments and the reason for preferring one of the other.

@benediktahrens Do you have an opinion on this matter?

benediktahrens:
I think what you are suggesting sounds very good!

FCsacsa:
I have also moved the two files to a new directory, to keep them together.

nmvdw (approved):

nmvdw:
I think this PR is ready to be merged, right?

FCsacsa:
Yes, think so too

PR #2032, opened 2025-04-22
Autogenerate docs

arnoudvanderleer:
I believe it would be good to automatically generate rocqdoc (and other kinds of derived content) from the code at `master`, and combine that with the `*.md` files to get a website, very much like the way it works for [the HoTT library](https://github.com/HoTT/Coq-HoTT/blob/master/.github/workflows/ci.yml).

The last step of this PR fails because of permission issues (actions triggered by a commit in my fork cannot have write access on the UniMath repository). However, this afternoon I triggered the same action from the newly created [UniMath/autogenerate-docs](https://github.com/UniMath/UniMath/tree/autogenerate-docs) branch, which correctly gathered everything we need (apart from the jekyll config) on the [gh-pages](https://github.com/UniMath/UniMath/tree/gh-pages) branch.

The current version of my full proposal (copied from the approach of the HoTT repository) is as follows:
1. People write code for UniMath, and commit it via pull requests. At some point, this results in a commit on the master branch.
2. Currently, this triggers a CI action to compile the code. My idea is to have this CI action also compile rocqdoc(, alectryon, a dependency graph, etc)
3. After the documentation has been compiled, the CI action pushes this via a single commit to the `gh-pages` branch. We can either configure it to save earlier versions in the git history, or to overwrite the commits each time and keep the history "clean".
4. Currently, [the UniMath site](https://unimath.github.io) is published from the `master` branch. My proposal is to publish it from the `gh-pages` branch. This means that every time a commit to `gh-pages` is made, the site will update with the new docs.

Do you agree that this would be a good thing to have, and that this is a good implementation?


arnoudvanderleer:
Ah, I will clarify this:
- When you go to the [gh-pages](https://github.com/UniMath/UniMath/tree/gh-pages) branch and download it, you currently get a 15 MB zip file. When extracted, it amounts to about 300 MB, which is mostly the generated rocqdoc.
- I added a full outline of my proposal at the top. That should clarify things a bit.
- GitHub actions can generate artifacts. We can configure those as we see fit, but in the "easiest" case, we will have at least
  - A zip with the compiled `vo` files;
  - A zip with the generated rocqdoc;
  - A zip with the generated alectryon files, if we get that working;
  - etc.
 
  I believe such artifacts are automatically deleted after 90 days, or when we have too many (then the oldest will be automatically deleted(?)).
  To summarize: only the current version will be published on the site. If you permanently want to publish a version, you can download it from the `gh-pages` branch. If we want older versions, we can make sure that artifacts will be available for a while (i.e. 90 days, or a handful of commits on `master`).

benediktahrens:
Some random comments reflecting the state of my understanding:

I think it would be useful to have rocqdoc (and alectryon?) files generated automatically, as an artifact. At least for rocqdoc, this should not add much (computer) work compared to what we already ask from CI? For alectryon, I don't know how it works and how much extra computing time it would add to CI runs. It might be good to measure that roughly.

As for publishing rocqdoc (or alectryon) to the web, I am not sure how useful it is. The web versions are useful for linking to in papers, for which we need *permanent* links - this is not what you are suggesting, if I understand correctly. I don't usually read UniMath code in the browser - but maybe others do? Maybe I will start doing it once this feature is implemented in our CI.

Saving older versions of rocqdoc in `gh-pages` will soon add up to a lot of storage used, is my vague feeling. It might be good to keep the size of the repository small.

Is it planned, or possible, for the satellite repositories to also be part of the whole process?

arnoudvanderleer:
> At least for rocqdoc, this should not add much (computer) work compared to what we already ask from CI? For alectryon, I don't know how it works and how much extra computing time it would add to CI runs. It might be good to measure that roughly.

From my tests, I can deduce that a full compilation of UniMath takes about 45 minutes. A full compilation of the docs takes another 38 minutes. And I don't know how long Alectryon takes. However, all of these have support for caching, so this is only a far upper bound. In practice, we expect the numbers to be much lower. Also, we only have to do all this for mutations on the master branch (unless you expect the artifacts from open pull requests to be useful). Therefore, we would have just a little extra work, which happens rather infrequently.

> As for publishing rocqdoc (or alectryon) to the web, I am not sure how useful it is. I don't usually read UniMath code in the browser - but maybe others do? Maybe I will start doing it once this feature is implemented in our CI.

I think that there are three minor reasons for publishing rocqdoc in the browser: In issues and on zulip, one can link to the generated documentation, instead of to the code, since the former is somewhat more intended for viewing. Also, the generated docs are hyperlinked, so when reading, one can quickly jump to definitions (and IIRC, there was something about automatically folding proofs). Lastly, if I search online for e.g. "Grothendieck topology", [the first result](https://unimath.github.io/doc/UniMath/c26d11b/UniMath.CategoryTheory.GrothendieckTopos.html) is one of the "archival" documentations. This should, of course, be the up-to-date documentation instead.

Also, Alectryon can be insightful for reading and "evaluating" proofs without ever downloading coq. Or without compiling that version. For example, when making a change somewhere that breaks a proof in a totally different file, I would like to have a reference for how the proof worked before everything broke down.

> The web versions are useful for linking to in papers, for which we need _permanent_ links - this is not what you are suggesting, if I understand correctly.

Fair enough. However, the workflow that I am setting up here wil yield a `gh-pages` branch, as well as artifacts, with the up-to-date documentation which can be downloaded and re-uploaded elsewhere.

> Saving older versions of rocqdoc in `gh-pages` will soon add up to a lot of storage used, is my vague feeling. It might be good to keep the size of the repository small.

Agreed.

> Is it planned, or possible, for the satellite repositories to also be part of the whole process?

Not necessarily planned, but we can make it work if we want. We can either build them separately, or together with the core repository.
One thing I have noticed is that, building the rocqdoc requires the cache of the compilation of the `vo` files, but also generates its own cache (in the same folder). Therefore, to have effective caching, if we generate the `vo` files and rocqdoc in different jobs, we need to juggle a bit with the cache. Also, Alectryon has its own cache, but needs the build `vo` files, so we have to move that between steps as well. Therefore, I am considering merging these steps, but this would make individual (larger) jobs less "transparent" maybe.
This might also be a good moment to re-evaluate what exactly we want to happen whenever we something is pushed to `master`.

arnoudvanderleer:
See [this run](https://github.com/UniMath/UniMath/actions/runs/14700907332) for what happens if you do caching the right way :-)

benediktahrens:
Interesting! 

I imagine that later we will activate the guards restricting building alectryon et al to pushes to `master`?

arnoudvanderleer:
That is indeed my proposal. I will add the guards now :-)

arnoudvanderleer:
@benediktahrens Do you think we can merge this now?

benediktahrens:
Yes, looks great to me - thanks a lot for setting this up!

arnoudvanderleer:
> I think it would be useful to have rocqdoc (and alectryon?) files generated automatically, as an artifact.

If we want, we can save space by deleting the rocqdoc and alectryon artifacts after they are generated, and refer people to the `gh-pages` branch where they can download a zip with all the documentation at the same time. But for now, we will save the artifacts.

benediktahrens:
> > I think it would be useful to have rocqdoc (and alectryon?) files generated automatically, as an artifact.
> 
> If we want, we can save space by deleting the rocqdoc and alectryon artifacts after they are generated, and refer people to the `gh-pages` branch where they can download a zip with all the documentation at the same time. But for now, we will save the artifacts, I guess.

The artifacts do not become part of the repository, I think - so I personally don't think we need to save space there. My main concern is the repository growing too much.

arnoudvanderleer:
> The artifacts do not become part of the repository, I think - so I personally don't think we need to save space there. My main concern is the repository growing too much.

I'm not entirely sure how storage works regarding artifacts. Still, having a `gh-pages` branch will grow the repository, but I'm not sure whether it will be much. That is very relative ^^

arnoudvanderleer:
Would you merge this branch?

PR #2035, opened 2025-04-28
Weak equivalence exponentials

KobeWullaert:
In this PR, it is proven that the Rezk completion for categories (with binary products) lifts to categories with exponentials.
Furthermore, a refactoring has happened for reflectors/universal arrows for (1-)categories. More precisely, the constructor for adjoints via "universal_arrow_from" and "left_adjoint_from_partial" has been separated from the file Adjunctions.Core. into Adjunctions.(Co)Reflections.


UniMath/CategoryTheory/Exponentials.v
```diff
@@ -0,0 +1,579 @@
+(**
+   Let C be a category with binary products.
+   In [CategoryTheory/exponentials.v], the theory of exponential objects has been developed using the language of adjunctions.
+   In this file, we develop exponentials using universal arrows.
+   This is used, e.g., to construct Rezk completions for cartesian closed categories.
+   The characterization presented in this file, allows to reason about exponents of pairs of objects, without assuming that an object is exponentiable (wrt to all objects).
```
arnoudvanderleer, line 6: To protect against a proliferation of different equivalent definitions: it seems that the core definition here is `IsExponentiableObject`, which is equivalent to `exponentials.is_exponentiable`. Is there a reason why you cannot formulate all this in terms of `is_exponentiable`, just adding a constructor and "accessors" for that definition?

UniMath/CategoryTheory/Exponentials.v
```diff
@@ -0,0 +1,579 @@
+(**
```
arnoudvanderleer, line 1: I think it is confusing to have two files named `exponentials` and `Exponentials`. That is sure to cause confusion in the future.

UniMath/CategoryTheory/WeakEquivalences/Creation/BinProducts.v
```diff
@@ -45,7 +45,7 @@ Proof.
   set (t := weak_equiv_preserves_binproducts Fw _ _ _ _ _ (pr2 p)).
   pose (PB := make_BinProduct _ _ _ _ _ _ t).
   exact (binproduct_of_isos PB i1 i2).
-Qed.
+Defined.
```
arnoudvanderleer, line 48: This is to be able to use it with univalence, I assume?

Would it, by the way, be possible to formulate this in terms of `weak_equiv_into_univ_creates_binproducts` or a similar lemma?

UniMath/CategoryTheory/WeakEquivalences/LiftPreservation/Exponentials.v
```diff
@@ -0,0 +1,564 @@
+(**
+  The universal property of the Rezk completion states that for every functor F : C₁ → C₂, with C₂ univalent, extends along the (unit of the) Rezk completion to a functor F' : RC(C₁) → C₂.
+  In this file, we proof that if F preserves exponentials, then so does F'.
```
arnoudvanderleer, line 3:
```suggestion
  In this file, we prove that if F preserves exponentials, then so does F'.
```

UniMath/Bicategories/RezkCompletions/StructuredCats/Exponentials.v
```diff
@@ -0,0 +1,71 @@
+(*
```
arnoudvanderleer, line 1:
```suggestion
(**
```

UniMath/CategoryTheory/WeakEquivalences/LiftPreservation/Exponentials.v
```diff
@@ -0,0 +1,564 @@
...
+  : preserves_exponentials E₂ E₃ (weak_equiv_lifts_preserves_binproducts C₂ C₃ α G_weq F_pP).
+Proof.
+  use weak_equiv_lifts_preserves_exponentials.
+  { exact P₁. }
+
+  set (t := preserves_exponentials_to_preserves_exponential_objects P₁ P₃ F_pP E₁ E₃ F_pE).
+  exact (preserves_exponential_objects'_to_preserves_exponential_objects P₁ P₃ F_pP t).
+Qed.
+
+(**********************************************)
```
arnoudvanderleer, line 408: There is a large block of commented code here.

UniMath/CategoryTheory/WeakEquivalences/Preservation/Exponentials.v
```diff
@@ -0,0 +1,432 @@
+(**
+   In this file, we show that an arbitrary weak equivalence F : C -> D preserves exponential objects.
+ *)
```
arnoudvanderleer, line 3: Can you document the sections in this file, as described in [the style guide](https://unimath.github.io/UniMath/documentation/contributing/Style-guide.html#add-comments-to-your-code)?

UniMath/CategoryTheory/WeakEquivalences/LiftPreservation/Exponentials.v
KobeWullaert: Yes, the one commented out is a partial proof for the same statement, but using the different notion of preservation. Either I'll delete this, or I can fill in the remaining proof. I'll come back to this.

UniMath/CategoryTheory/exponentials.v
```diff
@@ -900,3 +901,61 @@ Section ExpIndependent.
...
+    use make_nat_z_iso.
+    - use binproduct_nat_trans.
+      + use (nat_trans_comp _ _ _ _ α).
+        apply binproduct_nat_trans_pr1.
+      + use (nat_trans_comp _ _ _ _ β).
+        apply binproduct_nat_trans_pr2.
+    - intro.
+      use (pr2 (binproduct_of_z_iso (P_D _ _) (P_D _ _) (_,,_) (_,,_)))
+      ; apply nat_z_iso_pointwise_z_iso.
+  Defined.
```
arnoudvanderleer, line 926: This is just `binproduct_of_z_iso`, right? And if it is original, I think it should be upstreamed to `BinProducts`, because it is not about exponentials.

UniMath/CategoryTheory/exponentials.v
```diff
@@ -900,3 +901,61 @@ Section ExpIndependent.
...
+    {C : category}
+    {x y : C} (i : z_iso x y)
+    : nat_z_iso (constant_functor C C x) (constant_functor C C y).
+  Proof.
+    use make_nat_z_iso.
+    - use make_nat_trans.
+      + exact (λ _, i).
+      + exact (λ _ _ _, id_left _ @ ! id_right _).
+    - intro ; apply z_iso_is_z_isomorphism.
+  Defined.
```
arnoudvanderleer, line 938: This is a combination of `binproduct_of_z_iso` and the fact any functor, including the "constant functor functor", preserves isos.

UniMath/CategoryTheory/Exponentials.v
arnoudvanderleer: It is possible that even MacOS is confused by this: [the MacOS filesystem is case-insensitive by default](https://www.reddit.com/r/MacOS/comments/qkbasx/macos_filesystem_not_being_case_sensitive/)

nmvdw: The MacOS test in the CI is failing, so that might be the reason.

It is better to prevent that files have the same names.

UniMath/CategoryTheory/Exponentials.v
KobeWullaert: Yes and No. 
Exponentials.is_exponentiable does not allow to reason with exponentials for chosen 2 objects, which is necessary to prove preservation. It is somewhat subtle, but since we are reasoning with eso's we need stuff to be propositions. The definitions are here to have the alternative notion of preservation of exponentials.

UniMath/CategoryTheory/Exponentials.v
KobeWullaert: That is annoying. I could merge the two files, but I thought it was nicer to split it up. The dependency implies that the main file for exponentials is the "{e}xponentials".

UniMath/CategoryTheory/exponentials.v
KobeWullaert: This can indeed be defined via binproduct_of_z_iso in [C, D], I think. But you still need to add the appropriate conversions, say that [C, D] has binproducts, ... I don't think the code will be shorter or more optimized. 
Anyway, it should be moved more upstream indeed, thanks!

UniMath/CategoryTheory/exponentials.v
KobeWullaert: This comment refers to z_iso_of_constant_functors, right? I don't see how this is a combination herefore. Or are you referring to constprod_functor1_mod_iso?

UniMath/CategoryTheory/Exponentials.v
arnoudvanderleer: Can you explain in two sentences why you need the definition for pairs of objects to prove preservation?

KobeWullaert: The two sentences:
To prove is_exponentiable in the (univalent) codomain, you need to use eso'ness 3 times. 
is_exponentiable in terms of is a left adjoint, does not allow for this as you do not have a proposition.

Additional sentences: You are right that strictly speaking it is not necessary. I could just copy-paste the body everywhere, but then it is nicer to make this into a definition. I'm not suggesting that this definition of Exponents.is_exponentiable should be used in general (even though it would make more sense). But this domain-specific application one requires to have the correct setup, which is the purpose of this file.

arnoudvanderleer: Hm... Right. So you need that being a universal arrow is a proposition, and using that pointwise, you show that a functor preserves exponents. Right?

arnoudvanderleer: I believe that a lot of the material that you formulated about exponentials, is actually about adjunctions or universal arrows. A while back, I wrote some (not yet merged) material about universal arrows, which I called ([co](https://github.com/arnoudvanderleer/UniMath/blob/98deb505e498cf61cc93aa8d1271700703850c88/UniMath/CategoryTheory/Adjunctions/Coreflections.v))[reflections](https://github.com/arnoudvanderleer/UniMath/blob/98deb505e498cf61cc93aa8d1271700703850c88/UniMath/CategoryTheory/Adjunctions/Reflections.v) (after their main name on nLab). Would it be an idea if you
 - Adopt that material;
 - Extend it (and `Adjunctions/Core.v`) in the way that you require;
 - Add the exponential-specific material, if any remains, to `exponentials.v`?

If you want, I can help.

UniMath/CategoryTheory/Exponentials.v
```diff
@@ -170,25 +193,37 @@ Section UniquenessUpToEq.
...
     - use (isotoid _ C_univ).
       exact (exponentials_unique_up_to_iso P (pr22 ϕ₁) (pr22 ϕ₂)).
     - rewrite inv_isotoid, idtoiso_isotoid.
-      exact (! pr21 (pr22 ϕ₁ _ (pr12 ϕ₂))).
-  Qed.
+      (* Check (! pr21 (pr22 ϕ₁ _ (pr12 ϕ₂))). *)
+
+
+  (* Qed.*) Admitted. *)
+
```
benediktahrens, line 215: ?

UniMath/CategoryTheory/Exponentials.v
```diff
@@ -261,19 +296,18 @@ Section ExponentiabilityTransportsAlongIso.
...
+    intros [a f].
+    use (iscontrweqb' (pr22 e (a ,, _ · f))).
     - apply BinProductOfArrows.
       + exact i.
       + apply identity.
     - use weqfibtototal ; intro ; apply equiv_of_mor.
   Defined.
 
-End ExponentiabilityTransportsAlongIso.
+End ExponentiabilityTransportsAlongIso. *)
```
benediktahrens, line 310: Here, a long comment closes. Is the stuff in the comment useful?

UniMath/CategoryTheory/Exponentials.v
```diff
@@ -284,14 +318,14 @@ Section ExponentiabilityTransportsAlongIso'.
 
   Context (e : Exponent P x y).
 
-  Lemma equiv_of_mor'
+  (* Lemma equiv_of_mor'
```
benediktahrens, line 321: Comment?

UniMath/CategoryTheory/Exponentials.v
```diff
@@ -305,18 +339,21 @@ Section ExponentiabilityTransportsAlongIso'.
...
     exists (pr1 e).
     exists (pr12 e · j).
-    intros a f.
-    use (iscontrweqb' (pr22 e a (f · z_iso_inv j))).
-    use weqfibtototal ; intro ; apply equiv_of_mor'.
-  Defined.
+    intros [a f].
+    use (iscontrweqb' (pr22 e (a ,, f · z_iso_inv j))).
+    use weqfibtototal ; intro ; apply equiv_of_mor'. *)
+  (* Defined. *) Admitted.
```
benediktahrens, line 356: We should not have admitted results in the library.

benediktahrens (commented):
There seems to be a lot of commented code in this PR. Can we clarify its status before merging?

KobeWullaert:
> There seems to be a lot of commented code in this PR. Can we clarify its status before merging?

@benediktahrens The commented code is now gone. Also, the admit you referred to has been resolved. The admit popped up due to changes made after revision of this PR.

UniMath/CategoryTheory/WeakEquivalences/Reflection/Exponentials.v
```diff
@@ -0,0 +1,155 @@
...
+  Let f₁ : C₁⟦P₁ (F x₀) (F a₀), F y₀⟧ := weak_equiv_z_iso_binprod _ _ · #F f₀.
+
+  Let f₁_corefl : coreflection_data (F y₀) (constprod_functor1 P₁ (F x₀))
+      := make_coreflection_data (F := constprod_functor1 _ _) _ f₁.
+
+  Definition weak_equiv_reflect_exp_lam : C₀⟦a₀, e₀⟧
+    := (fully_faithful_inv_hom (pr2 F_weq) _ _  (pr11 (ev_uvp f₁_corefl))).
+
+  Lemma weak_equiv_reflect_exp_lam_helper (ψ : C₀⟦a₀, e₀⟧)
+    :  BinProductOfArrows C₁ (P₁ _ _) (P₁ _ _)
```
nmvdw, line 61: There are two spaces between the `:` and the `BinProductOfArrows` instead of 1

nmvdw (commented):
This PR contains a refactoring and new material. Could you describe what the refactoring contains? Like: what is happening with the file `Adjunctions/Core.v`? By looking at some concrete examples that changed, I am unconvinced that this refactoring is an actual improvement, because certain proofs become longer and they nest deeper. The removal of `left_adjoint_from_partial` is not a good change, in my opinion, because it gives nice way to construct adjunctions.

I would suggest to keep `left_adjoint_from_partial`, and to keep all proofs that use it in the same way. Its value is that it gives a completely unfolded way to construct adjunctions, and it doesn't require the user to unfold anything themselves. If the goal is to use coreflections for this, then sure, but then refactor `left_adjoint_from_partial` to make use of it.

I would also strongly suggest to make that a change for a different PR, because it is preferable to not put both refactorings and additions in a PR.

UniMath/CategoryTheory/exponentials.v
```diff
@@ -18,6 +18,11 @@
   4. Transport along an adjoint equivalence
```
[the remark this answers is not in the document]
arnoudvanderleer: Are there files where their meaning is unclear? Usually, the location of a file gives a lot of context to interpret the name, right?
 - `CategoryTheory/exponentials` is the file with the definition of exponentials
 - `WeakEquivalences/reflection/exponentials` should probably show that weak equivalence reflect exponentials
 - etc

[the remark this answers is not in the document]
nmvdw: What about names like:
- `ExponentialsBasics` ~> basic definitions for exponentials
- `PreservesExps` ~> preservation of exponentials
- `ReflectsExps` ~< reflection of exponentials
- `CatWithExps` ~> the bicategory of categories with exponentials
- `UnivCatWithExps` ~> the bicategory of univalent categories with exponentials

Or what about using `CCC`? Like `PreservesCCC`

UniMath/CategoryTheory/DisplayedCats/Codomain/CodDomain.v
```diff
@@ -106,13 +107,16 @@ Definition is_left_adjoint_slice_dom
            (x : C)
   : is_left_adjoint (slice_dom x).
 Proof.
-  use left_adjoint_from_partial.
-  - exact (λ z, pr_cod_fib BP x z).
-  - exact (λ z, BinProductPr2 _ (BP x z)).
-  - intros z yf g.
+  use coreflections_to_is_left_adjoint.
```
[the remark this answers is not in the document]
arnoudvanderleer: Hm, that was my doing. I still think that this is a change for the better:
- `left_adjoint_weq_reflections` is a better name than `left_adjoint_from_partial`, because it makes explicit that the "partial" data is given in the form of coreflections, and that this constitutes an equivalence.
- Formerly, we only had a definition about what it means for "something" to be a universal arrow. This PR makes it explicit what the "something" is (reflection data).
- The current structure is an attempt to bundle the data (`reflection_data`) in one definition, and the property (`is_reflection`) in another. I think that this is a good pattern to apply consistently in general. On the other hand, I agree that the current setup is a bit unrefined, and can be improved. Can you point to places where you think the shortcomings really show?

arnoudvanderleer:
> By looking at some concrete examples that changed, I am unconvinced that this refactoring is an actual improvement, because certain proofs become longer and they nest deeper. The removal of `left_adjoint_from_partial` is not a good change, in my opinion, because it gives nice way to construct adjunctions.

Hm, that was my doing. I still think that this is a change for the better:
- `left_adjoint_weq_reflections` is a better name than `left_adjoint_from_partial`, because it makes explicit that the "partial" data is given in the form of coreflections, and that this constitutes an equivalence.
- Formerly, we only had a definition about what it means for "something" to be a universal arrow. This PR makes it explicit what the "something" is (reflection data).
- The current structure is an attempt to bundle the data (`reflection_data`) in one definition, and the property (`is_reflection`) in another. I think that this is a good pattern to apply consistently in general.

On the other hand, I agree that the current setup is a bit unrefined, and can be improved. Can you point to places where you think the shortcomings really show? I can imagine that in these places, we will see that either, the current setup requires one to write code that is "better structured", or that it may be nice to have a second, more unfolded constructor for `coreflection` :-)

nmvdw:
> `left_adjoint_weq_reflections` is a better name than `left_adjoint_from_partial`, because it makes explicit that the "partial" data is given in the form of coreflections, and that this constitutes an equivalence.

Fair point. The name is not important to me

> Formerly, we only had a definition about what it means for "something" to be a universal arrow. This PR makes it explicit what the "something" is (reflection data).

Again, fair point. I do think that this is a good change

> The current structure is an attempt to bundle the data (`reflection_data`) in one definition, and the property (is_reflection) in another. I think that this is a good pattern to apply consistently in general.

Related to the previous: good change. This is an improvement, and actually helpful.

The point I am making is more about the usage. Before, one would just write
···
use right_adjoint_left_from_partial.
···
And one would only go 1 deep when constructing the data. Now this would become.
```
    apply reflections_to_is_right_adjoint.
    intro X.
    use make_reflection.
    - use make_reflection_data.
···
Here you need to go two deep, and you need to use more tactics.

For the laws, this change is an improvement, but for the data I think that one can improve it further.

**Comment by arnoudvanderleer** [comment]
> ```
>     apply reflections_to_is_right_adjoint.
>     intro X.
>     use make_reflection.
>     - use make_reflection_data.
> ```
> 
> Here you need to go two deep, and you need to use more tactics.
> 
> For the laws, this change is an improvement, but for the data I think that one can improve it further.

Fair enough. Note that formerly, the `intro X` would usually be repeated thrice: once for each component. Would it help if we add a constructor `make_reflection'` or `make_reflection_alt` that unfolds `reflection_data`?

File: UniMath/CategoryTheory/WeakEquivalences/Reflection/Exponentials.v [hunk 17]
```diff
@@ -0,0 +1,155 @@
...
+    refine (_ @ weak_equiv_reflect_exp_lam_helper weak_equiv_reflect_exp_lam).
+    apply maponpaths_2, maponpaths.
+    apply pathsinv0, functor_on_fully_faithful_inv_hom.
+  Qed.
+
+  Section WeakEquivReflectsExponentialsObjectsUniqueness.
+
+    Let A := (∑ f' : C₀ ⟦ a₀, e₀ ⟧, f₀ = # (constprod_functor1 P₀ x₀) f' · ev₀).
+
+    Lemma bla (ϕ : A)
```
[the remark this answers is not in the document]
KobeWullaert: Thanks for catching this, I had to come up with a better name, but I had forgotten about this..

UniMath/CategoryTheory/exponentials.v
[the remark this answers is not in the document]
KobeWullaert: This is also the case for the other structures. What Arnoud said was also how I was reasoning, and how it was setup. 
Do you propose to do this for the other structures as well? If so, should I make a separate PR for this, or do I do the all the changes in here?

arnoudvanderleer:
Okay, so the upshot is that we want an additional constructor, `make_(co)reflection'`, which already unfolds `coreflection_data`.

UniMath/CONTENTS.md
```diff
@@ -263,6 +263,8 @@ The packages and files are listed here in logical order: each file depends only
    - [HomotopicalCategory.v](CategoryTheory/HomotopicalCategory.v)
```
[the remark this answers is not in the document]
nmvdw: Remove the whole file

UniMath/CategoryTheory/exponentials.v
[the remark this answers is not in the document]
arnoudvanderleer: I am not sure how I feel about abbreviations in filenames. For example, every time I have to think whether a file name or constructor is about `displayed_categories`, `disp_cats` or `displayed_cats` (because files use one, and definitions another)

KobeWullaert:
The satellite largecatmodules no longer compiles, since it is dependent on universal_arrow_from/to.
Do I fix this after this PR has been merged?

UniMath/CategoryTheory/exponentials.v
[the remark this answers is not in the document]
nmvdw: > This is also the case for the other structures. What Arnoud said was also how I was reasoning, and how it was setup. Do you propose to do this for the other structures as well? If so, should I make a separate PR for this, or do I do the all the changes in here?

It is not fully uniform in UniMath, I think. 

> I am not sure how I feel about abbreviations in filenames. For example, I am often confused whether a file or definition is about displayed_categories, disp_cats or displayed_cats (because files use one, and definitions another).

What do you mean? Note: abbreviations in file names are used commonly.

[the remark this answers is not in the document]
KobeWullaert: I should have been more explicit. I meant the structures considered in the WeakEquivalences folder. Even though it is not fully uniform in UniMath, I think it would be good to have uniformity in the WeakEquivalence folder.

nmvdw:
What about capitalizing the file `exponentials.v` (only one without a capital) and then merging this PR?

UniMath/CategoryTheory/WeakEquivalences/Preservation/Exponentials.v
```diff
@@ -0,0 +1,444 @@
...
+    use (factor_through_squash _ _ (pr1 F_weq x₁)).
+    {
+      apply impred_isaprop ; intro.
+      apply isaprop_Exponent.
+      exact C₁_univ.
+    }
+    intros [x₀ ix].
+    intro y₁.
+    use (Exponent_transport_along_iso' P₁ ix (identity_z_iso y₁)).
+    (* apply (is_exponentiable_alt_closed_under_iso _ ix).
```
nmvdw: This lines can be removed

nmvdw (approved):

PR #2038, opened 2025-05-12
Dagger Monoidal Structure

chai770:
A file on dagger structures in monoidal categories. This file is part of the project to formalize parts of the book Categories for Quantum Theory by Heunen and Vicary.

### DIFF
--- a/UniMath/CONTENTS.md
+++ b/UniMath/CONTENTS.md
@@ -869,6 +869,7 @@ The packages and files are listed here in logical order: each file depends only 
    - [DaggerCategories/Univalence.v](CategoryTheory/DaggerCategories/Univalence.v)
    - [DaggerCategories/Functors.v](CategoryTheory/DaggerCategories/Functors.v)
    - [DaggerCategories/Transformations.v](CategoryTheory/DaggerCategories/Transformations.v)
+   - [Monoidal/DaggerMonoidal.v](CategoryTheory/Monoidal/DaggerMonoidal.v)
    - [DaggerCategories/FunctorCategory.v](CategoryTheory/DaggerCategories/FunctorCategory.v)
    - [DaggerCategories/Examples/Groupoids.v](CategoryTheory/DaggerCategories/Examples/Groupoids.v)
    - [DaggerCategories/Examples/Relations.v](CategoryTheory/DaggerCategories/Examples/Relations.v)
--- a/UniMath/CategoryTheory/Monoidal/DaggerMonoidal.v
+++ b/UniMath/CategoryTheory/Monoidal/DaggerMonoidal.v
@@ -81,6 +81,8 @@ Section DaggerMonoidal.
     := pr1 C.
 
   Coercion mon_dag_cat_to_pre_dag_mon_cat (C : monoidal_dagger_category): pre_dag_monoid_cat := monoidal_dagger_cat_to_pre_dag_mon_cat C.
+
+  End DaggerMonoidal.
   
   Section DaggerBiproducts.
 
@@ -366,3 +368,4 @@ Section DaggerMonoidal.
     : UU
     := ∑ (N : nat), ∏ (n : nat)(p: n > N), (sum_effects (prob_set S a) op) n = identity I_{C}.
 
+  End QuantumNotation.
--- a/UniMath/CategoryTheory/Monoidal/DaggerMonoidal.v
+++ b/UniMath/CategoryTheory/Monoidal/DaggerMonoidal.v
@@ -1,0 +1,368 @@
+(***************************************************************************
+
+ Dagger Monoidal Categories
+
+ In this file we introduce the dagger structure to monoidal categories. Then,
+ we delve into its interaction with biproducts, we define kernels, and finally
+ define part of the quantum computing notation from Heunen and Vicary. 
+ This file is part of a larger project to formalize parts of the book 
+ Categories for Quantum Theory: An Introduction (Heunen and Vicary 2019).
+
+ Note that in the section Quantum Notation a different definition is used for 
+ Born rule than that provided in Heunen and Vicary.
+
+ Contents
+ 
+ 1. Dagger Monoidal Categories
+ 2. Dagger Biproducts
+ 3. Kernels
+ 4. Quantum Notation
+ ***************************************************************************)
+
+
+Require Import UniMath.Foundations.All.
+Require Import UniMath.MoreFoundations.All.
+Require Import UniMath.CategoryTheory.Core.Categories.
+Require Import UniMath.CategoryTheory.Core.Functors.
+Require Import UniMath.CategoryTheory.Monoidal.Categories.
+Require Import UniMath.CategoryTheory.DaggerCategories.Categories.
+Require Import UniMath.CategoryTheory.Limits.Zero.
+Require Import UniMath.CategoryTheory.DaggerCategories.Unitary.
+Require Import UniMath.CategoryTheory.Monoidal.CategoriesForQuantum.
+Require Import UniMath.CategoryTheory.Limits.BinBiproducts.
+
+Import MonoidalNotations.
+
+Local Open Scope cat.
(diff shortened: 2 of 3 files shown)


UniMath/CONTENTS.md
```diff
@@ -869,6 +869,7 @@ The packages and files are listed here in logical order: each file depends only
    - [DaggerCategories/Univalence.v](CategoryTheory/DaggerCategories/Univalence.v)
    - [DaggerCategories/Functors.v](CategoryTheory/DaggerCategories/Functors.v)
    - [DaggerCategories/Transformations.v](CategoryTheory/DaggerCategories/Transformations.v)
+   - [Monoidal/DaggerMonoidal.v](CategoryTheory/Monoidal/DaggerMonoidal.v)
```
nmvdw, line 872: The `CONTENTS.md` file can be deleted. This will solve the merge conflict.

UniMath/CategoryTheory/Monoidal/DaggerMonoidal.v
```diff
@@ -0,0 +1,371 @@
...
+Require Import UniMath.CategoryTheory.DaggerCategories.Unitary.
+Require Import UniMath.CategoryTheory.Monoidal.CategoriesForQuantum.
+Require Import UniMath.CategoryTheory.Limits.BinBiproducts.
+
+Import MonoidalNotations.
+
+Local Open Scope cat.
+Local Open Scope moncat.
+
+Notation "{ f }_ C ^†" := (dagger_category_to_dagger C _ _ f).
```
nmvdw, line 39: It would be good to put this notation in a scope.

UniMath/CategoryTheory/Monoidal/DaggerMonoidal.v
```diff
@@ -0,0 +1,371 @@
...
+    {C : dagger_category}
+    (Z : Zero C)
+    {A B : C}
+    (f : A --> B)
+    : UU
+    := ∑(K : C)(k : K --> A)(k1 : is_isometry k), is_dagger_kernel Z k k1 f.
+  
+  (* Accessors for dagger kernel *)
+
+  Definition ker_obj
```
nmvdw, line 160: Would this one be better as a coercion?

UniMath/CategoryTheory/Monoidal/DaggerMonoidal.v
```diff
@@ -0,0 +1,371 @@
...
+    Proof.
+    exact (pr2 (pr222 T)).
+    Qed.
+
+  Lemma zero_dag
+    {C : dagger_category}
+    (Z : Zero C)
+    {A B : C}
+    : { ZeroArrow Z A B }_ C ^† = ZeroArrow Z B A.
+  Proof.
```
nmvdw, line 212: The indenting is inconsistent between this lemma and the previous definition

nmvdw (approved):
lgtm

UniMath/CategoryTheory/Monoidal/DaggerMonoidal.v
chai770: I'm a bit unsure how to do this. Would I have to go back to the file where the notation is introduced and change that file or should I create a new scope?

nmvdw: It would be fine to put it in the scope `cat`. If this notation is already introduced, then it is not necessary to define it here.

chai770: Ok, so the line should be like this right?

`Notation "{ f }_ C ^†" := (dagger_category_to_dagger C _ _ f): cat.`

Also the notation is already introduced in DaggerCategories/Categories.v, but if I don't have this line above included for some reason it doesn't recognize the notation.

nmvdw: > Ok, so the line should be like this right?
> 
> `Notation "{ f }_ C ^†" := (dagger_category_to_dagger C _ _ f): cat.`

Yes.

> Also the notation is already introduced in DaggerCategories/Categories.v, but if I don't have this line above included for some reason it doesn't recognize the notation.

This is because the notation is defined in a section, and it only is important if it is defined outside a section.

chai770: Ah ok thanks, I've made the change.

UniMath/CategoryTheory/Monoidal/DaggerMonoidal.v
```diff
@@ -0,0 +1,371 @@
...
+Require Import UniMath.CategoryTheory.DaggerCategories.Unitary.
+Require Import UniMath.CategoryTheory.Monoidal.CategoriesForQuantum.
+Require Import UniMath.CategoryTheory.Limits.BinBiproducts.
+
+Import MonoidalNotations.
+
+Local Open Scope cat.
+Local Open Scope moncat.
+
+Notation "{ f }_ C ^†" := (dagger_category_to_dagger C _ _ f): cat.
```
nmvdw, line 39: It might be better to put this notation in `UniMath.CategoryTheory.DaggerCategories.Categories.v`, because then it is available whenever one uses dagger categories. For the remainder, this PR is ready to be merged.

chai770: Yes, that sounds good. I've made the change in the file `UniMath.CategoryTheory.DaggerCategories.Categories.v` but I am still getting an error message of `Syntax Error: Lexer: Undefined token` when removing the notation line from this current file. This error occurs just when `^†` is introduced; maybe this could be fixed by creating a module for the notation?

nmvdw: Could you add the notation `Notation "{ f }_ C ^†" := (dagger_category_to_dagger C _ _ f): cat.` at the end of the file `DaggerCategories.Categories.v`? It currently only is introduced in a section, and then it is only available in that section. By introducing it outside the section, it becomes available globally.

nmvdw: Interestingly enough, I do not get such error. It could be related to the version of Coq/Rocq (I have 9.0.0).

chai770: I'm able to compile it fine when I open it with `coqide UniMath/CategoryTheory/Monoidal/DaggerMonoidal.v` on terminal. But if I copy and paste the contents of this file exactly into a new file in the same IDE, I get the `Syntax Error: Lexer: Undefined token` again. 

In the new file I also get an error if I try `Require Import UniMath.CategoryTheory.Monoidal.DaggerMonoidal.` with the error being:

`Cannot find a physical path bound to
logical path
UniMath.CategoryTheory.Monoidal.DaggerMonoidal.
`

nmvdw: @arnoudvanderleer do you have an idea?

arnoudvanderleer: Okay, so your coqide can process the file. But just to be sure: can you also _compile_ it? I.e. can you generate (`_build/default/`)`UniMath/CategoryTheory/Monoidal/DaggerMonoidal.glob` using `dune` or `coqc`? You can only import the (updated) contents of a rocq file after it has been compiled into a derived `.glob` file.

chai770: Yes, I ran `make clean` and then `make` to compile everything again, and I do generate a DaggerMonoidal `.glob`, `.v`, `.vok`, `.vos` files. Actually I think the problem was just that I was opening coqide without using this branch as a root, because I am no longer getting the same error. That was probably just my mistake then, sorry about that.

nmvdw: So everything is fine now?

PR #2047, opened 2025-06-19
Added proof that KFinite types with decidable equality are Bishop finite

klinashka:
This PR adds a proof that K-Finite types with decidable equality are Bishop finite. I have currently left the proof in the file `Combinatorics.DecidableEqualityKFiniteTypes`, since I am not sure yet where to place all of the proofs and the helper lemmas. Let me know if I should add an issue to correspond to this PR. I will also leave below a short sketch of the proof so that it is easier for you to follow the code. I omitted some implementation details in the proof for the sake of clarity. 

For the proof below I will denote `stn n` to be the `n`th standard finite set.  Additionally, I will abuse notation and say that the terms of `stn (S n)` are 0 ... n. 

Claim: K-Finite types (`kfinstruct`) with decidable equality are Bishop finite `finstruct`. 
Proof
Let `X` be a K-Finite type. Then there exists `n : nat` with a surjection from the `stn n` to `X`. 

The proof goes by induction on `n`, with the type `X` being generalized. 
In the base case, we have a surjection from an uninhabited type to `X`. Thus, `X` must also have no elements and therefore, `X` is equivalent with `stn 0`. 

In the inductive step, we assume that for any type `X` for which there exists a surjection from `stn n` to `X` and has decidable equality is Bishop finite. We have to show that if there exists a surjection `f` from `stn (S n)` to `X` and `X` has decidable equality, then `X` is Bishop finite. 
Let `g : stn n -> X` such that `g (x) := f x`.  If `X` has decidable equality, then it is decidable whether `f (n)` is included in the image of `g` (intuitively, since there is a finite number of terms in `stn n`, we just go through all of them and decide whether `f n` is included). 

We will thus proceed by case analysis on whether `f n` is included in the image of `g`. 
If `f n` is included in the image of `g`, then `g` must be a surjection as well. By the inductive hypothesis `X` is Bishop finite.
If `f n` is not included in the image of `g`, we will denote `y := f n`. We now consider the type `X / y` (pair X (\x -> x != y)) which is inhabited by terms different than `y`. First, we note that `X / y` has decidable equality. Secondly, we note that `X / y + unit` is equivalent to `X`. Lastly, we note that by restricting the codomain of `g` to `X / y`, we obtain a surjection `g1 : stn n -> X / y`. 
By the inductive hypothesis, the type `X / y` is Bishop finite, and thus equivalent to `stn m` for some `m : nat`. To conclude, we have the following chain of equivalences `X <-> X / y + 1 <-> stn m + 1 <-> stn (S m)`. Thus, `X` is Bishop finite.

### DIFF
--- a/UniMath/Combinatorics/KFiniteTypes.v
+++ b/UniMath/Combinatorics/KFiniteTypes.v
@@ -292,7 +292,7 @@ Section iskfinite_isdeceq_isfinite.
       induction (natlehchoice _ _ (natlthsntoleh _ _ lth)).
       + apply (tpair _ (m ,, a)).
         unfold stnfun_singleton_complement, fun_stnsn_to_stnn, make_stn.
-        apply subtypePath; try (apply isPredicate_singleton_complement); cbn.
+        apply subtypePath_prop; cbn.
         induction q. apply maponpaths, stn_eq, idpath. 
       + assert (H : (m ,, lth = lastelement)) by apply stn_eq, b.
         apply fromempty. induction neq. induction H. apply pathsinv0, q.
--- a/UniMath/Combinatorics/StandardFiniteSets.v
+++ b/UniMath/Combinatorics/StandardFiniteSets.v
@@ -2033,7 +2033,7 @@ Definition stnfun_singleton_complement {X : UU} {n : nat} (f : stn (S n) → X) 
       ¬ hfiber (fun_stnsn_to_stnn f) (f lastelement) → stn n → 
       (singleton_complement (f lastelement)).
 Proof.
-  intros.
+  intros X0 X1.
   exists (fun_stnsn_to_stnn f X1).
   intros contra. apply X0.
   eexists. apply contra.
--- a/UniMath/MoreFoundations/DecidablePropositions.v
+++ b/UniMath/MoreFoundations/DecidablePropositions.v
@@ -468,11 +468,6 @@ Qed.
 
   Definition singleton_complement {X : UU} (y : X) : hsubtype X := λ x, make_hProp (x != y) (isapropneg (x = y)).
 
-  Lemma isPredicate_singleton_complement {X : UU} (y : X) : isPredicate (singleton_complement y).
-  Proof.
-    intros x. apply isapropneg.
-  Qed.
-
   Lemma weq_singleton_complement_unit {X : UU} (y : X) : isdeceq X → carrier (singleton_complement y) ⨿ unit ≃ X.
   Proof.
     intros deceqx.
@@ -483,8 +478,7 @@ Qed.
     - intros [[x neq] | tt].
       + induction (deceqx x y).
         * apply fromempty. induction neq. assumption.
-        * cbn. apply maponpaths, subtypePath; 
-          [apply isPredicate_singleton_complement | apply idpath].
+        * cbn. apply maponpaths, subtypePath_prop, idpath. 
       + induction (deceqx y y).
         * induction tt. apply idpath.
         * apply fromempty, b, idpath.


UniMath/Combinatorics/DecidableEqualityKFiniteTypes.v
```diff
@@ -0,0 +1,183 @@
...
+           { unfold g, snton, make_stn. 
+             assert (lth = natlthtolths m n a) by apply proofirrelevance, propproperty. induction X0. apply idpath. }
+           induction X0. assumption.
+        -- apply b0. unfold lastelement. induction b1.
+           assert (lth = natgthsnn m).
+           { apply proofirrelevance, propproperty. }
+           induction X0. assumption.
+Qed.
+
+Lemma kfinstruct_dec_finstruct {X : UU} : kfinstruct X → isdeceq X → finstruct X.
```
nmvdw: Would it be sensible to split this into lemmas?

UniMath/Combinatorics/DecidableEqualityKFiniteTypes.v
```diff
@@ -0,0 +1,183 @@
...
+Proof.
+    intros [n [f surj]] deceqX.
+    generalize dependent X.
```
nmvdw: Would the following have worked as well?
```
intros [n [f surj]].
revert X.
```
In my eyes this seems a bit simpler (although the different is negligible.

UniMath/Combinatorics/DecidableEqualityKFiniteTypes.v
```diff
@@ -0,0 +1,183 @@
...
+    induction n; intros.
+    - exists 0.
+      assert (H : X → ∅).
+      { intros x.  
+        assert (hyp : ∥ hfiber f x ∥) by (exact (surj x)). 
+        eapply squash_to_prop.
+        + apply hyp.
+        + apply isapropempty. 
+        + intros [n eq].
+            apply negstn0; assumption.
```
nmvdw: indenting

UniMath/Combinatorics/DecidableEqualityKFiniteTypes.v
```diff
@@ -0,0 +1,183 @@
...
+        }
+        apply weqiff; [ | apply isapropifnegtrue, negstn0 | apply isapropifnegtrue, H ].
+      + split; [apply f | intros x; apply fromempty; apply H, x ].
+    - set (g := snton f).
+      set (y := f lastelement).
+      induction (isdeceq_isdecsurj g y deceqX).
+      + assert (H : issurjective g).
+        * intros y'.
```
nmvdw: It might be good to use `{ }` rather than `*`. This would make the proof a bit flatter and more readable.

UniMath/Combinatorics/DecidableEqualityKFiniteTypes.v
```diff
@@ -0,0 +1,183 @@
...
+             apply b, pathsinv0, bla.
+Qed.
+
+Definition snton {X : UU} {n : nat} (f : stn (S n) → X) : stn n → X.
+Proof.
+    intros [m lem].
+    apply f. apply (make_stn _  m). apply natlthtolths, lem.
+Defined.
+
+Definition helper_fun {X : UU} {n : nat} (f : stn (S n) → X) : ¬ hfiber (snton f) (f lastelement) → stn n → (∑ (x : X), ¬ (x = (f lastelement))).
```
nmvdw: It might be good to split this statement over multiple lines, so that this line becomes a bit shorter.

UniMath/Combinatorics/DecidableEqualityKFiniteTypes.v
```diff
@@ -0,0 +1,183 @@
...
+Proof.
+    intros.
+    exists (snton f X1).
+    intros contra.
+    apply X0.
+    eexists. apply contra.
+Defined.
+
+Lemma issurjective_helper_fun {X : UU} {n : nat} (f : stn (S n) → X) (nfib : ¬ hfiber (snton f) (f lastelement)) : issurjective f → issurjective (helper_fun f nfib).
```
nmvdw: It might be good to split this statement over multiple lines so that this line becomes shorter

UniMath/Combinatorics/DecidableEqualityKFiniteTypes.v
klinashka: The code you have given does not fully work, because `revert X` seems to attempt to revert only `X`, and not all the terms in the context that depend on `X`. What does work is

```
intros [n [f surj]].
revert f surj X.
```
This is indeed shorter. I used `generalize dependent` just because I felt it reflected better that I wanted to have `X` generalized in the proof.

UniMath/Combinatorics/DecidableEqualityKFiniteTypes.v
```diff
@@ -0,0 +1,183 @@
+Require Import UniMath.Foundations.All.
```
arnoudvanderleer: Since this file is less than 200 lines, I think you can safely add it (the non-generalizable parts, that is) to KFiniteTypes.

arnoudvanderleer: You wrote a very nice outline of the proof in the PR. I think it would be valuable to have a summary of that as documentation in the file. Although since I advised to make this a section of `KFiniteTypes`, I am not sure whether it would be wise to put the summary at the top of that file, or just above the relevant section.

UniMath/Combinatorics/DecidableEqualityKFiniteTypes.v
klinashka: I think there are some lemmas that can be extracted. 
For example, a lemma of type ` (f : stn 0 -> X) -> issurjective f -> X equiv stn 0`, which is what I essentially prove in the base case could seems generally helpful. 
Other than that, I am pretty sure the proof can be reduced at least a fair bit (e.g. I have many places where I show manually that two terms of type `stn n` are equal when they have the first projection equal, and I think there already is a lemma for that in UniMath.).

UniMath/Combinatorics/DecidableEqualityKFiniteTypes.v
klinashka: I think we can decide on that once I reorganize all the code.

nmvdw: If the proof is added to the documentation, then it might be better to keep it in a separate file. Else the file `KFiniteTypes` might become a bit too big.

UniMath/Combinatorics/DecidableEqualityKFiniteTypes.v
klinashka: I've tried to split it as much as I could. I don't think I can do more than this.

UniMath/Combinatorics/DecidableEqualityKFiniteTypes.v
klinashka: I've refactored the code a little and added more lemmas. Now I think some of them would logically belong in other places. (e.g. `isdeceq_total2'`, `surjfromstn0toneg`, `isdeceq_isdecsurj`)

arnoudvanderleer: > If the proof is added to the documentation, then it might be better to keep it in a separate file. Else the file `KFiniteTypes` might become a bit too big.

If so, what would you call the file?

UniMath/Combinatorics/DecidableEqualityKFiniteTypes.v
```diff
@@ -0,0 +1,161 @@
...
+      induction (natlehchoice _ _ (natlthsntoleh _ _ lth)).
+      + apply (tpair _ (m ,, a)).
+        unfold helper_fun, snton, make_stn.
+        apply subtypePath; try (apply isPredicateremoveterm); simpl.
+        induction q. apply maponpaths, stn_eq, idpath. 
+      + assert (m ,, lth = lastelement) by apply stn_eq, b.
+        apply fromempty, neq. induction X0. apply pathsinv0, q.
+Qed.
+
+Lemma isdeceq_total2' {X : UU} (P : X → UU) :
```
nmvdw: This lemma can be moved here:
https://github.com/UniMath/UniMath/blob/master/UniMath/MoreFoundations/DecidablePropositions.v

UniMath/Combinatorics/DecidableEqualityKFiniteTypes.v
```diff
@@ -0,0 +1,161 @@
...
+        [apply isPredicateremoveterm | apply idpath].
+    + induction (deceqx y y).
+      * induction tt. apply idpath.
+      * apply fromempty, b, idpath.
+  - intros x; cbn beta.
+    induction (deceqx x y); simpl;
+    try induction a; apply idpath.
+Qed.
+
+Definition snton {X : UU} {n : nat} (f : stn (S n) → X) : stn n → X.
```
nmvdw: What does `snton` stand for? Would this definition fit in https://github.com/UniMath/UniMath/blob/master/UniMath/Combinatorics/StandardFiniteSets.v?

UniMath/Combinatorics/DecidableEqualityKFiniteTypes.v
```diff
@@ -0,0 +1,161 @@
...
+
+Lemma isdeceq_total2' {X : UU} (P : X → UU) : 
+  isdeceq X → isPredicate P → isdeceq (∑ x : X, P x).
+Proof.
+  intros deceqX predP. apply isdeceq_total2.
+  + apply deceqX.
+  + intros x. apply isdeceqifisaprop, predP.
+Qed. 
+
+Lemma surjfromstn0toneg {X : UU} (f : ⟦ 0 ⟧ → X ) : (issurjective f) → ⟦ 0 ⟧ ≃ X.
```
nmvdw: Would this lemma fit in https://github.com/UniMath/UniMath/blob/master/UniMath/Combinatorics/StandardFiniteSets.v?

UniMath/Combinatorics/DecidableEqualityKFiniteTypes.v
```diff
@@ -0,0 +1,161 @@
...
+      * intros [[m lth] eq].
+        induction (natlehchoice _ _ (natlthsntoleh _ _ lth)).
+        -- apply b. apply tpair with (pr1 := (m ,, a)).
+           induction eq. unfold g, snton, make_stn.
+           apply maponpaths, stn_eq, idpath.
+        -- apply b0. unfold lastelement. induction b1, eq.
+           apply maponpaths, stn_eq, idpath.
+Qed.
+
+Lemma surjstnon {X : UU} {n : nat} (f : ⟦ S n ⟧ → X) : isdeceq X → issurjective f →
```
nmvdw: If I see correctly, this one only depends on `snton` (of the identifiers defined in this file). So, it might fit in https://github.com/UniMath/UniMath/blob/master/UniMath/Combinatorics/StandardFiniteSets.v

nmvdw (commented):
I added some suggestions on where some lemmas could be moved

UniMath/Combinatorics/DecidableEqualityKFiniteTypes.v
nmvdw: There could be a file focussed on decidable equality and finiteness called `DecidableEqualityAndFinite`. This is because there are many notions of finiteness, and generally they simplify if one assumes decidable equallity.

UniMath/Combinatorics/DecidableEqualityKFiniteTypes.v
klinashka: The name `snton` is bad. This function basically restricts the domain of the function. Similarly, the other helper function restricts the codomain. I still have to think of an appropriate name for them.

UniMath/Combinatorics/DecidableEqualityKFiniteTypes.v
klinashka: I was thinking either in the file you mentioned, or in [https://github.com/UniMath/UniMath/blob/master/UniMath/Combinatorics/FiniteSets.v](), as the equivalence can be rewritten as `nelstruct 0 X`, which would belong to the `FiniteSets` directory.

klinashka:
I have moved all of the proofs from the separate file throughout the library. I also added the informal proof of the theorem as a comment. If any of the results feel out of place, I can revert this commit and move them back as they were.

UniMath/Combinatorics/DecidableEqualityKFiniteTypes.v
arnoudvanderleer: Then `snton f` is just `f \circ dni lastelement` (or `dni_lastelement`), right?

UniMath/Combinatorics/KFiniteTypes.v
```diff
@@ -241,3 +242,115 @@ Section kfinite_definition.
     := hinhfun kfinstruct_finstruct.
 
 End kfinite_definition.
+
+Section iskfinite_isdeceq_isfinite.
```
nmvdw, line 246: I would indent in the section to indicate that you are in the section

UniMath/Combinatorics/KFiniteTypes.v
```diff
@@ -241,3 +242,115 @@ Section kfinite_definition.
...
+(*
```
nmvdw:
```suggestion
(**
```
By using two `*`, it becomes `RocqDoc` rather than a comment

UniMath/Combinatorics/KFiniteTypes.v
```diff
@@ -241,3 +242,115 @@ Section kfinite_definition.
...
+  - intros x; cbn beta.
+    induction (deceqx x y); simpl;
+    try induction a; apply idpath.
+Qed.
+
+Lemma issurjective_stnfun_excludefixed {X : UU} {n : nat} (f : stn (S n) → X) 
+      (nfib : ¬ hfiber (fun_stnsn_to_stnn f) (f lastelement)) : issurjective f → 
+      issurjective (stnfun_excludefixed f nfib).
+Proof.
+    intros surjf y.
```
nmvdw, line 285: The indenting is different in this function (previous function uses 2 spaces and this one uses 4)

UniMath/Combinatorics/DecidableEqualityKFiniteTypes.v
klinashka: Yea, that is correct. Should I then remove the definition? Or define the function in terms of `dni_lastelement`?

nmvdw:
This can be merged, right?

UniMath/Combinatorics/KFiniteTypes.v
```diff
@@ -347,10 +347,10 @@ Proof.
...
 
-Lemma iskfinitedectoisfinite {X : UU} : isdeceq X → iskfinite X → isfinite X.
-Proof.
-  intros deceqX. apply hinhfun, kfinstruct_dec_finstruct, deceqX.
-Qed.
+  Lemma iskfinitedectoisfinite {X : UU} : isdeceq X → iskfinite X → isfinite X.
+  Proof.
+    intros deceqX. apply hinhfun, kfinstruct_dec_finstruct, deceqX.
+  Qed.
 End iskfinite_isdeceq_isfinite.
```
arnoudvanderleer: GitHub warns that a couple of these files don't end in a newline.
Making sure that every file ends in a newline would be a good (albeit minor) good habit. If you don't do it, and someone wants to add something new to the file, they add a newline character at the end of your last line, and git will notice that that line has changed (even though there was no substantial change to that line, but only to the lines below).
Usually, you can configure an IDE (like vscode) to automatically make sure that every file ends in a newline.

UniMath/Combinatorics/KFiniteTypes.v
```diff
@@ -241,3 +242,115 @@ Section kfinite_definition.
...
+  y := f n. We now consider the type X / y (pair X (\x -> x != y)) which is inhabited by 
+  terms different than y. First, we note that X / y has decidable equality. Secondly, we note 
+  that X / y + unit is equivalent to X. Lastly, we note that by restricting the codomain of g 
+  to X / y, we obtain a surjection g1 : stn n -> X / y. By the inductive hypothesis, the type 
+  X / y is Bishop finite, and thus equivalent to stn m for some m : nat. 
+  To conclude, we have the following chain of equivalences 
+  X <-> X / y + 1 <-> stn m + 1 <-> stn (S m). Thus, X is Bishop finite.
+
+  *)
+  Definition removeterm {X : UU} (y : X) : X → UU := λ x, x != y.
```
arnoudvanderleer: Two comments about this one:
 - I believe `removeterm` should have type `hsubtype X`, and should be located in `MoreFoundations/Subtypes`
 - Can you make sure that the names of your definitions and lemmas consist of words, separated by underscores? So `removeterm` would be `remove_term`. Although maybe something like `singleton_complement` would be more appropriate?

UniMath/Combinatorics/KFiniteTypes.v
```diff
@@ -241,3 +242,115 @@ Section kfinite_definition.
...
+
+  Lemma isPredicateremoveterm {X : UU} (y : X) : isPredicate (removeterm y).
+  Proof.
+    intros x. apply isapropneg.
+  Qed.
+
+  Lemma removetermequiv {X : UU} (y : X) : isdeceq X → (∑ (x : X), ¬ (x = y)) ⨿ unit ≃ X.
```
arnoudvanderleer: Here again two comments:
 - Can you add underscores to the name here?
 - If `remove_term` or whatever it will be called is a `hsubtype`, then `(∑ (x : X), ¬ (x = y))` can also just be written as `remove_term y` (which uses the coercion `carrier : hsubtype -> UU`)

UniMath/Combinatorics/KFiniteTypes.v
```diff
@@ -241,3 +242,115 @@ Section kfinite_definition.
...
+  Proof.
+    intros deceqx.
```
arnoudvanderleer: Are you aware that the `intro` tactic also exists? (But `intros` is also fine, of course)

UniMath/Combinatorics/KFiniteTypes.v
```diff
@@ -241,3 +242,115 @@ Section kfinite_definition.
...
+    - exact (hfiber f x).
+    - exact (surjf x).
+    - apply propproperty.
+    - intros [[m lth] q]; clear surjf. apply hinhpr.
+      induction (natlehchoice _ _ (natlthsntoleh _ _ lth)).
+      + apply (tpair _ (m ,, a)).
+        unfold stnfun_excludefixed, fun_stnsn_to_stnn, make_stn.
+        apply subtypePath; try (apply isPredicateremoveterm); simpl.
+        induction q. apply maponpaths, stn_eq, idpath. 
+      + assert (m ,, lth = lastelement) by apply stn_eq, b.
```
arnoudvanderleer: `assert` introduces a variable `X0` that you use later on. Can you give it a name explicitly? Like `assert (X0 := m ,, lth = lastelement)`. That will be more future-proof.

UniMath/Combinatorics/KFiniteTypes.v
```diff
@@ -241,3 +242,115 @@ Section kfinite_definition.
...
+      apply surjfromstn0toneg with (f := f). assumption. 
+    - set (g := fun_stnsn_to_stnn f).
+      set (y := f lastelement).
+      induction (isdeceq_isdecsurj g y deceqX).
+      + apply IHn with (f := g); try apply surj_fun_stnsn_to_stnn; assumption.
+      + set (g' := stnfun_excludefixed f b).
+        set (surjg' := (issurjective_stnfun_excludefixed f b surj)).
+        specialize IHn with (f := g').
+        apply IHn in surjg'; 
+        [ | apply isdeceq_subtype; try assumption; intros x; apply isapropneg ].
```
arnoudvanderleer: @nmvdw What do you think about this way of flattening the proofs?

UniMath/MoreFoundations/DecidablePropositions.v
```diff
@@ -456,6 +456,18 @@ Definition natneq_DecidableProposition : DecidableRelation nat :=
 
 End DecidablePropositions.
 
+Section DecidableEquality.
+
+Lemma isdeceq_subtype {X : UU} (P : X → UU) : 
+  isdeceq X → isPredicate P → isdeceq (∑ x : X, P x).
```
arnoudvanderleer: You could maybe reformulate this with `P : hsubtype X` and `isdeceq P`.

.gitignore
```diff
@@ -29,3 +29,4 @@ _CoqProject
 /..coq_makefile_output.d
 /UniMath/.dir-locals.el
 /_build/
+*/CONTENTS.md
```
arnoudvanderleer: Why did you add `CONTENTS.md` to the gitignore?

UniMath/Combinatorics/.package/files
```diff
@@ -18,4 +18,4 @@ Tests.v
 DecSet.v
 Maybe.v
 MoreLists.v
-Tuples.v
+Tuples.v
```
arnoudvanderleer: This file also needs a newline at the end.

UniMath/Combinatorics/FiniteSets.v
```diff
@@ -618,3 +618,12 @@ Delimit Scope finset with finset.
 Notation "'∑' x .. y , P" := (FiniteSetSum (λ x,.. (FiniteSetSum (λ y, P))..))
                                (at level 200, x binder, y binder, right associativity) : finset.
 (* type this in emacs in agda-input method with \sum *)
+
+(* A surjection from stn 0 is an equivalence *)
+Lemma surjfromstn0toneg {X : UU} (f : ⟦ 0 ⟧ → X ) : (issurjective f) → ⟦ 0 ⟧ ≃ X.
+Proof.
```
arnoudvanderleer: Why did you put this in `FiniteSets` and not `StandardFiniteSets`?

UniMath/Combinatorics/KFiniteTypes.v
```diff
@@ -241,3 +242,115 @@ Section kfinite_definition.
...
+  We will thus proceed by case analysis on whether f n is included in the image of g.
+  If f n is included in the image of g, then g must be a surjection as well. By the inductive 
+  hypothesis X is Bishop finite. If f n is not included in the image of g, we will denote 
+  y := f n. We now consider the type X / y (pair X (\x -> x != y)) which is inhabited by 
+  terms different than y. First, we note that X / y has decidable equality. Secondly, we note 
+  that X / y + unit is equivalent to X. Lastly, we note that by restricting the codomain of g 
+  to X / y, we obtain a surjection g1 : stn n -> X / y. By the inductive hypothesis, the type 
+  X / y is Bishop finite, and thus equivalent to stn m for some m : nat. 
+  To conclude, we have the following chain of equivalences 
+  X <-> X / y + 1 <-> stn m + 1 <-> stn (S m). Thus, X is Bishop finite.
```
arnoudvanderleer: You can use the unicode notation `≃` in this line :-)

arnoudvanderleer (changes requested):
Looks good in general, but I have some questions and suggestions about your style :-)
(I'm sorry about the late review)

UniMath/Combinatorics/KFiniteTypes.v
klinashka: Yes, I am aware that `intro` tactic exists. I have just made it a habit to always use `intros`. I can change it if you think it is more appropriate to have `intro`.

arnoudvanderleer: Okay, it is a choice. Then it's fine :-)

UniMath/Combinatorics/FiniteSets.v
klinashka: This function constructs an equivalence from `stn 0` to `X`. This is actually the same as `nelstruct X 0` and thus this is a function that basically shows that `X` is a finite set (one aditional lemma for `finstruct X` and one for `isfinite X`. That is why I believed it is more appropriate to have it here instead of `StandardFiniteSets`.

.gitignore
klinashka: when I ran `make` locally it automatically added the `CONTENTS.md` to my file, which was registered as an addition in my commit. I then assumed that it is not necessarily intended to have the `CONTENTS.md` in the repository. That is why I added it to `.gitignore`. I'll revert this change if you actually want the `CONTENTS.md` file in the repo.

UniMath/Combinatorics/FiniteSets.v
arnoudvanderleer: Fair enough. Would it be an idea to then formulate this as
```suggestion
(* A surjection from stn 0 is an equivalence *)
Lemma surjfromstn0toneg {X : UU} (f : ⟦ 0 ⟧ → X ) : (issurjective f) → nelstruct 0 X.
Proof.
```

klinashka: Yes, I think that is a good idea. Can you change in the suggestion `nelstruct X 0` with `nelstruct 0 X`, so that I can commit the suggestion directly?

.gitignore
arnoudvanderleer: Ah righ. This is a remnant of a recent change. We should probably remove `CONTENTS.md` from the makefile.
Can you revert your change to the gitignore for now?

UniMath/Combinatorics/KFiniteTypes.v
klinashka: I've checked the proof term for this lemma. It does not contain any pattern matching. (the brackets just fill the goals that were done by the previous goal for this.) There are some `let` expressions coming from my use of `set` and `assert` probably. Additionally, (if I read correctly) in the old style guide it mentions to avoid brackets for `intros` tactics and nested brackets for `induction`. I'll look further into the other definitions to ensure I avoid introducing pattern matching in proof terms as much as possible.

arnoudvanderleer: Oh, sorry, that is not what I meant. There are two kinds of square brackets:
 - Destructuring brackets, which use `match` when you nest them, I believe. Nesting these is discouraged.
 - Branching brackets, which you use here. They make a proof flatter and denser, but this can reduce readability. I have to confess that I also sometimes write quite dense and unreadable proofs, but now you did this, I was wondering what Niels thought about it :-)

nmvdw:
What still need to be done before this can be merged?

UniMath/Combinatorics/KFiniteTypes.v
klinashka: I tried to have `removeterm` as an `hsubtype`, but I found that it added a lot of complexity, due to having to deal with an `hProp`, instead of the standard structure. However, if you try to standardize things and want to have it as an `hsubtype`, I will definitely try to do that.

arnoudvanderleer: Can you give one or two examples of this added complexity?

klinashka: These issues might also come from my inability to deal with coercions neatly and not being fully aware of all the lemmas of `hProp`, but here are some issues I encountered. 
- I could not find any lemma specifying that if the first projections of terms of type `hsubtype` are equal, then the terms should be equal. For the approach I have used, there is the lemma `subtypePath`, which does that. 
- I have found myself having to do induction on a term of type `isdeceq X x y`. It then it was easier to handle the contradictory cases with dealing with the type `x = y -> empty`, rather than having it wrapped around an `hProp`.
- I personally struggled with the `carrier` structure and using it in the same way I used the dependent pair.

arnoudvanderleer: - "terms of type `hsubtype` are equal". Do you mean that you want to show equality of `f, g: hsubtype X` (so `f, g : X -> hProp`?) Or do you mean equality of `s, t: carrier S` for `S : hsubtype X`? The latter has multiple proofs, for example `carrier_eq` or `subtypePath_prop`.
- If you have `x : carrier (singleton_complement y`, then `pr2 x : pr1carrier _ x != y`. I agree that this is one additional step, but not a large one (and bundling information is often useful for passing complicated data around).
- Hm. It is still a dependent pair, with some additional features (like `carrier_eq`, for example).

Does this help in any way?

klinashka: I have made the modifications to `hsubtype X`.

UniMath/MoreFoundations/DecidablePropositions.v
```diff
@@ -458,13 +458,40 @@ End DecidablePropositions.
...
+    intros deceqx.
+    use weq_iso.
+    - intros [[x neq] | tt]; [apply x | apply y].
+    - intros x. induction (deceqx x y); [right; apply tt | left].
+      apply tpair with (pr1 := x), b. 
+    - intros [[x neq] | tt].
+      + induction (deceqx x y).
+        * apply fromempty. induction neq. assumption.
+        * cbn. apply maponpaths, subtypePath; 
+          [apply isPredicate_singleton_complement | apply idpath].
```
arnoudvanderleer, line 487:
```suggestion
          [apply propproperty | apply idpath].
```
(and the same for the other occurrences of `isPredicate_singleton_complement` ;-))

arnoudvanderleer:
```suggestion
        * cbn. apply maponpaths, subtypePath_prop.
```
This should work, right?

UniMath/Combinatorics/KFiniteTypes.v
```diff
@@ -317,11 +291,11 @@ Section iskfinite_isdeceq_isfinite.
     - intros [[m lth] q]; clear surjf. apply hinhpr.
       induction (natlehchoice _ _ (natlthsntoleh _ _ lth)).
       + apply (tpair _ (m ,, a)).
-        unfold stnfun_excludefixed, fun_stnsn_to_stnn, make_stn.
-        apply subtypePath; try (apply isPredicate_singleton_complement); simpl.
+        unfold stnfun_singleton_complement, fun_stnsn_to_stnn, make_stn.
+        apply subtypePath; try (apply isPredicate_singleton_complement); cbn.
```
arnoudvanderleer:
```suggestion
        apply subtypePath_prop; cbn.
```

UniMath/Combinatorics/StandardFiniteSets.v
```diff
@@ -2029,9 +2029,9 @@ Defined.
 (* Functions with domain in Standard finite sets. *)
 Definition fun_stnsn_to_stnn {X : UU} {n : nat} (f : stn (S n) → X) : stn n → X := f ∘ dni_lastelement.
 
-Definition stnfun_excludefixed {X : UU} {n : nat} (f : stn (S n) → X) : 
+Definition stnfun_singleton_complement {X : UU} {n : nat} (f : stn (S n) → X) : 
       ¬ hfiber (fun_stnsn_to_stnn f) (f lastelement) → stn n → 
-      (∑ (x : X), ¬ (x = (f lastelement))).
+      (singleton_complement (f lastelement)).
 Proof.
   intros.
```
arnoudvanderleer: Sorry, I did not see this one before. Please explicitly name any variable you introduce (and use):
```suggestion
  intros X0 X1.
```

UniMath/Combinatorics/StandardFiniteSets.v
```diff
@@ -2025,3 +2025,55 @@ Lemma stn_ord_bij {n : nat} (f : ⟦ n ⟧ ≃ ⟦ n ⟧) :
...
+  - induction a. apply hinhpr, tpair with (pr1 := x), eq.
+  - apply (squash_to_prop (surj y')); try apply propproperty.
+    intros m. induction m as [m eq']; induction m as [m lth]. apply hinhpr.
+    induction (natlehchoice _ _ (natlthsntoleh _ _ lth)).
+    + apply tpair with (pr1 := m ,, a).
+      unfold fun_stnsn_to_stnn, make_stn, funcomp.
+      induction eq'. apply maponpaths, stn_eq, idpath.
+    + induction eq', b0. apply fromempty, b. 
+      apply maponpaths, stn_eq, idpath.
+Qed.
```
arnoudvanderleer: Although... Can you maybe also add a newline here?

PR #2054, opened 2025-07-16
massive overhaul of the refinement of an M type by the Rech construction

rmatthes:
the crucial element is the total rewrite of the proof of Lemma P_isaprop but there are many more changes described in the header info of the file

### DIFF
--- a/UniMath/Induction/M/ComputationalM.v
+++ b/UniMath/Induction/M/ComputationalM.v
@@ -9,6 +9,21 @@
 
     Author: Dominik Kirst (@dominik-kirst) and Ralph Matthes (@rmatthes)
 
+    massive overhaul by Ralph Matthes (May to July 2025):
+    - clearer mathematical display of what the refinement brings w.r.t.
+      the supposed final coalgebra
+    - computable elements are now called purely coiterative elements, and
+      their property is isolated
+    - more typing information given
+    - better use of documentation facilities
+    - new lemma destrM'_aux to encapsulate the equational part of destrM'
+    - more readable proof of eq_corecM0
+    - the crucial element of the overhaul is the total rewrite of the proof
+      of Lemma P_isaprop that now tries to generate the new goals as much as
+      possible, while the previous version announced the intermediate goals
+      through intermediate_weq with formulas conceived in the UniMath 2017
+      school formalization
+
 *)
 
 Require Import UniMath.Foundations.All.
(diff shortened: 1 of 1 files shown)


UniMath/Induction/M/ComputationalM.v
```diff
@@ -30,54 +45,79 @@ Require Import UniMath.Induction.M.Uniqueness.
...
   Context (B : A → UU).
-  Local Notation F := (polynomial_functor A B).
+
+  Local Definition F : type_precat ⟶ type_precat
+    := polynomial_functor A B.
 
   Variable M0 : coalgebra F.
-  Local Notation carrierM0 := (coalgebra_ob _  M0).
-  Local Notation destrM0 := (coalgebra_mor _ M0).
+  Local Definition carrierM0 : type_precat := coalgebra_ob F M0.
```
nmvdw, line 51: Would a `Let` also be fine here? One can also use a local definition outside of the section, but that is not so with a `Let`

UniMath/Induction/M/ComputationalM.v
```diff
@@ -221,28 +223,33 @@ Section Refinement.
...
+  Lemma corec_computation (C : coalgebra F) (c : coalgebra_ob F C) :
+    destrM (corecM C c) = # F (corecM C) (coalgebra_mor F C c).
   Proof.
     apply idpath.
   Defined.
 
-  (* The two carriers are equal *)
-
-  Lemma eq_corecM0 m0 :
+  (** for the moment only an experiment *)
```
nmvdw, line 224: Is this lemma meant to be proven in the future?

nmvdw (approved):
looks good to me (up to a couple of small comments)

UniMath/Induction/M/ComputationalM.v
```diff
@@ -9,6 +9,21 @@
...
+      their property is isolated
+    - more typing information given
+    - better use of documentation facilities
+    - new lemma destrM'_aux to encapsulate the equational part of destrM'
+    - more readable proof of eq_corecM0
+    - the crucial element of the overhaul is the total rewrite of the proof
+      of Lemma P_isaprop that now tries to generate the new goals as much as
+      possible, while the previous version announced the intermediate goals
+      through intermediate_weq with formulas conceived in the UniMath 2017
+      school formalization
```
arnoudvanderleer: Would it be an idea to have some sort of separation of concerns, where you add this information (via the description of the PR) to the git history instead of the code?

UniMath/Induction/M/ComputationalM.v
rmatthes: Only an adaptation of it - and this if I am lucky. I guess the name "Lemma" is misplaced then. Should I use "Goal"?

UniMath/Induction/M/ComputationalM.v
rmatthes: I agree that I would better focus on the essential part of the change in the header info. And that I think is still justified since this change is not by the original authors.

UniMath/Induction/M/ComputationalM.v
rmatthes: I expected to use the development of this file outside in applications.

UniMath/Induction/M/ComputationalM.v
nmvdw: I am fine with either `Lemma` or `Goal`. I was mainly wondering about the comment: `for the moment only an experiment`. It is just trying something out: maybe it gets deleted, maybe it survives

UniMath/Induction/M/ComputationalM.v
arnoudvanderleer: Thanks for changing this. Note that if you squash and merge, the individual commit messages disappear, so I'd recommend putting the extended description in the PR description, instead of, or in addition to, "the crucial element ... the file".

rmatthes: But cannot we do a normal merge (not squashing) for such a small sequence of commits?

arnoudvanderleer: I'm not sure whether there is a policy. I guess we can :man_shrugging:

rmatthes: We might express our opinions again in discussion #1678 .

rmatthes:
I'll try a "rebase and merge" now.

PR #2055, opened 2025-07-17
reverts the definition of adjunctions back to precategories

rmatthes:
this reverts the move from precategories to categories as parameters of adjunctions 
the change is limited to replacement of category by precategory, no further adaptations are made 
hence, it is the notion of adjunction that becomes again available for precategories, not the results
the motivation is the construction of an adjunction in the analysis of coalgebras for polynomial functors (endofunctors on type_precat that notoriously does not have homsets) - to be proposed very soon


rmatthes:
For such a trivial PR, I think I can myself merge. I just wanted to see if there is any strong reaction concerning the philosophy of UniMath.

PR #2062, opened 2025-09-24
Only run CI for non-draft PRs

nmvdw:
This is taken from @fredrik-bakke, and the explanation is as follosw

Clause by clause:
- `github.event_name == 'workflow_dispatch'`: the job will run if it was manually dispatched.
- `github.event_name == 'push'`: the job will run if it was triggered by a push commit, e.g., when a commit is made to the master branch (note that you must specify your triggers in the on block, e.g., if the workflow is triggered on pushes to master or any other branch.)
- `(github.event_name == 'pull_request' && github.event.pull_request.draft == false)`: the job will run if it is triggered by a pull request commit and the pull request is not a draft


nmvdw:
This seems to work. So, the changes are as follows:
- When a PR is a draft, the CI does not run
- When a PR is marked 'ready for review', the CI runs

We can use this to save CI minutes if we incorporate the following policy:
- PRs are draft by default
- Even if a PR is a draft, we still give feedback if the author wants so

.github/workflows/build-unimath.yml
```diff
@@ -5,6 +5,8 @@ on:
     branches: [master]
   pull_request:
     branches: [master]
+    types:
+      - ready_for_review
```
fredrik-bakke, line 9: You might also want to add
```suggestion
    types:
      - ready_for_review
      - opened
      - reopened
      - synchronize
```

Due to the `if`condition you add below, this will make it so if a pull request is opened or reopened in ready for review mode, or commits are pushed on a ready-for-review PR, then the CI is restarted.

nmvdw: Thank you :D

PR #2077, opened 2026-04-28
Ensure that some evars get unshelved

Tragicus:
https://github.com/rocq-prover/rocq/pull/21180 slightly changes the behavior of the shelving of evars. There are a few situations here where some evar used to be unshelved (which is debatable) and is not anymore. In the cases here, my PR should ideally not change anything but I could not find a way to keep the previous behavior. Considering how small the fix is, can we consider going through?


nmvdw (approved):

nmvdw:
@rmatthes Do you want to take a look as well?

rmatthes:
> @rmatthes Do you want to take a look as well?

Yes, thanks for asking. I'll look into it this afternoon (Paris time).

rmatthes:
Thank you for having given me the time to see what I should do differently in the future. However, I did not try out with the cutting edge upstream versions.

PR #2082, opened 2026-06-11
The presheaf model of type theory

nmvdw:
This PR contains a formalisation of the presheaf model of type theory together with various type formers (∏, ∑, =, subobject classifier, N). For this purpose, I also proved some theorems about Beck-Chevalley conditions that allows one to use chosen pullbacks rather than arbitrary pullbacks.


UniMath/CategoryTheory/Categories/CategoryOfElementsPsh.v
```diff
@@ -0,0 +1,149 @@
+(**
+
+ The category of elements
+
+ Every presheaf gives rise to a category of elements. If we have a presheaf `F` on `C`,
+ then the objects of the category of elements `∫ F` are pairs of objects `c : C` with
+ an elements `x : F c`.
```
rmatthes:
```suggestion
 an element `x : F c`.
```

rmatthes:
May I ask a silly question? It seems that `rewrite` does a lot here (in the Beck-Chevalley conditions). In the past, rewriting failed so often that applying lemmas became the standard. Is the present situation favorable, or has something happened to Rocq?

nmvdw:
> May I ask a silly question? It seems that `rewrite` does a lot here (in the Beck-Chevalley conditions). In the past, rewriting failed so often that applying lemmas became the standard. Is the present situation favorable, or has something happened to Rocq?

In my experience, rewriting is generally quite effective, especially for equational reasoning in category theory. When it fails, it is due to unification. I guess that  it works better now than in the past is due to improvements in the unification algorithm of Rocq, but I cannot confirm that.

rmatthes:
> > May I ask a silly question? It seems that `rewrite` does a lot here (in the Beck-Chevalley conditions). In the past, rewriting failed so often that applying lemmas became the standard. Is the present situation favorable, or has something happened to Rocq?
> 
> In my experience, rewriting is generally quite effective, especially for equational reasoning in category theory. When it fails, it is due to unification. I guess that it works better now than in the past is due to improvements in the unification algorithm of Rocq, but I cannot confirm that.

Thank you, one might then even try to revisit existing UniMath developments.

UniMath/CategoryTheory/Categories/CategoryOfElementsPsh.v
```diff
@@ -0,0 +1,149 @@
+(**
+
+ The category of elements
+
+ Every presheaf gives rise to a category of elements. If we have a presheaf `F` on `C`,
+ then the objects of the category of elements `∫ F` are pairs of objects `c : C` with
+ an element `x : F c`.
+
+ Note: the category of elements for a presheaf is also constructed in ElementsOp.v. The
```
benediktahrens, line 9: Can we add a pointer in `ElementsOp.v` to this file?

benediktahrens (approved):

UniMath/CategoryTheory/Categories/CategoryOfElementsPsh.v
nmvdw: I added a comment in that file

PR #2079, opened 2026-05-22
Modules over monoids and total category of monoids

hugo-s29:
As part of an internship I am doing with @benediktahrens, I added the definitions of modules (and the category $`\mathrm{Mod}(R)`$ they form) over some monoid $`R`$ in a fixed monoidal category $`\mathcal{C}`$, some general examples of modules and a (partial) proof that the category of modules (over some fixed monoid) inherits its colimits from the base category $`\mathcal{C}`$. 

Benedikt suggested me to file a PR with my (partial) proofs to have feedback on how the code could be improved. I have tried to stick with the style guide.

The proof of colimits in $`\mathrm{Mod}(R)`$ is not fully written yet (what remains should be quickly proven), but getting feedback on how code could be improved should help me quite a bit for the rest of the proof. Also, this proof required some general lemmas on colimits that I added to the `Colimits.v` file.

PR #2085, opened 2026-07-02
Module Signatures and Models of Modules Signatures

hugo-s29:
This PR is a follow-up to #2079, as part of my internship with @benediktahrens. It contains definitions and some results regarding module signatures and models of such signatures. Let me know if the code/documentation could be improved, I am open to any suggestion.

I have also renamed `Modules.v` to `RModules.v` (and similarly for `TotalCategoryOfRModules.v`), as discussed in #2079.
Also, `TotalCategoryOfRModules.v` is now using a displayed category.

It was discussed in #2079 to compare modules over a monoid to the modules over a monad. However, for monads, ***left***-modules are defined, while for monoids, ***right***-modules are defined, so the correspondance is awkward and requires swapping the order of the arguments for the monoidal product. I did not prove the correspondance with the "swapped" version, should I? Before noticing that issue, I added an equivalence of categories for the monoids/monad correspondance. 

Module signatures are (displayed) sections of the forgertful functor from TotalCategoryOfRModules to Monoids.
Models (of a model signature $`\Sigma`$) are monoids $`R`$ with a (right) module morphism $`\Sigma(R) \to \mathsf{trivial\_module}(R)`$ . I have added some results about models, e.g., initial models are fixpoints of the $`I + \Sigma(-)`$ funtor, and a result about modularity.

I added a proof that the category of right modules inherits its limits from C (in #2079, it was colimits). I think it should be sufficient to prove that module signatures also inherit their (co)limits from C. I tried proving these results for arbitrary displayed sections, but I was sadly not successful.

In the file about models, I had to use a lemma like `maponpaths` but for dependant functions (`maponpaths_dep`, in the file). It probably exists somewhere else in unimath but I was not able to find it. I will happily change the code to use some already existing lemma.

Also, the file about models is quite large, and if needed, it could be split in two files (definitions + the fixpoint result in one file and the total category and modularity in another). Let me know if I should.


hugo-s29:
I think when I synced my fork of UniMath, something went wrong and it re-added all the commits from my earlier pull-request.

UniMath/CategoryTheory/.package/files
```diff
@@ -799,13 +801,3 @@ MarkovCategories/State.v
 MarkovCategories/ProbabilitySpaces.v
 MarkovCategories/Couplings.v
 MarkovCategories/RelativeProduct.v
-MarkovCategories/AlternativeDefinitions/Stratified.v
```
nmvdw, line 819: These files are removed, but they should not be. For this reason, the checks are failing. Can you add these back?

hugo-s29: Sure thing, I've added them back

hugo-s29: The sanity checks is failing again and in the log, it mentions that "TotalCategoriesOfModules.v" was not found (and also "Modules.v"). However this was the file before it was renamed to "TotalCategoriesOfRModules.v". I looked in the code if "TotalCategoriesOfModules" was mentioned somewhere but did not find anything. Did I miss something?

hugo-s29: The merge/sync that duplicated my commits also re-added the files before they were renamed. Sorry about that. The sanity check is now successful

UniMath/CategoryTheory/Limits/Graphs/Limits.v
```diff
@@ -838,6 +882,7 @@ use (uniqueExists (limUnivProp CC _ _)).
 - simpl; now apply H.
 Qed.
 
+
```
nmvdw, line 885:
```suggestion
```
I think this can be removed?

UniMath/CategoryTheory/Monoidal/Examples/MonadsAsMonoidsElementary.v
```diff
@@ -26,9 +30,15 @@ Section FixACategory.
 
   Context {C : category}.
 
+  Let EndC : monoidal_cat := _ ,, monendocat_monoidal C.
```
nmvdw: Is there an identifier for this `monoidal_cat`? It would be useful to have one, because it is a common construction.

UniMath/CategoryTheory/Monoidal/Examples/MonadsAsMonoidsElementary.v
```diff
@@ -26,9 +30,15 @@ Section FixACategory.
...
+  Let Monoids := category_of_monoids_in_monoidal_cat EndC.
```
nmvdw: Can you add the type declaration for both `Monoids` and `Monads`?

UniMath/CategoryTheory/Monoidal/Examples/MonadsAsMonoidsElementary.v
```diff
@@ -49,36 +59,214 @@ Section MonoidToMonad.
...
   Qed.
 
-  Definition monoid_to_monad_CAT : Monad C
+  Definition monoid_to_monad_CAT : Monads
     := _ ,, _ ,, monoid_to_disp_Monad_laws_CAT.
 
+End OnObjects.
+
+  Lemma monoid_to_monad_map_is_monad_mor {M M' : Monoids} (f : M --> M')
+    : disp_Monad_Mor_laws (monoid_to_disp_Monad_data_CAT M)
```
nmvdw: I am a bit confused here: why is there `monoid_to_disp_Monad_data_CAT`, which should be `Monad_data` rather than something of type `Monad`?

The indenting on the next line could also be improved

UniMath/CategoryTheory/Monoidal/Examples/MonadsAsMonoidsElementary.v
```diff
@@ -49,36 +59,214 @@ Section MonoidToMonad.
...
+      (monoid_to_disp_Monad_data_CAT M') (pr1 f).
+  Proof.
+    induction f as [f [H1 H2]]; split; intro A; cbn in f, H1, H2 |- *.
+    - etrans.
```
nmvdw: It would be better to use `{ }` here to focus on the goal

UniMath/CategoryTheory/Monoidal/Examples/MonadsAsMonoidsElementary.v
```diff
@@ -49,36 +59,214 @@ Section MonoidToMonad.
...
+    split.
+    - intro M.
+      use invmap; [|use path_sigma_hprop|easy].
+      use isaprop_disp_Monad_Mor_laws.
+    - intros M1 M2 M3 f g.
+      use invmap; [|use path_sigma_hprop|easy].
+      use isaprop_disp_Monad_Mor_laws.
+  Qed.
+
+
```
nmvdw:
```suggestion
```

UniMath/CategoryTheory/Monoidal/Examples/MonadsAsMonoidsElementary.v
```diff
@@ -49,36 +59,214 @@ Section MonoidToMonad.
...
+  Lemma monad_to_monoid_CAT_laws :  monoid_laws EndC monad_to_monoid_CAT_data.
   Proof.
     split3; apply (nat_trans_eq C); intro c; cbn.
     - apply Monad_law2.
     - apply Monad_law1.
     - rewrite id_left. apply pathsinv0, Monad_law3.
   Qed.
 
-  Definition monad_to_monoid_CAT_disp : monoid (monendocat_monoidal C) (M : C ⟶ C)
+  Definition monad_to_monoid_CAT_disp : monoid EndC (pr1 M : C ⟶ C)
```
nmvdw: Why do you need to use both `pr1` and `: C ⟶ C`? Could one be removed? Would it be possible to use an accessor as well?

UniMath/CategoryTheory/Monoidal/Examples/MonadsAsMonoidsElementary.v
```diff
@@ -49,36 +59,214 @@ Section MonoidToMonad.
...
+  Definition monad_to_monoid_CAT : Monoids
     := _,,monad_to_monoid_CAT_disp.
 
+End OnObjects.
+
+  Lemma monad_to_monoid_map_is_monoid_mor {M M' : Monads} (f : M --> M')
+    : is_monoid_mor _ (monad_to_monoid_CAT_disp M) (monad_to_monoid_CAT_disp M') (pr1 f).
+  Proof.
+    induction f as [f [H1 H2]]; split.
+    - use invmap; [|use path_sigma_hprop|].
```
nmvdw, line 145: I find it a bit difficult to read this proof, because I feel that there are multiple open goals without focussing.

UniMath/CategoryTheory/Monoidal/Examples/MonadsAsMonoidsElementary.v
```diff
@@ -49,36 +59,214 @@ Section MonoidToMonad.
...
+      use isaprop_is_nat_trans; use homset_property.
+      use funextsec; intro A; cbn.
+      etrans; cycle 1.
```
nmvdw: What does `cycle 1` do?

UniMath/CategoryTheory/Monoidal/Examples/MonadsAsMonoidsElementary.v
```diff
@@ -49,36 +59,214 @@ Section MonoidToMonad.
...
+  use (total2_paths2_f (idpath _)).
+  use invmap; [|use path_sigma_hprop|easy].
+  use isaprop_disp_Monad_laws.
+Qed.
+
+Definition adjunction_monad_monoid 
+  : adjunction_data Monads Monoids.
+Proof.
+  use (make_adjunction_data monad_to_monoid_functor monoid_to_monad_functor);
+  use make_nat_trans.
```
nmvdw: It might be good to make a separate definition for this natural transformation. Perhaps it might also be nice to make separate definitions for the data and the laws.

UniMath/CategoryTheory/Monoidal/Examples/MonadsAsMonoidsElementary.v
```diff
@@ -49,36 +59,214 @@ Section MonoidToMonad.
...
+      [use isaprop_is_nat_trans; use homset_property|];
+      use funextsec; intro A; cbn;
+      now rewrite id_left, id_right
+    ).
+Defined.
+
+Lemma adjunction_monad_monoid_equiv : forms_equivalence adjunction_monad_monoid.
+Proof.
+  use make_forms_equivalence.
+  - intro M; use Isos.make_is_z_isomorphism; [|use Isos.make_is_inverse_in_precat].
```
nmvdw: What about importing the file `Isos`?

UniMath/CategoryTheory/Monoidal/ModuleSignatures.v
```diff
@@ -0,0 +1,189 @@
...
+  Lemma signature_evaluation_is_functor
+    : is_functor signature_evaluation_data.
+  Proof.
+    split.
+    - intro Σ; use invmap; [|use path_sigma_hprop|].
+      use isaprop_is_module_mor.
+      easy.
+    - intros Σ Σ' Σ'' f g; use invmap; [|use path_sigma_hprop|].
+      use isaprop_is_module_mor.
+      cbn; unfold Core.mor_disp; cbn.
```
nmvdw: What about importing `DisplayedCats.Core` so that you do not need to qualify `mor_disp`? Then it might also not be needed to unfold `mor_disp`, but I am not sure

UniMath/CategoryTheory/Monoidal/RModules.v
```diff
@@ -58,10 +58,10 @@ Section Modules.
   Coercion module_object : module_subst >-> ob.
 
   Definition module_laws_assoc {M : C} (p : module_subst M) : UU
-    := α^{m}_{M,R,R} · (M ⊗l μ) · p = p ⊗r R · p.
+    := α^{C}_{M,R,R} · (M ⊗l μ) · p = p ⊗r R · p.
```
nmvdw, line 61: I am wondering: there also is `mon_lassociator` and `mon_rassociator` for `monoidal_cats` (https://github.com/UniMath/UniMath/blob/master/UniMath/CategoryTheory/Monoidal/Categories.v#L1614), which can be more convenient to use. Are those usable here?

UniMath/CategoryTheory/Monoidal/TotalCategoriesOfRModules.v
```diff
@@ -0,0 +1,228 @@
+(***************************************************************************
+
+ Total categories of right modules
+
+ In this file, the total category of right modules in some fixed monoidal 
+ category C. Objects are pairs (R, M) where R is a monoid in C and M is
+ a right module over R.
```
nmvdw, line 7: Can you add a table of contents for this file? I do not see the connection between the header (which is about a total category) and the content (which is about pullback functors)

UniMath/CategoryTheory/Monoidal/ModelsOfModuleSignature.v
```diff
@@ -0,0 +1,943 @@
...
+  Local Definition arrow_from_prod_sum {X Y Z : C}
+    : (X ++ Y) ⊗ Z --> (X ⊗ Z ++ Y ⊗ Z)
+    := BinCoproductArrow 
+      (make_BinCoproduct _ _ _ _ _ _ (Hpres _ _ _ _ _ _ iscopr)) 
+      inl inr.
+
+  Local Lemma arrow_from_prod_sum_inl {X Y Z : C}
+    : inl ⊗r Z · @arrow_from_prod_sum X Y Z = inl.
+  Proof.
+    use (BinCoproductIn1Commutes _ _ _ (make_BinCoproduct _ _ _ _ _ _ (Hpres _ _ _ _ _ _ iscopr))).
```
nmvdw: It might be good to add an identifier to prevent having to write `make_BinCoproduct _ _ _ _ _ _ (Hpres _ _ _ _ _ _ iscopr))`. It can either be globally accessible (in the file `Preservation.v`) or as a local statement in this section

UniMath/CategoryTheory/Monoidal/ModelsOfModuleSignature.v
```diff
@@ -0,0 +1,943 @@
...
+    : (X ++ Y) ⊗l inl · @arrow_from_prod_sum X Y (Z ++ W) 
+    = arrow_from_prod_sum · BinCoproductOfArrows _ _ _ (_ ⊗l inl) (_ ⊗l inl). 
+  Proof.
+    use (
+      BinCoproductArrowsEq _ _ _
+      (make_BinCoproduct _ _ _ _ _ _ (Hpres _ _ _ _ _ _ iscopr))
+      (X ⊗ (Z ++ W) ++ Y ⊗ (Z ++ W))
+    ); cbn.
+    - etrans; etrans; try rewrite assoc; swap 3 4.
+      + use (maponpaths (λ x, x · _)); [shelve|].
```
nmvdw: Why do you use `shelve` here? What about using `refine` instead of `use`?

UniMath/CategoryTheory/Monoidal/ModelsOfModuleSignature.v
```diff
@@ -0,0 +1,943 @@
...
+
+    Local Lemma f_inr : inr · f = rM.
+    Proof.
+      use BinCoproductIn2Commutes.
+    Qed.
+
+
+    (* Three laws for M' to be a monoid *)
+
+    Local Lemma monoid_law_1 : η ⊗r M' · μ = lu^{ C }_{ M'}.
```
nmvdw: What about calling it something like `μ_lunit`? That might be more informative

UniMath/CategoryTheory/Monoidal/ModelsOfModuleSignature.v
```diff
@@ -0,0 +1,943 @@
...
+    Qed.
+
+    Local Lemma monoid_law_3 : α^{ C }_{ M', M', M'} · M' ⊗l μ · μ = μ ⊗r M' · μ.
+    Proof.
+      use (
+          BinCoproductArrowsEq _ _ _ 
+          (make_BinCoproduct _ _ _ _ _ _ (Hpres _ _ _ _ _ _ (Hpres _ _ _ _ _ _ iscopr)))
+          M'
+      ); cbn; repeat rewrite assoc.
+      - symmetry; etrans; [|symmetry;etrans;[etrans|]]; etrans.
```
nmvdw: Here tactics are chained in such a way that the code becomes more compact, but I find it harder to read. I am fine with keeping it as is, but what do you think of chaining less here with the costs of using more lines?

UniMath/CategoryTheory/Monoidal/ModelsOfModuleSignature.v
```diff
@@ -0,0 +1,943 @@
...
+        + rewrite id_left; use f_inl.
+        + use (maponpaths (λ x, x · _)); [|symmetry; use f_inl].
+        + symmetry; use (pr221 h).
+      - rewrite assoc, assoc; etrans; etrans; swap 3 4.
+        + use (maponpaths (λ x, x · _)); [|use BinCoproductOfArrowsIn2].
+        + rewrite <- assoc; use maponpaths; [|use f_inr].
+        + use (maponpaths (λ x, x · _)); [|symmetry; use f_inr].
+        + use (!pr2 h).
+    Qed.
+
```
nmvdw, line 477:
```suggestion
```
Less new lines can be used

UniMath/CategoryTheory/Monoidal/ModelsOfModuleSignature.v
```diff
@@ -0,0 +1,943 @@
...
+          simpl; rewrite transportf_total2;
+          simpl; rewrite transportf_const.
+          simpl; use (maponpaths pr1 (InitialArrowUnique HΣ₁₂ (pr1 Rr ,, _) (pr12 u ,, _))).
+          simpl; etrans; [use (pr22 u)|].
+          use (maponpaths (λ x, _ (pr1 x · _))).
+          exact (
+            maponpaths (λ x, pr1 x (pr1 Rr)) 
+            (PushoutArrowUnique _ _ _ _ _ (isPushout_Pushout H_pushout) _ _ _ (maponpaths pr1 H) _ (maponpaths pr1 H') (maponpaths pr1 H''))
+          ).
+  Qed.
```
nmvdw: I am wondering whether `Qed` is the best choice here. The universal property of pushouts contains data, namely the morphism that comes from the universal property, and now this morphisms cannot be computed. Is this behaviour intended/desirable?

nmvdw (approved):
Thank you for your contribution! Overall, I think it looks good, but I do have several comments.

UniMath/CategoryTheory/Monoidal/Examples/MonadsAsMonoidsElementary.v
hugo-s29: I looked for it but could not find it. The file `Monoidal/Examples/EndofunctorsMonoidalElementary.v` has a proof that `[C, C]` is monoidal but it doesn't define the `monoidal_cat` for `[C, C]`.
I can add this definition as `monendocat_monoidal_cat` in the "Monoidal/Examples/EndofunctorsMonoidalElementary.v" if you want.

UniMath/CategoryTheory/Monoidal/Examples/MonadsAsMonoidsElementary.v
hugo-s29: The type of `disp_Monad_Mor_laws` actually uses arguments of type `disp_Monad_data` instead of `Monad`, and for now, there is no coercion from `Monad` to `disp_Monad_data`.

Image

UniMath/CategoryTheory/Monoidal/Examples/MonadsAsMonoidsElementary.v
hugo-s29: There is a coercion from `Monad` to `C ⟶ C` so the `: C ⟶ C` that was there before was forcing this coercion. However, switching to the category of monads does not that coercion to take place anymore as it now has an object of type `ob category_Monad`, I think (even though `Monad` is defined as `ob category_Monad`).
I switched to using `functor_from_Monad` directly.

UniMath/CategoryTheory/Monoidal/Examples/MonadsAsMonoidsElementary.v
hugo-s29: Here, it's equivalent to `swap 1 2` but in general it rotates the order of the goals (the first goal is moved to be last). I switched to `swap`

UniMath/CategoryTheory/Monoidal/ModelsOfModuleSignature.v
hugo-s29: Thank you for suggesting `refine`.
I was using `shelve` as `use (maponpaths f)` to prove `f x = ?` will ask for the element `y` when the right-hand side is not yet known. 
I didn't know `refine` does not add the extra goal that `use` was adding.
I used that trick quite a few times in my earlier PR, but now I will change those instances to use `refine` instead

UniMath/CategoryTheory/Monoidal/Examples/MonadsAsMonoidsElementary.v
nmvdw: Yes, it would be good to add this

UniMath/CategoryTheory/Monoidal/ModuleSignatures.v
hugo-s29: The `DisplayedCats.Core` was already imported so I don't know why I added the `Core.` prefix. I tried removing the unfolding but it looks like it is necessary for `rewrite transportf_total2` to be successful.

UniMath/CategoryTheory/Monoidal/RModules.v
hugo-s29: Yes, I think using `mon_lassociator` (and others) would work here, but I think it wouldn't shorten/simplify the proofs by a lot (for this file, at least).

UniMath/CategoryTheory/Monoidal/ModelsOfModuleSignature.v
hugo-s29: I think it might be fine to use `Qed` here as `isPushout` is a proposition. Is it useful to get the exact construction used for this morphism? It should not matter too much as that morphism is unique.

I can extract the "center" in a definition and keep the rest in an opaque lemma, if you prefer. 
In `RModules.v`, I also used `Qed` for `isColimCocone` and `isLimCone` so I'll have to change those too.

nmvdw: It is indeed a proposition, and this morphism is determind uniquely. However, there might be certain cases where it might be convenient that one can calculate the center of contraction. I do not know whether it is the case for this definition, but it should be thought about. Specifically, do you think there will be cases where one uses this pushout square and has to calculate this morphism?

nmvdw:
Tell me when I should look at the code again

UniMath/CategoryTheory/Monoidal/ModelsOfModuleSignature.v
hugo-s29: I don't really know. I think you should be able to recover the construction by looking at each component of the morphism.
Just in case it may be useful later on, I made the construction of the induced morphism non-opaque, and kept the "uniqueness" part opaque.

benediktahrens: > I can extract the "center" in a definition and keep the rest in an opaque lemma, if you prefer.  In RModules.v, I also used Qed for isColimCocone and isLimCone so I'll have to change those too.

We usually keep the construction of the morphism transparent, and only make the proofs of the equations opaque. I would suggest to follow that convention.

hugo-s29:
I think I have finished applying your suggestions, so you should be able to read.
There were a lot of changes due to the switch from `use maponpaths` to `refine`.
Let me know if there is anything else to improve.

And thanks a lot for all your suggestions, by the way.

UniMath/CategoryTheory/Monoidal/ModelsOfModuleSignature.v
hugo-s29: > > I can extract the "center" in a definition and keep the rest in an opaque lemma, if you prefer.  In RModules.v, I also used Qed for isColimCocone and isLimCone so I'll have to change those too.
> 
> We usually keep the construction of the morphism transparent, and only make the proofs of the equations opaque. I would suggest to follow that convention.

Okay, in that case, I should also revisit the (co)-limits in RModules to make the morphism transparent.

nmvdw (approved):

nmvdw:
> I think I have finished applying your suggestions, so you should be able to read. There were a lot of changes due to the switch from `use maponpaths` to `refine`. Let me know if there is anything else to improve.
> 
> And thanks a lot for all your suggestions, by the way.

Looks good. Are you also planning to make changes regarding "Okay, in that case, I should also revisit the (co)-limits in RModules to make the morphism transparent."?

hugo-s29:
> Looks good. Are you also planning to make changes regarding "Okay, in that case, I should also revisit the (co)-limits in RModules to make the morphism transparent."?

Yes and I already made the changes earlier today, it should be part of the "[Non-opaque isLimCone and isColimCocone for RModules](https://github.com/UniMath/UniMath/pull/2085/commits/fe5200f4899452d6180f47cb00e3202379f90b1d)" commit (the latest one in this PR).

nmvdw:
> > Looks good. Are you also planning to make changes regarding "Okay, in that case, I should also revisit the (co)-limits in RModules to make the morphism transparent."?
> 
> Yes and I already made the changes earlier today, it should be part of the "[Non-opaque isLimCone and isColimCocone for RModules](https://github.com/UniMath/UniMath/pull/2085/commits/fe5200f4899452d6180f47cb00e3202379f90b1d)" commit (the latest one in this PR).

Nice. Then it can be merged

PR #2087, opened 2026-07-15
Signatures with strength for general monoidal categories, and constructing an initial model

hugo-s29:
In this PR, there is the definition of a signature with strength for a monoidal category, given some simple example of those, and shown that they are closed under (co)limits. There is also the definition of a functor from signature with strength into model signatures and proof that it maps the example constructions to known constructions of model signatures.
Then there is the definition of an heterogeneous substitution system for a signature with strength. An HSS generates a model for that signature with strength (or more precisely a model for the module signature induced by the given signature with strength). There is a construction for an HSS for a given signature with strength using a generalized Mendler iteration. And finally, there is the proof that the model generated by the constructed HSS is an initial model of the given signature with strength.

After doing all these proofs, I noticed that some of the code was redundant with things in `SubstitutionSystems/`, for example the definition of an HSS in `SubstitutionSystems/GeneralizedSubstitutionSystems.v` seems to match my definition (though I'm not sure if the definition of strength I am using matches the definition used in this file). Similarly, there are results about generalized Mendler iteration in `SubstitutionSystems/` (the `..._alt.v` version).

So, for now this is a draft PR, I will have to remove the redundant code and move some files, though I am not sure where everything should go. I am open to suggestions.


rmatthes:
> though I'm not sure if the definition of strength I am using matches the definition used in this file

What creates the suspicion that there is a mathematical difference between the definition of strength used previously (and described in an FSCD'24 paper - in particular Section 2.2) and the one proposed here? Admittedly, the UniMath library has it in a more generic setting, but the laws governing `pointedtensorialstrength` are spelt out concretely as well. A precise comparison will have to overcome that the pointed argument now comes second in the strength. And that the strength for unit and for the product were previously not isolated on one side of the equational laws but that the use of inverses was minimized in those laws. I also wonder why the UniMath construction of the monoidal category of those pointed objects (`monoidal_pointed_objects`) is not used in the present PR. `pointed_unit` and `pointed_prod` are now given without establishing the monoidal laws. I cannot yet judge if their verification will then be deep within proofs of technical lemmas concerning strength. So, the notion of pointed object may be defined identically, but nothing points any longer to those monoidal laws.

Such preliminaries distract from looking at the essential new stuff in this PR.

Edit: the FSCD paper was '24 not '25.

hugo-s29:
Thanks for your comment. Looking at the paper was very helpful to me. 

I do not know why I did not think to use `monoidal_pointed_objects` instead of defining them directly, the definition seems to match exactly. However, it will not matter as I will switch to using the existing definition for an HSS (except instantiating it with the "swapped" monoidal product).

I think my PR contains a lot of duplicate proofs from what was mentioned in your paper. For example, there is already the construction of the HSS using Mendler iteration, and of an initial $`(H, \theta)`$-monoid from this generated HSS. Models for the module signature constructed from a signature with strength are precisely $`(H, \theta)`$-monoids, so I will add this correspondance and rely on the existing construction.

So, in terms of "new" content, there is only the following:
- (co)limits for signatures with strength
- the functor from signatures with strength to module signatures
I will adapt these two things to the "swap" for the arguments in the strength.

rmatthes:
@hugo-s29 Even if there are duplicates, you might have done proofs better (I already see you did simplifications in the upstream material), you might have given clarifying comments, etc. Please try to incorporate them if this does not take too much time. It may well worth giving a second proof of the same lemma (with a telling name) or maybe even just replace the old proof. Making a correspondence of vocabulary is generally virtuous.

Personal comment: my reaction time may be slow since we are suffering a very long heat wave in the region I live in

rmatthes:
@hugo-s29 I see major progress towards a non-draft PR. Next week, I could have a closer look - better give the signal by declaring it a PR. You can nevertheless ask for not merging it if you still want to add stuff.

hugo-s29:
@rmatthes Yes, thanks for noticing, I should have finished adapting what I had proved so I will mark this PR as ready for review. It can be merged if the review goes well. 
I also want to prove an adjunction result for the initial models/$`\Sigma`$-monoids but I think it would be better in a later (smaller) PR.

Regarding the initial model/$`\Sigma`$-monoid, the construction I used is exactly the same, so I simply split the proof of initiality in smaller lemmas.
There was one lemma (`μ · h = Mendler_arrow`) that is an instance of the fusion law for Mendler iteration, and I relied on the fusion law in my own proof. However, the existing proof – that does not use on it – it is a lot simpler (the commutation  of $`\Phi/\psi/\psi'`$ involves some coproduct simplifications), so I kept it and added a comment pointing that out.

benediktahrens:
What is the status of this PR? Is it mergeable?

rmatthes:
I had the firm intention to review this PR. The time since has been rather difficult, in particular concerning the weather, but also because of an incident. If possible, I would still try to go through this work in the next few days.

rmatthes:
I still have 32 degrees Celsius in my office. I am struggling with the prerequisites for proper review. Who is still using ProofGeneral? I cannot make the functions Diffs and Show Proof (Diffs) in the Coq menu of ProofGeneral work. I normally make heavy use of those for code review. (I am on Ubuntu 24.04 LTS, with GNU Emacs 29.3 and ProofGeneral together with company-coq from MELPA).
A minor technical comment: In the file `ModuleSignatures.v`, the bulleting is not always done properly. This can be automatically revealed by adding the line `Require Import UniMath.Tactics.EnsureStructuredProofs.`. The variant with `Require Export` is used several times in the package `SubstitutionSystems` but also elsewhere, in particular to impose proper bulleting on everything involving bicategories.

rmatthes:
I am still not through the entire formalization. Generally, this code is of high quality. I could only make some small suggestions for improvement that I put into a branch of my fork, see the comparison through https://github.com/hugo-s29/UniMath/compare/signatures_with_strength...rmatthes:UniMath:PR2087ToBeReviewed?expand=1 .
The documentation is helpful. Of course, it can by itself not replace a paper presentation, especially concerning where the ideas came from. I wondered if there is something more general behind `signature_evaluation` in `ModuleSignatures.v`, but I did not see something emerge that would make the proof shorter.

I have a doubt on style, seen in `MonadsAsMonoidsElementary.v`: some categories are now locally given names in plural form, such as Monoids and Monads. This does not convey a good intuition when using these identifiers to type a construction. E.g., `monoid_to_monad_CAT : Monads` should be read that a monad is constructed, not monads. I see this differently when packages formalize a structure of related things, and so the package name is good in plural form.

`ModelsOfModuleSignature.v`: the equation governing the model morphisms in `models_of_module_signatures_disp_cat_ob_mor` is not encapsulated in a definition and is then written out in several instances (where only the name of the lemma is telling the nature of the result).

Language: fix-point is not good, fixed point and fixpoint are considered acceptable (but fixed point is better).

`StrengthToModelSignature.v`: in l.7 (in the documentation header), example should be examples

UniMath/SubstitutionSystems/CategoryOfSignaturesWithStrength.v
```diff
@@ -0,0 +1,753 @@
+(***************************************************************************
+
+ Category of signatures with (pointed tensorial) strength
+
+ In this file, we define the category of signatures with a pointed tensorial
```
rmatthes, line 5: It would be good to clarify the relation with existing material. The sister file `SignatureCategory.v` already defines the category of signatures with strength, but not on the abstract level of monoidal categories. Hence, the present file gives more general results. Please drop a note on that.
On the other hand, strength is an instance of an action morphism, and their morphisms are thus the two-cells of the bicategory of actions. Does the preservation of limits and colimits depend on the specific instance of action morphisms used for strength? This would not need to be expressed in bicategorical terms, just with lax lineators and their linear natural transformations.

A more generic question on bicategories would be: if one defines "by hand" (e.g. through a displayed category) a category C, and if there is a bicategory where the objects of C are the 1-cells for fixed source and target and the morphisms of C are their 2-cells, in what sense is C the same as the hom-category for this source-target-pair?

In the concrete situation here:

```
Require Import UniMath.Bicategories.Core.Bicat.
Require Import UniMath.Bicategories.MonoidalCategories.BicatOfActegories.
Require Import UniMath.Bicategories.Core.Examples.BicatOfCats.

  Local Definition mybicat : bicat := actbicat (monoidal_pointed_objects Mon_V).
  Local Definition myhomcat : category
    := hom(C:=mybicat) (V,,actegory_with_canonical_pointed_action Mon_V) (V,,actegory_with_canonical_pointed_action Mon_V).
```

What is the relation between `pointedtensorialstrength_cat` and `myhomcat`? They should be isomorphic.

rmatthes: In fact, it is fairly straightforward to establish `catiso pointedtensorialstrength_cat myhomcat`, hence these two categories are even equal. The practical advantage of `pointedtensorialstrength_cat` is that it does not make use of the bicategories package.

rmatthes:
A comment on so far the latest commit in my repository with suggestions for changes to the present PR: https://github.com/rmatthes/UniMath/commit/3422ca5fb7766a1e443a19f89b94e675d82b2b4b : the problem of the prior commits was solved by not appealing to `cbn` before playing back the equations to those between the first projections of the LHS and RHS (I forgot to mention this in the log message).

I have studied the entire PR and approve it modulo all the comments I already made here and those in my branch https://github.com/rmatthes/UniMath/commits/PR2087ToBeReviewed/ . As I wrote earlier, the code quality is very good. There are questions of style, though. This concerns in particular the aforementioned playing back equations to those between the first projections. The PR uses a recurrent idiom `use invmap; [|use path_sigma_hprop|].`. Interestingly, it looks like more often applicable than the usual `apply subtypePath.` since the latter needs to prove that the entire predicate is propositional, not just the used instance. Still, for me, there is too much boilerplate, and the proof scripts become less type-directed. That is to say, since no specific lemmas are applied, the types of the goals get out of sight. And I see numerous small "bulleting mistakes" related to that practice. In my branch, I solved many of those stylistic problems, and there is definitely no need to deal with all of them. But I looked at this also from the perspective of the entire UniMath library - also I could be mistaken about more recent stylistic choices.

rmatthes:
> Regarding the initial model/ Σ -monoid, the construction I used is exactly the same, so I simply split the proof of initiality in smaller lemmas. There was one lemma (`μ · h = Mendler_arrow`) that is an instance of the fusion law for Mendler iteration, and I relied on the fusion law in my own proof. However, the existing proof – that does not use on it – it is a lot simpler (the commutation of Φ / ψ / ψ ′ involves some coproduct simplifications), so I kept it and added a comment pointing that out.

I highly appreciate the breakdown of the former proof of `SigmaMonoidFromInitialAlgebra_is_initial` into all the small steps providing the data and the intermediary results.

rmatthes (approved):
Good work, but see my numerous comments in this thread that should be addressed before merging the PR.

hugo-s29:
@rmatthes: Thank you for your feedback. I am very sorry for not replying sooner, I was not able to look at this PR for the last week. I am hoping to be able to read read and correct everything by the end of the week, or maybe during the weekend.

rmatthes:
@hugo-s29 : I have fixed the non-compiling satellites and re-ran CI for them. From now on, we will again expect that all the checks have passed (so that the latest commit is marked with a green hook). Am I right that you merged by PR to your fork but that you did not yet address the other (coding-related, not mathematical) issues?

hugo-s29:
> @hugo-s29 : I have fixed the non-compiling satellites and re-ran CI for them.
> From now on, we will again expect that all the checks have passed (so that
> the latest commit is marked with a green hook). Am I right that you merged by
> PR to your fork but that you did not yet address the other (coding-related,
> not mathematical) issues?

Yes, I was a bit too optimistic with my last message. As you noticed, I rebased
your changes onto this PR, but did not answer your comments yet. Between mid-
August and the start of September, I did not have a lot of time to look at this PR.
Thanks for your patience.

> A comment on so far the latest commit in my repository with suggestions for
> changes to the present PR: rmatthes@3422ca5 : the problem of the prior
> commits was solved by not appealing to cbn before playing back the equations
> to those between the first projections of the LHS and RHS (I forgot to
> mention this in the log message).

> I have studied the entire PR and approve it modulo all the comments I already
> made here and those in my branch
> https://github.com/rmatthes/UniMath/commits/PR2087ToBeReviewed/ . As I wrote
> earlier, the code quality is very good. There are questions of style, though.
> This concerns in particular the aforementioned playing back equations to
> those between the first projections. The PR uses a recurrent idiom use
> invmap; [|use path_sigma_hprop|].. Interestingly, it looks like more often
> applicable than the usual apply subtypePath. since the latter needs to prove
> that the entire predicate is propositional, not just the used instance.
> Still, for me, there is too much boilerplate, and the proof scripts become
> less type-directed. That is to say, since no specific lemmas are applied, the
> types of the goals get out of sight. And I see numerous small "bulleting
> mistakes" related to that practice. In my branch, I solved many of those
> stylistic problems, and there is definitely no need to deal with all of them.
> But I looked at this also from the perspective of the entire UniMath library -
> also I could be mistaken about more recent stylistic choices.

As part of contributing to UniMath, my PRs should try to follow the same style
as the rest of the library. I have seen that you removed most of the instances
of this `invmap` + `path_sigma_hprop`, there were some remaining, so I updated
them.

In the 3 PRs I made, I have relied on `use invmap; [|use path_sigma_hprop|]` to
prove equalities of pairs, as I did not know about `subtypePath`. In this case,
I could have relied on `subtypePath` as the predicate was always a proposition,
no matter the instances used. Similarly, there were a few tactics that would
have been very helpful to know ahead of time (`subtypePath` is one of them, but
also `cancel_pre/postcomposition`, for example), as they weren't (as far as I
know) included in the online tutorial/guides, so it might be worth adding them.

Thanks a lot for your feedback. I should have addressed all your comments. Let
me know if I missed something.

hugo-s29:
> I have a doubt on style, seen in MonadsAsMonoidsElementary.v: some categories
> are now locally given names in plural form, such as Monoids and Monads. This
> does not convey a good intuition when using these identifiers to type a
> construction. E.g., monoid_to_monad_CAT : Monads should be read that a monad
> is constructed, not monads. I see this differently when packages formalize a
> structure of related things, and so the package name is good in plural form.

Yes, I agree with your remark, the example `monoid_to_monad_CAT : Monads`
should be using `Monad`; I changed it.

> The documentation is helpful. Of course, it can by itself not replace a paper
> presentation, especially concerning where the ideas came from. I wondered if
> there is something more general behind signature_evaluation in
> ModuleSignatures.v, but I did not see something emerge that would make the
> proof shorter.

The goal of my internship with Benedikt was to help with an article, and the
PRs I made during this internship were there to provide the formalization of
the article. So, hopefully, there will be an article presenting these results.

For `signature_evaluation`, this is mostly to justify the use of
sections, I think. Looking at the proofs of the "main" results, I do not know
where this might help to shorten the proof.

UniMath/SubstitutionSystems/CategoryOfSignaturesWithStrength.v
hugo-s29: My knowledge of bicategories and actegories is very limited. For morphisms
between signatures with pointed strength, I relied on the definition given in
the paper you recommended. As, in the article my internship was focused on, we
rely on a functor from signatures with pointed strength to module signatures, I
defined the category of signatures with pointed strength in this way, not
knowing that there was a bicategorical interpretation of this category.

I have seen that you have added commented proofs of this isomorphism in the
file `CategoryOfSignaturesWithStrength.v`. Do you think it should be included
in the file? Or was it simply to know the relation with `pointedtensorialstrength_cat`?

rmatthes:
> Similarly, there were a few tactics that would
> have been very helpful to know ahead of time (`subtypePath` is one of them, but
> also `cancel_pre/postcomposition`, for example), as they weren't (as far as I
> know) included in the online tutorial/guides, so it might be worth adding them.

@hugo-s29 It would be most welcome if you created a GitHub issue on that point. Issues do not oblige people but they keep good suggestions in the right place.

I did not check if all the bulleting is fine for now. Anyway, that is a "nice to have". I am satisfied with all the answers to my questions and remarks and the latest commits.

PR #2089, opened 2026-08-07
Blacklist subproofs generated by abstract

nbvdkamp:
I noticed that Search with all of UniMath imported took 45 minutes on my machine. By adding some custom instrumentation to Rocq I found out that about 90% of the time is being spent on two types of objects:
- subproofs generated by `abstract` which have very hairy types and shouldn't be used manually according to the docs and UniMath's style guide. These are blacklisted (from Search) in Rocq's default prelude but because that's not used they can be found.
- 'Local' objects which are blacklisted by default but time is still being spent on them because Rocq checks if the query matches before checking if an object is blacklisted. I've made a [PR to fix that](https://github.com/rocq-prover/rocq/pull/22331).

This PR blacklists objects created by `abstract` in the preamble (copying [the default prelude](https://github.com/rocq-prover/rocq/blob/6df5ae331262750d9fc2d115dd2198a6373e4dd0/theories/Corelib/Init/Prelude.v#L43-L44)) and helps get the time down to under 4 minutes (once that Rocq PR is merged), however this also means the 22 normally defined lemmas in UniMath that contain "_subproof" in their name but aren't local won't be found anymore.


nmvdw:
Looks good and thank you for this contribution: it will make `Search` much more usable. I am wondering: why are `"_subterm"` and `"Private_"` blacklisted as well? I know that `abstract` generates `_subproof`, but what generates `_subterm` and `Private_`?

nmvdw (approved):

nbvdkamp:
Thanks for the merge

> why are `"_subterm"` and `"Private_"` blacklisted as well? I know that `abstract` generates `_subproof`, but what generates `_subterm` and `Private_`?

The [docs](https://rocq-prover.org/doc/master/refman/proofs/writing-proofs/proof-mode.html#rocq:tacn.abstract) say `transparent_abstract` is a variant that generates things with `_subterm` in their name. As far as I can tell `Private_` is just a convention for names that shouldn't appear in search results.

PR #2091, opened 2026-09-04
minimal changes from coq to rocq in dune build config

rmatthes:
Do not expect the satellites to compile well unless they are also modified (which I have prepared).

### DIFF
--- a/UniMath/dune
+++ b/UniMath/dune
@@ -1,11 +1,12 @@
 
 (include_subdirs qualified)
 
-(coq.theory
+(rocq.theory
  (name UniMath)
  (package coq-unimath)
+ ;; is this still the package name in 2026?
  (theories Ltac2)
- (coqdoc_flags
+ (rocqdoc_flags
         :standard
         -interpolate
         --charset utf-8)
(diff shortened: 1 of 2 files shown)


rmatthes:
I allow myself to merge, so that I can mend the satellites. In the worst case, this simple PR could be taken back.

nmvdw:
Just to check, the only change is that a newer version is used?

rmatthes:
Yes, a newer version that also requires renaming. In the satellites, I did a little bit more to the compilation flags.

nmvdw:
Thank you for updating it!

rmatthes:
I predicted that the satellites would not compile in our CI. This was true. But after merging my PRs for the satellites (three of them merged by @nmvdw), I re-ran the checks individually. So, this operation finally ends successfully.

One curious observation: when re-running a satellite's CI check, I see that also some files in main UniMath are being compiled. This should be examined for a "normal" CI run.

rmatthes:
A note on how demanding this move is: Dune needs to be at version 3.21 or later, available since mid-January 2026. For Coq-HoTT, the maintainers decided to require at least 3.22 (less than 6 months old), see https://github.com/HoTT/Coq-HoTT/pull/2358/changes .

PR #2093, opened 2026-09-05
avoid CI failure after September 23

rmatthes:
v4 -> v5 for first-party actions checkout, cache/save, cache/restore, upload-artifact, download-artifact

see https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/ for the announcement of removal of Node20 that seems to run the v4 actions (which is not spelt out in that post)


rmatthes:
Since this PR is purely administrative in nature, I allow myself to merge it.

PR #2095, opened 2026-09-13
Add reflexive graphs library

eutro:
This PR adds a new library for reflexive graphs, in a package called `IdentitySystems`.[^1] The literature has similar constructs under the name "identity system" (e.g. [the HoTT book](https://homotopytypetheory.org/book), Egbert Rijke's [Introduction to Homotopy Type Theory](https://www.doi.org/10.1017/9781108933568)). Importantly, reflexive graphs have displayed versions, for building up reflexive graphs in towers, similarly to how one might with displayed categories and bicategories. Reflexive graphs can be considered a much more lightweight and general way to characterise identifications than univalent (bi)categories, involving considerably less data, while still providing a more robust organising framework than manually writing out all the intermediate types and proving the equivalences by hand.

Features include:
- Reflexive graphs and displayed reflexive graphs (`RXGraph.v`).
- The "fundamental theorem of identity types". As a corollary, it is possible to prove that the canonical `id_to_XYZ : a = b -> a ≈ b` is an equivalence by providing any(!) equivalence `a = b ≃ a ≈ b`.
- Many simple examples of reflexive graphs corresponding to type constructors (`Examples.v`).
- Reflexive graph lenses, for making displayed reflexive graphs easily (`Lenses.v`).

This was originally written for use in my work on duploids in UniMath that I have yet to ready for upstreaming.[^2] Much of the work here, particularly the lenses, is based on Jon Sterling's (@jonsterling) paper "Reflexive graph lenses in Univalent Foundations" ([doi:10.1017/S0960129526100565](https://doi.org/10.1017/S0960129526100565), [arXiv:2404.07854](https://arxiv.org/abs/2404.07854)). This PR includes two examples of how to use the library: the first is with the reflexive graphs of (displayed) reflexive graphs themselves, and the second is a characterization of identifications of categories.

[^1]: The rationale for the package name is that it should be generic in case anyone wants to add something similar here that isn't exactly reflexive graphs.
[^2]: On that branch, they currently live in the `Combinatorics` package, which is a horrible place for them.

### DIFF
--- a/UniMath/IdentitySystems/Examples.v
+++ b/UniMath/IdentitySystems/Examples.v
@@ -1,0 +1,519 @@
+(********************************************************************************
+
+ Constructions of reflexive graphs
+
+ This file contains a zoo of reflexive graph constructions,
+ corresponding to various type constructors.
+
+ Contains:
+ - Reflexive graphs:
+   discrete (Δ), codiscrete (∇),
+   product (∏), binary product (×),
+   binary coproduct (⨿), universe ([UU_rxgraph]),
+   opposite (^op), total ([total_rxgraph]),
+   component (⟦⟧), sub- ({ ∣ }).
+ - Displayed reflexive graphs:
+   sigma, discrete, trivial, total opposite.
+ - Notations for the above.
+
+ Author: B. Szilvasy
+ February—September 2026
+
+ ********************************************************************************)
+
+Require Import UniMath.Foundations.All.
+Require Import UniMath.MoreFoundations.All.
+
+Require Import UniMath.IdentitySystems.RXGraph.
+
+Local Open Scope rxgraph.
+
+(** ** Reflexive graphs *)
+
+(** The discrete reflexive graph [Δ A] is the reflexive graph whose vertices are
+    [A] and whose edges are identifications.  It is always univalent. *)
+
+Definition discrete_rxgraph (A : UU) : univalent_rxgraph.
+Proof.
+  use make_univalent_rxgraph.
+  1: use make_rxgraph.
+  - exact A.
+  - exact (@paths A).
+  - exact idpath.
+  - apply is_univalent_from_iscontr_edges_from; intro.
+    exact (iscontr_paths_from _).
+Defined.
+
+(** The codiscrete reflexive graph [∇ A] has vertices [A]
+    and edges [unit].  It is univalent if [A] is a proposition. *)
+
+Definition codiscrete_rxgraph (A : UU) : rxgraph.
+Proof.
+  use make_rxgraph.
+  - exact A.
+  - exact (λ _ _, unit).
+  - easy.
+Defined.
+
+Definition is_univalent_codiscrete_rxgraph (A : UU)
+  (H : isaprop A)
+  : is_univalent (codiscrete_rxgraph A).
+Proof.
+  apply is_univalent_from_weq.
+  intros a b; apply weqcontrtounit, H.
+Qed.
+
+Definition prop_rxgraph (A : UU) (H : isaprop A) : univalent_rxgraph
+  := make_univalent_rxgraph _ (is_univalent_codiscrete_rxgraph A H).
+
+Definition hProp_rxgraph (A : hProp) : univalent_rxgraph
+  := prop_rxgraph A (propproperty A).
+
+(** A family of reflexive graphs [x : B ⊢ E[x]] gives rise to a univalent
+    reflexive graph whose vertices are [∏ x, E[x]].  It is univalent
+    if all [E] are (assuming function extensionality). *)
+
+Definition product_rxgraph {B : UU} (E : B -> rxgraph) : rxgraph.
+Proof.
+  use make_rxgraph.
+  - exact (∏ x, E x).
+  - intros f g.
+    exact (∏ x, f x ≈ g x).
+  - intros f x.
+    exact (refl (f x)).
+Defined.
+
+Definition is_univalent_product_rxgraph {B : UU} (E : B -> rxgraph)
+  (H : ∏ x, is_univalent (E x))
+  : is_univalent (product_rxgraph E).
+Proof.
+  use is_univalent_from_weq.
+  intros f g.
+  apply (weqcomp (weqtoforallpaths _ _ _)).
+  use weqonsecfibers; intro a.
+  apply (weq_id_to_edge (H a)).
+Qed.
+
+Definition product_rxgraph' {B : UU} (E : B -> univalent_rxgraph) : univalent_rxgraph
+  := make_univalent_rxgraph _ (is_univalent_product_rxgraph E (λ x, E x)).
+
+(** The binary product [A × B] of reflexive graphs [A] and [B] has the obvious
+    vertices and edges the binary products of [A]- and [B]-edges.  It is univalent
+    if both [A] and [B] are. *)
+
+Definition dirprod_rxgraph (A B : rxgraph) : rxgraph.
+Proof.
+  use make_rxgraph.
+  - exact (A × B).
+  - intros a b.
+    exact (pr1 a ≈ pr1 b × pr2 a ≈ pr2 b).
+  - intro a.
+    exact (refl (pr1 a),, refl (pr2 a)).
+Defined.
+
+Lemma is_univalent_dirprod_rxgraph (A B : rxgraph)
+  (HA : is_univalent A)
+  (HB : is_univalent B)
+  : is_univalent (dirprod_rxgraph A B).
+Proof.
+  apply is_univalent_from_weq.
+  intros aa bb.
+  apply (weqcomp pathsdirprodweq).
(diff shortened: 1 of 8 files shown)


nmvdw:
Thank you for the contribution. The reason why the CI fails, is because `CategoryRXGraph.v` is not listed in `CategoryTheory/.package/files`. I will review this PR today.

UniMath/IdentitySystems/Examples.v
```diff
@@ -0,0 +1,519 @@
...
+
+(** ** Reflexive graphs *)
+
+(** The discrete reflexive graph [Δ A] is the reflexive graph whose vertices are
+    [A] and whose edges are identifications.  It is always univalent. *)
+
+Definition discrete_rxgraph (A : UU) : univalent_rxgraph.
+Proof.
+  use make_univalent_rxgraph.
+  1: use make_rxgraph.
```
nmvdw: What about using `-` here instead of `1:`? One would need to nest a bit deeper, but one can then more clearly see which parts belong to constructing a reflexive graph and which ones contribute to univalence.

UniMath/IdentitySystems/Examples.v
```diff
@@ -0,0 +1,519 @@
...
+  : is_univalent (codiscrete_rxgraph A).
+Proof.
+  apply is_univalent_from_weq.
+  intros a b; apply weqcontrtounit, H.
+Qed.
+
+Definition prop_rxgraph (A : UU) (H : isaprop A) : univalent_rxgraph
+  := make_univalent_rxgraph _ (is_univalent_codiscrete_rxgraph A H).
+
+Definition hProp_rxgraph (A : hProp) : univalent_rxgraph
```
nmvdw: The name `hProp_rxgraph` suggests to me a reflexive graph whose type of vertices is `hProp`. WHat about `hProp_to_rxgraph`?

UniMath/IdentitySystems/Examples.v
```diff
@@ -0,0 +1,519 @@
...
+  intros x y.
+  apply (weqcomp (equality_by_case_equiv x y)).
+  induction x as [a₁ | b₁]; induction y as [a₂ | b₂]; cbn.
+  - apply (weq_id_to_edge HA).
+  - exact (idweq empty).
+  - exact (idweq empty).
+  - apply (weq_id_to_edge HB).
+Qed.
+
+Definition coprod_rxgraph'
```
nmvdw: I prefer the name `coprod_univalent_rxgraph`

UniMath/IdentitySystems/Examples.v
```diff
@@ -0,0 +1,519 @@
...
+Definition UU_rxgraph (* ℓ *) : rxgraph (* ℓ+1 *).
+Proof.
+  use make_rxgraph.
+  - exact (UU (* ℓ *)).
+  - exact weq.
+  - exact idweq.
+Defined.
+
+Goal univalenceStatement = is_univalent UU_rxgraph.
+  reflexivity.
```
nmvdw, line 240: It would be better to add `Proof` here (not putting `Proof` gives a warning).

UniMath/IdentitySystems/Examples.v
```diff
@@ -0,0 +1,519 @@
...
+
+Lemma is_univalent_opp_rxgraph (A : rxgraph)
+  (HA : is_univalent A)
+  : is_univalent (opp_rxgraph A).
+Proof.
+  apply (is_univalent_from_isaprop_edges_from (opp_rxgraph A)).
+  exact (is_univalent_to_isaprop_edges_to A HA).
+Defined.
+
+Definition opp_rxgraph' (A : univalent_rxgraph) : univalent_rxgraph
```
nmvdw: I think that the name `opp_univalent_rxgraph` would be more fitting

UniMath/IdentitySystems/RXGraph.v
```diff
@@ -0,0 +1,610 @@
...
+ ********************************************************************************)
+
+Require Import UniMath.Foundations.All.
+Require Import UniMath.MoreFoundations.All.
+
+Declare Scope rxgraph.
+Delimit Scope rxgraph with rxgraph.
+Local Open Scope rxgraph.
+
+Declare Scope rxgraph_spec.
```
nmvdw: Why are `rxgraph` and `rxgraph_spec` two different scopes?

UniMath/IdentitySystems/RXGraph.v
```diff
@@ -0,0 +1,610 @@
...
+    : rxgraph_edge_rect_right G H b P base b (refl b) = base.
+  Proof.
+    refine (maponpaths (λ p, transportb _ p base) (_ : _ = idpath _)).
+    apply proofirrelevancecontr, (is_univalent_to_isaprop_edges_to G H).
+  Defined.
+
+  (** Examples using the induction schemes with the [elim] and [induction]
+      tactics. *)
+
+  Goal ∏ (G : univalent_rxgraph) (Q : UU) (a b : G) (f : G -> Q) (e : a ≈ b), f a = f b.
```
nmvdw, line 368: Rocq gives a warning if a proof, including one under `Goal`, is not started by `Proof.`

UniMath/IdentitySystems/RXGraphOfRXGraphs.v
```diff
@@ -0,0 +1,341 @@
...
+Local Bind Scope rxgraph_spec with rxgraph.
+Local Bind Scope rxgraph_spec with univalent_rxgraph.
+
+(** ** Reflexive graph of graphs *)
+
+Definition graph_on_rxgraph (* ℓ *)
+  : contra_lens (* ℓ+1 *) (UU_rxgraph (* ℓ *)).
+Proof.
+  use make_contra_lens.
+  - intros A; exact (∏ (_ _ : A), UU_rxgraph (* ℓ *))%rxgraph_spec.
```
nmvdw: Would it be convenient to open scope `rxgraph_spec` in this file? Or does that lead to clashes in notation?

nmvdw (approved):
Thank you for the interesting contribution! It looks very good, and I think that it can be merged soon. My only comments are small. In the file `Examples.v`, I think that some names can be changed: the name `dirprod_univalent_rxgraph` is more informative than `dirprod_rxgraph'`, so I think the first would be preferable.

UniMath/IdentitySystems/RXGraphOfRXGraphs.v
eutro: Currently the type former notations like `∏` and `×` have conflicting bindings in `rxgraph_spec` and `type`. It might be better to just have syntaxes that don't conflict with `type` so that `rxgraph` and `rxgraph_spec` can be merged.

UniMath/IdentitySystems/Examples.v
eutro: Honestly I get lazy reindenting and changing bullets so I prefer to keep things flat, but I suppose you are right that it is clearer.

eutro:
I'll change all the univalent type formers to use `univalent` instead of `'`. They are more accessible through the syntax anyway.

UniMath/IdentitySystems/Examples.v
```diff
@@ -94,9 +102,24 @@ Proof.
...
 
+Notation "'∏~' x .. y , G" :=
+  (product_rxgraph (λ x, .. (product_rxgraph (λ y, G)) ..))
+    (at level 200, x binder, y binder, right associativity) : rxgraph.
+Notation "'∏!' x .. y , G" :=
+  (product_univalent_rxgraph (λ x, .. (product_univalent_rxgraph (λ y, G)) ..))
+    (at level 200, x binder, y binder, right associativity) : rxgraph.
+(* type in Emacs using agda-input with \prod *)
+
+Notation "A '~>' B" :=
```
jonsterling, line 116: I wonder if `⋔` might be a more appropriate operator, since it's a power rather than an exponential? But maybe it doesn't matter so much :)

eutro:
There is an ambiguous coercion path that will annoy me to no end so I'll have to fix it.

UniMath/IdentitySystems/RXGraph.v
eutro: I have combined them now and changed the notations not to clash with type formers.

UniMath/IdentitySystems/Examples.v
eutro: Maybe, I chose `~>` because I wanted it to look like the type former.

nmvdw:
Are there any additional changes you would like to do before I merge this PR?

eutro:
Thanks for the quick review. I don't think there's anything I want to change about the code, but I would like to ask that the commits be squashed for merging (as I think you tend to anyway).

nmvdw:
> Thanks for the quick review. I don't think there's anything I want to change about the code, but I would like to ask that the commits be squashed for merging (as I think you tend to anyway).

Done!

